PR #186, opened 2014-11-26
Adding an overload of ContainSingle() without arguments

abatishchev:
Adding to SelfReferencingCollectionAssertions an overload of ContainSingle() with no meaningful arguments meaning the collection has only 1 item:

```
public AndConstraint<TAssertions> ContainSingle(string because = "", params object[] reasonArgs)
```

It's useful in a case when method in the real code returns a collection (e.g. `IEnumerable<T>`) but in the test just a single item.

Currently it looks like this:

```
actual.Should().ContainSingle(x => x == expected);
```

or

```
actual.Should().HaveCount(1);
```

but is more explicit and readable (imo).

### DIFF
--- a/FluentAssertions.Core/Collections/SelfReferencingCollectionAssertions.cs
+++ b/FluentAssertions.Core/Collections/SelfReferencingCollectionAssertions.cs
@@ -270,6 +270,34 @@ namespace FluentAssertions.Collections
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
+        /// <summary>
+        /// Expects the current collection to contain only a single item.
+        /// </summary>
+        /// <param name="because">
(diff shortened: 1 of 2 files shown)


abatishchev:
What did I do wrong that this overload didn't appear in none of the recent releases? I mean did I miss something or it just wasn't releases yet?

vossad01:
There have no feature releases since this was integrated.  v3.2.2 was released after this was integrated, but it was minor release off the 'master' branch mainly to support the Xamarin packages (IIRC) and did not contain new features that had been added to the 'develop' branch.

If you would like earlier access, you can get a v3.3.0 pre-release build from [Fluent Assertions' MyGet](https://www.myget.org/gallery/fluentassertions)

Issue #157, opened 2014-09-17
Cyclic reference and ShouldBeEquivalentTo - is this a valid solution?


fnajera-rac-de:
Hi,

I want to deep-test two objects for equivalency, but they contain cycles so the library halts with "but it contains a cyclic reference.". I know it is possible to ignore those properties, but I want them tested.

The solution I have found is:

* EquivalencyValidator: remove the line

        if (!current.Get<ObjectTracker>("objects").IsCyclicReference(new ObjectReference(context.Subject, context.PropertyPath)))

* Create a new IAssertionRule class, and use it (surely this can use some refactoring / improvements):

        private class MyAssertionRule : IAssertionRule
        {

            private List<Tuple<object, object>> _list = new List<Tuple<object, object>>();

            public bool AssertEquality(IEquivalencyValidationContext context)
            {
                var exists = _list.Any(x => object.ReferenceEquals(x.Item1, context.Subject) && object.ReferenceEquals(x.Item2, context.Expectation));
                if (exists)
                {
                    return true;
                }
                else
                {
                    _list.Add(new Tuple<object, object>(context.Subject, context.Expectation));
                    return false;
                }
            }

        }

The reasoning is:

* The assertion rule is called for each pair of matching properties.

* By removing the "if" line, I don't fail on cycles but I don't ignore them either. Note that by doing so without adding the new rule you will certainly hit the maximum depth limit.

* In my rule, if it's the first time I see a given pair of values, then I cannot conclude if they match, so I return false, but I note down that they appeared before together.

    * If those values are indeed not equivalent, then the process will eventually stop with an error, so it finally won't matter that I noted them down.

    * If those values are found equivalent by another rule or by finishing the deep inspection, then they were indeed a match, and I have that match on my list, which connects with the following point.

* If I have seen this pair of values already, then I can return true, because either they were compared before and they were found to be equivalent - or - this specific pair of objects is being compared right now and this property specifies "the same cycle" in both objects.

* An important point is that any given pair of objects will be thoroughly tested once (because the first time I see them I return false).

Am I missing something here? If not, could this be a universal way of dealing with cycles - maybe applied by default?

Thanks and regards,

Fernando

dennisdoomen:
@vossad01 This is still ongoing work, isn't it?

vossad01:
[...]

Issue #134, opened 2014-07-08
ShouldBeEquivalentTo throws TargetInvocationException: Exception has been thrown by the target of an invocation with IPAddress


knguyen0928:
I found this in >= 2.2.0

Test:
```
void Main()
{
	var bar1 = new Bar {Address = IPAddress.Parse("1.2.3.4"), Word = "a"};
	var bar2 = new Bar {Address = IPAddress.Parse("1.2.3.4"), Word = "a"};
	bar1.ShouldBeEquivalentTo(bar2);
}

class Bar
{
	public IPAddress Address {get;set;}
	public string Word {get;set;}
}
```
Throws:
```
TargetInvocationException: Exception has been thrown by the target of an invocation.
```
Stacktrace:
```
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimePropertyInfo.GetValue(Object obj, BindingFlags invokeAttr, Binder binder, Object[] index, CultureInfo culture)
   at System.Reflection.RuntimePropertyInfo.GetValue(Object obj, Object[] index)
   at FluentAssertions.Equivalency.EquivalencyValidationContext.CreateForNestedProperty(PropertyInfo nestedProperty, PropertyInfo matchingProperty) in z:\Builds\work\d21bade4fe20a99b\FluentAssertions.Core\Equivalency\EquivalencyValidationContext.cs:line 91
   at FluentAssertions.Equivalency.ComplexTypeEquivalencyStep.Handle(EquivalencyValidationContext context, IEquivalencyValidator parent, IEquivalencyAssertionOptions config) in z:\Builds\work\d21bade4fe20a99b\FluentAssertions.Core\Equivalency\ComplexTypeEquivalencyStep.cs:line 48
   at FluentAssertions.Equivalency.EquivalencyValidator.AssertEqualityUsing(EquivalencyValidationContext context) in z:\Builds\work\d21bade4fe20a99b\FluentAssertions.Core\Equivalency\EquivalencyValidator.cs:line 74
[...]

Issue #22, opened 2013-09-26
ShouldBeEquivalentTo does not compare fields.


TorbenRahbekKoch:
It works fine on properties, but public fields are not taken into account. One can discuss whether this is a good thing or not, though ;)

Thanks for a very nice fluent api, with very few surprises.

vossad01:
Anyone care to weigh in on the behavior of this edge case?

``` c#
[TestMethod]
public void When_asserting_equivalence_including_only_fields_it_should_not_match_properties()
{
	//-----------------------------------------------------------------------------------------------------------
	// Arrange
	//-----------------------------------------------------------------------------------------------------------
	var onlyAField = new ClassWithOnlyAField { Value = 1 };
	var onlyAProperty = new ClassWithOnlyAProperty { Value = 1 };

	//-----------------------------------------------------------------------------------------------------------
	// Act
	//-----------------------------------------------------------------------------------------------------------
	Action act = () => onlyAField.ShouldBeEquivalentTo(
            onlyAProperty,
            opts => opts.IncludingAllDeclaredFields());

	//-----------------------------------------------------------------------------------------------------------
	// Assert
	//-----------------------------------------------------------------------------------------------------------
	???
}
```

My initial inclination is that it should fail because the matching member on the expectation object is a   property and we specified to only include fields.

However, we do not respect the Runtime/Declared distinction on the expectation object, so perhaps we should not be respecting the Property/Field distinction either.

dennisdoomen:
It should fail. Fields and properties are fundamentally different things and should not be confused with the Runtime/Declared distinction which is more about visibility.

dennisdoomen:
This was already completed in release 3.3.0.

Issue #15, opened 2013-09-06
Hang on ShouldBeEquivalent


wied03:
I'm not sure what happened, but in FluentAssertions 2.0.1, this code sample worked fine and it stopped working in 2.1.0.  I simplified as much as I could to reproduce the issue and it seems confined to use NodaTime's (version 1.1) LocalTime objects.

I have a ZIP file with the solution but GitHub won't let me attach it, so I'll just paste in the code here:

using System;
using System.Collections.Generic;
using System.Linq;
using System.Linq.Expressions;
using FluentAssertions;
using NodaTime;
using NUnit.Framework;

namespace HangProblem
{
    public class NestedClass
    {
        public LocalTime Start { get; private set; }
        public LocalTime End { get; private set; }
        public IsoDayOfWeek DayOfWeek { get; private set; }

```
    public NestedClass(IsoDayOfWeek dayOfWeek,
                       LocalTime start,
                       LocalTime end)
    {
        Start = start;
        End = end;
        DayOfWeek = dayOfWeek;
    }
}

public class Class1
{
    public string Name { get; set; }
    public IEnumerable<NestedClass> Nested { get; set; }

    public Class1(string name,
                  IEnumerable<NestedClass> nested)
    {
        Name = name;
        Nested = nested;
    }
}

[TestFixture]
public class SomeTest1
{
    private static readonly LocalTime Midnight = new LocalTime(hour: 0,
                                                               minute: 0);

    private static readonly LocalTime JustBeforeMidnight = new LocalTime(hour: 23,
                                                                         minute: 59,
                                                                         second: 59);

    #region Utility Methods

    private static IEnumerable<NestedClass> GetNestedObjects()
    {
        for (var index = 0; index <= 7; index++)
        {
            var day = (IsoDayOfWeek) index;
            // skip monday or none
            if (day == IsoDayOfWeek.Monday || day == IsoDayOfWeek.None)
            {
                continue;
            }
            var obj = new NestedClass(day,
                                      new LocalTime(hour: 0,
                                                    minute: 0),
                                      new LocalTime(hour: 23,
                                                    minute: 59,
                                                    second: 59));
            yield return obj;
        }
    }

    #endregion

    #region Tests

    [Test]
    public void Fails_properly()
    {
        var actual = new NestedClass(IsoDayOfWeek.Thursday,
                                    new LocalTime(hour: 0,
                                                    minute: 0),
                                      new LocalTime(hour: 23,
                                                    minute: 59,
[...]

PR #189, opened 2014-11-27
Starts to replace AssertionRules in favor of EquivalencySteps

adamvoss:
This is groundwork for a lot of things but should not make any changes impacting users (other than `AssertionRule` being deprecated, though I do not know why anyone would use that directly).  This is simpler than history might suggest, but I am sure it will need some discussion.

This is working to achieve a number of things.  The first reason is what got me started on this, but it ties in with a number of other things that have crossed my mind from time to time.
1. Eventually deprecate and remove the `IAssertionRule` interface.  It and `IEquivalencyStep` serve the same function but `IEquivalencyStep` is more robust, if one can be eliminated it will reduce duplication of effort in maintaining two similar interfaces.  This duplication as a burden becomes apparent when trying to address Issue #22.
2. Ultimately expose a `Using` IEquivalencyStep to allow users/library designers to provide their own additions to the Equivalency pipeline.  This could be things like a NodaTime step that better handles its types to mitigate things like Issue #15 or .NET types like in #134.  Not sure if it will actually be used, but it could be powerful.  Use of this extensibility will be more practical when there is central configuration as is coming out of #171.
3. Provide a way that configuration of the equivalence assertions can affect the root object.  Arguably, it could be seen as an abuse of `ShouldBeEquivalentTo` to use it where you need to define custom assertion behavior on the root object; instead you should be using `Should`.  However, given FA defines special behaviors for certain root object types with `ShouldBeEquivalentTo` I would imagine users could find value it as well and do not see the need to restrict it only to properties.
4. I could see value in also exposing the built-in steps in some fashion so they can be removed or reordered if desired by the user (ex. user never wants the conversion step), but I think it is premature to decide that is actually needed.  I am thinking maybe something like a the situation in Issue #157 with wanting very different dictionary might start to point towards this flexibility.

This does not deprecate `IAssertionRule` yet or expose `IEquivalencyStep` because I think Issue #22 cause changes to some of the interfaces.

When debugging the Equivalence validation, this also makes it a little easier to follow the impact of custom configuration because there are no longer two levels of check (Assertion Rule and Equivalency Step).

The good news is I do not think I did anything with boolean expressions in this one, so I am probably no missing parenthesis in those again :)

This is based on #177, will either need to be rebased or merged after that one. Of course, let me know if there is something you want me to do.


vossad01:
I finished this deprecation, https://github.com/vossad01/fluentassertions/tree/DeprecateIAssertionRule, at this point I don't have a strong reason to do this for the next release though.

PR #204, opened 2015-01-04
Include fields in equivalency assertions

vossad01:
This should resolve Issue #22.

In addition to including Fields in `ShouldBeEquivalentTo` it also makes the Binary and XML Serialization assertions check fields.

Not sure if there is any etiquette or practices around this, but I did not include the full history of my work thinking this was cleaner.

### DIFF
--- a/FluentAssertions.Core/Equivalency/EquivalencyAssertionOptions.cs
+++ b/FluentAssertions.Core/Equivalency/EquivalencyAssertionOptions.cs
@@ -101,7 +101,7 @@ namespace FluentAssertions.Equivalency
         public EquivalencyAssertionOptions()
         {
             IncludingNestedObjects();
-            IncludingAllDeclaredProperties();
+            IncludingAllDeclaredMembers();
         }
     }
 }
--- a/FluentAssertions.Core/Equivalency/EquivalencyAssertionOptions.cs
+++ b/FluentAssertions.Core/Equivalency/EquivalencyAssertionOptions.cs
@@ -46,7 +46,7 @@ namespace FluentAssertions.Equivalency
         /// </summary>
         public EquivalencyAssertionOptions<TSubject> Excluding(Expression<Func<TSubject, object>> expression)
         {
-            AddSelectionRule(new ExcludeMemberByPathSelectionRule(expression.GetPropertyPath()));
+            AddSelectionRule(new ExcludeMemberByPathSelectionRule(expression.GetMemberPath()));
             return this;
         }
 
@@ -88,7 +88,7 @@ namespace FluentAssertions.Equivalency
         public EquivalencyAssertionOptions<TSubject> WithStrictOrderingFor(
             Expression<Func<TSubject, object>> expression)
         {
-            orderingRules.Add(new PathBasedOrderingRule(expression.GetPropertyPath()));
+            orderingRules.Add(new PathBasedOrderingRule(expression.GetMemberPath()));
             return this;
         }
     }
--- a/FluentAssertions.Core/Collections/GenericCollectionAssertions.cs
+++ b/FluentAssertions.Core/Collections/GenericCollectionAssertions.cs
@@ -74,7 +74,7 @@ namespace FluentAssertions.Collections
                     .ForCondition(unordered.SequenceEqual(expectation))
                     .BecauseOf(because, args)
                     .FailWith("Expected collection {0} to be ordered by {1}{reason} and result in {2}.",
-                        Subject, propertyExpression.GetPropertyPath(), expectation);
+                        Subject, propertyExpression.GetMemberPath(), expectation);
             }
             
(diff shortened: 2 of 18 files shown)


FluentAssertions.Core/Equivalency/EquivalencyAssertionOptions.cs
```diff
@@ -101,7 +101,7 @@ public class EquivalencyAssertionOptions : EquivalencyAssertionOptionsBase<Equiv
         public EquivalencyAssertionOptions()
         {
             IncludingNestedObjects();
-            IncludingAllDeclaredProperties();
+            IncludingAllDeclaredMembers();
```
vossad01: Is it recognized that this changes the default behavior to always include fields and properties?

dennisdoomen: No, and I think that would be a serious breaking change. Fields should represent internal state, so including them in the assertion would only make sense in exceptional situations,

vossad01: I do not disagree with you on the role field should generally have in design.  What is in the balance here is not how users of the library should design their code, but how FA should respond to the code used with it.

I have mentioned before how `ShouldBeEquivalentTo` is one of those features has a "just works" expectation.  If a object does publicly expose a field (even if design wise that is exceptional) it would be expected that it is included.  If that field is not checked then I think false positives are subject to occur in this case because users are used to it working.  If they are lucky they write a test that should fail and have to go though the head-scratching, confusion, and debugging as to what is wrong before finally realizing it is a bug/limitation in FA and not an issue with their code.

dennisdoomen: Yeah, I think you're right after all. I didn't realize that we were talking about the public fields only. That's what you get when you read those discussions between the other work or during breakfast :-)

dennisdoomen:
So can I merge?

vossad01:
That is fine with me.  There is that minor fix-up you pointed out and dropping or reverting bbf75d2b3098308da18460674eeeb863502e0767 if you still don't want it.  I can do those things and push, but if you are waiting to merge, you can probably do it just as quick and easily locally.

FluentAssertions.Core/Equivalency/EquivalencyAssertionOptions.cs
vossad01: I should note that it follows the same "NonPublic" behavior used with properties.  It my knowledge `ShouldBeEquivalentTo` does not have any built in support for private members.  What I don't know off the top of my head is how you classified `internal` and `protected` with respect to non-public.  I can check quick.

vossad01: Looks like `private`, `internal`, and `protected internal` are Private, and `public` and `protected` are Public.

dennisdoomen: I would be surprised if `protected` would be included as well.

vossad01: Ahh, you are right.  I forgot to scroll up to look at the binding flags

```
private const BindingFlags PublicMembersFlag =
        BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance;
```

dennisdoomen:
No, just do that minor thing and keep bbf75d2. But somebody should really spend some time on the documentation for the next release.

vossad01:
I think you are right about that.  I think I neglected to up the doc-comments on the Serialization methods.

I can take a look at `ShouldBeEquivalentTo` this weekend I think.  I have known there was some outdated material in the wiki a while back (before I started making any changes in the area) and I do not know if anyone else has already fixed that up.

What is the state of the organization there?  The README.md redirects to that new website but there is also content in the GitHub wiki.

dennisdoomen:
The new web site is just the landing page and has references to the documentation, the nuget package site and the binaries. I would like to have to present the wiki documentation directly from www.fluentassertions.com, but I haven't found the time to invest in that Jekyll stuff yet.

FluentAssertions.Core/Equivalency/EquivalencyAssertionOptions.cs
vossad01: I had expected this was going to need some discussion.  I am glad to hear this can make it in >400,000 downloads on NuGet is nothing to scoff at; I would hate to bring about changes or instability that would turn people away.  But at the same am trying to make it the best I can going forward too.

While I was thinking this would not make it I thought through a 'next-best-option' and an it is still stuck in my mind that it might still be a good idea in addition to this.  I will start a discussion on that elsewhere.

PR #210, opened 2015-01-11
Different API for including fields

vossad01:
Switches to a more granular API for specifying runtime/declared typing and inclusion of properties/fields in ShouldBeEquivalentTo.

The thinking is this API will work better with the global configuration being introduced for #137.  The idea being global configuration could specify `ExcludingFields()` and a test just wanting runtime typing only needs to specify `RespectingRuntimeTypes()` and not need to possibly override the `ExcludingFields()` configuration.

### DIFF
--- a/FluentAssertions.Core/Equivalency/EquivalencyAssertionOptionsBase.cs
+++ b/FluentAssertions.Core/Equivalency/EquivalencyAssertionOptionsBase.cs
@@ -198,7 +198,7 @@ namespace FluentAssertions.Equivalency
         /// </remarks>
         public TSelf IncludingAllDeclaredProperties()
         {
-            RespectingDeclaredType();
+            RespectingDeclaredTypes();
 
             ExcludingFields();
             IncludingProperties();
@@ -216,7 +216,7 @@ namespace FluentAssertions.Equivalency
         /// </remarks>
         public TSelf IncludingAllRuntimeProperties()
         {
-            RespectingRuntimeType();
+            RespectingRuntimeTypes();
 
             ExcludingFields();
             IncludingProperties();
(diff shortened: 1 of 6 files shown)


FluentAssertions.Core/Equivalency/EquivalencyAssertionOptionsBase.cs
```diff
@@ -67,18 +72,55 @@ protected EquivalencyAssertionOptionsBase(IEquivalencyAssertionOptions defaults)
             includeProperties = defaults.IncludeProperties;
             includeFields = defaults.IncludeFields;
 
-            selectionRules.AddRange(defaults.SelectionRules);
+            selectionRules.AddRange(defaults.SelectionRules.Where(IsNotStandardSelectionRule));
+
+            if (IncludesIncludingSelectionRule(defaults.SelectionRules))
```
vossad01: I do not think there is a practical way for this code to be hit at this point in time.  It more is an attempt to future proof for the global confgiuration

FluentAssertions.Net40/ObjectAssertionsExtensions.cs
```diff
@@ -47,8 +47,8 @@ public static class ObjectAssertionsExtensions
             {
                 object deserializedObject = CreateCloneUsingBinarySerializer(assertions.Subject);
 
-                EquivalencyAssertionOptions<T> defaultOptions = new EquivalencyAssertionOptions<T>()
-                    .IncludingAllRuntimeMembers();
+                EquivalencyAssertionOptions<T> defaultOptions = EquivalencyAssertionOptions<T>.Default()
```
vossad01: With the advent of global configuration you will need to decide if you want this respecting that configuration or not.  Changing it this way at the moment so that it at least matches XML serialization.

PR #216, opened 2015-01-17
Added StyleCop

konstmsu:
I've added `StyleCop` to all projects, it runs on every build and issues warnings when there are coding standards issues. I've also fixed typos and some other issues that StyleCop reported.
I haven't noticed much difference in compile time with and without StyleCop so it's enabled by default. This is also convenient because every contributor will see the warnings without any extra effort.

There are some potential improvements:
* To slightly improve build time only add StyleCop to a subset of projects (Net45 but not Net40, for example). Not sure if we won't miss anything if we do this
* Treat StyleCop warnings as errors. Would need to be done at project level or at TeamCity level. Instructions are here: http://stackoverflow.com/questions/24804315/warnings-as-errors-does-not-apply-to-stylecop-warnings
* Run StyleCop only on TeamCity or only in Release mode. We can discuss these this additionally if anybody thinks it's a good idea.

PS. Thumbs up for having "Treat Warnings as Errors" enabled by default!


dennisdoomen:
Well done man!

Does it also run on release builds? I don't see any Stylecop references in the TeamCity build.

konstmsu:
I've updated `default.ps1` to fail the build when there are StyleCop issues. I assumed that `default.ps1` is what is used by TeamCity, let me know if it should be something else.
Also I've changed encoding of `default.ps1` to be UTF-8 as otherwise GitHub client was thinking that it's a binary file and didn't show diff correctly.

dennisdoomen:
Seems like reverting the date time offset changes also affected this pull request.

dennisdoomen:
In order for the Visual Studio build to fail on StyleCop warnings, I have to re-enable _Warnign as Errors_. Any objections?

konstmsu:
I thought this might be too obtrusive in development. Say you add a public
method and without a xml-doc you won't even be able to run the code.
So in my opinion it's best to leave failures to a later stage - TeamCity
build.

dennisdoomen:
I could live with that. Can you see the details of such a build in TeamCity? In other words, when the PR is marked as _failed_, can you as a contributor figure out what's wrong?

vossad01:
Yep, I can confirm once you go to TeamCity you can tell what is wrong.  It pulls out the relevant text from the log.  The error does not propagate back to github though.

http://teamcity.codebetter.com/viewLog.html?buildTypeId=bt1236&buildId=178778

vossad01:
Thanks for doing this by the way.  While it may not be fun to be called out on typos, it certainly benefits FA with better documentation.

PR #217, opened 2015-01-20
Adds `BeAssignableTo` implementation for `Type` objects.

vossad01:
In the absence of this new hiding definition, FA would try to assert the Type, `Type`, was assignable to the type.  An assertion having no real value.

This also updates the formatting on the specs of the hidden method.

This also corrects a slight inaccuracy in the failure message of the hidden method.

### DIFF
--- a/FluentAssertions.Shared.Specs/TypeAssertionSpecs.cs
+++ b/FluentAssertions.Shared.Specs/TypeAssertionSpecs.cs
@@ -363,6 +363,56 @@ namespace FluentAssertions.Specs
 
         #endregion
 
+        #region BeAssignableTo
+
+        [TestMethod]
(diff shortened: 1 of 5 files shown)


FluentAssertions.Shared.Specs/TypeAssertionSpecs.cs
```diff
@@ -363,6 +363,56 @@ public void When_type_is_not_equal_to_the_same_type_using_generics_it_should_thr
 
         #endregion
 
+        #region BeAssignableTo
+
+        [TestMethod]
+        public void When_asserting_an_object_is_assignable_its_own_type_it_should_succeed()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange / Act / Assert
```
vossad01: Is it acceptable to combine all 3 in simple assertions like this?

PR #218, opened 2015-01-20
New extension methods to make filtering of `Type`s functionality more accessible

vossad01:
This is done on the premise that these methods are rather useful; but, hard to find.

These extension methods are still on a fairly specialized type so they  should not cause much of a pollution problem.  Even then these might be fairly common things to want to do on that type so I still think it would not be pollution.

### DIFF
--- a/FluentAssertions.Core/Core.csproj
+++ b/FluentAssertions.Core/Core.csproj
@@ -206,6 +206,7 @@
     <Compile Include="Specialized\ExceptionAssertions.cs" />
     <Compile Include="Specialized\IExtractExceptions.cs" />
     <Compile Include="TimeSpanConversionExtensions.cs" />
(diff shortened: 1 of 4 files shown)


konstmsu:
I would argue that this is out of scope of FluentAssertions. These methods are definitely very useful but some might be tempted to reference FluentAssertions from non-test projects just to get them.

vossad01:
I see your point.  From my point of view the [feature is already in FA](https://github.com/dennisdoomen/fluentassertions/wiki#type-method-and-property-assertions) but can be hard to find.  So long as the feature is there, should make it as convenient as possible for users.

As it happens, [people already do reference FA for other things](https://github.com/dennisdoomen/fluentassertions/issues/71) though maybe not for these methods in particular.

PR #229, opened 2015-02-12
Better support for rendering a TimeSpan.MinValue/MaxValue. 

### LINKED ISSUE #212: TimeSpanValueFormatter throws on OverflowException when its value contains TimeSpan.MinValue
When an assertion fails, and its object contains a TimeSpan field which has a value of TimeSpan.MinValue, an OverflowException is thrown.
This is because TimeSpanValueFormatter.ToString() calls TimeSpanValueFormatter.GetNonZeroFragments(), which calls timeSpan.Duration(), which throws OverflowException exception if its Ticks equals to TimeSpan.MinValue.Ticks.

dennisdoomen:
Should fix #212

### DIFF
--- a/FluentAssertions.Core/Formatting/TimeSpanValueFormatter.cs
+++ b/FluentAssertions.Core/Formatting/TimeSpanValueFormatter.cs
@@ -37,6 +37,16 @@ namespace FluentAssertions.Formatting
         {
             var timeSpan = (TimeSpan) value;
 
+            if (timeSpan == TimeSpan.MinValue)
+            {
+                return "min-value";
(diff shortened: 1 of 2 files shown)


splintor:
Thanks for fixing. Might be a good idea if the returned string contained information that related it to time span, as someone who reads it, might not be aware that this is for a TimeSpan object (e.g. if it is a member of a class with many other members). I'd use **min-TimeSpan-value** and **max-TimeSpan-value**.

vossad01:
Regarding formatting and type awareness, see also Issue #4

PR #261, opened 2015-04-21
Resolve FailWith overload resolution when only constant-string is passed

slogen:
(This is my first github pull request, so please bear with me)

I was unable to compile the develop branch in 26be9f6295d9b3ab6cbca8b4ee800fd9ca5674ae due to overload resolution on FailWith.

### DIFF
--- a/FluentAssertions.Core/Execution/GivenSelector.cs
+++ b/FluentAssertions.Core/Execution/GivenSelector.cs
@@ -113,5 +113,24 @@ namespace FluentAssertions.Execution
 
             return new ContinuationOfGiven<T>(this, parentScope);
         }
+
+        /// <summary>
+        /// Sets the failure message when the assertion is not met, or completes the failure message set to a 
+        /// prior call to to <see cref="WithExpectation"/>.
+        /// </summary>
+        /// <remarks>
+        /// If an expectation was set through a prior call to <see cref="WithExpectation"/>, then the failure message is appended to that
+        /// expectation. 
+        /// </remarks>
+        /// <param name="message">The format string that represents the failure message.</param>
+        public ContinuationOfGiven<T> FailWith(string message)
+        {
+            if (evaluateCondition)
+            {
+                parentScope.FailWith(message, new object[0]);
+            }
+
+            return new ContinuationOfGiven<T>(this, parentScope);
+        }
     }
 }


FluentAssertions.Core/Execution/GivenSelector.cs
```diff
@@ -113,5 +113,24 @@ public ContinuationOfGiven<T> FailWith(string message, params object[] args)
...
+        /// <summary>
+        /// Sets the failure message when the assertion is not met, or completes the failure message set to a 
+        /// prior call to to <see cref="WithExpectation"/>.
+        /// </summary>
+        /// <remarks>
+        /// If an expectation was set through a prior call to <see cref="WithExpectation"/>, then the failure message is appended to that
+        /// expectation. 
+        /// </remarks>
+        /// <param name="message">The format string that represents the failure message.</param>
+        public ContinuationOfGiven<T> FailWith(string message)
```
dennisdoomen: How is that different from the `FailWith` that takes a `params` array?

slogen: When I compile 26be9f6 in visual studio 2015 i get 6 compiler errors of the type:

```
Show Details	Severity	Code	Description	Project	File	Line
	Error	CS0121	The call is ambiguous between the following methods or properties: 'GivenSelector<T>.FailWith(string, params Func<T, object>[])' and 'GivenSelector<T>.FailWith(string, params object[])'	Core	CollectionAssertions.cs	1074
```

Caused by code invoking .FailWith with only the string parameter, as here in CollectionAssert.cs line 1074:

```C#
        public AndConstraint<TAssertions> StartWith(object element, string because = "", params object[] becauseArgs)
        {
            Execute.Assertion
                .WithExpectation("Expected {context:collection} to start with {0}{reason}, ", element)
                .ForCondition(!ReferenceEquals(Subject, null))
                .FailWith("but the collection is {0}.", (object)null)
                .Then
                .Given(() => Subject.Cast<object>())
                .ForCondition(subject => subject.Any())
                .FailWith("but the collection is empty.") // Line 1074 <<<<<<<<<<< Ambiguous 
                .Then
                .Given(objects => objects.FirstOrDefault())
                .ForCondition(first => first.IsSameOrEqualTo(element))
                .FailWith("but found {0}.", first => first); // <<<<<<<<<<< No problem

            return new AndConstraint<TAssertions>((TAssertions) this);
        }
```

This is (of course) because the compiler cannot choose between the two available FailWith implementations:

```C#
        public ContinuationOfGiven<T> FailWith(string message, params Func<T, object>[] args);
        public ContinuationOfGiven<T> FailWith(string message, params object[] args);
```

as the type of the 2nd `params T args` cannot be deduced from the 1st parameter.

The ambiguity can be resolved by:

1. Refactoring the call-sites to call with an empty array of objects `.FailWith("static msg", new object[0])`
2. Creating an overload with a single parameter, which will have precedence over the previous, and thus resolve the ambiguity

Option 2 looked most attractive to me, however I don't have enough "feel" for the code-style in the project to make an *informed* suggestion :)

dennisdoomen:
Thanks for the elaborate explanation. I noticed that VS2015 is more strict about ambigious overloads, but I didn't fully switch yet.

Issue #214, opened 2015-01-12
AssertionFailedException does not cross AppDomain boundaries


ECrank:
I think this Exception class should be serializable so that you can run tests in a separate AppDomain and a failure exception will be reported in the unit test AppDomain. I am using the following pattern to perform tests that rely on uninitialized static state:
```C#
AppDomain domain = AppDomain.CreateDomain("TestSingletonTrue", AppDomain.CurrentDomain.Evidence, AppDomain.CurrentDomain.SetupInformation);
try
{
    domain.DoCallBack(() =>
    {
        // Do Test
    });
}
finally
{
    AppDomain.Unload(domain);
}
```
On failure, I get the following message: 
>System.Runtime.Serialization.SerializationExceptionType 'FluentAssertions.Execution.AssertionFailedException' in assembly 'FluentAssertions, Version=2.0.1.0, Culture=neutral, PublicKeyToken=33f2691a05b67b6a' is not marked as serializable.

vossad01:
I see no reason that `AssertionFailedException` cannot be made serializable.

Do you think you could prepare a Pull Request for that change?

In case you are not familiar with the behavior, I should mention that the type of exception thrown will vary depending on the test framework (NUnit, MSTest, etc.) that was detected or configured.  So long as FA is using the [`FallbackTestFramework`](https://github.com/dennisdoomen/fluentassertions/blob/5b3eaa336e2fb347afe2cfe29f9b3c5a1153b8f6/Shared/Execution/FallbackTestFramework.cs) it will use `AssertionFailedException`.

konstmsu:
Microsoft guidelines suggest exceptions should be  serializable so this is
definitely the right thing to do
http://msdn.microsoft.com/en-us/library/vstudio/ms229064%28v=vs.100%29.aspx

dennisdoomen:
Fixed by PR #228.

florianwittmann:
Actually PR #228 is missing the serialization constructor as described here: http://stackoverflow.com/questions/94488/what-is-the-correct-way-to-make-a-custom-net-exception-serializable
Therefore currently it still can not be used cross-appdomain as deserialization is failing.

PR #262, opened 2015-04-22
Add serialization constructor for AssertionFailedException

florianwittmann:
Currently the  AssertionFailedException is missing a serializable constructor. Therefore it throws an error when deserializing the exception.

See: http://stackoverflow.com/questions/94488/what-is-the-correct-way-to-make-a-custom-net-exception-serializable

Addition to implement #214 properly

### DIFF
--- a/FluentAssertions.Shared.Specs/EventAssertionSpecs.cs
+++ b/FluentAssertions.Shared.Specs/EventAssertionSpecs.cs
@@ -2,9 +2,12 @@
 using System.ComponentModel;
 using FluentAssertions.Events;
 using FluentAssertions.Formatting;
+
+#if NET40 || NET45
 using System.IO;
 using System.Runtime.Serialization.Formatters.Binary;
 using FluentAssertions.Execution;
+#endif
 
 #if !OLD_MSTEST
 using Microsoft.VisualStudio.TestPlatform.UnitTestFramework;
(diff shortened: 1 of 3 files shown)


dennisdoomen:
Thanks. Unfortunately, it contains some StyleCop warnings (that are treated as errors these days). Could you please fix those?

florianwittmann:
Sure, working on it.

florianwittmann:
Okay the unit test is failing (Like in manual test)
Test method 

>Type 'FluentAssertions.Execution.AssertionFailedException' in Assembly 'FluentAssertions, Version=3.4.0.4, Culture=neutral, PublicKeyToken=33f2691a05b67b6a' is not marked as serializable

Actually it is:
>  #if NET40 || NET45
     [Serializable]
 #endif

Can't get the error here?
Commenting out the conditionals in AssertionFailedException, will make the test green for Net40 and Net45, but will have the other platforms to throw compile errors.

dennisdoomen:
Shouldn't you wrap that unit test in `#if NET40 || NET45`

FluentAssertions.Shared.Specs/EventAssertionSpecs.cs
```diff
@@ -578,7 +583,6 @@ public void When_a_monitored_class_in_not_referenced_anymore_it_should_be_garbag
         }
 
 #if NET40 || NET45
```
florianwittmann: Should not this 
> #if NET40 || NET45
> [...]
> #endif

 block already span both unit tests?

florianwittmann:
But yes, sure it's for NET40 and NET45 only.

dennisdoomen:
I found the problem. The .NET 4.0 assembly doesn't define the NET40 symbol, so the `[Serializable]` attribute is not included.

florianwittmann:
Okay, passing now, but a lots of commits, I would prepare a clean fresh pull request.

PR #266, opened 2015-05-04
Resolves #265 - FluentAssertions.Dnx

### LINKED ISSUE #265: Port to .Net Core
As mentioned in gitter, I would like to see FluentAssertions ported to .Net Core. Hopefully it will be a nice and easy port. I'm new to .Net Core myself, but happy to give a hand where I can.

smudge202:
This addresses issue #265.

I created a new DNX Project which will selectively compile sources from other projects in order to produce a `FluentAssertions.Dnx` NuGet package.

Due to the current solution structure, these outputs will generate one directory **below** the solution in an artifacts folder. I leave it up to @dennisdoomen as to whether the solution structure should be updated to fit more [current methodologies](https://gist.github.com/davidfowl/ed7564297c61fe9ab814) or whether the artifacts folder should be redirected (no idea how to be honest).

Most changes are very minor, just tedious `ifdef`s.

Hopefully enough to go from.

I have not looked into merging DNX support with the primary FluentAssertions package yet. Unless that can be done fairly quickly, I'm hoping this DNX package can be published to the wild as a beta package (DNX is also beta), with the intent to merge DNX compatibility into the main NuGet package when DNX comes out of beta?


smudge202:
The PR will obviously require changes made to CI in order to build. @mattridgway can probably provide some TeamCity tips regarding vNext builds if we ask nicely.

There is an issue with this code:

```
No platform-specific assembly detected
```

This occurs in the `ProbingAdapterResolver` so clearly my hack & slash ifdefs didn't work there. I'm going to drop this whole PR and create a new one tomorrow using the new TypeInfo mechanics throughout instead of ifdef'ing them in (if def reverse compatibility, not the other way round like I have here).

For example, the recent commits to the [Compose Port](https://github.com/smudge202/compose/commits/CoreCLR).

However, if anyone has time to look into the _platform-specific assembly_ issue, that'd be greatly appreciated.

One last note, @davidfowl does not like the FluentAssertions._Dnx_ naming, though after a brief chat I don't believe it would be possible right now to fully port the solution given all the profiles supported. (Feel free to correct me if I'm wrong).

dennisdoomen:
Great work and thanks so far.

PR #267, opened 2015-05-06
Core clr2

smudge202:
I almost bricked my windows installing the Win80 sdk so I simply can't do all the projects.

However, if you're happy with the pattern I've gone with here @dennisdoomen this should be very easy to expand upon.

Considerable change; I'm happy to continue working on the components I can and any changes required as a result of PR review.

I'll try to remain very active in the [gitter channel](https://gitter.im/dennisdoomen/fluentassertions) for any questions.

I'll also continue adding remaining profiles (those that I can get SDKs for compatible with Win10!).


smudge202:
For what it's worth, @mattrigway and @herecydev are generally responsible for reviewing my code, so they may help with general code cleanup.

FluentAssertions.Core/TypeInfo.cs
```diff
@@ -0,0 +1,47 @@
...
+			this.type = type;
+		}
+
+		public Assembly Assembly { get { return type.Assembly; } }
+
+		public CustomAttributeData[] CustomAttributes
+		{
+			get
+			{
+#if CUSTOM_ATTRIBUTEDATA
```
herecydev: If this isn't defined then you won't have a return value for the get. Therefore no compilation.

src/FluentAssertions/project.lock.json
```diff
@@ -0,0 +1,910 @@
+{
```
herecydev: This shouldn't be included

herecydev:
Looking good to me @smudge202, I haven't checked behavior but I'm sure there are plenty of tests covering this.

smudge202:
Thanks for taking a look @herecydev and nice spots.

I haven't added behaviour so hopefully it makes sense to have not added tests. The intent was to not modify behaviour so simply comply with existing tests. Unfortunately I failed at that (there are 3 failing tests as mentioned in [gitter](https://gitter.im/dennisdoomen/fluentassertions)).

dennisdoomen:
Now that @onovotny has completed #288. Is this now fully covered? I will release 4.0 beta in a couple of days.

smudge202:
Yes, this PR can die now.

dennisdoomen:
OK. Regardless, thank for trying to make this happen.

PR #273, opened 2015-06-17
Added support for recursively comparing two multi-dimensional arrays …

### LINKED ISSUE #260: ShouldBeEquivalentTo does not detect difference in multi-dimensional array
The test cases below show that ShouldBeEqualTo does not detect differences in multi-dimensional arrays.

```C#
    public class FluentAssertionArrayTest
    {
        [Test]
        public void When_applying_ShouldBeEquivalentTo_to_equal_multi_dimensional_array_it_should_not_throw()
        {
            var x = new[,] { { 1, 2, 3 }, { 4, 5, 6 } };
            var y = new[,] { { 1, 2, 3 }, { 4, 5, 6 } };
            Action act = () => x.ShouldBeEquivalentTo(y);
            act.ShouldNotThrow();
        }
        [Test]
        public void When_applying_ShouldBeEquivalentTo_to_a_multi_dimensional_array_with_diff_it_should_throw()
        {
            var x = new[,] {{1, 2, 3}, {4, 5, 6}};
            var y = new[,] { { 1, 2, 3 }, { 4, -5, 6 } };
            Action act = () => x.ShouldBeEquivalentTo(y);
            act.ShouldThrow<Exception>();
            // Fails: "Expected a <System.Exception> to be thrown, but no exception was thrown."
        }
        [Test]
        public void When_applying_ShouldBeEquivalentTo_to_a_single_dimensional_array_with_diff_it_should_throw()
        {
            var x = new[] {4, 5, 6};
            var y = new[] { 4, -5, 6 };
            Action act = () => x.ShouldBeEquivalentTo(y);
            act.ShouldThrow<Exception>();
        }
        [Test]
        public void When_applying_ShouldAllBeEquivalentTo_to_a_single_dimensional_array_with_diff_it_should_throw()
        {
            var x = new[] { 4, 5, 6 };
            var y = new[] { 4, -5, 6 };
            Action act = () => x.ShouldAllBeEquivalentTo(y);
            act.ShouldThrow<Exception>();
        }
    }
```

dennisdoomen:
…for equivalency. Should fix #260.

Fixed a bug when chaining multiple assertions together.
During equivalency assertions, a declared type of object is still processed as a run-time type.

### DIFF
--- a/FluentAssertions.Core/Core.csproj
+++ b/FluentAssertions.Core/Core.csproj
@@ -65,6 +65,7 @@
     <Compile Include="Common\NullReflector.cs" />
     <Compile Include="Common\PlatformAdapter.cs" />
     <Compile Include="Common\ProbingAdapterResolver.cs" />
+    <Compile Include="Equivalency\MultiDimensionalArrayEquivalencyStep.cs" />
     <Compile Include="Equivalency\Selection\AllPublicFieldsSelectionRule.cs" />
     <Compile Include="EquivalencyStepCollection.cs" />
     <Compile Include="Equivalency\AssertionContext.cs" />
--- a/FluentAssertions.Core/Equivalency/EnumerableEquivalencyValidator.cs
+++ b/FluentAssertions.Core/Equivalency/EnumerableEquivalencyValidator.cs
@@ -108,7 +108,7 @@ namespace FluentAssertions.Equivalency
         {
             using (var scope = new AssertionScope())
             {
-                parent.AssertEqualityUsing(context.CreateForCollectionItem(expectationIndex, subject, expectation));
+                parent.AssertEqualityUsing(context.CreateForCollectionItem(expectationIndex.ToString(), subject, expectation));
 
                 return scope.Discard();
             }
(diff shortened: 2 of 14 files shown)


vossad01:
"During equivalency assertions, a declared type of object is ***still*** processed as a run-time type" (emphasis mine)

Isn't that actually a change that largely reverts #161 rather than a preservation of behavior?

dennisdoomen:
In my opinion it's a small correction. The problem is that without it, non-generic collections and dictionaries will never be able to benefit from `ShouldBeEquivalentTo`. I couldn't even add support for multi-dimensional arrays without it. Functionally, I would be surprised if an `object` property doesn't get compared at all. The main reason for making a distinction between the declared and the run-time type was to support comparing interfaces and only including the members the interface defines.

vossad01:
The non-generic collections were still able to benefit from `ShouldBeEquivalentTo`,  a user just needs to indicate their intention by calling `RespectingRuntimeTypes()`.  In my opinion, `ShouldBeEquivalentTo` is complex enough and "special case" behaviors like treating collections differently with respect to runtime vs declared types makes it harder for users to know how it is going to behave.  Yes, per #116 any two objects are found equivalent, when `RespectingDeclaredTypes()`, though an error is given if the root object is type `object`.  It is possible to implement multi-dimensional array support while preserving the runtime/declared type behavior that was in place.

dennisdoomen:
v3.4 will still respect the declared type all the way down (or up) the object graph, _unless_ a property or element is of type `object`. In that case, it will use the run-time type. This evaluation happens for every evaluation of the `CompileTimeType` in the [EquivalencyValidationContext](https://github.com/dennisdoomen/fluentassertions/blob/develop/FluentAssertions.Core/Equivalency/EquivalencyValidationContext.cs#L90). IMHO, it doesn't make sense to throw an exception just because some nested property is defines as `object.

Issue #271, opened 2015-05-27
Assertions to verify assembly and type structure


piotrpMSFT:
I would like to expand the capabilities of the existing reflection assertions [Assembly, Type, MethodInfo, PropertyInfo] and add a few related assertions [ConstructorInfo, MemberInfo, MethodBase] to enable reasoning about the structure of assemblies and the types they contain.

**AssemblyAssertions**
AssemblyAssertions should expose the ability to assert that an assembly exposes a particular type:

```csharp
this.GetType().Assembly
    .Should().DefineType("MyNamespace", "MyTestClass")
         .Which.Should().BeDecoratedWith<TestClassAttribute>();
```

**TypeAssertions**
TypeAssertions should allow us to assert that the type has expected members, does not have other members, is derived from a base class, and implements some interfaces:

```csharp
typeof(MyClass)
    .Should().Implement(typeof(IEnumerable), "Because I like those linq extension methods")
       .And.Implement<IEnumerable>("Because this syntax is often more convenient!")
       .And.NotImplement(typeof(IOrderedEnumerable), "Because I intentionally don't guarantee order")
       .And.NotImplement<IOrderedEnumerable>("Because this syntax is often more convenient!")
       .And.BeDerivedFrom(typeof(MyBaseClass), "Because I'm a specialization of that class")
       .And.BeDerivedFrom<MyBaseClass>("Because this syntax is often more convenient!")
       .And.HaveIndexer(new [] { typeof(int) }, "Because I'm a collection of some sort")
       .And.NotHaveIndexer(new [] { typeof(string) }, "Because that would be inefficient")
       .And.HaveProperty(typeof(string), "MyStringProperty", "Because my users expect it")
       .And.HaveProperty<string>("MyStringProperty", "Because this syntax is often more convenient!")
       .And.NotHaveProperty(typeof(string), "YourStringProperty", "Because it belongs on a different class")
       .And.NotHaveProperty<string>("YourStringProperty", "Because this syntax is often more convenient!")
       .And.HaveMethod("Jump", new [] { typeof(long) }, "Because how high?")
       .And.NotHaveMethod("Jump", new [] { typeof(string) }, "Because AtoI is not our concern here")
       .And.HaveExplicitProperty(typeof(IHaveAName), "Name", "Because properties are sometimes implemented explicitly to avoid name collisions")
       .And.HaveExplicitProperty<IHaveAName>("Name", "Because this syntax is often more convenient!")
[...]

PR #277, opened 2015-06-23
Extended Type Assertions

piotrpMSFT:
Addresses Issue #271


dennisdoomen:
Now that's what I call a pull request ;-)

But on a serious not, well done. Thanks for that. I'll have to look at those changes a bit more and clean-up the code a bit, so it'll probably take me a couple of days to merge the changes. 

Any change you could provide me with an updated version of that section in the wiki so that I can copy it over as soon as 3.5 is released?

piotrpMSFT:
Dennis, take a lo K at the associates issue. Is that good for the wiki?

dennisdoomen:
Should be good enough. Thanks.

PR #279, opened 2015-06-30
NumericAssertions small fix

GDoronin:
- NumericAssertions BePositive XML-Doc comment fixed.
- Few tests added for BePositive/BeNegative when asserting zero value.


dennisdoomen:
Thanks for the fix. Would you be willing to update those specs to match the [current style](https://github.com/dennisdoomen/fluentassertions/blob/develop/FluentAssertions.Shared.Specs/NumericAssertionSpecs.cs#L93)?

GDoronin:
Sure, can fix specs.
Can you please provide more details?

By [current style](https://github.com/dennisdoomen/fluentassertions/blob/develop/FluentAssertions.Shared.Specs/NumericAssertionSpecs.cs#L93) you mean using  Arrange/Act/Assert sections?

Do you want me to fix spec methods I've just added to match the current style?

Should there be Arrange section or not? Since I don't arrange anything.

dennisdoomen:
Yes, the AAA structure as well as the naming style. If you just update the new ones you added, i'll look at those others late ron.

GDoronin:
Okay, let's keep this pull request open then.
Will update it.

Name should be something like `When_asserting_zero_value_to_be_positive_it_should_throw`?

And the last question is still unclear.
Should there be Arrange section or not? Since I don't arrange anything.

```csharp
// Arrange

// Act
Action action = () => (0).Should().BePositive();

// Assert
act.ShouldThrow<AssertFailedException>();
```

dennisdoomen:
Oh sorry, no, if there's nothing to arrange, leave it out.

GDoronin:
Migrated to AAA structure.
It only remains to rename methods.

dennisdoomen:
Thanks. You're going to do that as well?

GDoronin:
Yes, I will.

GDoronin:
Renaming also has been done.

dennisdoomen:
Thanks!

PR #285, opened 2015-07-12
Added null-check for when Subject is null

sietsevdschoot:
I found an issue in the nullable NumericAssertions where FluentAssertions would throw a NullReferenceException if the subject would be null and the expected value would be not null. 

I've added a test and a small code fix.

Best regards,


dennisdoomen:
Merged manually into *develop*

PR #297, opened 2015-09-17
Allow SelfReferencingCollectionAssertions to compare collections of different types using predicate

AlexanderSher:
Sometimes it is required to verify collection of objects that have no public constructor and not abstracted by interface, so there is no way to create an expected collection of objects of exactly the same type. In this case, it is useful to compare it with collection of anonymous objects that contain all required data:
```c#
IList<TypeWithPrivateCtor> actual = GetActual();
var expected = new [] 
{
   new { P1 = "a", P2 = 0 },
   new { P1 = "b", P2 = 1 },
   new { P1 = "c", P2 = 2 },
}
actual.Should().Equal(expected, (a, e) => a.P1 == e.P1 && a.P2 == e.P2);
```

### DIFF
--- a/Src/Core/Collections/CollectionAssertions.cs
+++ b/Src/Core/Collections/CollectionAssertions.cs
@@ -201,12 +201,12 @@ namespace FluentAssertions.Collections
         /// </param>
         public AndConstraint<TAssertions> Equal(IEnumerable expected, string because = "", params object[] reasonArgs)
         {
-            AssertSubjectEquality<object>(expected, (s, e) => s.IsSameOrEqualTo(e), because, reasonArgs);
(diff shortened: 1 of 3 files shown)


dennisdoomen:
Thanks for the PR. But isn't this what `ShouldBeEquivalentTo` is doing as well?

AlexanderSher:
AFAIK `ShouldBeEquivalentTo` doesn't allow to chain constraints and doesn't allow to use custom equality comparers (which can be more complicated than just comparing properties or fields).

dennisdoomen:
Ah yes. Good point.

dennisdoomen:
Looks good. Waiting for appveyor to build it.

PR #300, opened 2015-09-29
NotBeDecoratedWith attribute assertion added

levenleven:
Used it to test my web api methods are not decorated with Authorize attribute


dennisdoomen:
Thanks!

PR #321, opened 2015-12-08
Added failing test for parallel execution of event monitoring.

frankebersoll:
Unfortunately, event monitoring seems to use unsynchronized access to static members. This leads to multithreading problems whenever a test runner happens to execute tests in parallel, as xUnit 2 does for instance. 

This pull requests adds a unit test that shows the behavior by running `.MonitorEvents()` 1000 times, using Parallel LINQ. On a default Azure A2 development machine, this would reproduce the issue every time.


dennisdoomen:
Oh shoot, removing develop also closed the PR. Sorry @frankebersoll

frankebersoll:
Not a problem, just added one for Master: #323

PR #325, opened 2015-12-12
Event monitoring will now work in multi-threaded unit test frameworks…

### LINKED ISSUE #320: Bug in EventRecordersMap.cs
While Unittesting with XUnit, some tests failed, because EnventRecordersMap is not thread safe.
Also this issue just apear as we run the tests within the Test Explorer of Visual Studio 2013. 
NCrunsh shows all tests as passed.

Can somebody confirm that?

Thanks and regards,
Akuzo

dennisdoomen:
… such as XUnit. Fixes #320.

### DIFF
--- a/Src/Shared/AssertionExtensions.cs
+++ b/Src/Shared/AssertionExtensions.cs
@@ -7,7 +7,6 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Xml.Linq;
-
 using FluentAssertions.Collections;
 using FluentAssertions.Equivalency;
 using FluentAssertions.Events;
@@ -17,9 +16,9 @@ using FluentAssertions.Reflection;
 using FluentAssertions.Specialized;
 using FluentAssertions.Types;
 using FluentAssertions.Xml;
-
 #if NET45 || WINRT || CORE_CLR
 using System.Threading.Tasks;
+
 #endif
 
 namespace FluentAssertions
@@ -126,7 +125,7 @@ namespace FluentAssertions
         /// </summary>
         public static Action Enumerating<T>(this Func<IEnumerable<T>> enumerable)
         {
-            return () => ForceEnumeration(() => (IEnumerable)enumerable());
+            return () => ForceEnumeration(() => (IEnumerable) enumerable());
         }
 
         private static void ForceEnumeration(Func<IEnumerable> enumerable)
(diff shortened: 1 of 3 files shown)


Src/Shared/AssertionExtensions.cs
```diff
@@ -617,20 +616,21 @@ public static PropertyInfoSelectorAssertions Should(this PropertyInfoSelector pr
...
+            EventRecorder[] recorders = eventSource
+                .GetType()
+                .GetEvents()
+                .Select(@event => CreateEventHandler(eventSource, @event)).ToArray();
 
             if (!recorders.Any())
             {
                 throw new InvalidOperationException(
-                    string.Format("Type {0} does not expose any events.", eventSource.GetType().Name));
+                    $"Type {eventSource.GetType().Name} does not expose any events.");
```
Meir017: This is a C# 6 feature, I'm not sure you want use it since all of the project doesn't use it's features (yet)

dennisdoomen: Why not? You can only build FluentAssertions with Visual Studio 2015, and it doesn't affect the target assemblies.

Meir017: Just wanted to be sure it wasn't just a resharper hint you used

dennisdoomen: Ah, no, but thanks for mentioning this though.

Tests/FluentAssertions.Shared.Specs/EventAssertionSpecs.cs
```diff
@@ -311,6 +312,33 @@ public void When_the_event_parameters_do_match_it_should_not_throw()
...
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //----------------------------------------------------------------------------------------------------------
+            Action<int> action = _ =>
+            {
+                EventRaisingClass subject = new EventRaisingClass();
+                subject.MonitorEvents();
+                subject.RaiseEventWithSender();
+                subject.ShouldRaise("PropertyChanged");
```
Meir017: could use ````nameof(EventRaisingClass.PropertyChanged)```` instead of ````"PropertyChanged"````

dennisdoomen: :+1:

PR #330, opened 2015-12-17
monitor events for implemented interface or base class only

gstamac:
I added a new generic extension method MonitorEvents&lt;T&gt;. It works simmilar to MonitorEvents but it will only monitor events of type T. It is useful when you don't want to monitor all events of an object. It is very useful when monitoring events of an object created using System.Reflection.Emit. In this case MonitorEvents will not work because the events added using TypeBuilder will not be located using non-generic version of MonitorEvents. With generic version of MonitorEvents you can provide the interface implemented using TypeBuilder. Check the tests to see what I mean.

### DIFF
--- a/Src/Shared/AssertionExtensions.cs
+++ b/Src/Shared/AssertionExtensions.cs
@@ -611,9 +611,10 @@ namespace FluentAssertions
 
 #if !SILVERLIGHT && !WINRT && !PORTABLE && !CORE_CLR
         /// <summary>
-        ///   Starts monitoring an object for its events.
+        ///   Starts monitoring <paramref name="eventSource"/> for its events.
         /// </summary>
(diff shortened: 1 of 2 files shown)


dennisdoomen:
:+1:

dennisdoomen:
@gstamac I have some minor code review remarks. Any change that you're willing to handle these?

dennisdoomen:
I'll squash the commits myself and then merge it.

gstamac:
Great. Thanks. And thank you for the library.

dennisdoomen:
Merged through #332.

dennisdoomen:
@gstamac Would you be willing to update the wiki as well?

gstamac:
Added a description in Wiki. Please check it out.

dennisdoomen:
:+1:

PR #331, opened 2015-12-19
Added properties for the access modifier of getter and setter on SelectedMemberInfo

tomachristian:
I added properties to get the access modifiers of the getter and setter on `SelectedMemberInfo`.
For field members it uses the same access modifier for both the getter and setter.

```csharp
CSharpAccessModifier GetAccessModifier { get; }
// and
CSharpAccessModifier SetAccessModifier { get; }
```

Also added extension methods on `ISubjectInfo` for checking the access modifier of the setter or getter in a fluent manner:

- `subjectInfo.WhichSetterHas(CSharpAccessModifier)`
- `subjectInfo.WhichSetterDoesNotHave(CSharpAccessModifier)`
- `subjectInfo.WhichGetterHas(CSharpAccessModifier)`
- `subjectInfo.WhichGetterDoesNotHave(CSharpAccessModifier)`

Usage sample

```csharp
subject.ShouldBeEquivalentTo(expected, config =>
                config.Excluding(ctx => ctx.WhichGetterHas(CSharpAccessModifier.Internal) ||
                                        ctx.WhichGetterHas(CSharpAccessModifier.ProtectedInternal)))
```

The `GetAccessModifier` and `SetAccessModifier` properties can also be directly used from the equivalency check configurator as follows:

```csharp
subject.ShouldBeEquivalentTo(expected, config =>
                config.Excluding(ctx => ctx.SelectedMemberInfo.GetAccessModifier == CSharpAccessModifier.Internal ||
                                        ctx.SelectedMemberInfo.SetAccessModifier == CSharpAccessModifier.ProtectedInternal)
```

### DIFF
--- a/Src/Core/Equivalency/FieldSelectedMemberInfo.cs
+++ b/Src/Core/Equivalency/FieldSelectedMemberInfo.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using System.Reflection;
+using FluentAssertions.Common;
 
 namespace FluentAssertions.Equivalency
 {
@@ -29,6 +30,16 @@ namespace FluentAssertions.Equivalency
         public override Type MemberType
         {
             get { return fieldInfo.FieldType; }
+        }
+
+        internal override CSharpAccessModifier GetAccessModifier
+        {
+            get { return fieldInfo.GetCSharpAccessModifier(); }
+        }
+
+        internal override CSharpAccessModifier SetAccessModifier
+        {
+            get { return fieldInfo.GetCSharpAccessModifier(); }
         }
     }
 }
--- a/Src/Core/Common/CSharpAccessModifierExtensions.cs
+++ b/Src/Core/Common/CSharpAccessModifierExtensions.cs
@@ -35,6 +35,36 @@ namespace FluentAssertions.Common
             return CSharpAccessModifier.InvalidForCSharp;
         }
 
+        internal static CSharpAccessModifier GetCSharpAccessModifier(this FieldInfo fieldInfo)
+        {
+            if (fieldInfo.IsPrivate)
+            {
+                return CSharpAccessModifier.Private;
+            }
+
+            if (fieldInfo.IsFamily)
+            {
+                return CSharpAccessModifier.Protected;
+            }
+
+            if (fieldInfo.IsAssembly)
+            {
+                return CSharpAccessModifier.Internal;
+            }
+
+            if (fieldInfo.IsPublic)
+            {
+                return CSharpAccessModifier.Public;
+            }
+
+            if (fieldInfo.IsFamilyOrAssembly)
+            {
+                return CSharpAccessModifier.ProtectedInternal;
+            }
+
+            return CSharpAccessModifier.InvalidForCSharp;
+        }
+
         internal static CSharpAccessModifier GetCSharpAccessModifier(this Type type)
         {
             if (type.GetTypeInfo().IsNestedPrivate)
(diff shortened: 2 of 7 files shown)


tomachristian:
As I'm not very familiar with the desired structure of the project, I require input on the following:
- Are the names and docs used for the properties correct ?
- Is the implementation for this feature tackled correctly ?
- Should there be a `SelectionRule` for this ?
- What should the specs cover and where should they be placed ? Are the current specs enough ?
- Should there be any extra selection methods on `EquivalencyAssertionOptions` ? Like

```csharp
config => config.ExcludingMembersOf(CSharpAccessModifier.Internal)
```

or is this too specific ?
- Is the usage of `CSharpAccessModifier` in this correct, or should a different type be used ?

Src/Core/Equivalency/FieldSelectedMemberInfo.cs
```diff
@@ -30,5 +31,15 @@ public override Type MemberType
         {
             get { return fieldInfo.FieldType; }
         }
+
+        public override CSharpAccessModifier GetAccessModifier
```
dennisdoomen: Rather than adding a property, wouldn't it not make more sense to replace them with methods, e.g. WhichSetterHasModifier(CSharpAccessModifier.Internal);

dennisdoomen:
Thanks for the initiative. 

Since this is the first time I've seen the need for this feature, just supporting these kind of filters on the `ISubjectInfo` is sufficient for now.  

One thing I have in mind is to merge the `ISubjectInfo` and the `SelectedMemberInfo` to prevent too many chained properties. But that'll be a breaking change, so I'll postpone that as much as possible.

Src/Core/Equivalency/FieldSelectedMemberInfo.cs
tomachristian: Indeed this looks like it would point the user into the right direction.
- Wouldn't `WhichHasSetterModifier` be better ? Or is this the followed convention, with the `Which` ?
- What would the negated version of this be named ? (This is the one I personally need)
- So no more public property on SelectedMemberInfo for the access modifier.

tomachristian:
Will move to `ISubectInfo` as methods, please see note above with additional questions.

Src/Core/Equivalency/FieldSelectedMemberInfo.cs
dennisdoomen: * Well, if you would state the same thing in English, you would say something like `which setter has modifier Internal`.
* WhichSetterDoesNotHaveModifier. On second thought, maybe remove the `Modifier` suffix from all methods.  
* I would remove the public property.

tomachristian:
On holiday, will continue when back online

dennisdoomen:
@tomachristian still on a holiday?

abatishchev:
@tomachristian ping

tomachristian:
@dennisdoomen Hey, sorry for the late response, but not much time on my hands right now.

I rebased the branch and amended the last commit to match what you requested. Am I on the right track ?

If I am, I will proceed in adding summary comments, and more specs for the negated versions of the methods), although I don't feel comfortable having that code in each implementation of `ISubjectInfo`. I would make extension methods out of those `WhichSetterHasModifier` and `WhichGetterHasModifier` methods.

What do you think ?

dennisdoomen:
Looks good!

dennisdoomen:
@tomachristian you gave up?

tomachristian:
Finally had a bit of time to clean this up.

- Moved access modifier check methods from ISubjectInfo to extension methods.
- Cleaned code and added comments.
- Updated PR description.
- Rebased branch.

Please take a look.

PR #333, opened 2015-12-27
Fixed (non-)optional 'because' and 'reasonArgs' for Dictionaries

jbattermann:
GenericDictionaryAssertions<,>.ContainKeys(..) had a non-optional 'because' argument and
so did .ContainValuesAndWhich(..) - more precisely, the later for 'reasonArgs'. Both were/are inconsistent with the rest (looked like an oversight - these two parameters are used optionally / as params[] everywhere else).

Cheers,
-J


dennisdoomen:
:+1:

jbattermann:
\o/

PR #335, opened 2015-12-30
Added GenericDictionaryAssertions .Contain(..) and .NotContain() that take an IEnumable<KeyValuePair<,,>>

jbattermann:
The GenericDictionaryAssertions.cs file already contained .Contain() and .NotContain() methods that took a single KeyValuePair < TKey, TValue > but I am current writing tests against a custom IDictionary implementation and wanted to check more fluently whether more than one KeyValuePair is or is not in a given Dictionary.. so I figured why not adding it to FluentAssertions :)

Besides the two Assertion methods themselves, I took the existing tests targeting the single-key/value pair methods and used them as base for the added tests.

Cheers,
-Jörg

### DIFF
--- a/Src/Core/Collections/GenericDictionaryAssertions.cs
+++ b/Src/Core/Collections/GenericDictionaryAssertions.cs
@@ -610,8 +610,8 @@ namespace FluentAssertions.Collections
                 {
                     Execute.Assertion
                         .BecauseOf(because, reasonArgs)
-                        .FailWith("Expected {context:dictionary} to contain {0}{reason}, but {1} differs at key {2}.",
-                            expectedKeyValuePairs, Subject, keyValuePairsNotSameOrEqualInSubject.Select(keyValuePair => keyValuePair.Key));
+                        .FailWith("Expected {context:dictionary} to contain {0}{reason}, but {context:dictionary} differs at keys {1}.",
+                            expectedKeyValuePairs, keyValuePairsNotSameOrEqualInSubject.Select(keyValuePair => keyValuePair.Key));
                 }
                 else
                 {
--- a/Src/Core/Collections/GenericDictionaryAssertions.cs
+++ b/Src/Core/Collections/GenericDictionaryAssertions.cs
@@ -551,7 +551,7 @@ namespace FluentAssertions.Collections
         /// Asserts that the current dictionary contains the specified <paramref name="expected"/>.
         /// Keys and values are compared using their <see cref="object.Equals(object)" /> implementation.
         /// </summary>
-        /// <param name="expected">The expected <see cref="KeyValuePair{TKey,TValue}">key value pairs</see></param>
+        /// <param name="expected">The expected key/value pairs.</param>
         /// <param name="because">
         /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
         /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
@@ -578,7 +578,7 @@ namespace FluentAssertions.Collections
             {
                 Execute.Assertion
                     .BecauseOf(because, reasonArgs)
-                    .FailWith("Expected {context:dictionary} to contain key value pairs {0}{reason}, but found {1}.", expected, Subject);
+                    .FailWith("Expected {context:dictionary} to contain key/value pairs {0}{reason}, but found {1}.", expected, Subject);
             }
 
             var expectedKeys = expectedKeyValuePairs.Select(keyValuePair => keyValuePair.Key).ToArray();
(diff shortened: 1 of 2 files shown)


Src/Core/Collections/GenericDictionaryAssertions.cs
```diff
@@ -551,6 +551,86 @@ public GenericDictionaryAssertions(IDictionary<TKey, TValue> dictionary)
...
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="reasonArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<GenericDictionaryAssertions<TKey, TValue>> Contain(IEnumerable<KeyValuePair<TKey, TValue>> expected,
+            string because = "", params object[] reasonArgs)
+        {
+            if (expected == null)
```
dennisdoomen: A lot of these edge cases are not covered by tests.

Src/Core/Collections/GenericDictionaryAssertions.cs
```diff
@@ -551,6 +551,86 @@ public GenericDictionaryAssertions(IDictionary<TKey, TValue> dictionary)
...
+            }
+
+            if (ReferenceEquals(Subject, null))
+            {
+                Execute.Assertion
+                    .BecauseOf(because, reasonArgs)
+                    .FailWith("Expected {context:dictionary} to contain key/value pairs {0}{reason}, but found {1}.", expected, Subject);
+            }
+
+            var expectedKeys = expectedKeyValuePairs.Select(keyValuePair => keyValuePair.Key).ToArray();
```
dennisdoomen: I prefer not to use `var` unless the type is visible. See [AV1520](https://github.com/dennisdoomen/csharpguidelines/blob/c1a45dacb2ff8b48f8b3adbd9ad2a173204e8da1/Guidelines/1500_MaintainabilityGuidelines.md#only-use-var-when-the-type-is-very-obvious-av1520-)

Src/Core/Collections/GenericDictionaryAssertions.cs
```diff
@@ -551,6 +551,86 @@ public GenericDictionaryAssertions(IDictionary<TKey, TValue> dictionary)
...
+            var missingKeys = expectedKeys.Where(key => !Subject.ContainsKey(key));
+
+            if (missingKeys.Any())
+            {
+                if (expectedKeyValuePairs.Count() > 1)
```
dennisdoomen: If you want, you can do some pretty advanced chaining to keep your code smaller. See https://github.com/dennisdoomen/fluentassertions/blob/master/Src/Core/Collections/CollectionAssertions.cs#L1169

dennisdoomen:
Thanks for this PR. I added some review comments, of which only the missing tests are blocking the PR. Would you be willing to add those?

jbattermann:
Sure thing Dennis - I'll add the tests and let you know when I'm done.

jbattermann:
Dennis, one related question here: I noticed before that one or two ArgumentException messages in GenericDictionaryAssertions.cs are a bit confusing to read because they reference a 'dictionary' to be empty, but the parameter that causes the exception is not a dictionary, see for example https://github.com/dennisdoomen/fluentassertions/blob/master/Src/Core/Collections/GenericDictionaryAssertions.cs#L326 or https://github.com/dennisdoomen/fluentassertions/blob/master/Src/Core/Collections/GenericDictionaryAssertions.cs#L463 so when using those assertion methods, it is a bit misleading - not the dictionary is (necessarily) empty, but the ienumerable set is.

Should I just go ahead and fix that naming & tests (if available), too.. or keep that separate in another PR because it logically has nothing to do with those two new methods I added?

dennisdoomen:
I don't understand. Both of your references are indeed mentioning an empty dictionary and checking for one.

jbattermann:
The check(s) at line 324 & 461 do not check for an empty dictionary but whether the IEnumerable<TKey> parameter is empty. That parameter is not a dictionary but a set / collection / list of TKey instances - that's what makes the ArgumentException misleading stating "an empty dictionary".

I ran into this exact exception with my own tests and was confused at first because my dictionary was not empty, but the List<TKey>.. the parameter was, yet the exception said 'dictionary'.

jbattermann:
Alright - added bunch of test covering all previously untested if/else branches.

(oh and - happy new year! :fireworks:)

dennisdoomen:
:+1:

PS. Happy New Year to you too. But it's 22:46 over here. Still 2015 ;-)

dennisdoomen:
Replaced by #342

jbattermann:
:+1:

Issue #354, opened 2016-02-11
Semantic equivalence for XmlNode


AndersAbel:
I'm working with a lot of Xml generation in a library that uses the classical XmlDocument/XmlElement API rather than XDocument/XElement, so I'm thinking about implementing some comparison helpers.

I'll add a new XmlNodeAssertions class containing BeEquivavelentTo(XmlNode) and BeEquivalentTo(XmlNode, string, params object[]).

The comparison will make a semantical comparison that:
- Checks that fully qualified names of elements matches (regardless of what namespace prefixes are used)
- Ignores any namespace declaration attributes.
- Checks that attribute names and values match, regardless of which order they appear in.
- Ignores any whitespace between elements tags.
- Checks text contents of elements.

This is somewhat more lax than the canonical XML standard that requires the namespace prefixes to be the same, but to me two XML docs are equivalent even if they use different ways to express the same namespace.

dennisdoomen:
That would be awesome. Issue #3 has been discussing this a long time ago, but I never found the time to implement it correctly.

AndersAbel:
I'll see if I can make it use an XmlReader internally. That way the logic can be reused to solve #3 too as there is an XNode.CreateReader() method.

dennisdoomen:
:+1:

If you would honor the [contribution guidelines](https://github.com/dennisdoomen/FluentAssertions/blob/master/CONTRIBUTING.md), it would speed up the merge process.

udlose:
is this issue still open? i have code that i extended based on the work of Eric White (who worked in the xml group on .NET)  - https://blogs.msdn.microsoft.com/ericwhite/2009/01/27/equality-semantics-of-linq-to-xml-trees/

All of my work has unit tests

AndersAbel:
@udlose The basic functionality is in place, I've apparently had [three PRs merged](https://github.com/fluentassertions/fluentassertions/pulls?q=is%3Apr+author%3AAndersAbel+is%3Aclosed) for Xml stuff. But there are still more scenarios that could be added. I don't know if @dennisdoomen has kept this issue open because he wants further functionality added or if the issue simply just didn't get closed.

dennisdoomen:
@AndersAbel did we use the algorithm from that article?

AndersAbel:
@udlose @dennisdoomen No, I wrote my own algorithm. I read the article and the biggest difference is that the code in the article is based on the assumption that there is an xml schema available. My code is based only on the XML contents.

[...]

PR #356, opened 2016-02-14
Xml node assertions

AndersAbel:
XmlDocument/XmlElement assertions as discussed in #354. 

Only available in .NET4.0 and .NET4.5 projects, as those are the only that appears to have XmlDocument support. The real comparison work is done in the XmlReaderValidator which I think only references stuff available on all platforms, so it should be possible to move to a shared library and reuse for XDocument/XElement comparisons (they can create an XmlReader too).

The comparison is done for each xml node in document order. A basic path to the current element is included in the error message to make it easier to track down where the difference is.

Example of assertion messages produced:
> Expected local name of element at "/xml/child" to be "expected" because we want to test the failure message, but found "subject".

> Expected local name of element at "/" to be "expected" because we want to test the failure message, but found "subject".

>Expected namespace of element "data" at "/xml/child" to be "urn:b" because we want to test the failure message, but found "urn:a".

Please comment on the approach before I've got too far. I'll continue working on the set up and get test coverage before asking for a merge.

### DIFF
--- a/Src/FluentAssertions.Net40/Xml/XmlNodeAssertions.cs
+++ b/Src/FluentAssertions.Net40/Xml/XmlNodeAssertions.cs
@@ -1,0 +1,95 @@
+﻿using FluentAssertions.Execution;
+using FluentAssertions.Primitives;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
+using System.Xml;
+
+namespace FluentAssertions.Xml
+{
+    /// <summary>
+    /// Contains a number of methods to assert that an <see cref="XmlNode"/> is in the expected state.
+    /// </summary>
+    [DebuggerNonUserCode]
+    public class XmlNodeAssertions : ReferenceTypeAssertions<XmlNode, XmlNodeAssertions>
+    {
+        public XmlNodeAssertions(XmlNode xmlNode)
+        {
+            Subject = xmlNode;
+        }
+
+        /// <summary>
+        /// Asserts that the current <see cref="XmlNode"/> is equivalent to the <paramref name="expected"/> element.
+        /// </summary>
+        /// <param name="expected">The expected element</param>
+        public AndConstraint<XmlNodeAssertions> BeEquivalentTo(XmlNode expected)
+        {
+            return BeEquivalentTo(expected, string.Empty);
+        }
+
+        /// <summary>
+        /// Asserts that the current <see cref="XmlNode"/> is equivalent to the <paramref name="expected"/> node.
+        /// </summary>
+        /// <param name="expected">The expected node</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="reasonArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<XmlNodeAssertions> BeEquivalentTo(XmlNode expected, string because, params object[] reasonArgs)
+        {
+            using (XmlNodeReader subjectReader = new XmlNodeReader(Subject))
+            using (XmlNodeReader expectedReader = new XmlNodeReader(expected))
+            {
+                new XmlReaderValidator(subjectReader, expectedReader, because, reasonArgs).Validate(true);
+            }
+
+            return new AndConstraint<XmlNodeAssertions>(this);
+        }
+
+        /// <summary>
+        /// Asserts that the current <see cref="XmlNode"/> is not equivalent to
+        /// the <paramref name="unexpected"/> node.
+        /// </summary>
+        /// <param name="unexpected">The unexpected node</param>
+        public AndConstraint<XmlNodeAssertions> NotBeEquivalentTo(XmlNode unexpected)
+        {
+            return NotBeEquivalentTo(unexpected, string.Empty);
+        }
+
+        /// <summary>
+        /// Asserts that the current <see cref="XmlNode"/> is not equivalent to
+        /// the <paramref name="unexpected"/> node.
+        /// </summary>
+        /// <param name="unexpected">The unexpected node</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="reasonArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        /// <returns></returns>
+        public AndConstraint<XmlNodeAssertions> NotBeEquivalentTo(XmlNode unexpected, string because, params object[] reasonArgs)
+        {
+            using (XmlNodeReader subjectReader = new XmlNodeReader(Subject))
+            using (XmlNodeReader unexpectedReader = new XmlNodeReader(unexpected))
+            {
+                new XmlReaderValidator(subjectReader, unexpectedReader, because, reasonArgs).Validate(false);
+            }
+
+            return new AndConstraint<XmlNodeAssertions>(this);
+        }
+
+        /// <summary>
+        /// Returns the type of the subject the assertion applies on.
+        /// </summary>
+        protected override string Context
+        {
+            get { return "Xml Node"; }
+        }
+    }
+}
--- a/Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
+++ b/Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
@@ -1,0 +1,239 @@
+﻿using FluentAssertions.Execution;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Xml;
+
+namespace FluentAssertions.Xml
+{
+    internal class XmlReaderValidator
+    {
+        AssertionScope assertion;
+        XmlReader subjectReader, otherReader;
+
+        string CurrentLocation
+        {
+            get { return "/" + string.Join("/", locationStack.Reverse()); }
+        }
+
+        Stack<string> locationStack = new Stack<string>();
+
+        public XmlReaderValidator(XmlReader subjectReader, XmlReader otherReader, string because, object[] reasonArgs)
+        {
+            assertion = Execute.Assertion.BecauseOf(because, reasonArgs);
+        
+            this.subjectReader = subjectReader;
+            this.otherReader = otherReader;
+        }
+                
+        class ValidationResult
+        {
+            public ValidationResult(string formatString, params object[] formatParams)
+            {
+                FormatString = formatString;
+                FormatParams = formatParams;
+            }
+
+            public string FormatString { get; }
+
+            public object[] FormatParams { get; }
+        }
+
+        public void Validate(bool expectedEquivalence)
+        {
+            ValidationResult validationResult = Validate();
+
+            if(expectedEquivalence && validationResult != null)
+            {
+                assertion.FailWith(validationResult.FormatString, validationResult.FormatParams);
+            }
+            if(!expectedEquivalence && validationResult == null)
+            {
+                assertion.FailWith("Expected Xml to not be equivalent{reason}, but it is.");
+            }
+        }
+
+        private ValidationResult Validate()
+        {
+            subjectReader.MoveToContent();
+            otherReader.MoveToContent();
+            while (!subjectReader.EOF && !otherReader.EOF)
+            {
+                if (subjectReader.NodeType != otherReader.NodeType)
+                {
+                    return new ValidationResult("Expected node of type {0} at {1}{reason}, but found {2}.",
+                        otherReader.NodeType, CurrentLocation, subjectReader.NodeType);
+                }
+
+                ValidationResult validationResult = null;
+
+                switch (subjectReader.NodeType)
+                {
+                    case XmlNodeType.Element:
+                        validationResult = ValidateStartElement();
+                        if(validationResult != null)
+                        {
+                            return validationResult;
+                        }
+                        locationStack.Push(subjectReader.LocalName);
+                        validationResult = ValidateAttributes();
+                        break;
+                    case XmlNodeType.EndElement:
+                        // No need to verify end element, if it doesn't match
+                        // the start element it isn't valid XML, so the parser
+                        // would handle that.
+                        locationStack.Pop();
+                        break;
+                    case XmlNodeType.Text:
+                        validationResult = ValidateText();
+                        break;
+                    default:
+                        throw new NotSupportedException($"{subjectReader.NodeType} found at {CurrentLocation} is not supported for equivalency comparison.");
+                }
+
+                if(validationResult != null)
+                {
+                    return validationResult;
+                }
+
+                subjectReader.Read();
+                otherReader.Read();
+
+                subjectReader.MoveToContent();
+                otherReader.MoveToContent();
+            }
+
+            if (!otherReader.EOF)
+            {
+                return new ValidationResult("Expected {0}{reason}, but found end of document.",
+                    otherReader.LocalName);
+            }
+
+            if (!subjectReader.EOF)
+            {
+                return new ValidationResult("Expected end of document{reason}, but found {0}.",
+                    subjectReader.LocalName);
+            }
+
+            return null;
+        }
+
+
+        class AttributeData
+        {
+            public AttributeData(string namespaceUri, string localName, string value, string prefix)
+            {
+                NamespaceUri = namespaceUri;
+                LocalName = localName;
+                Value = value;
+                Prefix = prefix;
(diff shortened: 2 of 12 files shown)


AndersAbel:
Now I've completed a first version, with test coverage for all stuff that's implemented. What's the bare minimum you will accept for merging?

Currently supports:
* Be/NotBe and BeEquivalentTo/NotBeEquivalentTo.
* Elements, Attributes and Text nodes.
* Namespaces - prefixes are not compared, but equivalence is checked on Namespace-URI+LocalName of node.

Things that might be needed:
* More element types (CData etc...)
* More methods (HaveAttribute, etc)

After this is merged, updating the existing XElement/XDocument tests to use the same validation engine is just a few lines of code, but I'd prefer to get the XmlDocument functionality in place first.

dennisdoomen:
Looks quite promising. Will do a full code-review shortly. Couple of questions.
1. What about support for `XDocument`? I thought you wanted to support that as well?
2. When two XML document are not fully equivalent, does it properly report the differences?
3. Would it be possible to extend the structural equvalency API to support `XDocument` and `XmlDocument` as well? Technically it would require a simple implementation of `IEquivalencyStep`. I'm trying to get some consistency in the code-base, so it would be unexpected to have a separate API for XML document that is not included in a structural equivalency check.

Src/FluentAssertions.Net40/Xml/XmlNodeAssertions.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="reasonArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        /// <returns></returns>
+        public AndConstraint<XmlNodeAssertions> NotBeEquivalentTo(XmlNode unexpected, string because, params object[] reasonArgs)
+        {
+            new XmlReaderValidator(new XmlNodeReader(Subject), new XmlNodeReader(unexpected), because, reasonArgs).Validate(false);
```
dennisdoomen: Don't you have to dispose those readers at some point?

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
```diff
@@ -0,0 +1,239 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Xml;
+
+namespace FluentAssertions.Xml
+{
+    class XmlReaderValidator
```
dennisdoomen: I prefer to explicitly scope all classes.

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
```diff
@@ -0,0 +1,239 @@
...
+            }
+
+            public string FormatString { get; }
+
+            public object[] FormatParams { get; }
+        }
+
+        public void Validate(bool expectedEquivalence)
+        {
+            var validationResult = Validate();
```
dennisdoomen: Please don't use `var`, unless the type is explicitly specified.

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
```diff
@@ -0,0 +1,239 @@
...
+                        locationStack.Pop();
+                        break;
+                    case XmlNodeType.Text:
+                        validationResult = ValidateText();
+                        break;
+                    default:
+                        throw new NotImplementedException();
+                }
+
+                if(validationResult != null)
```
dennisdoomen: So it will bail out with the first mismatch that it finds?

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
```diff
@@ -0,0 +1,239 @@
...
+                        // No need to verify end element, if it doesn't match
+                        // the start element it isn't valid XML, so the parser
+                        // would handle that.
+                        locationStack.Pop();
+                        break;
+                    case XmlNodeType.Text:
+                        validationResult = ValidateText();
+                        break;
+                    default:
+                        throw new NotImplementedException();
```
dennisdoomen: What other elements would it not support? Trying to understand if the current subset is useful enough.

Next to that, it should at least tell me *what* it doesn't support.

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
```diff
@@ -0,0 +1,239 @@
...
+
+        public XmlReaderValidator(XmlReader subjectReader, XmlReader expectedReader, string because, object[] reasonArgs)
+        {
+            assertion = Execute.Assertion.BecauseOf(because, reasonArgs);
+        
+            this.subjectReader = subjectReader;
+            this.expectedReader = expectedReader;
+        }
+                
+        class ValidationResult
```
dennisdoomen: Instead of collecting these validation errors yourself, you can also use an `AssertionScope` like [this](https://github.com/dennisdoomen/FluentAssertions/blob/master/Tests/FluentAssertions.Shared.Specs/AssertionScopeSpecs.cs#L54) to have `Execute.Assertion` add all assertion failures together.

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
```diff
@@ -0,0 +1,239 @@
...
+            {
+                return new ValidationResult("Expected end of document{reason}, but found {0}.",
+                    subjectReader.LocalName);
+            }
+
+            return null;
+        }
+
+
+        class AttributeData
```
dennisdoomen: I prefer to keep every class in a single file.

Tests/FluentAssertions.Net40.Specs/XmlNodeAssertionSpecs.cs
```diff
@@ -0,0 +1,489 @@
...
+            // Act
+            //-------------------------------------------------------------------------------------------------------------------
+            Action act = () =>
+                doc.Should().BeSameAs(otherNode, "we want to test the failure {0}", "message");
+
+            //-------------------------------------------------------------------------------------------------------------------
+            // Assert
+            //-------------------------------------------------------------------------------------------------------------------
+            act.ShouldThrow<AssertFailedException>()
+                .WithMessage("Expected object to refer to <otherDoc /> because we want to test the failure message, but found object <doc />.");
```
dennisdoomen: Would be nicer to see `document` or `xml fragment`

AndersAbel:
> 1 .What about support for XDocument? I thought you wanted to support that as well?

Yes, but I'll do that as a separate PR after this is merged. It's just a few lines of code that need to be changed. But there are a lot of XPlat issues that need to be looked into as XDocument is supported on more platforms.

> 2. When two XML document are not fully equivalent, does it properly report the differences?

Yes, see example messages in the description of the PR.

>3. Would it be possible to extend the structural equvalency API to support XDocument and XmlDocument as well? Technically it would require a simple implementation of IEquivalencyStep. I'm trying to get some consistency in the code-base, so it would be unexpected to have a separate API for XML document that is not included in a structural equivalency check.

Don't know, will have to look into that.

Src/FluentAssertions.Net40/Xml/XmlNodeAssertions.cs
AndersAbel: Not really, they are just in-memory and not backed by any unmanaged resource, but I'll add a using to it.

dennisdoomen:
Fair enough.

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
AndersAbel: Other node types are those derived from XmlNode. Things that I think might occur in a document are comments (can just be ignored) and CData. Less likely are entities and notations that I think are DTD related. Comments should just be ignored and CDAta would be good to add support for.

And of course, the NotImplementedException should include the node type.

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
AndersAbel: You mean adding `internal`?

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
AndersAbel: Changed all `var` to explicit types.

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
AndersAbel: Is there a simple way to limit the number of reported errors if using nested scopes? To make multiple error reports useful I'm also afraid some more complex logic will be needed to get things in sync when elements are found in one of the readers that are not present in the other. Think of `<xml><a/><b/></c></xml>` vs `<xml><b/></c></xml>`. Doing good reports on that would require detection that a is missing, but b and c are correct.

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
AndersAbel: This class is only used internally in `XmlReaderValidator` and never visible to the outside. I put it here to make it private to the class. Do you prefer it in an separate file as `internal`?

Tests/FluentAssertions.Net40.Specs/XmlNodeAssertionSpecs.cs
AndersAbel: I don't get what you mean. Can you please elaborate?

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
dennisdoomen: Whatever is suitable, but I always specify the scope.

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
dennisdoomen: It's not build in, but you could extend `AssertionScope` to accept a maximum number of failures before it throws.

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
dennisdoomen: No, just ignore my comment.

Tests/FluentAssertions.Net40.Specs/XmlNodeAssertionSpecs.cs
dennisdoomen: Sorry. I meant the reference to `object` in the failure message.

dennisdoomen:
BTW, you can just amend your existing commit to keep the PR to a single commit. Or you can squash just before I merge.

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
AndersAbel: Would it be okay to start with the current bail-out-at-first-error approach for now?

Tests/FluentAssertions.Net40.Specs/XmlNodeAssertionSpecs.cs
AndersAbel: I'd love to change it, but I can't figure out how to use the context to get the right thing there. What should I do to make the test in `ReferenceTypeAssertion` expand `{context:object}` into something I specify?

dennisdoomen: Just change it to `{context:xml document}`. The context is used in nested assertions such as the `EquivalencyValidator` does.

Src/FluentAssertions.Net40/Xml/XmlReaderValidator.cs
dennisdoomen: Sure, but it's one of the USPs of FA to really help the developer with detailed failures.

AndersAbel: Yes, that's why I use FA in the first place :-)

If I find the time, I'd love to write a better error aggregation, but I think that already as this is, it's a vast improvement for Xml work over the existing features.

dennisdoomen: Agreed

Tests/FluentAssertions.Net40.Specs/XmlNodeAssertionSpecs.cs
AndersAbel: Now it's relying on the implementation in the `ReferenceTypeAssertions` base class, that contains the string constant with `{context:object}`. Would I need to provide my own implementation in `XmlNodeAssertions` to change the text? I assumed there would be a way to somehow inject the name, but I might have made an incorrect assumption.

dennisdoomen: I don't get it. What is it in `ReferenceTypeAssertions` you're relying on?

AndersAbel: There is no implemenation of `BeSameAs` in `XmlNodeAssertions`. This test calls the `ReferenceTypeAssertions.BeSameAs()` method. So maybe this test shouldn't even exist in the first place ;-)

Or should I remove the inheritance of `ReferenceTypeAsseritions<>` completely and re-implement `BeSameAs` in `XmlNodeAssertins`?

dennisdoomen: Well, being able to compare references is useful, so I wouldn't remove the base-class here and override the `Context property instead.

AndersAbel: Now I finally think I have found the reason:

Looking at `ReferenceTypeAssertion<>.BeNull()` the failure message is formatted with:

````
.FailWith("Expected {context:" + Context + "} to be <null>{reason}, but found {0}.", Subject);
````

It uses the `Context` property. But looking at ReferenceTypeAssertion<>.BeSameAs()` the failure message is formatted with:

````
.FailWith("Expected {context:object} to refer to {0}{reason}, but found object {1}.", expected, Subject);
````

Shouldn't that be changed to pick up the name from `Context` instead of hard coding it to "object"?

````
.FailWith("Expected {context:" + Context + "} to refer to {0}{reason}, but found object {1}.", expected, Subject);
````

If that change is done, I've already overloaded `Context` in the `XmlNodeAssertions` class, so it will be presented as "XmlNode".

dennisdoomen: Ah, yes, that's a bug ;-)

AndersAbel: Then I understand why I didn't get it working. I can fix it in this PR, or do you rather want a separate PR for the bug fix first?

dennisdoomen: No, not at all. Just fix it.

AndersAbel:
Now I think all issues are resolved. I've rebased and squashed it to one commit.

dennisdoomen:
Great work!

So if I merge this, what are you short-term plans for completing the functionality? E.g.
* Adding `XElement`/`XDocument` support
* FA already has an extensive set of XML assertions on document, elements and attributes. For consistency reasons, I think we need to have a similar set for `XmlDocument` and `XmlElement`
* The `ShouldBeEquivalent` API also needs support for comparing XML documents.

AndersAbel:
* I'll get back with a PR to replace usage of `DeepEquals` in `XElement`/`XDocument` with the `XmlReaderValidator`.
* Implementations of `HaveAttribute()` and other methods available on `XElement`/`XDocument` would be nice - I'll see if I can find the time.
* Implementing `IEquivalencyStep` looks quite straight forward. I'll se what I can do. Is there an implementation for `XElement`/`XDocument` or should that be implemented too?

dennisdoomen:
> Implementing  IEquivalencyStep  looks quite straight forward. I'll se what I can do. Is there an implementation for  XElement / XDocument  or should that be implemented too?

No, simply because FA didn't have support for comparing XML documents in an equivalent way. Now you've introduced that, it would be a nice addition.

dennisdoomen:
So it's ready to be merged?

AndersAbel:
Yes, it is. Unless you've found any more issues in the code ;-)

AndersAbel:
Great! When is the party?

dennisdoomen:
Probably this weekend. I need to finish fixing one bug. After that, each commit should result in a new release.

dennisdoomen:
Now that 4.3.0 is released, I would really like to make the API consistent as soon as possible:
* Adding XElement/XDocument support
* FA already has an extensive set of XML assertions on document, elements and attributes. For consistency reasons, I think we need to have a similar set for XmlDocument and XmlElement

Any change you can play a role here?

AndersAbel:
Yes, it's on my todo list, but I need to find some available time to do it.

PR #370, opened 2016-03-07
Add SelfReferencingCollectionAssertions.StartWith and SelfReferencingCollectionAssertions.EndWith that support sequence of elements

AlexanderSher:
Also added unit tests and extracted some shared code into extension methods


Src/Core/Collections/CollectionAssertions.cs
```diff
@@ -209,65 +207,33 @@ public AndConstraint<TAssertions> Equal(IEnumerable expected, string because = "
         protected void AssertSubjectEquality<TActual, TExpected>(IEnumerable expectation, Func<TActual, TExpected, bool> predicate,
             string because = "", params object[] reasonArgs)
         {
-            AssertionScope assertion = Execute.Assertion.BecauseOf(because, reasonArgs);
-
-            bool subjectIsNull = ReferenceEquals(Subject, null);
-            bool expectationIsNull = ReferenceEquals(expectation, null);
 
+            var subjectIsNull = ReferenceEquals(Subject, null);
```
dennisdoomen: Please avoid using `var`, unless the type is explicitly mentioned. Same for all other changes in this PR.

dennisdoomen:
Thanks @AlexanderSher. I'll will review your PR shortly.

Src/Core/Collections/SelfReferencingCollectionAssertions.cs
```diff
@@ -130,6 +130,64 @@ public AndConstraint<TAssertions> Equal(params T[] elements)
...
+        /// A predicate the is used to determine whether two objects should be treated as equal.
+        /// </param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="reasonArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> StartWith<TExpected>(
```
dennisdoomen: Wouldn't it make sense to have an overload that doesn't require that predicate?

Src/Core/Common/EnumerableExtensions.cs
```diff
@@ -0,0 +1,48 @@
+using System;
+using System.Collections.Generic;
+
+namespace FluentAssertions.Common
+{
+    public static class EnumerableExtensions
```
dennisdoomen: Can't this be `internal`?

Src/Core/Collections/CollectionAssertions.cs
dennisdoomen: Please avoid using `var`, unless the type is explicitly stated. I'd like to have the code base meet https://github.com/dennisdoomen/csharpguidelines

Src/Core/Collections/CollectionAssertions.cs
```diff
@@ -1096,21 +1063,26 @@ public AndConstraint<TAssertions> StartWith(object element, string because = "",
...
+        {
             Execute.Assertion
-                .BecauseOf(because, becauseArgs)
-                .WithExpectation("Expected {context:collection} to end with {0}{reason}, ", element)
-                .ForCondition(!ReferenceEquals(Subject, null))
-                .FailWith("but the collection is {0}.", (object)null)
+                .BecauseOf(because, reasonArgs)
+                .WithExpectation("Expected {context:collection} to end with {0}{reason}, ", expected)
+                .Given(() => actual)
+                .AssertCollectionIsNotNullOrEmpty()
```
dennisdoomen: I like the way you extend the fluent API here although I'm not entirely sure the naming is correct since the API is there to build up an assertion, e.g. *execute an assertion because of some reason with expectation*. Unfortunately, I can't come up with anything better.

dennisdoomen:
Done reviewing. Only a suggestion and some minor remarks.

Src/Core/Collections/CollectionAssertions.cs
AlexanderSher: Sure, though "the type is very obvious from the same statement" is a bit vague. ReferenceEquals is a well-known method, though there is no real difference in maintainability between 'bool' and 'var', so it can what is easier to read. Another example - `ToArray()` produces an array, but type is unknown, though you can see the type because it is `Cast<TExpected>().ToArray()'.
Do you want me to replace all the `var` usages in the changeset?

Src/Core/Collections/SelfReferencingCollectionAssertions.cs
AlexanderSher: Sure

AlexanderSher:
Ok, I'll send an update later today.

dennisdoomen:
:+1:

Src/Core/Collections/CollectionAssertions.cs
dennisdoomen: The one with the `Cast<TExpected>().ToArray()` is an example where the type is explicitly stated. In all other cases, I prefer not using var at all.

dennisdoomen: The one with the `Cast<TExpected>().ToArray()` is an example where the type is explicitly stated. In all other cases, I prefer not using var at all.

AlexanderSher:
@dennisdoomen, can you please review the update?
Thank you!

dennisdoomen:
Awesome. Now it just needs squashing. If you don't have time, I'll have to do that tonight using a manual merge.

dennisdoomen:
Why did you delete it?

AlexanderSher:
Accidentally.

PR #375, opened 2016-03-15
360 add json support

mkoertgen:
This adds `FluentAssertions.Json` depending on `Newtonsoft.Json 6+` with support for `net45, dotnet & pcl`


Src/Json/Shared.Json/JTokenAssertions.cs
```diff
@@ -0,0 +1,234 @@
...
+                .FailWith("Expected JSON document not to be {0}{reason}.", unexpected);
+
+            return new AndConstraint<JTokenAssertions>(this);
+        }
+
+        /// <summary>
+        ///     Asserts that the current <see cref="JToken" /> is equivalent to the <paramref name="expected" /> element,
+        ///     using its <see cref="JToken.DeepEquals(JToken, JToken)" /> implementation.
+        /// </summary>
+        /// <param name="expected">The expected element</param>
```
dennisdoomen: Can you add a `<remarks>` section explaining what *equivalent* means?

Src/Json/Shared.Json/JTokenAssertions.cs
```diff
@@ -0,0 +1,234 @@
...
+        /// <param name="because">
+        ///     A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        ///     is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="reasonArgs">
+        ///     Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<JTokenAssertions> Be(JToken expected, string because, params object[] reasonArgs)
```
dennisdoomen: Since a couple of days, all `reasonArgs` parameters are now called `becauseArgs`.

Src/Json/Shared.Json/JTokenAssertions.cs
```diff
@@ -0,0 +1,234 @@
...
+        {
+            Execute.Assertion
+                .ForCondition(JToken.DeepEquals(Subject, expected))
```
dennisdoomen: Isn't there any information on what are the differences? This has been the ultimate USP of FluentAssertions.

Src/Json/Shared.Json/JTokenAssertions.cs
```diff
@@ -0,0 +1,234 @@
...
+        ///     A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        ///     is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="reasonArgs">
+        ///     Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<JTokenAssertions> BeEquivalentTo(JToken expected, string because,
+            params object[] reasonArgs)
+        {
+            var diff = ObjectDiffPatch.GenerateDiff(Subject, expected);
```
dennisdoomen: Please avoid `var` unless the type is explicitly visible.

Src/Json/Shared.Json/JTokenAssertions.cs
```diff
@@ -0,0 +1,234 @@
...
+        ///     A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        ///     is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="reasonArgs">
+        ///     Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndWhichConstraint<JTokenAssertions, JToken> HaveElement(string expected, string because,
+            params object[] reasonArgs)
+        {
+            var jToken = Subject[expected];
```
dennisdoomen: Consider rewritting as `JToken token = Subject[expected];`.

Src/Json/Shared.Json/ObjectDiffPatch.cs
```diff
@@ -0,0 +1,378 @@
+﻿#region *   License     *
+/*
+    SimpleHelpers - ObjectDiffPatch
```
dennisdoomen: JSON.NET doesn't contain anything like this OOB? Or isn't there a NuGet package for this code? I'm a bit wary that we have to maintain this now.

dennisdoomen:
Awesome work! 

Some remarks:
* FA currently directly supports .NET 4.0. You could switch your PCL to support 4.5 and use the explicit project to target 4.0 (not 4.0.3)
* It would make me very happy if you could use the same focused style for the unit tests as I do, e.g. the explicit arrange/act/assert sections and a single assertion per test.
* The final PR still can be squashed to a single commit

mkoertgen:
https://github.com/dennisdoomen/FluentAssertions/pull/375#issuecomment-197168711

- net4: I copied settings from existing project (`Core.Pcl`). Moving net45 into the pcl project would allow us to squash 3 projects down to 2 projects, right? Great!

- AAA-style & single commit: Ok.

Src/Json/Shared.Json/JTokenAssertions.cs
mkoertgen: That makes sense and is a (continuous) improvement ;-)

Src/Json/Shared.Json/JTokenAssertions.cs
mkoertgen: Sure

Src/Json/Shared.Json/JTokenAssertions.cs
mkoertgen: Hm, right. The DiffHelper is added only to `BeEquivalentTo` and works much like the `DictionaryAssertions`, i.e. showing the first differing key.

Maybe there is no distinct use for `Be`. Should we just remove it then?

Src/Json/Shared.Json/ObjectDiffPatch.cs
mkoertgen: Sadly, json.net has only `JToken.DeepEquals`. The NuGet package is https://www.nuget.org/packages/SimpleHelpers.ObjectDiffPatch/. I can factor this out if you want. However, it's a MiniMod, i.e. will be under FA source control.

Src/Json/Shared.Json/JTokenAssertions.cs
dennisdoomen: Good point. What *is* the difference between `Be` and `BeEquivalentto`?

Src/Json/Shared.Json/JTokenAssertions.cs
dennisdoomen: Yeah, we should, although that will break all extensions people have been building on top of FA. So let's keep it as it is.

Src/Json/Shared.Json/JTokenAssertions.cs
dennisdoomen: Now that I think of it, maybe you should rename `Be` to `Equal`. That's the name I've used for collection and dictionary assertions as well. So `Equal` means they contain the same items in the same order. `Equivalent` means that they contain the same items in any order. The only exception is that there's no `Equivalent` for dictionaries because dictionaries don't define any order.

Src/Json/Shared.Json/ObjectDiffPatch.cs
dennisdoomen: Well, that NuGet package is a source-only package, so it makes sense to directly use the package. However, that code got me thinking. Wouldn't it be more useful to use the structural equivalency API (`ShouldBeEquivalentTo`) to compare the properties and fields of two `JToken` objects. I suppose those tokens contain hierarchical dictionaries that FA can inspect and compare. Thoughts?

dennisdoomen:
>  I copied settings from existing project (Core.Pcl). Moving net45 into the pcl project would allow us to squash 3 projects down to 2 projects, right? Great!

No, we still need to support .NET 4.0.

Src/Json/Shared.Json/JTokenAssertions.cs
dennisdoomen: Please!

Src/Json/Shared.Json/ObjectDiffPatch.cs
mkoertgen: You mean this https://github.com/dennisdoomen/fluentassertions/wiki#object-graph-comparison ? Yes, that would be the best place to fit in. 

And you're right: it boils down to comparing property graphs just as dictionaries. So, there is a good chance we can use some of FA's built-in stuff and leave out the package. I will look into this.

mkoertgen:
> No, we still need to support .NET 4.0.

Ok then. I will add a net4 project.

dennisdoomen:
> Ok then. I will add a net4 project.

I think you don't need a separate .NET 4.5 project. You're PCL can target that for you.

mkoertgen:
> I think you don't need a separate .NET 4.5 project. You're PCL can target that for you.
Probably yes. I cross-checked how `AutoFac` and `Newtonsoft.Json` handle this. 
`Autofac` provides
- `net40`
- `portable-net4+sl5+netcore45+wpa81+wp8+MonoAndroid1+MonoTouch1`
When consuming from a `net45`-project the `net40` gets referenced (hmm).

Newtonsoft.Json provides
- `net20, net30, net35, net40, net45`
- `portable-net40+sl5+wp80+win8+wpa81`
- `portable-net45+wp80+win8+wpa81+dnxcore50`
When consuming from a `net45`-project the `net45` gets referenced (ok).

For now let's keep it.

Src/Json/Shared.Json/ObjectDiffPatch.cs
mkoertgen: I did a quick test, plainly using

            var subject = JToken.Parse("{child:{subject:'foo'}}");
            var expected = JToken.Parse("{child:{expected:'bar'}}");
            subject.ShouldBeEquivalentTo(expected);

which unexpectedly did not fail. 

The equivalency steps handling the JTokens were

`CanHandle == true`:

- `TryConversionEquivalencyStep`
- `ReferenceEqualityEquivalencyStep`
- `GenericEnumerableEquivalencyStep`
- `EnumerableEquivalencyStep`
- `StructuralEqualityEquivalencyStep`

`Handle == true`:

- `ValueTypeEquivalencyStep` (not sure where this one came from)
- `StructuralEqualityEquivalencyStep`

`Any/FirstOrDefault( CanHandle && Handle)`:

- GenericEnumerableEquivalencyStep

Right now i want to stop digging on the graph equivalency API as it seems too complicated for me, especially the expression based `GenericEnumerableEquivalencyStep` and the coupling between steps, validator & context.

mkoertgen: So finally i don't see a quick & clean way to use the structural equivalency API for `JToken` - which may be due to my lack of understanding ;-)

My current suggestion is to reduce the risk of maintenance by cutting down the `ObjectDiffPatch` code down to the bare minimum needed for FA.

dennisdoomen: I agree. Thanks for trying anyway.

mkoertgen:
Ok then. If you are fine with the code i will clean up & squash the commits now.

mkoertgen:
I realized that the **ObjectDiffPatch**-code caused a serious drop in test coverage. My tests only covered about 65% of it. Whereas the other json-assertions-code is 100% covered.

I removed the uneeded **Patch**-code (we only need the **Diff** part) hoping to raise coverage. But there remained some untested code paths regarding edge cases like array handling etc. which were cumbersome to test.

So i decided to add the tests from the original repo instead, taking the (unused) **Patch** code back in. No we have 95.48% coverage on **ObjectDiffPatch** and 96.56% total coverage on the Json code.

Why not use the original source package? Well, the source needs modification anyway because the source class is **public** but we don't want it to leak out into the public API.

If this is Ok for you, i will now squash commits

mkoertgen:
What about documentation?

mkoertgen:
>> Ok then. I will add a net4 project.

> I think you don't need a separate .NET 4.5 project. You're PCL can target that for you.

I did a small test project to see how the NuGet package would be consumed. It targets net45 but when adding the dependency NuGet will add the net40-version. Did i miss something? If this is the way it is, i will rather add back the net45-project back.

dennisdoomen:
> So i decided to add the tests from the original repo instead, taking the (unused) Patch code back in

Not entirely happy with those tests, but we can always refactor that later on. 

> What about documentation?

If you can, please add them to this PR. I'll copy the documentation when this is released.

> I did a small test project to see how the NuGet package would be consumed. It targets net45 but when adding the dependency NuGet will add the net40-version. I will rather add back the net45-project.

Hmm, that's weird indeed.

mkoertgen:
I will do some more testing to verify consuming the package results i a good first experience. 

The API should span a broad range of versions, i.e. Newtonsoft.Json [6,8). Turns out i forgot to set `SpecificVersion=false` which may result in `CS1705`.

mkoertgen:
Sidenote: Just tested a portable lib consuming the package and containing tests.

As documented & expected, NCrunch & ReSharper do not support running NUnit tests within PCL. 

But XUnit works flawlessly. Cool!

mkoertgen:
>> What about documentation?

> If you can, please add them to this PR. I'll copy the documentation when this is released.

Documentation already added within Xml-Docs. Any markdown/wiki i could add?

dennisdoomen:
> Documentation already added within Xml-Docs. Any markdown/wiki i could add?

I meant something I can paste into the wiki.

dennisdoomen:
Thanks. I'll merge this one manually so that I can rebase it on *master*.

mkoertgen:
Here is some documentation for the wiki

    ## JSON classes ##

    The package `FluentAssertions.Json` supports assertions on the [LINQ-to-JSON](http://www.newtonsoft.com/json/help/html/LINQtoJSON.htm) classes of `Newtonsoft.Json`, i.e. everything that inherits [JToken](http://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_Linq_JToken.htm).

    Here are some examples:

    ```csharp
    var jToken = JToken.Parse("{'id':42}");
    jToken.Should().HaveElement("id");
    jToken["id"].Should().HaveValue("42");
    ```

    You can also perform a deep comparison between two elements like this:

    ```csharp
    jObject.Should().BeEquivalentTo(JObject.Parse("{'configuration':{'item':'value'}}"));
    ```

mkoertgen:
I wanted to help you and already rebased on current master. Please review

dennisdoomen:
Continued through #391.

PR #385, opened 2016-04-03
ShouldAllBeEquivalentTo a single object enhancement

SaroTasciyan:
###### Description:
Asserting _all elements in given collection should be equivalent to given object_. 
Implemented method basically provides a short way for using a single object repeated collection and asserting using known `ShouldAllBeEquivalentTo` method.

###### Changes:
- `ShouldAllBeEquivalentTo` new overload accepting single object and asserting all elements of given collection is equivalent to given object is implemented
- Testing for implementation
- Included new general tests for `IEnumerable` overload in order to see if both work as expected

Please note that no new `AssertFailedException` messages are included. All exception messages are expected to be the same as they were.

###### Wiki:
Following sample can be added under [Collections and Dictionaries](https://github.com/dennisdoomen/FluentAssertions/wiki#collections-and-dictionaries) heading.

```C#
orderDto.ShouldAllBeEquivalentTo(order, options => options.Including(o => o.Customer.Name));
```

Alternatively, more detailed sample can be:

```C#
var someObjects = new[]
{
    new { Id = 1, Name = "SomeName", Target = "Target" },
    new { Id = 2, Name = "SomeName", Target = "Target" }
};

someObjects.ShouldAllBeEquivalentTo(new { Target = "Target" }, options => options.Including(x => x.Target));
```


dennisdoomen:
Thanks for taking the time to submit this pull request. I'll evaluate it as soon as possible.

Src/Shared/AssertionExtensions.cs
```diff
@@ -483,6 +483,84 @@ public static PropertyInfoSelectorAssertions Should(this PropertyInfoSelector pr
         }
 
         /// <summary>
+        /// Asserts that all elements in a collection of objects is equivalent to given object.
```
dennisdoomen: Typo. It's "*are* equivalent to a given object"

Tests/FluentAssertions.Shared.Specs/CollectionEquivalencySpecs.cs
```diff
@@ -17,6 +17,128 @@ namespace FluentAssertions.Specs
     [TestClass]
     public class CollectionEquivalencySpecs
     {
+        #region General
```
dennisdoomen: Probably better to rename this region to "Single Object" or something like that.

Tests/FluentAssertions.Shared.Specs/CollectionEquivalencySpecs.cs
```diff
@@ -17,6 +17,128 @@ namespace FluentAssertions.Specs
...
+
+        [TestMethod]
+        public void When_ShouldAllBeEquivalentTo_subject_is_null_it_should_throw()
```
dennisdoomen: All these specs use the same method, so please keep their name functional

Tests/FluentAssertions.Shared.Specs/CollectionEquivalencySpecs.cs
```diff
@@ -17,6 +17,128 @@ namespace FluentAssertions.Specs
...
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => subject.ShouldAllBeEquivalentTo(Enumerable.Empty<object>());
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act.ShouldThrow<AssertFailedException>().WithMessage(
+                "Expected subject to be*, but found <null>*");
```
dennisdoomen: What does it report for the expectation?

dennisdoomen:
Looks pretty good. Thanks for that. Are you willing to do the rework I suggested?

SaroTasciyan:
Hi @dennisdoomen, my pleasure. I appreciate your feedback and support. Fair enough,  I will try to find some time to add my _line notes_ and start working in the following days.

dennisdoomen:
You're not forgetting it, are you?

SaroTasciyan:
Nope, of course not. Thanks for reminding (:
I ve been very busy last week.

Tests/FluentAssertions.Shared.Specs/CollectionEquivalencySpecs.cs
SaroTasciyan: There are six tests in this _region_. Three of them are "single object" overload tests while others are not.

Please check:
- Included new general tests for `IEnumerable` overload in order to see if both work as expected

I think those new tests (subject-null-throw, expectation-null-throw, suject-expectation-null-succeed) add value to the fixture escpecially with the new single object tests which doesn't have the same expectation for all given cases.

How should I arrange them? I can move new single object tests to a new region called "Single Object" and keep the rest (above) in "General" region.

Tests/FluentAssertions.Shared.Specs/CollectionEquivalencySpecs.cs
SaroTasciyan: This is not the "single object" overload test. The report is:

> Expected subject to be a collection, but found \<null\>.
> 
> With configuration:
> - Use declared types and members
> - Compare enums by value
> - Include all non-private properties
> - Include all non-private fields
> - Match member by name (or throw)
> - Be strict about the order of items in byte arrays

Same applies for "single object" overload as tested below:
```c#
[TestMethod]
public void When_ShouldAllBeEquivalentTo_subject_is_null_and_expectation_is_nonenumerable_it_should_throw()
{
    // Arrange
    IEnumerable<DateTime> subject = null;

    // Act
    Action act = () => subject.ShouldAllBeEquivalentTo(DateTime.Now);

    // Assert
    act.ShouldThrow<AssertFailedException>().WithMessage(
        "Expected subject to be*, but found <null>*");
}
```

Expectation report for a failed "single object" `ShouldAllBeEquivalentTo` will contain repeat of single object that was passed in the API. Please check the following example:
```c#
var subject = new int[] { 1, 2, 3 };
subject.ShouldAllBeEquivalentTo(1);
```

> Expected item[1] to be 1, but found 2.
> Expected item[2] to be 1, but found 3.
> 
> With configuration:
> ...

I would appreciate your feedback.

Tests/FluentAssertions.Shared.Specs/CollectionEquivalencySpecs.cs
SaroTasciyan: > 11. Don't refer to technical names such as the names of methods, variables and UI elements in the name of a test. In my opinion, the name of a test should explain the functional scenario, whereas the cause and effect should clearly illustrate the technicalities of that scenario. If some class or member is renamed, it shouldn't affect the functional name of the test

I suppose the following tests in `CollectionEquivalencySpecs` mislead me:

`When_ShouldAllBeEquivalentTo_has_selection_rules_configured_they_should_be_evaluated_from_right_to_left`
`When_ShouldAllBeEquivalentTo_utilizes_custom_assertion_rules_the_rules_should_be_respected`

I can rename them in addition to the ones I added if you like

Tests/FluentAssertions.Shared.Specs/CollectionEquivalencySpecs.cs
dennisdoomen: What are you trying to say?

Tests/FluentAssertions.Shared.Specs/CollectionEquivalencySpecs.cs
dennisdoomen: Probably from a different PR :wink:

SaroTasciyan: Alright, I've them fixed aswell

Tests/FluentAssertions.Shared.Specs/CollectionEquivalencySpecs.cs
SaroTasciyan: Well, the point is not all the tests in that region tests "Single Object" overload. I splitted them into two regions called "Null (Collection)" and "Null (Single Object" if it's fine.

PR #386, opened 2016-04-04
Semantic (Not)BeEquivalent for XDocument/XElement

AndersAbel:
Also fixed a bug where the location for xml comparisons was incorrect after a selfclosed <xml /> element had been compared.

The tests for `XDocument/XElement.Should().BeEquivalent()` are not very covering. They basically just checks that semantical equivalency is used. The tests for `XmlNode` are much more extensive, as those are what I used to test the `XmlReaderValidator` implementation. How do you want the tests? I can think of three options.

1. Kept as it is
2. Also do full tests/specs for `XDocument/XElement` (will be a lot of copied tests)
3. Move the detailed tests to a separate class that tests `XmlReaderValidator` directly

### DIFF
--- a/Tests/FluentAssertions.Shared.Specs/XDocumentAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Shared.Specs/XDocumentAssertionSpecs.cs
@@ -297,8 +297,7 @@ namespace FluentAssertions.Specs
             // Assert
             //-------------------------------------------------------------------------------------------------------------------
             act.ShouldThrow<AssertFailedException>()
-                .WithMessage("Expected XML document <parent>...</parent> to be equivalent to <parent>...</parent>" +
-                    " because we want to test the failure message.");
+                .WithMessage("Expected node of type EndElement at \"/parent\" because we want to test the failure message, but found Element.");
         }
 
         [TestMethod]
@@ -320,8 +319,7 @@ namespace FluentAssertions.Specs
             // Assert
             //-------------------------------------------------------------------------------------------------------------------
             act.ShouldThrow<AssertFailedException>()
-                .WithMessage("Expected XML document <parent>...</parent> to be equivalent to <parent>...</parent>" +
-                    " because we want to test the failure message.");
+                .WithMessage("Expected node of type Element at \"/parent\" because we want to test the failure message, but found EndElement.");
         }
 
         [TestMethod]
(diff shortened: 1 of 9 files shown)


dennisdoomen:
Please go for option 2. We've tried to avoid that before by trying to be smart with shared code or base-classes, but it always resulted in a lot of pain.

AndersAbel:
Updated with tests, copied the namespace related equivalency tests from XmlDocument.

AndersAbel:
Looks like something is broken in the CI. It says "All checks have passed", but the build contains failed tests: https://ci.appveyor.com/project/dennisdoomen/fluentassertions/build/4.3.2+0%20(Build%20151)

dennisdoomen:
Yep, noticed that too on *master*. It's the Silverlight test. Will check out the build script to make sure it will fail on that.

dennisdoomen:
I've fixed the problems on *master*, so I'm about to merge this PR. However, you could have saved yourself some time by not putting terms like `XElement` in the names of tests. I follow these [principles](http://www.continuousimprover.com/2015/11/12-tips-to-write-unit-tests-that-dont.html).

AndersAbel:
I did that renaming to be able to see directly from the test name if it concerns XDocument or XmlDocument - but if you want I can roll it back and follow the linked principles.

dennisdoomen:
Would you?? :confounded:

Tests/FluentAssertions.Shared.Specs/XDocumentAssertionSpecs.cs
```diff
@@ -17,7 +17,7 @@ public class XDocumentAssertionSpecs
         #region Be / NotBe
 
         [TestMethod]
-        public void When_asserting_a_xml_document_is_equal_to_the_same_xml_document_it_should_succeed()
+        public void When_asserting_a_XDocument_is_equal_to_the_same_XDocument_it_should_succeed()
```
AndersAbel: I assume these are the renames you're thinking of. I read the principles and I think that in this case it still makes sense to name the test after the class. The principle (which is sound) is to not tie test names to *internal* names of the system, which in this case is FA. But here I'm tying names to the *external* interface being tested - i.e. the .NET API names (which will not be changed).

But that's just my 2c, so if you want I can spend 2 minutes reverting the renames.

dennisdoomen: Actually, the principle is to not tie test names to *any* literal name of a code element or anything alike. Next to that, the name should be functional. So in this particular case, I think I would name the test `When_an_xml_document_is_equal_to_another_it_should_not_throw`. But the problem is that I don't really know what *equals* means in this context. Otherwise I would make it more specific.

AndersAbel: I've reverted the renames of those tests and also renamed the other tests I've created to follow the same pattern.

For the question about what *equals* means in this case, I haven't looked into it. It's an existing test that I didn't modify.

AndersAbel:
Now I think this one is ready to be merged. Do you have anything more you want me to fix?

I've got another one coming with `HaveAttribute` etc. for XmlElements. I've based it on this commit, so I can't create a PR for it independently. As soon as this one is merged I'll submit the other one.

AndersAbel:
The additional assertions for XmlElements are now pushed to https://github.com/AndersAbel/FluentAssertions/tree/XmlElementAssertions. I'll make a PR for that when this one is merged.

PR #392, opened 2016-04-12
HaveInnerText, HaveElement and HaveAttribute for XmlElement

AndersAbel:
- Follows the principles of the corresponding methods on XDocument/XElement.
- It doesn't make sense to add these on XmlDocument as the DocumentElement
  is an explicit property there (XDocument behaves implicitly as the document
  XElement).
- Named the test HaveInnerText because the Value property is always null on XmlElements.

### DIFF
--- a/Src/FluentAssertions.Net40/Xml/XmlElementAssertions.cs
+++ b/Src/FluentAssertions.Net40/Xml/XmlElementAssertions.cs
@@ -1,0 +1,226 @@
+﻿using FluentAssertions.Common;
+using FluentAssertions.Execution;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using System.Xml;
+
+namespace FluentAssertions.Xml
+{
+    /// <summary>
+    /// Cointains a number of methods to assert that an <see cref="XmlElement"/>
+    /// is in the expected state./>
+    /// </summary>
+    [DebuggerNonUserCode]
+    public class XmlElementAssertions : XmlNodeAssertions<XmlElement, XmlElementAssertions>
+    {
+        /// <summary>
+        /// Initialized a new instance of the <see cref="XmlElementAssertions"/>
+        /// class.
+        /// </summary>
+        /// <param name="xmlElement"></param>
+        public XmlElementAssertions(XmlElement xmlElement)
+            : base (xmlElement)
+        { }
+
+        /// <summary>
+        /// Asserts taht the current <see cref="XmlElement"/> has the specified
+        /// <paramref name="expected"/> inner text.
+        /// </summary>
+        /// <param name="expected">The expected value.</param>
+        public AndConstraint<XmlElementAssertions> HaveInnerText(string expected)
+        {
+            return HaveInnerText(expected, string.Empty);
+        }
+
+        /// <summary>
+        /// Asserts that the current <see cref="XmlElement"/> has the specified
+        /// <paramref name="expected"/> inner text.
+        /// </summary>
+        /// <param name="expected">The expected value.</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<XmlElementAssertions> HaveInnerText(string expected, string because, params object[] becauseArgs)
+        {
+            Execute.Assertion
+                .ForCondition(Subject.InnerText == expected)
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected XML element {0} to have value {1}{reason}, but found {2}.",
+                    Subject.Name, expected, Subject.InnerText);
+
+            return new AndConstraint<XmlElementAssertions>(this);
+        }
+
+        /// <summary>
+        /// Asserts that the current <see cref="XmlElement"/> has an attribute
+        /// with the specified <paramref name="expectedName"/>
+        /// and <paramref name="expectedValue"/>.
+        /// </summary>
+        /// <param name="expectedName">The name of the expected attribute</param>
+        /// <param name="expectedValue">The value of the expected attribute</param>
+
+        public AndConstraint<XmlElementAssertions> HaveAttribute(string expectedName, string expectedValue)
+        {
+            return HaveAttribute(expectedName, expectedValue, string.Empty);
+        }
+
+        /// <summary>
+        /// Asserts that the current <see cref="XmlElement"/> has an attribute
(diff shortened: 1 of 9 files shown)


AndersAbel:
# Documentation #

Suggested documentation update for the Xml classes heading. Describes both the functionality in this PR as well as the other two that are already merged.

## XML classes ##

Fluent Assertions has support for assertions on several of the LINQ-to-XML classes and classic Xml classes:

```csharp
xDocument.Should().HaveRoot("configuration");
xDocument.Should().HaveElement("settings");

xElement.Should().HaveValue("36");
xElement.Should().HaveAttribute("age", "36");
xElement.Should().HaveElement("address");

xmlElement.Should().HaveInnerText("36");
xmlElement.Should().HaveAttribute("age", "36");
xmlElement.Should().HaveElement("address");
```

Those two last assertions also support `XName` parameters:

```csharp
xElement.Should().HaveAttribute(XName.Get("age", "http://www.example.com/2012/test"), "36");
xElement.Should().HaveElement(XName.Get("address", "http://www.example.com/2012/test"));

xAttribute.Should().HaveValue("Amsterdam");
```

For the classic xml classes namespaces are supported through the functions having `WithNamespace` suffix.

```csharp
xmlElement.Should().HaveAttributeWithNamespace("age", "http://www.example.com/2012/test"), "36");
xmlElement.Should().HaveElementWithNamespace("address", "http://www.example.com/2012/test");`
```

You can also perform a deep comparison between two elements/documents. `Be()` (only available on Linq-to-XML
classes) performs a strict element-by-element comparison through `XNode.DeepEquals()`. `BeEquivalentTo()`
performs a semantic equivalence check:

* Element name (including namespaces) should match.
* What prefix (if any) that is used for the names is ignored.
* Namespaces declarations (xmlns attributes) are completely ignored, as long as the resulting
  namespace on the element/attribute name match.
* Text contents should match.
* Comments are ignored.
* Order of attributes is ignored.

```csharp
xDocument.Should().Be(XDocument.Parse("<configuration><item>value</item></configuration>"));
xElement.Should().Be(XElement.Parse("<item>value</item>"));	

xDocument.Should().BeEquivalentTo(XDocument.Parse("<configuration><item>value</item></configuration>"));
xElement.Should().BeEquivalentTo(XElement.Parse("<item>value</item>"));	
xmlDocument.Should().BeEquivalentTo(XDocument.Parse("<configuration><item>value</item></configuration>"));
xmlElement.Should().BeEquivalentTo(XElement.Parse("<item>value</item>"));
```

Chaining additional assertions on top of a particular (root) element is possible through this syntax. 

```csharp
xDocument.Should().HaveElement("child").Which.Should().BeOfType<XElement>().And.HaveAttribute("attr", "1");
```

Src/FluentAssertions.Net40/Xml/AndWhichConstraint.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace FluentAssertions.Xml
+{
+    public class AndWhichConstraint<T>
```
dennisdoomen: Why do you need this? Isn't it in the core?

Src/FluentAssertions.Net40/Xml/XmlElementAssertions.cs
```diff
@@ -0,0 +1,190 @@
...
+                .ForCondition(attribute.Value == expectedValue)
+                .BecauseOf(because, becauseArgs)
+                .FailWith(
+                    "Expected XML attribute {0} to have value {1}{reason}, but found {2}.",
+                    expectedFormattedName, expectedValue, attribute.Value);
+
+            return new AndConstraint<XmlElementAssertions>(this);
+        }
+
+        public AndWhichConstraint<XmlElementAssertions, XmlElement> HaveElement(string expectedElement)
```
dennisdoomen: No docs?

Tests/FluentAssertions.Net40.Specs/XmlElementAssertionSpecs.cs
```diff
@@ -0,0 +1,509 @@
...
+        public void When_asserting_xml_element_is_equivalent_to_another_xml_element_with_same_contents_it_should_succeed()
+        {
+            // This test is basically just a check that the BeEquivalent method
+            // is available on XmlElementAssertions, which it should be if
+            // XmlElementAssertions inherits XmlNodeAssertions.
+
+            //-------------------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-------------------------------------------------------------------------------------------------------------------
+            var element = ParseElement(@"<user>grega</user>");
```
dennisdoomen: I prefer not to use any private methods. They tend to get all kinds of magic eventually.

Tests/FluentAssertions.Net45.Specs/Net45.Specs.csproj
```diff
@@ -84,6 +84,9 @@
     <Otherwise />
   </Choose>
   <ItemGroup>
+    <Compile Include="..\FluentAssertions.Net40.Specs\XmlElementAssertionSpecs.cs">
```
dennisdoomen: I just noticed that the XML stuff is only in the .NET 4.5 version. Is there  a reason for this? The XDocument stuff is in the Shared project.

Src/FluentAssertions.Net40/Xml/AndWhichConstraint.cs
AndersAbel: I don't need it. It's a stupid mistake (Ctrl+. going wild). I'll remove it. Sorry for the mess.

Src/FluentAssertions.Net40/Xml/XmlElementAssertions.cs
AndersAbel: Sorry, I think I was a bit too much in a hurry when trying to finish this. Adding all the missing docs.

dennisdoomen: Don't worry about that. The goal of the review is just to make sure the quality of the *master* branch is up to par.

Tests/FluentAssertions.Net45.Specs/Net45.Specs.csproj
AndersAbel: As far as I could find out, XmlDocument/XmlElement is not supported on portable etc. that is served by the Shared project. Looking at the docs it's only supported in full framework and UWP 10.

AndersAbel:
Pushed an update commit with fixes after the review.

Tests/FluentAssertions.Net45.Specs/Net45.Specs.csproj
dennisdoomen: You're right. The purpose of that shared project has gone beyond my initial goal by all the refactoring.

AndersAbel:
Thanks for reviewing and merging (I know that reviews takes time - a PR is a gift that takes time to accept). Now there's only one step remaining on our original list: the `IEquivalency` implementation. Unfortunately I won't have time to work on it in the near future. Is it something that you can do? Or should I add an issue for it and hope that someone picks it up?

PR #396, opened 2016-04-20
BeDataContractSerializable was added

borismod:


dennisdoomen:
You have failing unit tests.

borismod:
These tests were failing before my changes.

dennisdoomen:
You're right. And the PR that caused it was green... :confused:

borismod:
@dennisdoomen could you ping me, when nuget package is published with this change?

dennisdoomen:
@borismod [ping](https://www.nuget.org/packages/FluentAssertions/4.5.0)

PR #408, opened 2016-05-20
Fix: NullReferenceException occurs from AsyncFunctionAssertions.Shoul…

gyuwon:
Fix: NullReferenceException occurs from AsyncFunctionAssertions.ShouldNotThrow() when exception doen not have inner exception

When an async action throws any exception directly ShouldNotThrows() method throws not AssertFailedException but System.NullReferenceException.

```csharp
Func<Task> action = () =>
{
    throw new InvalidOperationException();
};

action.ShouldNotThrow(); // Throws System.NullReferenceException!
```

### DIFF
--- a/FluentAssertions.Net45.Specs/AsyncFunctionExceptionAssertionSpecs.cs
+++ b/FluentAssertions.Net45.Specs/AsyncFunctionExceptionAssertionSpecs.cs
@@ -183,6 +183,29 @@ namespace FluentAssertions.Net45.Specs
             action.ShouldThrow<AssertFailedException>()
                 .WithMessage("Did not expect System.ArgumentException, but found one*");
         }
+
+        [TestMethod]
+        public void When_async_method_throws_exception_without_inner_exception_it_should_throw_AssertFailedException()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            Func<Task> asyncAction = () =>
+            {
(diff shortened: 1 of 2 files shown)


gyuwon:
Oh, there is some conflict but I cannot find what it is exactly. Someone help me please.

FluentAssertions.Net45.Specs/AsyncFunctionExceptionAssertionSpecs.cs
```diff
@@ -183,6 +183,29 @@ public void When_async_method_throws_exception_expected_not_to_be_thrown_it_shou
             action.ShouldThrow<AssertFailedException>()
                 .WithMessage("Did not expect System.ArgumentException, but found one*");
         }
+
+        [TestMethod]
+        public void When_async_method_throws_exception_without_inner_exception_it_should_throw_AssertFailedException()
```
dennisdoomen: I don't get the rationale. Do you mean that it should not cause an internal error? 

PS. See also point 11 from [this post](http://www.continuousimprover.com/2015/11/12-tips-to-write-unit-tests-that-dont.html).

gyuwon: @dennisdoomen My poor English causes this confusing. I mean the situation an async method throws an exception that doesn't have an inner exception. Can you suggest a better test case name?

dennisdoomen:
Superseded by #415

Issue #366, opened 2016-03-03
support for NotBeOfType


SurajGupta:
This is request for adding the assertion NotBeOfType, just the negation of BeOfType

dennisdoomen:
Sure. I'd welcome a pull request if you will ;-)

somewhatabstract:
https://github.com/dennisdoomen/FluentAssertions/pull/413

snboisen:
Shouldn't this be closed with the merge of #413 ?

dennisdoomen:
@snboisen  You're right

PR #413, opened 2016-06-02
#366 support for NotBeOfType

somewhatabstract:
Added two methods for `NotBeOfType` and two associated tests (see issue #366).

-----

For the documentation, I can [update this page](https://github.com/dennisdoomen/fluentassertions/wiki#type-method-and-property-assertions) to include a couple examples as with the other `NotBe...`s.


Src/Core/Primitives/ReferenceTypeAssertions.cs
```diff
@@ -146,6 +164,31 @@ public AndConstraint<TAssertions> BeOfType(Type expectedType, string because = "
...
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> NotBeOfType( Type expectedType, string because = "", params object[] becauseArgs )
+        {
+            Execute.Assertion
+                .ForCondition( !ReferenceEquals( Subject, null ) )
+                .BecauseOf( because, becauseArgs )
+                .FailWith( "Expected type not to be {0}{reason}, but it is.", expectedType );
```
dennisdoomen: Consider replacing `type` with `{context:type}`. When your extension is using within an `AssertionScope`, you'll get information about the context.

Src/Core/Primitives/ReferenceTypeAssertions.cs
```diff
@@ -146,6 +164,31 @@ public AndConstraint<TAssertions> BeOfType(Type expectedType, string because = "
...
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> NotBeOfType( Type expectedType, string because = "", params object[] becauseArgs )
+        {
+            Execute.Assertion
+                .ForCondition( !ReferenceEquals( Subject, null ) )
```
dennisdoomen: There's no spec for this scenario.

dennisdoomen:
Thanks for the PR. Would you be willing to do the rework I suggested?

somewhatabstract:
Thanks for the comments. I can look at that rework. I copied the tests from the `BeOfType` specs, so it looks like there may be a missing spec for that too. I'll see if I can add that at the same time, unless you would prefer that be a separate PR.

Src/Core/Primitives/ReferenceTypeAssertions.cs
somewhatabstract: Sounds good; do you want me to do this for the `BeOfType` methods too?

dennisdoomen:
Yeah, some of the older specs don't meet the current quality guidelines.

somewhatabstract:
@dennisdoomen I have made some changes as suggested. I removed the `null` special case from the `NotBeOfType` since there is no real special case there, right? A `null` value is not of the given type. Let me know if you think that needs to go back in or if there needs to be a test for that scenario.

I have also added a test for `BeOfType` when the value is null and added the `{context:type}` part to the special case handling of `null` as suggested.

Src/Core/Primitives/ReferenceTypeAssertions.cs
somewhatabstract: Turns out I did not understand what this was really doing in `BeOfType`. It was not needed for the `NotBeOfType` so it's gone now.

dennisdoomen: Well, somebody *could* pass a `null` into that method. And then you don't want to end up with an internal error.

somewhatabstract: In that circumstance, is it a pass or a fail? Technically, null is not of the type given.

somewhatabstract: Thinking about it, it should probably fail if the item is null, since null can't have a type. I'll make the appropriate changes.

somewhatabstract:
@dennisdoomen OK, I've addressed the null situation. Part of me thought maybe null should be a pass for `NotBeOfType`, but I think it makes a little more sense as a fail. Let me know if I forget anything. Hopefully we're almost there. :+1:

Src/Core/Primitives/ReferenceTypeAssertions.cs
```diff
@@ -118,22 +118,40 @@ public AndConstraint<TAssertions> NotBeSameAs(TSubject unexpected, string becaus
...
+		/// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+		/// </param>
+		/// <param name="becauseArgs">
+		/// Zero or more objects to format using the placeholders in <see cref="because" />.
+		/// </param>
+		public AndWhichConstraint<TAssertions, T> NotBeOfType<T>( string because = "", params object[] becauseArgs )
+		{
+			NotBeOfType( typeof( T ), because, becauseArgs );
+
+			return new AndWhichConstraint<TAssertions, T>( (TAssertions)this, (T)(object)Subject );
```
snboisen: Sorry to comment on an already merged PR, but there is surely a copy/paste bug on this line. You have just asserted that Subject is not of type T, yet you attempt to cast it to T here. The return type of this method should probably just be AndConstraint, since you're not much wiser about the type of Subject after the test.

PR #415, opened 2016-06-05
Fix: NullReferenceException occurs from AsyncFunctionAssertions.Shoul…

dennisdoomen:
…dNotThrow() when exception doen not have inner exception

Rebased version of #408

### DIFF
--- a/Src/Core/Specialized/AsyncFunctionAssertions.cs
+++ b/Src/Core/Specialized/AsyncFunctionAssertions.cs
@@ -72,14 +72,14 @@ namespace FluentAssertions.Specialized
                 Task task = Subject();
                 task.Wait();
             }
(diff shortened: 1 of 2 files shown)


Src/Core/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -72,14 +72,14 @@ public void ShouldNotThrow(string because = "", params object[] becauseArgs)
                 Task task = Subject();
                 task.Wait();
```
tomachristian: Side note: Maybe look into [Task.WaitAndUnwrapException() extension ](https://gist.github.com/tomachristian/83cbf6618cdcd60949e90787347b9c36), it keeps the code cleaner IMO and you don't have too look into InnerException.

Issue #348, opened 2016-02-01
Allow MonitorEvents to be reset


stuartbrownstanbury:
If you need to check that multiple actions will trigger an event you need to create lots of small tests or recreate the unit under test and call .MonitorEvents() on it repeatedly. This request is for a way to clear out old events so that, for example, .ShouldRaisePropertyChangeFor() can be called over and over with the new "clear function" called in between.

dennisdoomen:
In other words, a `Clear` method on the `EventRecorder` returned from `MonitorEvents`?

stuartbrownstanbury:
Maybe a static method like monitor events. So you can do something like this:

`
[Test]
public void Test_Step()
{
    Step UUT = new Step();

    UUT.MonitorEvents();

    TerminationCondition TC1 = new TerminationCondition();
    TerminationCondition TC2 = new TerminationCondition();
    
    UUT.TerminationConditions.Add(TC1);
    UUT.ShouldRaisePropertyChangeFor(S => S.TerminationConditions);

    UUT.ClearEvents();
    UUT.TerminationConditions.Add(TC2);
    UUT.ShouldNotRaisePropertyChangeFor(S => S.TerminationConditions);
}
`

The new "ClearEvents" allows you to test that the event is raised for each insertion. Without it you can only say that the event was raised once. I need to check that it is raised twice (or more).

somewhatabstract:
The concept in Jasmine, where the history of calls on spies is cleared uses the nomenclature `reset:

http://jasmine.github.io/2.4/introduction.html#section-38

However, I think `ResetEvents()` and `ClearEvents()` don't really describe what's actually being done. In jasmine, it's clear what's being reset because it is a method off of `calls`. To that end, I think either an extension like `ResetMonitoredEvents()` or a `Reset()` call off the `EventRecorder` makes sense.

dennisdoomen:
That's pretty neat. I was thinking in the same direction. It will also work much better in parallel XUnit tests.

somewhatabstract:
OK, so I'll see about adding both the `Dispose` pattern approach and the `Reset` call. I think that covers the use cases that people need.

dennisdoomen:
This could also solve #477

bitbonk:
@dennisdoomen How could that solve #477? Is it because now that you explicitly dispose you do not need to use WeakReferences anymore ?

dennisdoomen:
[...]

PR #417, opened 2016-06-07
Issue #348: Adding ability to reset event monitoring

somewhatabstract:
This change overhauls the event monitoring somewhat so that the `MonitorEvents` method returns an object off which tracked events can be reset. This also adds a `Reset` call to the individual event recorders too. The event monitoring setup code has been moved into an `EventMonitor` class to encapsulate things a little more cleanly. Also, calling `MonitorEvents` multiple times with different interfaces for the same object adds to the existing events being tracked.

This is to address issue #348.


somewhatabstract:
Documentation changes would need to be made to cover the return value of `MonitorEvents`, the new `IEventMonitor` interface calls (`Reset`, `Attach`, and `GetEventRecorder`), and the new `Reset` method on `IEventRecorder`.

dennisdoomen:
You've been busy ;-)

What about support for a `using` block?

somewhatabstract:
I looked at that and found it to be unnecessary. It just didn't really add any value whatsoever. What would it actually do? What gets disposed of? I just couldn't see a useful scenario for it in the end.

dennisdoomen:
Well, it will make the scope of monitoring really explicit. And if the monitored code crashes, the reset will automatically happen without causing side-effect on other (parallel) tests.

somewhatabstract:
Yes, I suppose the `Dispose` would remove the event recorders for the specific monitor instead of just resetting them. That makes sense.

Would you mind if I cover that in a separate PR once this one is cleared through?

PR #432, opened 2016-07-04
Not contains fix

MortenBoysen:
This should fix issue https://github.com/dennisdoomen/FluentAssertions/issues/426. I also added more NotContain methods to align Contain and NotContain overloads more.

Being completely new to Git and Github, just let me know if I need to change something

### DIFF
--- a/Tests/FluentAssertions.Shared.Specs/CollectionAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Shared.Specs/CollectionAssertionSpecs.cs
@@ -1529,6 +1529,27 @@ namespace FluentAssertions.Specs
         }
 
         [TestMethod]
+        public void When_asserting_collection_does_contain_item_against_null_collection_it_should_throw()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            IEnumerable collection = null;
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => collection.Should()
+                .Contain(1, "because we want to test the behaviour with a null subject");
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act.ShouldThrow<AssertFailedException>().WithMessage(
+                "Expected collection to contain {1} because we want to test the behaviour with a null subject, but found <null>.");
+        }
+
+        [TestMethod]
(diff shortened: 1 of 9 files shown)


Tests/FluentAssertions.Shared.Specs/CollectionAssertionSpecs.cs
```diff
@@ -1580,6 +1601,13 @@ public void Should_succeed_when_asserting_collection_does_not_contain_an_item_th
         }
 
         [TestMethod]
+        public void Should_succeed_when_asserting_collection_does_not_contain_any_items_that_is_not_in_the_collection()
```
dennisdoomen: Please use the same naming conventions and AAA structure as the other specs.

Tests/FluentAssertions.Shared.Specs/GenericDictionaryAssertionSpecs.cs
```diff
@@ -917,6 +917,17 @@ public void Should_succeed_when_asserting_dictionary_does_not_contain_a_key_that
         }
 
         [TestMethod]
+        public void Should_succeed_when_asserting_dictionary_does_not_contain_multiple_keys_from_the_dictionary()
```
dennisdoomen: Please use the same naming conventions and AAA structure as the other specs.

dennisdoomen:
Nice work! 

I only have some minor comments. Are you willing to fix those? If so, I'll squash the commits during merging after you're done.

Tests/FluentAssertions.Shared.Specs/GenericDictionaryAssertionSpecs.cs
MortenBoysen: I actually copied both the structure and name from the existing test for Contain. That is why I thought it was correct. I simply changed _contains_ to _does_not_contain_ in the name.

MortenBoysen:
I can fix them, but if those tests are incorrectly named and structured, then there are several tests for Contain that are also incorrect - and perhaps others as well. Should we change all of them?

BTW: Could you give an example of how you would name these tests? The structure is easy to fix.

dennisdoomen:
Yeah, I know. These are old tests. I fix them when I or others touch them. Wrt the naming, I meant the `When_......_it_should_....` mantra.

MortenBoysen:
Ok. I will have a look and fix them. :-)

MortenBoysen:
I hope this fixes the issues. However, I have no idea if I am doing the right thing in Github to actually push the latest name fixes to you.

dennisdoomen:
Looks good. I'll merge and quash your changes.

PR #440, opened 2016-07-08
implemented BeOnOf for DateTime

jeroenpot:
Included the option for datetime to be one of specified values.


jeroenpot:
Oops.. let me fix that test first :-)

jeroenpot:
For the wiki, on this page
https://github.com/dennisdoomen/fluentassertions/wiki#dates-and-times

Include:
theDatetime.Should().BeOneOf(1.March(2010).At(22, 15), 1.February(2010));

dennisdoomen:
Nice. But can you do the same for `DateTimeOffset`. The assertions are supposed to be symmetrical.

jeroenpot:
Tests are running ok on local machine, but i think i found i see what the issue is.

jeroenpot:
Ok, fixed it.
If I can find some time next week I would like to built in BeCloseToOneOf

PR #445, opened 2016-07-13
Numbers are formatted in such a way that their type is visible...

### LINKED ISSUE #4: Print x.0 when x is is a double/float representing a whole number
See original CodePlex [issue](https://fluentassertions.codeplex.com/workitem/12493)

IharBury:
Numbers are formatted in such a way that their type is visible. Floating point numbers are formatted with their full precision.

Fixes #4


dennisdoomen:
Thanks for submitting this contribution. However, I'd prefer to use the normal C# post-fixes as described in [this post](http://www.blackwasp.co.uk/CSharpNumericLiterals.aspx). Would you be willing to change that?

IharBury:
There are two problems with that:

1. There are no post-fixes in C# for Byte, SByte, Int16 and UInt16
2. Numeric literal formats are different in different programming languages. FluentAssertions is not just for C#, right?

dennisdoomen:
> There are no post-fixes in C# for Byte, SByte, Int16 and UInt16

For those cases it might be acceptable to display them like you're current proposal does, but check out the [F# literals](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/literals-%5Bfsharp%5D). They are more complete than those of C#. 

> Numeric literal formats are different in different programming languages. FluentAssertions is not just for C#, right?
I'm not aware of anybody using FA in any other language than C#, but see above.

IharBury:
I've updated the PR to use C# and F# literal post-fixes where possible

PR #457, opened 2016-08-12
Upgraded to netstandard1.5

no1melman:
Updated from netstandard1.3 to netstandard1.5

### DIFF
--- a/Src/FluentAssertions.Core.Dotnet/project.json
+++ b/Src/FluentAssertions.Core.Dotnet/project.json
@@ -6,7 +6,7 @@
     "System.Reflection.TypeExtensions": "4.1.0" 
   },
   "frameworks": {
-    "netstandard1.3": {
+    "netstandard1.5": {
 
     }
   }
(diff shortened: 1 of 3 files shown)


no1melman:
I was seeing if it was that simple, but I don't have the UAP stuff installed on my VS, so couldn't open those projects and build with those. It works for all the standard stuff. Maybe worth just closing this PR

dennisdoomen:
It's not that simple. We carefully choose 1.3 so that we don't cause breaking changes for existing users.

no1melman:
Yeah, I didn't think it was going to be that simple. Is it possible to use this in a dotnet aspnet core app that targets 1.5?

dennisdoomen:
Sure. Libraries that target higher standards can always use lower versions.

dennisdoomen:
So what do we do wit this?

no1melman:
I'm not sure, shall I just close it... I would need to have all the other tools installed to be able to do the other upgrades......

dennisdoomen:
Well, I think this will be superseded by #463

Issue #467, opened 2016-08-22
ShouldBeEquivalentTo does not handle enum values > long.MaxValue


msackton:
When using ShouldBeEquivalentTo, for example to compare to object graphs, the assertion fails if any enumeration has a value greater than long.MaxValue.

Failing test:

 public enum EnumBig : ulong
    {
        ValueOne = UInt64.MaxValue, 
        ValueTwo = Int64.MaxValue,
        ValueThree = 0
    }

        public void ShouldBeEquivalentTo_should_succeed_when_asserting_large_enum_equals_large_enum()
        {
            Action act = () => EnumBig.ValueOne.ShouldBeEquivalentTo(EnumBig.ValueOne);
            act.ShouldNotThrow();
        }

Changing the EnumEqualityStep.Handle method to use Convert.ToDecimal rather than Convert.ToInt64 will fix this.

I would create a PR but I'm not sure where to add the test.

Updated to add: Using options.ComparingEnumsByName() does work as expected here, so that is an easy workaround for my use case (I don't actually need to compare by value).

dennisdoomen:
Fixed by PR #470

PR #470, opened 2016-08-23
Switch enum underlying type comparison to use decimal rather than long

msackton:
Fixes issue #467 by using decimal rather than long. I didn't see any place to add enum comparision tests so I made a new file.


dennisdoomen:
Thanks for the PR. Would you be willing to adapt the specs to use the AAA syntax and naming conventions? E.g. like https://github.com/dennisdoomen/FluentAssertions/blob/master/Tests/FluentAssertions.Shared.Specs/ComparableSpecs.cs#L57

msackton:
Sure, I probably won't get to it today, but happy to fix the naming.

dennisdoomen:
Hi @msackton, you've abandoned your PR?

msackton:
Sorry, I was out of town and forgot about this. I've cleaned up the tests.

PR #483, opened 2016-10-02
Non-generic BeAssignableTo enhancement

SaroTasciyan:
###### Description:
Asserting _subject object or Type should be assignable to given (non-generic) type parameter_. 

###### Changes:
- `ReferenceTypeAssertions` new overload for BeAssignableTo accepting (non-generic) type parameter
- `TypeAssertions` new overload for BeAssignableTo accepting (non-generic) type parameter
- Testing for overload implementation

###### Wiki:
I can add the samples under [Basic Assertions](https://github.com/dennisdoomen/FluentAssertions/wiki#basic-assertions) and [Type, Method and Property Assertions](https://github.com/dennisdoomen/FluentAssertions/wiki#type-method-and-property-assertions) if necessary.

###### Note:
_ReferenceTypeAssertion.BeAssignableTo_ overload does not return _AndWhichContraint_ since the generic type parameter is not present. Instead of returning and-which contraint with object, it returns _AndContraint_.

### DIFF
--- a/Tests/FluentAssertions.Shared.Specs/ObjectAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Shared.Specs/ObjectAssertionSpecs.cs
@@ -454,6 +454,68 @@ namespace FluentAssertions.Specs
             act.ShouldThrow<AssertFailedException>().WithMessage("*Expected*Other*Actual*");
         }
 
+        [TestMethod]
+        public void When_asserting_an_object_is_assignable_its_own_type_using_type_instance_it_should_succeed()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            var someObject = new DummyImplementingClass();
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act / Assert
+            //-----------------------------------------------------------------------------------------------------------
+            someObject.Should().BeAssignableTo(typeof(DummyImplementingClass));
+        }
+
+        [TestMethod]
+        public void When_asserting_an_object_is_assignable_to_its_base_type_using_type_instance_it_should_succeed()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            var someObject = new DummyImplementingClass();
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act / Assert
+            //-----------------------------------------------------------------------------------------------------------
+            someObject.Should().BeAssignableTo(typeof(DummyBaseClass));
+        }
+
+        [TestMethod]
+        public void When_asserting_an_object_is_assignable_to_an_implemented_interface_type_using_type_instance_it_should_succeed()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            var someObject = new DummyImplementingClass();
+
+            //-----------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


Tests/FluentAssertions.Shared.Specs/ObjectAssertionSpecs.cs
```diff
@@ -454,6 +454,68 @@ public void When_object_is_assignable_to_the_expected_type_it_should_cast_the_re
...
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act / Assert
+            //-----------------------------------------------------------------------------------------------------------
+            someObject.Invoking(
+                x => x.Should().BeAssignableTo(assignableCheckType, "because we want to test the failure {0}", "message"))
+                .ShouldThrow<AssertFailedException>()
+                .WithMessage(string.Format(
+                    "Expected object to be assignable to {1} because we want to test the failure message, but {0} is not",
+                    typeof(DummyImplementingClass), assignableCheckType));
```
dennisdoomen: I prefer to write these as:

```
// Act
Action act = () => someObject.Should().BeAssignableTo(assignableCheckType, "because we want to test the failure {0}", "message");

// Assert
act.ShouldThrow<AssertFailedException>()
   .WithMessage($"*assignable to {assignableCheckType}*failure message*{typeof(DummyImplementingClass)} is not*"));
```

This solves two problems:
1. `Invoking` is an old construct and is IMO a little less readable.
2. The exact exception message is irrelevant, as long as the important parts are there.

Tests/FluentAssertions.Shared.Specs/ObjectAssertionSpecs.cs
```diff
@@ -454,6 +454,68 @@ public void When_object_is_assignable_to_the_expected_type_it_should_cast_the_re
             act.ShouldThrow<AssertFailedException>().WithMessage("*Expected*Other*Actual*");
         }
 
+        [TestMethod]
+        public void When_asserting_an_object_is_assignable_its_own_type_using_type_instance_it_should_succeed()
```
dennisdoomen: Don't repeat the `When_asserting_an_object_is_assignable`. I'd rather put all the methods in a region and keep the names short and clear.

Tests/FluentAssertions.Shared.Specs/TypeAssertionSpecs.cs
```diff
@@ -413,6 +413,53 @@ public void When_asserting_an_object_is_assignable_to_an_unrelated_type_it_fails
...
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            Type someType = typeof(DummyImplementingClass);
+            Type assignableCheckType = typeof(DateTime);
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act / Assert
+            //-----------------------------------------------------------------------------------------------------------
+            someType.Invoking(
```
dennisdoomen: Same remark as above.

Tests/FluentAssertions.Shared.Specs/ObjectAssertionSpecs.cs
SaroTasciyan: It's my pleasure. Thanks for sharing the article, Maybe I am too focused on sticking with the existing conventions and style of the code base.

For the two comments below, I totally agree with you and I didn't like the `Invoking()` usage in the first place. However, the tests I added are already placed within a region (BeAssignableTo). For instance I can rename the first test to  `When_its_own_type_it_succeeds` removing the part you specified. Maybe `When_subjects_own_type_it_succeeds`

PR #486, opened 2016-10-08
MSTest V2 support

AbhitejJohn:
Adding in support for MSTest V2 which is the next version of the MSTest framework as detailed [here](https://blogs.msdn.microsoft.com/visualstudioalm/2016/06/17/taking-the-mstest-framework-forward-with-mstest-v2/). It supports .Net 4.5+, UWP and dotnet core.

### DIFF
--- a/FluentAssertions.sln
+++ b/FluentAssertions.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.25123.0
+VisualStudioVersion = 14.0.25420.1
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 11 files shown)


dennisdoomen:
Awesome. I'm trying to fix one more bug and then I'll release this to the general public.

AbhitejJohn:
Thanks Dennis.

AbhitejJohn:
@dennisdoomen : How do we add this(MSTest V2) to the set of officially supported frameworks in the documentation?

dennisdoomen:
Just did ;-)

AbhitejJohn:
Thanks. Would that surface here too: http://www.fluentassertions.com/

dennisdoomen:
Now it does...;-)

AbhitejJohn:
Thanks a bunch :)

PR #492, opened 2016-10-25
Allow to fluently specify the DateTimeKind when creating DateTimeValues

mpopdam:


Tests/FluentAssertions.Shared.Specs/FluentDateTimeSpecs.cs
```diff
@@ -185,12 +185,43 @@ public void When_fluently_specifying_a_date_and_time_it_should_return_the_correc
...
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            DateTime dateTime = 10.December(2011).At(09, 30, 45, 999).AsUtc();
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            Assert.AreEqual(new DateTime(2011, 12, 10, 9, 30, 45, 999), dateTime);
```
dennisdoomen: Why don't you use FA itself here?

dennisdoomen: Would you be willing to change that?

Tests/FluentAssertions.Shared.Specs/DateTimeAssertionSpecs.cs
```diff
@@ -1908,6 +1908,68 @@ public void When_asserting_subject_null_datetime_should_not_have_second_should_t
...
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            DateTime subject = new DateTime(2009, 12, 31, 23, 59, 00, DateTimeKind.Local);
+            DateTimeKind expectation = DateTimeKind.Local;
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => subject.Should().Represent(expectation);
```
dennisdoomen: What do you think of `dateTime.Should().BeIn(DateTimeKind.Local)`

mpopdam: I was not sure what a proper name would be. The [MSDN documentation for DateTimeKind](https://msdn.microsoft.com/en-us/library/shx7s921(v=vs.110).aspx) says 

> Specifies whether a DateTime object represents a local time, a Coordinated Universal Time (UTC), or is not specified as either local time or UTC.

That's how I came up with the `Should().Represent` method. But I can live with your suggestion of `BeIn`.

Issue #428, opened 2016-06-24
Implement ShouldAllBeOfType() and ShouldAllBeOfExactType for enumerables


jon-armen:
Say I have an instance of AggregateException.  I want to say aggEx.InnerExceptions.ShouldAllBeOfExactType<MyException>().  This would ensure that all objects in the collection are exactly MyException, and not derived.  I could also use the ShouldAllBeOfType to allow for derived classes.  Attempting to keep this request in line with how actions use ShouldThrow or ShouldThrowExactly.

dennisdoomen:
So using exceptions here is just an example?

jon-armen:
Right, another example might be if I have a collection such as List<object>, IEnumerable<IInterface>, or IList<BaseType>.  For any of these, It would be nice to say ShouldAllBeOfType.

dennisdoomen:
Released in 4.18.0

PR #507, opened 2016-11-15
fix for issue #428;  implement Should AllBeOfType and Should AllBeOfE…

ben-m-lucas:
…xactType for CollectionAssertions

### DIFF
--- a/Src/Core/Collections/CollectionAssertions.cs
+++ b/Src/Core/Collections/CollectionAssertions.cs
@@ -1279,6 +1279,90 @@ namespace FluentAssertions.Collections
         }
 
         /// <summary>
+        /// Asserts that all items in the collection are of the specified type <typeparamref name="T" />
+        /// </summary>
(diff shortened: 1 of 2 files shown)


dennisdoomen:
Thanks for providing this PR. I've noticed that you've used the extra overload with *exact* in its name to denote the exactness of the comparison. But if you compare these methods with the  methods on the [ReferenceTypeAssertions](https://github.com/dennisdoomen/FluentAssertions/blob/master/Src/Core/Primitives/ReferenceTypeAssertions.cs#L136), you'll notice that latter has two methods: `BeOfType` and `BeAssignableTo`. The former does an exact match, whereas the latter only verifies if the subject type is a sub-class of the provided type. So, your PR does the opposite of what the rest of FA is doing. I'd rather keep a consistent API. Do you agree?

ben-m-lucas:
That makes sense to keep a consistent API.  I'll update the PR to be consistent with the API.

Issue #446, opened 2016-07-18
Additional GenericDictionaryAssertions


sbrickey:
I found these to be useful:

```
public static void ContainKeyWithValue<TKey, TValue>(this GenericDictionaryAssertions<TKey, TValue> dict, TKey key, TValue value, string because = "", params object[] becauseArgs)
{
    dict.ContainKey(key, because, becauseArgs);
    dict.Subject[key].Should().Be(value, because, becauseArgs);
}

public static void ContainKeyWithValue<TKey, TValue>(this GenericDictionaryAssertions<TKey, TValue> dict, TKey key, Func<TValue, TValue> valueModifier, TValue value, string because = "", params object[] becauseArgs)
{
    dict.ContainKey(key, because, becauseArgs);
    valueModifier(dict.Subject[key]).Should().Be(value, because, becauseArgs);
}
```

benefit being that I can now do this...

```
actual.Should().ContainKeyWithValue("Item1", "a"); // default
actual.Should().ContainKeyWithValue("Item2", v => v.ToLower(), "b"); // using valueModifier
```

yes, it's only a tiny bit of code savings... but the formatting of the assertions is far easier to follow

sbrickey:
note: I suppose the second method (with modifier) should probably contain TModifiedValue, should the modifier return a different type of object... though i start to wonder how much that deviates towards additional ACTions instead of focusing on the ASSERTions.

dennisdoomen:
Looks like a decent improvement. Care to provide me with a PR?

sbrickey:
I'll let someone else take care of that, as well as unit test coverage, naming and other conventions to be consistent with the existing codebase, etc... I just wanted to provide the idea and an example implementation.

ben-m-lucas:
Note that I ended up not implementing the second method that used the valueModifier parameter.  When I was trying to write the failure message for this method, I found that no matter how I wrote the message, it was confusing.  

For example:

a dictionary with {[1, "One"], [2, "Two"]} and the call to ContainKeyWithValue(1, x => x.ToLower(), "two");

has the potential failure message:
Expected dictionary {[1, "One"], [2, "Two"]} to contain [1, "two"], but found [1, "one"]

The language implies that the dictionary should contain [1, "one"] but it doesn't.  The best I could think of was to modify the message as follows:
Expected dictionary {[1, "One"], [2, "Two"]} to contain modified value [1, "two"], but found modified value[1, "one"]

[...]

PR #508, opened 2016-11-15
implement ContainKeyWithValue as requested in issue #446

ben-m-lucas:


dennisdoomen:
How is that different from https://github.com/dennisdoomen/FluentAssertions/blob/master/Src/Core/Collections/GenericDictionaryAssertions.cs#L794?

ben-m-lucas:
You're right.  It's doing the same as the Contain method override.  I looked back over the issue and it looks like there were two parts.  One was to implement the method I implemented which is really just the same as the Contain method.  

The other was an override that took a function to apply to the dictionary value allowing the user to check some modified version of the dictionary value.  When I tried to implement the second method that doesn't exist, I found in writing the unit tests that I couldn't create an error message that accurately described what was going on and decided against implementing it.  

As it stands, this PR isn't needed.  If you agree with me that the other method is confusing, then maybe we should close that issue.  Thoughts?

PR #513, opened 2016-12-02
Support timespans shorter than 1 millisecond

jamezor:
Currently if an assertion dealing with a sub 1 millisecond timespan fails,
the error message describes the timespan as 'default'. Which makes it hard
to figure out what is going on.

This commit extends the timespan formatter to support timespans between
0.1 microseconds and 1 millisecond.


dennisdoomen:
Would you be willing to add a unit test?

jamezor:
Unit tests added, sorry for not including them originally!

Issue #30, opened 2013-10-21
Support WithArgs assertion for .NET 4.5 EventHandlers


timbussmann:
The restriction of EventHandler<TEventArgs> for TEventArgs to be of type EventArgs has fallen away with .NET 4.5 (see http://msdn.microsoft.com/en-us/library/db0etb8x(v=vs.110).aspx).

The WithArgs<TEventArgs>(..) assertion still requires TEventArgs to be of type EventArgs. This prevents us from making argument assertions for EventHandlers with TEventArgs other than EventArgs (e.g. EventHandler<string>).

Tdue21:
This would make my life much simpler as well :)

okobylianskyi:
I'm almost ready to provide PR for this issue, but I need to know the decision on @vossad01 question about repeated type parameters. My implementation currently sticks with second option. Should I rework it or it is enough just to assert that value was present among arguments of the same type?

dennisdoomen:
You still want to wrap this up?

dennisdoomen:
Nobody has been asking for this in three years, so we're going to close it.

PR #517, opened 2016-12-05
Fix for issue #30: Allow not EventArgs based event handlers

okobylianskyi:
Fork contains two branches: **master** and **alternative-idea**.

1. Master. Contains more clear and simple code but has one (minor?) issue. If delegate behind event has repeated generic types (like ```Action<string, string>```), assertion will be successful if expected value is present in ANY of the parameters (first string or second string).

2. Alternative-idea. Contains more strict but less pretty for end user implementation. The idea is to accept ```params``` of predicates in ```WithArgs```. So, if you need strict assertions for two input strings, you can specify two predicates, each corresponding to particular parameter in order. The one drawback is about the case when you wish to skip first N parameters and then assert the next. In this case you should specify ```null``` or ```args => true``` for the first N parameters.

### DIFF
--- a/Src/Shared/PropertyChangedSourceExtensions.cs
+++ b/Src/Shared/PropertyChangedSourceExtensions.cs
@@ -143,10 +143,9 @@ namespace FluentAssertions
         }
 
         /// <summary>
-        /// Asserts that at least one occurrence of the event had an <see cref="EventArgs"/> object matching a predicate.
+        /// Asserts that at least one occurrence of the event had at least one of the arguments matching a predicate.
         /// </summary>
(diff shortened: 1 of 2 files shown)


dennisdoomen:
Can't you provide both overloads?

okobylianskyi:
Done

somewhatabstract:
ASIDE: is 13926834546 a spam user? That looks suspiciously like a phone number.

dennisdoomen:
Noticed it as well. Not sure.

okobylianskyi:
I'm also confused with this "approve".
So what about this pull request? Is it ok or I should change smth?

somewhatabstract:
I'll let @dennisdoomen finish his review, but FYI, I reported that weird user as a spam account, since none of us know what it is.

PR #521, opened 2016-12-28
Reimplement FluentAssertions.com using Jekyll

vossad01:
I ended not being as ambitious as I thought and creating an entirely new website.  I learning all design things I needed to create a theme from scratch just wasn't something for me to prioritize right now.

By the time I realized that effort was bigger than I thought, I knew enough Jekyll to easily re-implement the existing website with the same theme only using the GitHub pages Jekyll build support.

While not flashy, this will make management and modification easier because the content is primarily in Markdown.  It is also then better positioned for changing themes, adding more pages, etc. if those ever are desired down the road.

I was also able add a couple new touches like reincorporating the old logo from the Codeplex days.

It would be easy enough for me to make minor modifications to this if there is something you would like to see.  However, big design changes are simply outside of my current skill set.

screen shot 2016-12-27 at 22 17 21-fullpage

The maintenance link at the bottom is tied to the GitHub account hosting the website, which is the only reason it lists me.
Live preview available here: http://vossad01.github.io/fluentassertions/


dennisdoomen:
Awesome!

dennisdoomen:
Now that you've done the groundwork, I can thing of a couple of follow up actions:
* Move the logo in the title bar
* Add the contents of the wiki to the site 
* Add the latest release notes 
* Twitter mentions
* News and references to other posts about FA.

vossad01:
> *   Move the logo in the title bar

This can probably be done.  I started there but was not immediately coming up with a layout that looked as good.  The black text on black gradient was still visible but not great.  I also thought there was some visual mis-match between the organic shape of the logo and the blocky style of the header.

> *   Add the contents of the wiki to the site

I gave up on this because [GitHub Pages only supports a limited number of  plugins](https://pages.github.com/versions/) and [the one that processes wiki pages to be compatible with Jekyll is not included](https://github.com/djacquel/WikiToJekyll).

I just realized that the lack of Front Matter on the pages is why the plugin is needed, so tried adding Front Matter manually to a page on the wiki.  I think I broke your wiki doing that because now `Edit` returns 404 for all pages.  I tried pushing a revert of my change via command line, but that gave a 403 for my user.

If the Front Matter is what broke it (and it is not just a coincidental system outage) you will probably need to try pushing via command line or contact GH Support to fix it.  Sorry. Since I saw this my account is now "flagged" for some reason, so that may also be why I can't edit.

> *   Add the latest release notes

So long as it is not trying to go within the Markdown provided content, this should be as easy as building the HTML to display it.  [The relevant information is available to the templates.](https://github.com/blog/1996-releases-metadata-for-github-pages)

> *   Twitter mentions

I'm guessing that is this: https://twitter.com/settings/widgets/new/search?query=FluentAssertions

> *   News and references to other posts about FA.

I am guessing this is manually populated?

dennisdoomen:
> I also thought there was some visual mis-match between the organic shape of the logo and the blocky style of the header.

Maybe I should ask the guy that designed the logo for my blog to create a new logo for FA.

> I gave up on this because GitHub Pages only supports a limited number of plugins and the one that processes wiki pages to be compatible with Jekyll is not included.

We could decide to move the wiki into an .md file in the repo. 

> I'm guessing that is this: https://twitter.com/settings/widgets/new/search?query=FluentAssertions

I typically use `fluentassertion OR "fluent assertion" OR fluentassertions OR "fluent assertions" `

> I am guessing this is manually populated?

Yes.

vossad01:
> We could decide to move the wiki into an .md file in the repo.

I think there is probably a good argument for moving the wiki pages into the GH pages branch and disabling the wiki.  I don't think disabling the wiki was always possible on GitHub, but it is now.

That said, it turns out that as of early December it would have been possible to include the wiki pages thanks the the availability of the Optional Front Matter plugin and the Jekyll Titles from Headings plugin [(announcement)](https://github.com/blog/2289-publishing-with-github-pages-now-as-easy-as-1-2-3).

Issue #459, opened 2016-08-15
Should().BeOfType<T>() does not support value types


RaYell:
`Should.BeOfType<T>()` supports only reference types. If you try something like

```
var x = 1;
x.Should().BeOfType<int>();
```

the code will not compile as `BeOfType` is defined only for reference types.

The workaround is something like `x.GetType().Should.Be(typeof(int))` but I think it'd be easier to just allow value types to be used with this method.

dennisdoomen:
Thanks for reporting this. It's funny nobody has noticed this over the course of the five years FA exists.

ankurMalhotra:
Can I pick this up if no one else is working on it?

dennisdoomen:
Sure. Go ahead.

raghav710:
I think this issue can be closed. Can someone please do the honours :)

PR #526, opened 2017-01-07
Added support for BeOfType and NotBeOfType for Value type objects  #459

DennisNerush:
Fixing issue #459


Tests/FluentAssertions.Shared.Specs/ObjectAssertionSpecs.cs
```diff
@@ -261,7 +261,83 @@ public void When_object_type_is_exactly_equal_to_the_specified_type_it_should_no
...
+
+            //-------------------------------------------------------------------------------------------------------------------
+            // Act
+            //-------------------------------------------------------------------------------------------------------------------
+            Action act = () => valueTypeObject.Should().BeOfType(typeof(double));
+
+            //-------------------------------------------------------------------------------------------------------------------
+            // Assert
+            //-------------------------------------------------------------------------------------------------------------------
+            act.ShouldThrow<AssertFailedException>();
```
dennisdoomen: Please also include some checks on the message so it's clear that the expected and actual types are included.

DennisNerush:
added the required assert

dennisdoomen:
> Added the required assert

You added it on the 2nd spec. Was that on purpose?

DennisNerush:
what do you mean? I've added it the ObjectAssertionSpec, in the test that you asked

DennisNerush:
My bad :) take a look now

Issue #527, opened 2017-01-10
Missing NotBeDecoratedWith assertion for types


zkestenbaum:
There does not appear to be any way to assert that a type is NOT decorated with a certain attribute. 

The wiki shows how to test for the presence of an attribute (see [https://github.com/dennisdoomen/fluentassertions/wiki#type-method-and-property-assertions](url)), but there is no way to test for the absence of an attribute. `NotBeDecoratedWith` already exists as a `MemberInfoAssertion `and a `PropertyInfoAssertion`, but not as a `TypeAssertion`.

Also see this related question on StackOverflow - [http://stackoverflow.com/questions/24070898/fluent-assertions-thatarenotdecoratedwith](url).

I would propose adding `NotBeDecoratedWith` and `ThatAreNotDecoratedWith` for types.

dennisdoomen:
You're right. We forgot about that one. I would welcome a PR for that ;-)

zkestenbaum:
Just submitted a pull request - my first one on GitHub, so I welcome any feedback.

dennisdoomen:
Hi @zkestenbaum. Are you still going to finish your PR?

zkestenbaum:
Yes, thanks for following up. Will finish this shortly.

PR #528, opened 2017-01-11
Added NotBeDecoratedWith to TypeAssertions, TypeSelectorAssertions, P…

zkestenbaum:
…ropertyInfoSelectorAssertions and MethodInfoSelectorAssertions.

Added unit tests for above additions.

Resolves (part of) issue #527 (did not add `ThatAreNotDecoratedWith`, only added `NotBeDecoratedWith`).

### DIFF
--- a/Tests/FluentAssertions.Shared.Specs/MethodInfoSelectorAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Shared.Specs/MethodInfoSelectorAssertionSpecs.cs
@@ -73,9 +73,9 @@ namespace FluentAssertions.Specs
                 .WithMessage("Expected all selected methods" +
                              " to be virtual because we want to test the error message," +
                              " but the following methods are not virtual:\r\n" +
-                             "Void FluentAssertions.Specs.ClassWithNonVirtualPublicMethods.PublicDoNothing\r\n" +
                              "Void FluentAssertions.Specs.ClassWithNonVirtualPublicMethods.InternalDoNothing\r\n" +
-                             "Void FluentAssertions.Specs.ClassWithNonVirtualPublicMethods.ProtectedDoNothing");
+                             "Void FluentAssertions.Specs.ClassWithNonVirtualPublicMethods.ProtectedDoNothing\r\n" +
+                             "Void FluentAssertions.Specs.ClassWithNonVirtualPublicMethods.PublicDoNothing");
         }
 
         [TestMethod]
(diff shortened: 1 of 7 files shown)


dennisdoomen:
Thanks for this contribution. It seems that some of the unit tests fail.

zkestenbaum:
Fixed the failed unit test (which was passing on my machine). There was a (preexisting) problem where some unit tests were checking for the exact exception message, but that message was indeterminate since it depended on an array of method descriptions (or property descriptions, or type descriptions) whose order is indeterminate. I added an OrderBy to the private methods that generates the descriptions summary so that it is determinate.

[Type.GetProperties](https://msdn.microsoft.com/en-us/library/kyaxdd3x%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396):
> The GetProperties method does not return properties in a particular order, such as alphabetical or declaration order. Your code must not depend on the order in which properties are returned, because that order varies.

[Type.GetMethods](https://msdn.microsoft.com/en-us/library/4d848zkb(v=vs.110).aspx):
> The GetMethods method does not return methods in a particular order, such as alphabetical or declaration order. Your code must not depend on the order in which methods are returned, because that order varies.

Tests/FluentAssertions.Shared.Specs/MethodInfoSelectorAssertionSpecs.cs
```diff
@@ -141,9 +141,104 @@ public void When_asserting_methods_are_decorated_with_attribute_but_they_are_not
...
+            // Act
+            //-------------------------------------------------------------------------------------------------------------------
+            Action act = () =>
+                methodSelector.Should().NotBeDecoratedWith<DummyMethodAttribute>("because we want to test the error {0}", "message");
+
+            //-------------------------------------------------------------------------------------------------------------------
+            // Assert
+            //-------------------------------------------------------------------------------------------------------------------
+            act.ShouldThrow<AssertFailedException>()
+                .WithMessage("Expected no selected methods to be decorated with" +
```
dennisdoomen: The `WithMessage` takes a wildcard pattern, so I would recommend verifying only the relevant parts of the message (as explained in [this blog post](http://www.continuousimprover.com/2015/11/12-tips-to-write-unit-tests-that-dont.html)). IMO, you don't want a failing test just because the wording of the message has changed.  Same for the other specs.

dennisdoomen:
@zkestenbaum Did you see my code review comments?

PR #529, opened 2017-01-14
Support Excluding/Including with expressions using indexers on types other than arrays

duncanbrown:
Currently, `foo.ShouldBeEquivalentTo(bar, opts => opts.Excluding(x => x.Baz[0].Quux)` will throw an `ArgumentException` if `Baz` is not an array, but is rather a list, dictionary etc.. This adds support for including/excluding using such non-array indexer expressions.

### DIFF
--- a/Src/Core/Common/ExpressionExtensions.cs
+++ b/Src/Core/Common/ExpressionExtensions.cs
@@ -96,6 +96,8 @@ namespace FluentAssertions.Common
                 throw new NullReferenceException("Expected an expression, but found <null>.");
             }
 
+            var unsupportedExpressionMessage = $"Expression <{expression.Body}> cannot be used to select a member.";
+
             while (node != null)
             {
                 switch (node.NodeType)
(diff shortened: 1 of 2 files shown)


Src/Core/Common/ExpressionExtensions.cs
```diff
@@ -129,9 +131,17 @@ public static class ExpressionExtensions
                         node = null;
                         break;
 
+                    case ExpressionType.Call:
+                        var methodCallExpression = node as MethodCallExpression;
```
dennisdoomen: If the `NoteType` is `Call`, can the `node` be anything else than a `MethodCallExpression`? Because the `as` operator does kind of imply that, as well as the `?` operator when you try to get the `Method`.

Src/Core/Common/ExpressionExtensions.cs
```diff
@@ -129,9 +131,17 @@ public static class ExpressionExtensions
...
+                        if (methodCallExpression?.Method.Name != "get_Item" || methodCallExpression.Arguments.Count != 1 || !(methodCallExpression.Arguments[0] is ConstantExpression))
+                            throw new ArgumentException(unsupportedExpressionMessage);
```
dennisdoomen: Why don't you inline the actual exception message here? 

By the way, I prefer to always use braces, per [these](https://github.com/dennisdoomen/CSharpGuidelines/blob/master/Src/Guidelines/2400_LayoutGuidelines.md) coding conventions

Src/Core/Common/ExpressionExtensions.cs
duncanbrown: Good point - I'd assumed it could be something else, but doesn't look like that's the case. Updated accordingly.

Src/Core/Common/ExpressionExtensions.cs
duncanbrown: To avoid duplication, because the exception message is now used in a couple of places (here and the `default` case).

Updated to add braces.

PR #541, opened 2017-03-05
FluentAssertions.com example -- correction for clarity

SeanKilleen:
Hi all,

I know FV is under heavy refactoring but I wanted to submit a minor fix for the home page; I won't be offended if you reject. :) 

The home page current has an example where a list of 3 items is created but the example expects four. The explanation reads "because we thought we added three items", but it really means "because we thought we added four items." This will make the failure message make more sense.

Would have submitted an issue first, but this seemed minor enough. Thanks for your consideration!

### DIFF
--- a/index.md
+++ b/index.md
@@ -19,12 +19,12 @@ To verify that a collection contains a specified number of elements and that all
 
 ```c#
 IEnumerable collection = new[] { 1, 2, 3 };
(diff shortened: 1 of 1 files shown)


dennisdoomen:
Thanks @SeanKilleen

PR #549, opened 2017-04-01
Move documentation to website, Move website to docs :-)

vossad01:
This does the following (derived from https://github.com/fluentassertions/fluentassertions/pull/521#issuecomment-269437351):

 - [x] Adds the contents of the wiki (ie. documentation) to the website
 - [x] Moves the website from its own repository to the `docs` folder
 - [x] Adds release notes
 - [x] Embeds Twitter mentions

This required a large rework of the website.  The base-theme was changed to [minima](https://github.com/jekyll/minima) to facilitate but [styling was added](https://github.com/vossad01/jekyll-minima-flavor-slate) to make it appear like the previous [slate](https://github.com/pages-themes/slate/) theme.  The main changes with respect to theme are the header and footer.
You can preview the website at: http://vossad01.github.io/fluentassertions/

**The move to the Docs folder:**
Some features of the website (release information) will not work unless it is hosted from this repository; and having the separate repository has not been without its own issues.  A `gh-pages` branch like before would work, but I always thought that (and the wiki) reduced visibility of the documentation and required separate changes.  GitHub allows for Pages to be sources from the `docs` folder and this way a single PR can implement a feature and add relevant documentation.

Because the preview is not hosted from this repository, the `News` section and the `Releases` page will not function (they depend on the release information from the hosting repository.

Here is what they look like:
image
^ Always shows the latest release (with whatever frequency GH rebuilds pages, may need to watch that if it is too infrequent)

image
^ This page lists all releases on GitHub (going back to 2.1)

The Wiki should be disabled if this is accepted.
https://github.com/fluentassertions/fluentassertions.github.io should be deleted if this is accepted (it is fully merged here at the time of PR creation).
PULL_REQUEST_TEMPLATE.md should be updated to remove reference to the Wiki if this is accepted.
This continues from and supersedes dennisdoomen/fluentassertions#1


dennisdoomen:
Awesome work @vossad01. 

So just to be sure, after merging this PR, I should switch the DNS entry back to its original setting? Are you sure the entire website can be served from the Docs folder?

vossad01:
> So just to be sure, after merging this PR, I should switch the DNS entry back to its original setting?

I don't know the specifics here, but I'm guessing not quite.  Originally it would have been pointing to http://dennisdoomen.github.io/fluentassertions.  Currently it is to http://fluentassertions.github.io.  After this is merged we would want it pointing at http://fluentassertions.github.io/fluentassertions.

> Are you sure the entire website can be served from the Docs folder?

image

With the above settings change, yes.  Only it is the `docs` folder not `Docs`.  Full disclosure: that may be a little bit of a pain point for a little while for existing clones on Windows (unless something has changed since I last did development with Git on Windows).  Since windows treats the file system as case-insensitive, git does not actually rename the folder so existing clones will continue to be `Docs` on the fs but `docs` to Git.  This works okay for modifications for existing files, but any **new** files created under in a "Docs" folder will be added to Git under `Docs` which is a different folder from `docs`.  There is really no way around this and I figured most users are not creating new files under "Docs" so the issue will probably be minimal.

dennisdoomen:
@SeanKilleen any opinions on this? It would undo the work you've done recently.

SeanKilleen:
@dennisdoomen sorry for the late reply; have some personal events this weekend. I likely have an opinion but will want to look at this more before offering it :) will try to put eyes on it this evening.

SeanKilleen:
Hi all,

Sorry for the delay. My thoughts -- 

@vossad01 this is an awesome effort -- really cool work! I wish I was aware it was going on at the time, we probably could have teamed up more. 

This setup still accomplishes what I set out to do originally:

✅ Makes it clear where the documentation lives
✅ Allows easy editing of docs via the PR
✅ Merges the .com site with the live github repo
✅ Doesn't make docs live on an entirely separate timeline from code

I actually like this solution better, because code & doc changes can be merged at the same time in one PR. Wasn't aware the option existed, so I learned a cool thing too. 👍 👍 

I think the steps will be something along the lines of:

* Merge this PR
* Delete CNAME from the other repo (or delete the repo entirely)
* Add the CNAME file to this repo
* Change the github pages settings to `master /docs` as @vossad01 specified.

I think at this point, we may actually be OK. The CNAME settings are the same for your DNS, it will just be a matter of github picking up on the CNAME file changes. After that, github will take over serving from that repo, and therefore I don't think we'll need to adjust anything (not 100% sure of this, of course).

I haven't reviewed the PR for content / code, but process-wise & idea-wise, this seems solid.

Sorry again for the delay! Let me know anything I can do to help

dennisdoomen:
@vossad01 does it always take the *master* branch as its source, even though *develop* is currently the default?

SeanKilleen:
@dennisdoomen it may use the default branch. What does your GitHub pages section of the repo settings look like? The drop down may say "develop/docs" instead.

vossad01:
Somewhat surprisingly, given most of GitHub seems to defer to the default branch rather than _master_, yes.  The [docs](https://help.github.com/articles/configuring-a-publishing-source-for-github-pages/) make no mention of the default branch.  I tried it on my repo, changing the default branch and even going as far as deleting _master_, and indeed the option only becomes grayed out rather than offering to use the default branch.

SeanKilleen:
Bummer. @dennisdoomen, at what point do you normally merge to master? Would it ever be a problem to hold up docs for that? Doesn't seem like it's be a huge impact, IMO. Any crucial docs updates that couldn't wait for master could be cherry-picked into it, perhaps.

dennisdoomen:
@vossad01 I'm fine with merging to *master*. I'm not yet close to finalizing 5.0 (which is on *develop*) anyway.

vossad01:
@dennisdoomen I think this PR is ready when you are.  If the _master_/_develop_ things becomes an issue again in the future, an option would be to use a little different branching strategy such as _stable_/_master_ so the latest bits can still provide the documentation but the last stable HEAD is just as accessible.

I have added updating PULL_REQUEST_TEMPLATE.md to the list of things to do if this is merged (at the bottom of the original post).  Since that file lives in `develop`, this PR can't touch it.

dennisdoomen:
We're live!!

Thanks for all the hard work @vossad01 and @SeanKilleen)

SeanKilleen:
:+1: :+1: sweet!

dennisdoomen:
@vossad01 and @SeanKilleen, do you guys have Twitter accounts?

SeanKilleen:
Sure! I'm [@sjkilleen](http://twitter.com/sjkilleen).

vossad01:
I do not.

Issue #533, opened 2017-02-06
Logo contest for FluentAssertions


SaroTasciyan:
I wonder whether there's an official logo for _FluentAssertions_ or not? I know that the official web site recently renewed and following logo was introduced (I'm not quite sure if it's new):

![FluentAssertionsLogo](http://www.fluentassertions.com/images/logo.png)

However, I don't see it being used for _NuGet_. 
If there's no official logo and one is to be designed; Have you guys ever considered starting a logo contest as [WebAssembly](https://github.com/WebAssembly/design/issues/112) did?

dennisdoomen:
Sounds like a great idea. As soon as I've moved the repo to the new organization, I'll post a message on twitter.

IUsername:
![fluent_assertions](https://cloud.githubusercontent.com/assets/2133735/22858674/29bb0ed0-f081-11e6-9d72-be7f4ebfff08.png)

I have used Fluent Assertions for many years on numerous project. It is the least I can do to throw a design into the ring. Here are some alternative layouts of this design:

![fluent_assertions_alts](https://cloud.githubusercontent.com/assets/2133735/22858692/737278ba-f081-11e6-84cf-fb4ebd2d7d43.png)

dennisdoomen:
@IUsername wow, simply wow!

umuralpay:
![fluent2](https://cloud.githubusercontent.com/assets/747647/22872688/2c995c00-f1cc-11e6-93c6-7cfff24ee979.png)

Can be something retro, the word "fluent" gives me that mood.

bitbonk:
When  choosing logo, we should also consider what it will look like as the icon for the nuget package, Especially in the nuget package manager of visual studio. This is the place where it will be seen most,

![light theme unselected](https://cloud.githubusercontent.com/assets/418030/23340962/bffd6518-fc3f-11e6-816a-d90a8b9d5e5a.PNG)
![light theme selected](https://cloud.githubusercontent.com/assets/418030/23340968/d3cf935e-fc3f-11e6-94fd-7e7eaccbfc40.PNG)
![dark theme unselected](https://cloud.githubusercontent.com/assets/418030/23340970/e38e0cc6-fc3f-11e6-829e-88cffdf57641.PNG)
![blue theme unselected](https://cloud.githubusercontent.com/assets/418030/23340976/f3ba532a-fc3f-11e6-9025-0ff577e0773f.PNG)
[...]

PR #573, opened 2017-06-01
Merge 'master' into 'develop'

adamvoss:
Done primarily so the website/documentation exists on the develop branch so changes related to FA 5.0 can be included.


dennisdoomen:
Looks like some code got merged in as well

adamvoss:
Is **master** -> **develop** not a merge path?  I assumed it was since **master** is indicated as where bugs should be fixed.

If it is not, I could merge be4714cb940be55add2a58c1c56f5ad0025e8940 and cherry-pick latter website-only changes, but I don't typically like doing that to history.  Perhaps it would be better to close this and wait until **develop** becomes **master**.  I was mainly just looking to make it so **develop** would be set with the new logo per https://github.com/fluentassertions/fluentassertions/issues/533#issuecomment-305571836.

dennisdoomen:
It is, so I don't get why it doesn't compile.

adamvoss:
I didn't at first either.  Turns out the reason was above all the red text: some added unit tests needed to be adapted for the XUnit switch.

PR #575, opened 2017-06-02
Attribution for new logo

adamvoss:
- logo attribution on website
- logo attribution in history

**Special request:** Please push this to **develop** (it will fast-forward) rather than merging it via GitHub UI (which will create a merge commit).  That will put these new commits on the left-hand side of history (accessible from HEAD following first-parent links) which for most implementations will cause 60c3652ade666eab322ce672e70252d1e32b0746 to be chosen before 6a3a3f1b5fb047779240bfffac74b039c933ed74 in things like `git-blame`


adamvoss:
Thanks!

Issue #594, opened 2017-07-22
Support for NotBeAssignableTo


jnyrup:
There does not seem to be support for the negation of `BeAssignableTo<T>`.

I propose addition of `NotBeAssignableTo<T>()`.

If the proposal is welcomed, I'll be happy to submit a PR once 5.0 is merged into master (unless somebody beats me to it).

dennisdoomen:
Well, we need to merge *master* to *develop* anyway, so go ahead with that PR on *master.

dennisdoomen:
Scratch that. If you can wait a couple of weeks and are fine with a temporary 5.0 version, you can target the PR at *develop*

jnyrup:
I already have some implementations and tests ready for my own use, so I'm in no hurry.
I would just like to avoid having to rebase PRs after a large merge. 
So whatever causes me the fewest conflicts I'm happy with.

By "wait a couple of weeks " do then mean:
* "wait a couple of weeks before branching out", 
* "branch out now and wait a couple of weeks extra before it will be released",
* or something else?

dennisdoomen:
"branch out now and wait a couple of weeks extra before it will be released",

jnyrup:
closed in #601

PR #601, opened 2017-07-23
Not be assignable to

jnyrup:

I would say it's done, but I'm unsure about one thing.
Like `BeAssignableTo<T>` returns an `AndWhichConstraint` where `Which` points to casted type, I wanted `NotBeAssignableTo` to return the Subject as an `AndConstraint` for chaining. 
But as the type is lost in `ObjectAssertions Should(this object actualValue)` `NotAssignableTo<T>().And` is of type `object``.

This gives some ugly code, where one needs to cast `someObject` back to `Exception` for chaining.
```csharp`
var someObject = new Exception("Actual Message");

Action act = () => someObject.Should().NotBeAssignableTo<DummyImplementingClass>()
                .And.Subject.Should().BeOfType<Exception>()
                .Which.Message.Should().Be("Other Message");
``` 

Is there a way to retain the type of `someObject`,
 or would it just be so confusing that it should just be a void function?


dennisdoomen:
I wouldn't worry about that. I simply can't define an assertion extension on `T`, because it would override all other extensions.

jnyrup:
Good to hear, I just needed a second opinion on this.
Then it should be all good to go.

Yo GitHub, this closes #594

PR #608, opened 2017-07-30
Report all indices offending NotContainNulls

jnyrup:
When asserting using `NotContainNulls` the current behavior is to only report the first index of a null item.
This PR changes that to report indices of all null items instead.

### DIFF
--- a/Src/FluentAssertions/Collections/CollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/CollectionAssertions.cs
@@ -162,14 +162,25 @@ namespace FluentAssertions.Collections
                     .FailWith("Expected {context:collection} not to contain nulls{reason}, but collection is <null>.");
             }
 
-            object[] values = Subject.Cast<object>().ToArray();
-            for (int index = 0; index < values.Length; index++)
+            IEnumerable<int> indices = Subject
+                .Cast<object>()
+                .Select((item, index) => new { Item = item, Index = index })
+                .Where(e => ReferenceEquals(e.Item, null))
+                .Select(e => e.Index);
+
+            if (indices.Any())
             {
-                if (ReferenceEquals(values[index], null))
+                if (indices.Count() > 1)
                 {
                     Execute.Assertion
                         .BecauseOf(because, becauseArgs)
-                        .FailWith("Expected {context:collection} not to contain nulls{reason}, but found one at index {0}.", index);
+                        .FailWith("Expected {context:collection} not to contain nulls{reason}, but found several at indices {0}.", indices);
+                }
+                else
+                {
+                    Execute.Assertion
+                        .BecauseOf(because, becauseArgs)
+                        .FailWith("Expected {context:collection} not to contain nulls{reason}, but found one at index {0}.", indices.First());
                 }
             }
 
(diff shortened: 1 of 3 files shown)


Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -162,14 +162,25 @@ public AndConstraint<TAssertions> NotContainNulls(string because = "", params ob
...
+            IEnumerable<int> indices = Subject
+                .Cast<object>()
+                .Select((item, index) => new { Item = item, Index = index })
+                .Where(e => ReferenceEquals(e.Item, null))
+                .Select(e => e.Index);
+
+            if (indices.Any())
             {
-                if (ReferenceEquals(values[index], null))
+                if (indices.Count() > 1)
```
dennisdoomen: This will cause multiple iterations of the LINQ statement returned by `indexes`.

Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -162,14 +162,25 @@ public AndConstraint<TAssertions> NotContainNulls(string because = "", params ob
...
+                {
+                    Execute.Assertion
+                        .BecauseOf(because, becauseArgs)
+                        .FailWith("Expected {context:collection} not to contain nulls{reason}, but found several at indices {0}.", indices);
```
dennisdoomen: Maybe spell _nulls_ as `<null>s` or something like that? This is how a _null_ is normally rendered in FA.

Tests/Shared.Specs/CollectionAssertionSpecs.cs
```diff
@@ -2329,6 +2329,26 @@ public void When_collection_contains_nulls_that_are_unexpected_it_should_throw()
...
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => collection.Should().NotContainNulls("because they are {0}", "evil");
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act.ShouldThrow<XunitException>().WithMessage(
+                "Expected collection not to contain nulls because they are evil, but found several at indices {1, 3}.");
```
dennisdoomen: I tend to use wildcards (`*`) to avoid asserting the exact message.

Src/FluentAssertions/Collections/CollectionAssertions.cs
jnyrup: There are 15 places where we do the check `Count() > 1` only to know it a sequence has multiple items. (About 8 of them is just using `Count()` on an array, so they will use the `Length` property)

Would you prefer materializing the `IEnumerable` in an array or make an extension method like `HasMultipleItem()`?
That extension method could either be as simple as `list.Skip(1).Any()` or more sophisticated like
https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq/CountMethods.cs

dennisdoomen: We had some performance reports for some specific methods. However, this specific one will not suffer from that. But a simple `ToArray` would be fine.

Issue #585, opened 2017-07-10
Best practices


jnyrup:
When I converted our test suite to Fluent Assertions I of course used the documentation a lot, but several times a had an assertion which it was not immediately clear to me how to convert.
So I think it would be nice (in addition to the documentation) to have a wiki or guidelines, i.e. tips on how to:
* convert common assertions into fluent assertions
* rewrite assertions to improve readability
* rewrite of assertions to improve error messages (e.g CollectionAssertions instead of foreach + assertion)

I have made a list of many of the conversions I made.
Ideally those could be (wishful thinking) converted into an analyzer like https://github.com/xunit/xunit.analyzers

```
Assert.IsTrue(b) => b.Should().BeTrue();
Assert.IsFalse(b) => b.Should().BeFalse();
Assert.IsNull(o) => o.Should().BeNull()
Assert.IsNotNull(o) => o.Should().NotBeNull()
Assert.AreEqual(b, a) => a.Should().Be(b)
Assert.AreEqual(b, a, delta) => a.Should().BeApproximately(b, delta)
Assert.AreNotEqual(other, obj) => obj.Should().NotBe(other)
Assert.AreSame(other, obj) => obj.Should().BeSameAs(other)
Assert.AreNotSame(b, a) => a.Should().NotBeSameAs(b)
Assert.IsInstanceOfType(a, typeof(T)) => a.Should().BeOfType<T>()
Assert.IsNotInstanceOfType(a, typeof(T)) => a.Should().NotBeOfType<T>()
Assert.IsTrue(a == b) => a.Should().Be(b)
Assert.IsFalse(a == b) => a.Should().NotBe(b)
Assert.IsTrue(a != b) => a.Should().NotBe(b)
Assert.IsFalse(a != b) => a.Should().Be(b)
Assert.IsTrue(a > b) => a.Should().BeGreaterThan(b)
Assert.IsFalse(a > b) => a.Should().BeLessOrEqualTo(b)
Assert.IsTrue(a >= b) => a.Should().BeGreaterOrEqualTo(b)
Assert.IsFalse(a >= b) => a.Should().BeLessThan(b)
Assert.IsTrue(a < b) => a.Should().BeLessThan(b)
Assert.IsFalse(a < b) => a.Should().BeGreaterOrEqualTo(b)
Assert.IsTrue(a <= b) => a.Should().BeLessOrEqualTo(b)
Assert.IsFalse(a <= b) => a.Should().BeGreaterThan(b)

[...]

PR #614, opened 2017-08-05
Migration from MSTest and improved assertions

jnyrup:
As promised in #585 here is a list of 104 examples about:
* Converting from MSTest to Fluent Assertions.
* Rewriting assertions for clearer intentions and better failure messages.

Comments on the preface and the layout are very welcome.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -7,10 +7,11 @@ github_username:  fluentassertions
 google_analytics: UA-99995670-1
 
 header_pages:
-  - about.md
-  - documentation.md
   - examples.md
+  - documentation.md
+  - tips.md
   - releases.md
+  - about.md
 
 theme: minima
 gems:
--- a/docs/_data/mstest-migration/assert.yml
+++ b/docs/_data/mstest-migration/assert.yml
@@ -1,0 +1,330 @@
+- old: |
+    Assert.IsTrue(actual);
+
+  new: |
+    actual.Should().BeTrue();
+
+  old-message: |
+    Assert.IsTrue failed.
+
+  new-message: |
+    Expected True, but found False.
+
+- old: |
+    Assert.IsFalse(actual);
+
+  new: |
+    actual.Should().BeFalse();
+
+  old-message: |
+    Assert.IsFalse failed.
+
+  new-message: |
+    Expected False, but found True.
+
+- old: |
+    Assert.IsNull(actual);
+
+  new: |
+    actual.Should().BeNull();
+
+  old-message: |
+    Assert.IsNull failed.
+
+  new-message: |
+    Expected object to be <null>, but found System.Object (HashCode=51904525).
+
+- old: |
+    Assert.IsNotNull(actual);
+
+  new: |
+    actual.Should().NotBeNull();
+
+  old-message: |
+    Assert.IsNotNull failed.
+
+  new-message: |
+    Expected object not to be <null>.
+
+- old: |
+    Assert.AreEqual(expected, actual);
+
+  new: |
+    actual.Should().Be(expected);
+
+  old-message: |
+    Assert.AreEqual failed. Expected:<SomeProperty: 2, OtherProperty: expected>. Actual:<SomeProperty: 1, OtherProperty: actual>.
+
+  new-message: |
+    Expected object to be SomeProperty: 2, OtherProperty: expected, but found SomeProperty: 1, OtherProperty: actual.
+
+- old: |
+    Assert.AreEqual(expected, actual, delta);
+
+  new: |
+    actual.Should().BeApproximately(expected, delta);
+
+  old-message: |
+    Assert.AreEqual failed. Expected a difference no greater than <0.5> between expected value <2> and actual value <1.25>.
+
+  new-message: |
+    Expected value 1.25 to approximate 2.0 +/- 0.5, but it differed by 0.75.
+
+- old: |
+    Assert.AreNotEqual(expected, actual);
+
+  new: |
+    actual.Should().NotBe(expected);
+
+  old-message: |
+    Assert.AreNotEqual failed. Expected any value except:<SomeProperty: 1, OtherProperty: expected>. Actual:<SomeProperty: 1, OtherProperty: expected>.
+
+  new-message: |
+    Did not expect object to be equal to SomeProperty: 1, OtherProperty: expected.
+
+- old: |
+    Assert.AreNotEqual(expected, actual, delta);
+
+  new: |
+    // Will be available in Fluent Assertions 5.0
+    actual.Should().NotBeApproximately(expected, delta);
+
+  old-message: |
+    Assert.AreNotEqual failed. Expected a difference greater than <0.5> between expected value <2> and actual value <2>.
+
+  new-message: |
+    Expected value 2.0 to not approximate 2.0 +/- 0.5, but it only differed by 0.0.
+
+- old: |
+    Assert.AreSame(expected, actual);
+
+  new: |
+    actual.Should().BeSameAs(expected);
+
+  old-message: |
+    Assert.AreSame failed.
+
+  new-message: |
+    Expected object to refer to 
+    SomeProperty: 1, OtherProperty: actual, but found 
+    SomeProperty: 1, OtherProperty: actual.
+
+- old: |
+    Assert.AreNotSame(expected, actual);
+
+  new: |
+    actual.Should().NotBeSameAs(expected);
+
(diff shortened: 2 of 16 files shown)


dennisdoomen:
Thanks for the great work. 

Isn't this just about migrating from MSTest?

Where do we want to display this? As a header on the landing page?

jnyrup:
The first 44 examples, are about migrating from MSTest.
The last 60 examples are on rewriting FA assertions.

Yes, I thought next to documentation or examples.

dennisdoomen:
I see now. So what's next? I see it's still WIP.

jnyrup:
I have implemented failing tests for all examples, and thought about including the failure message for all examples to give proof how the rewrites improves.

Meir017:
are there plan for writing Analyzers for the rewrites?

jnyrup:
@Meir017 _I_ don't have any concrete plans.
But only because I don't have any experience in writing analyzers.

dennisdoomen:
I'll keep ignoring this PR as long as it's WIP. Tell me if this is no longer the case.

jnyrup:
I still want to add the failure messages to each example, so it's still WIP.

@SaroTasciyan, you mentioned some examples you wanted to add.

SaroTasciyan:
Almost all of the items in my list (_NUnit_) is covered. Missing ones are assertions regarding; string _equivalency_, DateTime and some others which might not have an equivalent in MSTest. If that's the case, I don't see the value it would add since they are already covered in the documentation.

Regarding the _best practices_ part; It can be extended (in time) with _chained assertions_ (And, Which), `ShouldBeEquivalentTo`, string equivalency / wildcard and `FleuntDateTimeExtensions` demonstration.

jnyrup:
They don't need to have a MSTest equivalent, as long as there is a "wrong" and a "right" way to write it.

When you're new to FA, even after reading the documentation (for the first time), you might not be familiar with how to write idiomatic assertions.
These docs give examples, where you can search by fragments of your smelly assertion to find better alternatives.

SaroTasciyan:
In my opinion, _they_ don't need to go into this part.
> Converting from MSTest to Fluent Assertions.

That's why I listed them in this part (which I assume is _best practices_)
> Rewriting assertions for clearer intentions and better failure messages.

docs/BestPractices.md
```diff
@@ -0,0 +1,1820 @@
...
+<td>
+CollectionAssert.AllItemsAreInstancesOfType failed. Element at index 0 is not of expected type. Expected type:<UnitTests2.MyIdenticalClass>. Actual type:<UnitTests2.MyClass>.
+</td>
+<td>
+Expected collection to contain only items of type UnitTests2.MyIdenticalClass, but item SomeProperty: 1, OtherProperty: item at index 0 is of type UnitTests2.MyClass.
+</td>
+</tr></table>
+
+
+## [ExpectedException]
```
AbhitejJohn: Might be worth adding the Assert.ThrowsException equivalent as well, that is if this also applies for MSTest V2.

jnyrup: Didn't even know MSTest V2 had that assertion.
I think it should be included.
It also great advertisement why to upgrade to MSTest V2

In fact, this is my favorite example, as the MSTest V1 way is quite ugly.

AbhitejJohn: Yes it is. Here is the documentation for MSTest V2: https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting. 
Just putting this out here as well: https://github.com/Microsoft/testfx-docs/blob/master/RFCs/002-Framework-Extensibility-Custom-Assertions.md

jnyrup:
I have finished adding the examples and failure messages I had planned.
The remaining part would be layout.
I'm not familiar with how GitHub markdown turns out as html, but when looking at the .md file, the are multiple things I'm unsatisfied with:
* The margins, that you have to scroll horizontally.
* That snippets are not top+left aligned.
* Missing whitespace conservation.
* Better syntax highlighting?
* A Table of Contents.

Layout is really not my strong side, so could use ideas/help, if this should become prettier.
@adamvoss You seem to be an expert?

adamvoss:
@jnyrup I am really sorry I keep letting this slip through the cracks.  I will try looking into this tomorrow, and if not then for sure this coming week.

adamvoss:
I lied, I just took a quick peak now.  It looks like there is an error in the HTML because I get a nesting table, but didn't take time to track it down.

Try this for inspiration (its not great but should give some ideas on things you can do):

```
<table>
<thead>
<tr>
    <th>
        NUnit
    </th>
    <th>
        FluentAssertions
    </th>
</tr>
<tr>
    <th>
        Failure Message
    </th>
    <th>
        Failure Message
    </th>
</tr>
</thead>
<tbody>
<tr>
<td markdown="1">
```c#
Assert.IsTrue(actual);
```
</td>
<td markdown="1">
```c#
actual.Should().BeTrue();
```
</td>
</tr><tr>
<td markdown="1">
> Assert.IsTrue failed.
</td>
<td markdown="1">
> Expected True, but found False.
</td>
</tr>
</tbody></table>
```

Some of the problems will need to be fixed via CSS.  Regarding the page itself, if wanted to structure data into CSVs, I can help you use Liquid templates to build populate a table and avoid the sort of issue it has now with an error somewhere in the table.

TOC can be done automatically too (see the documentation page):

```
## Table of Contents ##
* TOC
{:toc}
```` 

Good work!

**Comment by jnyrup** [comment]
Thanks for the tips @adamvoss that helped *a lot*
The pages are now generated with Jekyll using Liquid templates to fetch the data from yaml files.

As there to my knowledge isn't a way to see the generated page here without pushing to master, here are two screenshots:
untitled

untitled2

**Comment by adamvoss** [comment]
Wow, I ran it locally and that was a great step!

The next part is really out of my area of expertise.  Does anyone reading this have design or CSS experience?

Someone who really knows this stuff could probably do some responsive design stuff that makes it flow well onto mobile too.

If it is just me trying to figure this out, my approach would probably be:
 - Make the tables a fixed width (based on the current SASS variable for content width)
 - Make the messages word-wrap when they would be too long for the box
 - Manually add line breaks in the assertions multi-line if they would be too long on their own.

File: docs/BestPractices.md [hunk 2]
```diff
@@ -3,6 +3,8 @@ title: "Best Practices"
 layout: page
 ---
 
+<link href="/assets/css/syntax.css" rel="stylesheet" >
```
adamvoss: What is different or better in this syntax highlighting over the existing highlighting in this project?

adamvoss:
I have proposed two pull request to this which hopefully will bring it close to completion:

https://github.com/jnyrup/fluentassertions/pull/1
https://github.com/jnyrup/fluentassertions/pull/2

adamvoss:
I have made some final suggestions at https://github.com/jnyrup/fluentassertions/pull/3.  Once those changes are decided on I'd recommend squashing the history. (I don't know if you have strong opinions on squashing or if @dennisdoomen will require it, this just looks like history that is better squashed to me).  Then I'd recommend @dennisdoomen review it for possible inclusion.

My only nit-pick would be that the horizontal scroll bar on code blocks should never appear when the windows is maximized (on a sufficiently large screen).  But, since all the examples are "close enough" I am not motivated enough by the issue to recommend blocking on that issue.

jnyrup:
@adamvoss GitHub has a squash-and-merge function, so no worries.

adamvoss:
> GitHub has a squash-and-merge function, so no worries.

It is generally not a big deal and I cannot speak for Dennis on this front, but I can give the perspective from being as a package maintainer elsewhere.  While GitHub added the squash option, when contributors expect the maintainer to use it shifts the responsibility for writing a good commit message to the maintainer.  Instead of everything presented as-desired so all you have to do is confirm its good and click the merge button, you have to stop and think up a good message.

There is plenty of room on GitHub for different usage patterns, so choosing not to squash is still a correct/valid decision.

dennisdoomen:
I prefer properly squashed and documented commits myself. This is what I require in my professional projecs and which I practice myself. So if somebody offers me a PR with three commits, I assume this person carefully squashed his work into these three commits and will merge them as is. If the PR only includes a single commit, I'll squash-merge it.

jnyrup:
Thanks for the clarifications both of you, it makes perfect sense.
My laziness should not shift the burden onto anyone else.

I don't have any outstanding issues left, so from my perspective it's done.

dennisdoomen:
Well, then somebody should remove the WIP prefix.

PR #620, opened 2017-08-15
Add missing/negations type assertions/selectors

jnyrup:
Adds lot `Not` Assertions/Selectors for types/methods/properties/.

The deletions are:
* Renaming of existing tests to fix ambiguity conflict with new tests of Not variant.
* double white space typo
* typo in summary for ThatAreUnderNamespace


dennisdoomen:
Awesome work @jnyrup. How can I contact you?

jnyrup:
If you mean by e.g email I have a gmail. It's the same as my github username.

Src/FluentAssertions/Types/MethodBaseAssertions.cs
```diff
@@ -37,6 +37,28 @@ public abstract class MethodBaseAssertions<TSubject, TAssertions> : MemberInfoAs
...
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> NotHaveAccessModifier(
+            CSharpAccessModifier accessModifier, string because = "", params object[] becauseArgs)
+        {
+            Execute.Assertion.ForCondition(accessModifier != Subject.GetCSharpAccessModifier())
```
dennisdoomen: Purely for readability, I prefer the `ForCondition` on the next line

Src/FluentAssertions/Types/MethodInfoAssertions.cs
```diff
@@ -44,6 +44,31 @@ public MethodInfoAssertions(MethodInfo methodInfo)
...
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<MethodInfoAssertions> NotBeVirtual(
+            string because = "", params object[] becauseArgs)
+        {
+            string failureMessage = "Expected method " +
```
dennisdoomen: Why not put this on a single line, preferably using string interpolation? Same for similar uses elsewhere in this PR.

Src/FluentAssertions/Types/MethodInfoAssertions.cs
```diff
@@ -64,6 +89,26 @@ public AndConstraint<MethodInfoAssertions> BeAsync(string because = "", params o
...
 
         /// <summary>
+        /// Asserts that the selected method is not async. 
+        /// </summary>
+        /// <param name="because">A formatted phrase as is supported by <see cref="M:System.String.Format(System.String,System.Object[])"/> explaining why the assertion
+        ///             is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.</param>
+        /// <param name="becauseArgs">Zero or more objects to format using the placeholders in <see cref="!:because"/>.</param>
+        public AndConstraint<MethodInfoAssertions> NotBeAsync(string because = "", params object[] becauseArgs)
+        {
+            string failureMessage = "Expected subject " +
```
dennisdoomen: Should probably say *method* rather than *subject*

jnyrup:
While I feel the expression "expected something to not be" is more strict, I do prefer your suggestion.
Without being an expert (at all) in the English language your suggestion does sound more correct.
It also makes it easier to distinguish a `Be` and `NotBe` as the negation is in the beginning of the message instead of in the middle somewhere.

jnyrup:
Now I'm a bit unsure. 

There are around 124 other places that uses the construct _Expected something to not be something_ or _Expected something not to be something_.
E.g `Expected {context:collection} not to be empty{reason}.`, which does become weaker if changing it to 
`Did not expect {context:collection} to be empty{reason}.` Now it sounds as if we didn't explicitly asserted it not to be empty.

I quick guess is that all instances of _Expected something to not be something_ should be changed to _Expected something not to be something_?

dennisdoomen:
OK, just ignore that comment.

PR #624, opened 2017-08-19
Add netstandard2.0 TFM support

am11:
* Make XML formatter compile.
* Update couple of instances where NETSTANDARD1_X required runtime APIs
  that are now deprecated (e.g. RuntimeLibrary).
* Potentially, I tried to make the NETSTANDARD2_0 implementation close
  to NET45.

---


dennisdoomen:
Nice work. I suppose you forgot about the https://github.com/fluentassertions/fluentassertions/blob/develop/Src/FluentAssertions.nuspec

am11:
Yup, I missed the nuspec 8-)

Would it make sense to reorganize `NET45.Specs` directory/project so the name is not tied to NET45, in order to accommodate netstandard2.0+ TFMs? IMO:

* all the conditional compiles scenarios under `Src`, where the whole file is wrapped under `#if NET45` can be conditionally included/excluded from the csproj.
* NET45.Specs can be removed and we can use xunit traits to control the test executions. For instance, a decorator after a `[Fact]` suggesting `NotSupportedPlatform: netcoreapp1.1` etc.

dennisdoomen:
Well, the purpose of these dedicated `.Specs` projects is to *really* make sure all assertions work correctly for a particular platform. I've seen some weird stuff here and there that made me believe I still need them. In your case, adding a .NET Core 2.0 Specs project would make more sense.

am11:
I started taking a stab at `NetCore20.Specs` as a fork of `Net45.Specs`, but it turned out the test dependency `Chill` which are used by Net45 doesn't support .NET Standard TFMs at all.
Moreover, Autofac does support v1.6 of standard, but the more close TFM support is preferred. I tried updating autofac (https://github.com/am11/Autofac/commit/a36578eb7b5eb326ce7129ef6b311e4c946e720a), but realized the hard way that ther are certain VS15.3 Analyzers issue, which is blocking Autofac as they have reported those upstream: https://github.com/dotnet/roslyn-analyzers/issues/1286.

What that being said, should we move ahead and add the STD 2.0 XML Specs at a later point? :)

dennisdoomen:
I would rather have that fork, but wrap the stuff you can't get to work in `#if` blocks.

am11:
Added a separate NetCore20.Specs project for .NET Core 2.0: https://github.com/fluentassertions/fluentassertions/pull/624/commits/127422f178a8f8f9e37b62922aea569788d80291?w=1. Also update Net45.Specs csproj to modern format.

dennisdoomen:
image

Are these dependencies still needed for 2.0?

Please also update the summary and descriptions in the .nuspec?

am11:
> Are these dependencies still needed for 2.0?

Yes, removing any one of these cause compilation error(s). Configuration Manager was ported to netstandard2.0 and missing in 1.x versions, Reflection.Emit for `AssemblyBuilder` and Reflection.Emit.Lightweight for `DynamicMethod` usages.

jnyrup:
Would it make sense to upgrade to `NetStandard.Library 1.6.1` as well?
xunit 2.3.0-beta4 changed that when upgrading `System.Reflection.TypeExtensions`
https://github.com/xunit/xunit/compare/2.3-beta3...2.3-beta4

dennisdoomen:
> Would it make sense to upgrade to NetStandard.Library 1.6.1 as well?

Upgrade to what?

jnyrup:
Upgrading
`<dependency id="NETStandard.Library" version="1.6.0" />` 
to 
`<dependency id="NETStandard.Library" version="1.6.1" />`

am11:
Good point; I have updated the nuspec.

Tests/Net45.Specs/Net45.Specs.csproj
```diff
@@ -1,157 +1,26 @@
...
-    </ProjectReference>
-    <ProjectReference Include="..\AssemblyB\AssemblyB.csproj">
-      <Project>{d9cccaf6-ff9e-4688-9d96-645e27616ac4}</Project>
-      <Name>AssemblyB</Name>
-    </ProjectReference>
+    <PackageReference Include="Autofac" Version="4.6.1" />
+    <PackageReference Include="Chill" Version="3.0.1.0" />
+    <PackageReference Include="FakeItEasy" Version="3.4.2" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
```
jnyrup: Duplicate reference

am11: Ouch! Removed and fixed up. :)

thomaslevesque:
@dennisdoomen when do you plan to make a new release with these changes? I want to add a netstandard2.0 target to FakeItEasy, but I need `BeBinarySerializable`, which is not available in FluentAssertions for netstandard1.3...

dennisdoomen:
@thomaslevesque well, the bulk of the breaking changes has been done. There's still some of those left around event monitoring, plus a couple of [new features](https://github.com/fluentassertions/fluentassertions/milestone/10) that I need to do to finalize 5.0. However, I could create a first beta release if you're willing to take the change.

thomaslevesque:
@dennisdoomen thanks. I could use a beta to at least test my changes.

dennisdoomen:
@thomaslevesque See https://github.com/fluentassertions/fluentassertions/releases/tag/5.0.0-beta.1

thomaslevesque:
Awesome, thanks @dennisdoomen !

PR #633, opened 2017-09-02
string.ContainAllOf and string.ContainAnyOf

beeleebow:
One thing I always find myself doing is asserting the contents of a string multiple times. Usually I am wanting to assert that something like a log message contains some important information in no particular order (so as to not have an overly rigid test). For example:

```csharp
logMsg.Should().Contain(customerNumber);
logMsg.Should().Contain("success");
logMsg.Should().Contain(timestamp);
```
I thought it might be nice if I could write this instead:

```csharp
logMsg.Should().ContainAllOf(new []{ customerNumber, "success", timestamp"});
```
I also thought a similar `ContainAnyOf` could be useful, so I added it while I was there.

I updated the CONTRIBUTING.MD document as well, adding the disclaimer about the v 5 rework and suggesting PRs be targeted at develop (it said master before).

### DIFF
--- a/Src/FluentAssertions/Primitives/StringAssertions.cs
+++ b/Src/FluentAssertions/Primitives/StringAssertions.cs
@@ -669,6 +669,93 @@ namespace FluentAssertions.Primitives
         }
 
         /// <summary>
+        /// Asserts that a string contains all values present in <paramref name="values"/>.
+        /// </summary>
+        /// <param name="values">
+        /// The values that should all be present in the string
+        /// </param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<StringAssertions> ContainAll(IEnumerable<string> values, string because = "", params object[] becauseArgs)
+        {
+            ThrowIfValuesNullOrEmpty(values);
+
+            var missing = values.Where(v => !Contains(Subject, v, StringComparison.Ordinal)).ToArray();
+            Execute.Assertion
+                .ForCondition(values.All(v => Contains(Subject, v, StringComparison.Ordinal)))
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected {context:string} {0} to contain the strings: {1}{reason}.", Subject, missing);
+
+            return new AndConstraint<StringAssertions>(this);
+        }
+
+        /// <summary>
+        /// Asserts that a string contains all values present in <paramref name="values"/>.
+        /// </summary>
+        /// <param name="values">
+        /// The values that should not be present in the string
+        /// </param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<StringAssertions> ContainAll(params string[] values)
+        {
+            return ContainAll(values, because: string.Empty);
+        }
+		
+		/// <summary>
+        /// Asserts that a string contains at least one value present in <paramref name="values"/>,.
+        /// </summary>
+        /// <param name="values">
+        /// The values that should will be tested against the string
+        /// </param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<StringAssertions> ContainAny(IEnumerable<string> values, string because = "", params object[] becauseArgs)
+        {
+            ThrowIfValuesNullOrEmpty(values);
+
+            Execute.Assertion
+                .ForCondition(values.Any(v => Contains(Subject, v, StringComparison.Ordinal)))
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected {context:string} {0} to contain at least one of the strings: {1}{reason}.", Subject, values.ToArray());
+
+            return new AndConstraint<StringAssertions>(this);
+        }
+
+        /// <summary>
+        /// Asserts that a string contains at least one value present in <paramref name="values"/>,.
+        /// </summary>
+        /// <param name="values">
+        /// The values that should will be tested against the string
+        /// </param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion 
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<StringAssertions> ContainAny(params string[] values)
+        {
+            return ContainAny(values, because: string.Empty);
+        }
+
+        /// <summary>
         /// Asserts that a string does not contain another (fragment of a) string.
         /// </summary>
         /// <param name="expected">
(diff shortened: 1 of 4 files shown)


jnyrup:
Nice additions!

I would rename `ContainAllOf` to be an overload of `Contain`.
And maybe rename `ContainsAnyOf` to `ContainAny`?
By having a glimpse on the other string assertions, that would make the naming more consistent. 

To make it complete you could add the `EquivalentOf` and `Not` versions.

jnyrup:
If staying with the name `ContainAll‘ we could have the convenient  `ContainAll(params string[] values)` overload. 

On a side note:
Your example could be shortened by utilizing `.And`:
`logMsg.Should().Contain().And.Contain();`

dennisdoomen:
Ah, now I understand what you're trying to accomplish here. Did you know you can also do logMsg.Should().Match($"*{customerNumber}*success*{timestamp}*")l

SaroTasciyan:
I think `ContainAllOf()` and `ContainAnyOf()` adds value since it's functionally different than `Match()` (order independent, works with shared sub-strings).
I'd go for `params` instead of `IEnumerable<string>` tho.

beeleebow:
Okay, great. Thanks for the feedback guys. 

With regards to the `params` overloads, that was what I had in mind initially, but then saw the common `params object[] becauseArgs` pattern that was in place. 

But we could have an extra overload that takes only a `params string[]`, which probably satisfies most use cases. You just can't use a "because" with that overload.

With that said, here is what I am thinking:

**Containment**
* `ContainAll(IEnumerable<string> values, string because, params object[] becauseArgs)` and `ContainAll(params string[] values)`
*  `ContainAny(IEnumerable<string> values, string because, params object[] becauseArgs)` and `ContainAny(params string[] values)`
*  `NotContainAny(IEnumerable<string> values, string because, params object[] becauseArgs)` and `NotContainAny(params string[] values)`

We could also add `NotContainAll`, but I'm not sure how valuable that is and the presence of both "not" and "all" could cause confusion about the semantics. Happy to add it if you guys think it's worthwhile.

**Equivalency**
From what I understand the "equivalent" methods for strings ignore whitespace and casing when doing comparisons. As @jnyrup suggests, we could add the following for consistency:

* `ContainEquivalentOfAll(IEnumerable<string> values, string because, params object[] becauseArgs)` and `ContainEquivalentOfAll(params string[] values)`
* `ContainEquivalentOfAny(IEnumerable<string> values, string because, params object[] becauseArgs)` and `ContainEquivalentOfAny(params string[] values)`
* `NotContainEquivalentOfAny(IEnumerable<string> values, string because, params object[] becauseArgs)` and `NotContainEquivalentOfAny(params string[] values)`

There are similar arguments to be made about `NotContainEquivalentOfAll`.

**Matching**
Taking the consistency argument to it's exteme conclusion we could also add:
* `MatchAll(IEnumerable<string> patterns, string because, params object[] becauseArgs)` and `MatchAll(params string[] patterns)`
*  `MatchAny(IEnumerable<string> patterns, string because, params object[] becauseArgs)` and `MatchAny(params string[] patterns)`
*  `NotMatchAny(IEnumerable<string> patterns, string because, params object[] becauseArgs)` and `NotMatchAny(params string[] patterns)`

(There is also the `MatchEquivalentOf` family of assertions...)

So that all said, it seems there is a **balancing act to be struck between consistency and size/noise of the API**. You guys are probably much better judges than me of where that line is so I'm happy to take your direction here.

As @jnyrup suggests the approach of chaining assertions via `And` is already available and is perhaps more in the spirit of the fluent style. So if we want to back out of this completely, then that is fine too :)

dennisdoomen:
You'll end up with lots of extension methods. I see that as well, but on the other hand, they do add value.

jnyrup:
`NotContainAll` _can_ fill out a gap between `ContainAny` and `ContainAll`
```csharp
logMsg.Should().NotContainAll(new[] {"A", "B"}, "Logging message should contain A or B but not both.").
```
But that is only if it works as "should contain any but not all".

Strictly speaking `logMsg` is allowed to not contain _any_ of the provided values and still succeed.
Or put in code, the following would is valid according to set theory.
```
"C D E".Should().NotContainAll("A", "B")
```

I think the first case do make sense in a testing framework, while I share the concern that second can be confusing whether or not `logMsg` should contain any of the provided values.

This is similar to the case for 
```csharp
var list = new List<int>();

list.Should().OnlyContain(e => e == 0)
``` 
which fails on empty collections.
Even though an empty collection strictly speaking only has elements equal to zero.

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -669,6 +669,93 @@ public AndConstraint<StringAssertions> ContainEquivalentOf(string expected, stri
...
+        /// </param>
+        public AndConstraint<StringAssertions> ContainAll(IEnumerable<string> values, string because = "", params object[] becauseArgs)
+        {
+            ThrowIfValuesNullOrEmpty(values);
+
+            var missing = values.Where(v => !Contains(Subject, v, StringComparison.Ordinal)).ToArray();
+            Execute.Assertion
+                .ForCondition(values.All(v => Contains(Subject, v, StringComparison.Ordinal)))
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected string {0} to contain strings: {1}{reason}.", Subject, missing);
```
dennisdoomen: Use the phrase `Expected {context:string} {0} to contain....`. This allows wrapping this call in an `AssertionScope` (and an upcoming feature that will try to get the `[CallerMemberName]`)

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -1773,6 +1774,222 @@ public void When_the_actual_string_equivalent_to_the_expected_but_with_trailing_
...
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => "a".Should().ContainAll(null);
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act
+                .ShouldThrow<ArgumentException>()
+                .WithMessage("Cannot assert string containment of values in null collection");
```
dennisdoomen: I prefer to use wildcards in `WithMessage` so that only the relevant parts are verified. Also see point 10 from http://www.continuousimprover.com/2015/11/12-tips-to-write-unit-tests-that-dont.html

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -1773,6 +1774,222 @@ public void When_the_actual_string_equivalent_to_the_expected_but_with_trailing_
...
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => "a".Should().ContainAll(new string[0] { });
+            Action actWithParams = () => "a".Should().ContainAll();
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act
```
dennisdoomen: Please verify only a single condition per test case.

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -1773,6 +1774,222 @@ public void When_the_actual_string_equivalent_to_the_expected_but_with_trailing_
 
         #endregion
 
+        #region ContainAll
+
+        [Fact]
+        public void When_null_collection_containall_should_throw_argument_exception()
```
dennisdoomen: Please don't repeat the name of a method or member in the name of the test case. See point 11 of http://www.continuousimprover.com/2015/11/12-tips-to-write-unit-tests-that-dont.html

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -1773,6 +1774,222 @@ public void When_the_actual_string_equivalent_to_the_expected_but_with_trailing_
...
+            var because = Guid.NewGuid().ToString();
+            Action actWithBecause = () =>
+                "red green yellow".Should().ContainAll(new[] { "blue", "orange", "pink" }, because: because);
+            Action actWithParams = () =>
+                "red green yellow".Should().ContainAll("blue", "orange", "pink");
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            const string expectedErrorMsg = "Expected string \"red green yellow\" to contain strings: " +
```
dennisdoomen: Why don't you use interpolation strings?

beeleebow: No problems. Happy to make the changes. Do you prefer the changes be added in new commits or rolled into the original commits (i.e. rewrite git history)? Also, what C# language level is acceptable? 7? (I played it cautious with string.Format etc. because I wasn't sure what your build process etc. was)

dennisdoomen: I prefer an amend. And yes, C# 7 is perfectly fine.

beeleebow:
@dennisdoomen have performed the requested changes. Let me know if I have missed anything. Note that this PR contains only `ContainAll`, `ContainAny`, `NotContainAll`, and `NotContainAny`. I may revisit the "equivalent" and "match" versions at a later date, in their own PRs.

PR #637, opened 2017-09-17
net451 tests are not discovered

jnyrup:
When running `build.ps1` net451 tests are not discovered due to a typo in the path pattern.


jnyrup:
btw, is `MSTestV2.Net45.Specs` supposed to be run when executing "Run-Unit-Tests"?

dennisdoomen:
Yes, it should. However, I'm afraid I accidentially deleted MSTest support when I switched the code-base to XUnit

Issue #636, opened 2017-09-15
Change Should(Not)Throw() into Should().(Not)Throw


dennisdoomen:
To align the recent changes to `ShouldBeEquivalenTo` and `ShouldRaise`, for consistency reason, I'd like to change the exception assertions to also change into a method that is fluently following `Should()`.

dennisdoomen:
This is now done. Thanks @omaraloraini.

PR #648, opened 2017-10-02
Should().(Not)Throw #636

omaraloraini:
#636


dennisdoomen:
Wow, thanks @omaraloraini. 

@jnyrup I hope you were not working on the same thing...

NoofSaeidh:
Hi guys. I tried to update my references and I was terrified about errors in my solutions.
Why are don't use [Obsolete] attribute instead of removing public methods?
You just bring much pain to your vendors. :(

dennisdoomen:
Hi @NoofSaeidh. I understand your concerns, but that's why we've been postponing these changes for a while. Please read https://www.continuousimprover.com/2018/02/fluent-assertions-50-best-unit-test.html for a better understanding of the rationale of those.

Issue #433, opened 2016-07-05
ExceptionAssertions.Where does not apply to InnerException


lukas-ais:
Given is the following test code:

``` csharp
  [Fact]
  public void Test()
  {
    var innerException = new InvalidOperationException("InnerMessage");
    var exception = new ArgumentException("OuterMessage", innerException);
    exception.Invoking(e => { throw e; })
      .ShouldThrowExactly<ArgumentException>()
      .Where(e => e.Message == exception.Message)
      .WithInnerExceptionExactly<InvalidOperationException>()
      .Where(i => i.Message == innerException.Message);
  }
```

The test will fail because the last line does not match.

I assume, that this is not the intended use for the Where method.
But this looks fine for a fluent API.

How to check more details on inner exception than inner message (which is available using WithInnerMessage)? I would like to check other properties of a custom exception.

dennisdoomen:
What does the `Where` return? The `InvalidOperationException` or the `ArgumentException`?

lukas-ais:
Sorry for the strong delay.
I missed the response notification and then it when out of scope... :|

The problem is, that the `Where` method applies again to the `ArgumentException` instead of the `InvalidOperationException` which is expected according the fluent API.
Is there any other method available to check the details of the **inner exception**?

Here is the output of the test execution:

> FluentAssertions.Execution.AssertionFailedException : Expected exception where (i.Message == value([...schnipp...]Tests+<>c__DisplayClass1_0).innerException.Message), but the condition was not met by:
> 
> {System.ArgumentException with message "OuterMessage"
>      bei [...schnipp...]Tests.<>c.<Test>b__1_0(ArgumentException e) in [...schnipp...]Tests.cs:Zeile 37.
>      bei FluentAssertions.AssertionExtensions.<>c__DisplayClass10_0`1.<Invoking>b__0() in C:\projects\fluentassertions-vf06b\Src\Shared\AssertionExtensions.cs:Zeile 38.
>      bei FluentAssertions.Specialized.ActionAssertions.InvokeSubjectWithInterception() in C:\projects\fluentassertions-vf06b\Src\Core\Specialized\ActionAssertions.cs:Zeile 112.
> }

dennisdoomen:
> This breaks the logic of When_asserting_with_an_aggregate_exception_type_the_asserts_should_occur_against_the_aggregate_exception
> 
> new AggregateException("Outer Message", new Exception("Inner Message"))
> 
[...]

PR #656, opened 2017-10-07
Chaining `Where` on `WithInnerExceptionExactly` should work on the correct type

dennisdoomen:


jnyrup:
This breaks the logic of `When_asserting_with_an_aggregate_exception_type_the_asserts_should_occur_against_the_aggregate_exception`

```c#
new AggregateException("Outer Message", new Exception("Inner Message"))

act.Should().Throw<AggregateException>()
                .WithMessage("Outer Message*")
                .WithInnerException<Exception>()
                .WithInnerMessage("Inner Message");
```

With this change, the test above fails when switching from `WithInnerException` to `WithInnerExceptionExactly`, as `WithInnerMessage` now asserts on the `Exception` instead of `AggregateException`.

dennisdoomen:
Hmm, very good point.

dennisdoomen:
Continue discussion in #433

PR #659, opened 2017-10-07
Use method instead of property to avoid mistakes

jnyrup:
The `BestResults` property violates the [MS guidelines](https://msdn.microsoft.com/en-us/library/ms229054%28v=vs.100%29.aspx) about when to prefer methods over properties.

> Do use a method, rather than a property, in the following situations.`
> The operation returns an array.

Probably because the use of a property made it look like a cheap operation, it was called twice, causing double enumeration and array allocation.

This is almost identical to the second example in the link above.

### DIFF
--- a/Src/FluentAssertions/Equivalency/AssertionResultSet.cs
+++ b/Src/FluentAssertions/Equivalency/AssertionResultSet.cs
@@ -30,25 +30,27 @@ namespace FluentAssertions.Equivalency
         /// </remarks>
         public string[] SelectClosestMatchFor(object key = null)
         {
-            if (!ContainsSuccessfulSet)
+            if (ContainsSuccessfulSet)
             {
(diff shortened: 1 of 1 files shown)


Src/FluentAssertions/Equivalency/AssertionResultSet.cs
```diff
@@ -30,25 +30,27 @@ public void AddSet(object key, string[] failures)
...
+            {
+                return bestResultSets[0].Value;
             }
 
-            return new string[0];
+            return bestMatch.Value;
         }
 
-        private KeyValuePair<object, string[]>[] BestResultSets
+        private KeyValuePair<object, string[]>[] BestResultSets()
```
dennisdoomen: If this is turned into a method, it probably should be named `GetBestResultSets` or something alike

PR #660, opened 2017-10-07
Use methods instead of mutating properties

jnyrup:
`ThatArePublicOrInternal`, `ThatReturnVoid` and `ThatDoNotReturnVoid` are properties even though they have observable side-effects.

Changing them to methods also aligns them with all other type/method/property selectors


dennisdoomen:
Hmm, this would cause another breaking change. Not sure if it is worth it.

dennisdoomen:
Decided to keep the existing API for this. The arguments do not outweigh the breaking changes.

jnyrup:
Fair enough.
Alternatively the side effects can be eliminated by returning a new selector object instead of modifying itself.

PR #669, opened 2017-10-09
Add support for microseconds and nanoseconds

ChristopherHaws:
I didn't realize that you were only accepting bug fix PR's when I did this. I'm opening the PR anyways since I have already done the work. If you decide not to take it then oh well... In the process of adding support for microseconds and nanoseconds, I did find some bugs with the formatters and fixed them. I also added unit tests for everything I touched.

- Fixed a visibility issue with DateTimeOffsetFormatter and TimeSpanFormatter. Previously, nanoseconds and microseconds were not visible in the assertion messages, which made it difficult to find issues. You would end up having to debug the code or add additional logging to get this visibility.
- Fixed an issue in TimeSpanFormatter where microseconds only showed up if it was the highest precision.
- To make working with microseconds and nanoseconds consistent, I added fluent extensions for them, as well as unit tests.
- I added an editorconfig file. Feel free to remove/modify the file as you see fit. My personal preference is tabs, however rather than changing all of my VS settings everytime I opened the project, I added the editorconfig file so that I know I am following the projects standards.


dennisdoomen:
No worries. We do accept improvements on the 5.0 release branch.

jnyrup:
@ChristopherHaws Any particular reason why `Nanosecond` and `Millisecond` are named in singular?
This seems to be inconsistent with e.g. `Seconds`, `Milliseconds`, etc.
I tried renaming them to plural and all tests still passed.

ChristopherHaws:
@jnyrup Just an oversight for the `TimeSpan` extensions. .NET is pretty inconsistent in this regard. `DateTime` and `DateTimeOffset` use singular tense and `TimeSpan` uses plural. After working on this PR, I proposed some new API's in .NET to make this out of the box functionality: https://github.com/dotnet/corefx/issues/24555.

jnyrup:
@ChristopherHaws Thanks for the clarification 👍 
I didn't even consider checking up on the `DateTime` and `TimeSpan` API.
I created #703 to fix the extension methods for `TimeSpan`s.

jnyrup:
@ChristopherHaws Do you recall why you added the `== 0` guards?

ChristopherHaws:
@jnyrup I do not recall. I'm guessing it was because if it is equal to zero, you don't need to actually do anything, so I just return the current value instead of adding 0 to the current value.

jnyrup:
Thanks, I also couldn't see it had anything to do with correctness.

Issue #674, opened 2017-10-12
DateTime extensions collision with NodaTime.Testing


bcowdery:
I'm running into a problem using FluentAssertions in conjuction with the NodaTime.Testing library. NodaTime.Testing provides a 'FakeClock' and other testing utilities useful in time based applications. Both libraries add extension methods to allow fluent construction of dates. For example,

```csharp
var date = 2.April(2017);
```

Attempting to use FluentAssertions and Noda testing results in the following error:

> The call is ambiguous between the following methods or properties: 'FluentAssertions.FluentDateTimeExtensions.April(int, int)' and 'NodaTime.Testing.Extensions.LocalDateConstruction.April(int, int)'

The issue is that the fluent assertion extension methods reside in the base "FluentAssertions" namespace and cannot be excluded. There is no way for me to separate the fluent style assertions from the extension methods and other syntax sugar. 

NodaTime's extension methods at least reside in a "NodaTime.Testing.Extensions" namespace allowing users to selectively use the extension methods. Without a similar pattern in fluent assertions there is no option for users to use both libraries in their test suites.

Is it possible to move fluent assertions extension methods into a similar "Extensions" namespace or would this cause too much disruption in the community? As it is now, there is no convention in CSharp for selecting one set of extension methods over another (explicit using statements won't help here). Any collision of extension methods will force users to select one library over the other.

dennisdoomen:
Yeah, we can introduce some breaking changes for 5.0

bcowdery:
Sounds good.

Would you like me to create a pull request for this change?

dennisdoomen:
Absolutely. If you target the `release-5.0` branch, it'll become part of the next beta.

dennisdoomen:
@bcowdery this is resolved by #675, right?

bcowdery:
Yes, this should be resolved.

PR #675, opened 2017-10-16
Add an Extensions namespace for helper extension methods.

bcowdery:
Fix for Issue #674, 

Moves DateTime and TimeSpan extensions into the "FluentAssertions.Extensions" namespace. I've also renamed the `TimeSpanConversionExtensions` to `FluentTimeSpanExtensions` to follow the naming conventions used in other "helper" extensions.

### DIFF
--- a/Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
+++ b/Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
@@ -19,7 +19,7 @@ namespace FluentAssertions.Extensions
     /// 12.Hours().And(30.Minutes()).
     /// </example>
     /// <seealso cref="FluentDateTimeExtensions"/>
-    public static class TimeSpanConversionExtensions
+    public static class FluentTimeSpanExtensions
     {
         /// <summary>
         /// Represents the number of ticks that are in 1 microsecond.
--- a/Src/FluentAssertions/Extensions/FluentDateTimeExtensions.cs
+++ b/Src/FluentAssertions/Extensions/FluentDateTimeExtensions.cs
@@ -1,7 +1,7 @@
 ﻿using System;
 using System.Diagnostics;
 
-namespace FluentAssertions
+namespace FluentAssertions.Extensions
 {
     /// <summary>
     /// Extension methods on <see cref="int"/> to allow for a more fluent way of specifying a <see cref="DateTime"/>.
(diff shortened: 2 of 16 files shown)


Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
```diff
@@ -65,7 +65,7 @@ public static int Nanosecond(this TimeSpan self)
         /// </remarks>
         public static TimeSpan Nanoseconds(this int nanoseconds)
         {
-            return ((long)Math.Round(nanoseconds * TicksPerNanosecond)).Ticks();
+            return ((long)Math.Round((double)(nanoseconds * TicksPerNanosecond))).Ticks();
```
jnyrup: Why this cast?

Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
```diff
@@ -78,7 +78,7 @@ public static TimeSpan Nanoseconds(this int nanoseconds)
         /// </remarks>
         public static TimeSpan Nanoseconds(this long nanoseconds)
         {
-            return ((long)Math.Round(nanoseconds * TicksPerNanosecond)).Ticks();
+            return ((long)Math.Round((double)(nanoseconds * TicksPerNanosecond))).Ticks();
```
jnyrup: Why this cast?

bcowdery: I used ReSharpers refactor tools to move the class into a new package, I think it got a bit overzealous and decided to add casts. I'll back out the change.

Sorry about that!

jnyrup: No need to feel sorry :)
I just read through the diff and noticed they were introduced, so i spun up VS to see they were marked as redundant.

bcowdery: Yep. Marked as redundant as on mine as well. Not sure why VS introduced it, I might have fat-fingered my new keyboard this morning and done it by accident or maybe ReSharper was trying to be cleaver. 

In either case, I've removed the casts and created new pull request #676. If it's easier to just make the change in the code base just kill the PR and I'll merge the change back into my fork later.

Cheers!

PR #691, opened 2017-11-17
Include generic type in failure message

### LINKED ISSUE #687: ShouldBeEquivalentTo seems to report type mismatch incorrectly
I have a following assertion:
```csharp
                auctionFromDb.ShouldBeEquivalentTo(
                    auction,
                    options => options
                        .Excluding(ctx => ctx.SelectedMemberPath.EndsWith("Id"))
                        .Excluding(ctx => 
                            Regex.IsMatch(ctx.SelectedMemberPath, @"AllowedModels\[\d+\]\.Auction"))
                        .Using<string>(ctx => ctx.Subject.Should().Be(ctx.Expectation + 1))
                        .When(info => info.SelectedMemberPath.EndsWith(".Version"))
                );
```
and since the .Version is of type int32 I get an error, but the error message says:
_Xunit.Sdk.XunitException: Expected member AuctionHouse.Brand.Version to be a System.Int32, but found a System.Int32_
When I change `.Using<string>` to` .Using<int>` the assertion passed.
I'm guessing that the error should say something about string not matching int32.

jnyrup:
This PR changes the failure message, when the type of a property is not equal or assignable to `TSubject`, to say that the property was not convertible to `TSubject`.

This is the first time I touch the object comparison code in Fluent Assertions, but I hope I haven't overlooked any (obvious) case(s).
E.g. are there any usages of `Using<TSubject>()` where `context.Expectation` or `context.Subject` are allowed to not be equal or assignable to `TSubject`?

This fixes #687

### DIFF
--- a/Src/FluentAssertions/Equivalency/AssertionRule.cs
+++ b/Src/FluentAssertions/Equivalency/AssertionRule.cs
@@ -45,15 +45,23 @@ namespace FluentAssertions.Equivalency
         {
             if (predicate(context))
             {
-                bool expectationisNull = ReferenceEquals(context.Expectation, null);
+                bool subjectIsNull = ReferenceEquals(context.Subject, null);
 
-                bool succeeded =
+                bool subjectIsValidType =
                     AssertionScope.Current
-                        .ForCondition(expectationisNull || context.Expectation.GetType().IsSameOrInherits(typeof(TSubject)))
-                        .FailWith("Expected " + context.SelectedMemberDescription + " to be a {0}{reason}, but found a {1}",
-                            !expectationisNull ? context.Expectation.GetType() : null, context.SelectedMemberInfo.MemberType);
+                        .ForCondition(subjectIsNull || context.Subject.GetType().IsSameOrInherits(typeof(TSubject)))
+                        .FailWith("Expected " + context.SelectedMemberDescription + " from subject to be a {0}{reason}, but found a {1}.",
+                            typeof(TSubject), context.Subject?.GetType());
 
-                if (succeeded)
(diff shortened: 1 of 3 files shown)


Tests/Shared.Specs/ExtensibilityRelatedEquivalencySpecs.cs
```diff
@@ -180,7 +180,100 @@ public SelectedMemberInfo Match(SelectedMemberInfo subjectMember, object expecta
         #endregion
 
         #region Assertion Rules
-        
+
+        [Fact]
+        public void when_property_of_other_is_incompatible_with_generic_type_the_message_should_include_generic_type()
```
dennisdoomen: Method should start with upper-case letters.

Src/FluentAssertions/Equivalency/AssertionRule.cs
```diff
@@ -45,20 +45,25 @@ public bool AssertEquality(IEquivalencyValidationContext context)
         {
             if (predicate(context))
             {
-                bool expectationisNull = ReferenceEquals(context.Expectation, null);
+                bool subjectSucceeded =
```
dennisdoomen: I had a hard time understanding the algorithm here. I suspect it's in the naming of things. For instance, both `subjectSucceeded` and `expectationSucceeded ` doesn't say much about what the assertion is trying to do. Same for the other changes in this PR.

jnyrup: These two commits helped me in trying to understand what was going on:
92fd4e10a7c77201ff2006b41ed1d7e01f384013
7914d73769e9bd6562022215456773eec55aebd6

PR #695, opened 2017-11-28
Fix DateTime formatting for near zero values

jnyrup:
This PR fixes formatting of `DateTime` when the time part only differs from zero by less than a second.


Tests/Shared.Specs/DateTimeAssertionSpecs.cs
```diff
@@ -258,6 +258,69 @@ public void Should_fail_when_asserting_datetime_value_is_equal_to_the_different_
                     "Expected date and time to be <2012-03-11>*failure message, but found <2012-03-10>.");
         }
 
+        [Fact]
+        public void Should_fail_when_asserting_datetime_value_is_equal_to_the_by_nanoseconds_different_value()
```
dennisdoomen: We generally try to use the `When_something_it_should_something`. Do you mind fixing that for consistency?

jnyrup:
I've updated the PR and inlined the unit test to cases of existing tests in `DateTimeOffsetValueFormatterSpecs.cs` instead.

Issue #588, opened 2017-07-21
Use the caller context / stackframe to preset the context


dennisdoomen:
* [x] Fix all unit tests
* [x] Rebase on the release branch and fix the remaining unit tests
* [x] Fix Xunit runner incompatibilities
* [x] Allow extensions to also participate gracefully in the caller context logic.

tavisca-pushkar:
@dennisdoomen Is this related to #498 ?

dennisdoomen:
@tavisca-pushkar yes, it is.

dennisdoomen:
Documentation will be updated separately. 

An important change is that this only works correctly if the unit tests are build without optimization and inlining (i.o.w. debug builds).

PR #705, opened 2017-12-09
Use the C# identifier from the calling code in the failure message instead of a generic name

dennisdoomen:
See #588 

*Order of precedence*
1. Context (for nested object graph assertion)
2. Identifier extracted from the calling source code
3. Fallback identifier specified in the `{identifier:fallback}` placeholder
4. Generic fallback provided through `WithDefaultIdentifier`


Src/FluentAssertions/CallerIdentifier.cs
```diff
@@ -69,6 +70,16 @@ private static string ExtractVariableNameFrom(StackFrame frame)
             return caller;
         }
 
+        private static bool UsesNewKeyword(string candidate)
+        {
+            return new Regex(@"new(\s?\[|\s?\{|\s\w+)").IsMatch(candidate);
```
jnyrup: Just nitpicking, but the static [`Regex.IsMatch`](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.ismatch?view=netframework-4.7.1#System_Text_RegularExpressions_Regex_IsMatch_System_String_System_String_) caches the regex.

dennisdoomen: Good point. I still have to clean up the code. Currently focusing on getting all tests green.

jnyrup:
Is `CallerIdentifier.logger` intentional or an just oversight before committing?

dennisdoomen:
I used it for diagnostic purposes and plan to keep it in in case it misbehaves. Have to think of a way to make it available as part of the tracing functionality of the `AssertionScope`.

jnyrup:
Ok. 
I mostly just noticed that there was a public static lowercase field in a public class, which was never reassigned.

dennisdoomen:
Yeah, thanks for reminding me about that one.

PR #710, opened 2017-12-20
[WIP] Feature/568 add readonly collections support

shatl:
Adding support for IReadOnlyDictionary<>.

Main issue that `IDictionary<>` does not implement `IReadOnlyDictionary<>`. 

Adding overloads of AssertionExtensions.Should() for IDictionary and IReadOnlyDictionary resulted in ambiguous reference error because dictionaries implemented in System.Collections.Generic implement both interfaces. So as work-around more overloads were added (for Dictionary, SortedList, and SortedDictionary) - I agree, this is  not an elegant solution.

Another option would be to introduce extension method  `IReadOnlyDictionary<TKey, TValue> ReadOnly<TKey, TValue>(this IDictionary<TKey, TValue> dictionary)` and have only one `Should()` method accepting IReadOnlyDictionary.

### DIFF
--- a/Src/FluentAssertions/AssertionExtensions.cs
+++ b/Src/FluentAssertions/AssertionExtensions.cs
@@ -235,50 +235,15 @@ namespace FluentAssertions
         /// current <see cref="IDictionary{TKey, TValue}"/>.
         /// </summary>
         [Pure]
-        public static GenericReadOnlyDictionaryAssertions<TKey, TValue> Should<TKey, TValue>(this Dictionary<TKey, TValue> actualValue)
-        {
-            return new GenericReadOnlyDictionaryAssertions<TKey, TValue>(actualValue);
-        }
-
-        /// <summary>
-        /// Returns an <see cref="GenericReadOnlyDictionaryAssertions{TKey, TValue}"/> object that can be used to assert the
-        /// current <see cref="SortedDictionary{TKey, TValue}"/>.
-        /// </summary>
-        [Pure]
-        public static GenericReadOnlyDictionaryAssertions<TKey, TValue> Should<TKey, TValue>(this SortedDictionary<TKey, TValue> actualValue)
-        {
-            return new GenericReadOnlyDictionaryAssertions<TKey, TValue>(actualValue);
-        }
-
-        /// <summary>
-        /// Returns an <see cref="GenericReadOnlyDictionaryAssertions{TKey, TValue}"/> object that can be used to assert the
-        /// current <see cref="SortedList{TKey, TValue}"/>.
-        /// </summary>
-        [Pure]
-        public static GenericReadOnlyDictionaryAssertions<TKey, TValue> Should<TKey, TValue>(this SortedList<TKey, TValue> actualValue)
-        {
-            return new GenericReadOnlyDictionaryAssertions<TKey, TValue>(actualValue);
-        }
-
-
-        /// <summary>
-        /// Returns an <see cref="GenericReadOnlyDictionaryAssertions{TKey, TValue}"/> object that can be used to assert the
-        /// current <see cref="IDictionary{TKey, TValue}"/>.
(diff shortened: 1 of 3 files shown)


dennisdoomen:
Awesome work @shatl.

It seems that the build is failing though.

jnyrup:
I haven't tried the code yet, but wouldn't this introduce ambiguity for existing code using a custom dictionary like this?
```csharp
public class MyDictionary<TKey, TValue> : IDictionary<TKey, TValue>, IReadOnlyDictionary<TKey, TValue>
```

shatl:
@jnyrup Unfortunately yes,  this won't compile. 
We may need to introduce either extension methods with different names (`ShouldDictionary()` and `ShouldReadOnlyictionary()`) or cast it custom dictionary to interface.

Please let me know if you have better idea.

shatl:
@dennisdoomen 
Not sure what's going on..
Test name says `DictionaryEquivalencySpecs.When_an_object_implements_two_IDictionary_interfaces_it_should_fail_descriptive` and code fails with exception. Why test checks for no exception? Am I missing something.

jnyrup:
@shatl I'm not aware of any pretty way to accomplish.

Src/FluentAssertions/AssertionExtensions.cs
```diff
@@ -230,13 +231,60 @@ public static StringCollectionAssertions Should(this IEnumerable<string> @this)
...
-        public static GenericDictionaryAssertions<TKey, TValue> Should<TKey, TValue>(this IDictionary<TKey, TValue> actualValue)
+        public static GenericReadOnlyDictionaryAssertions<TKey, TValue> Should<TKey, TValue>(this IDictionary<TKey, TValue> actualValue)
         {
-            return new GenericDictionaryAssertions<TKey, TValue>(actualValue);
+            IReadOnlyDictionary<TKey, TValue> dictionary = null;
+            if (actualValue != null)
+            {
+                dictionary = actualValue as IReadOnlyDictionary<TKey, TValue>;
+                if (dictionary == null)
+                    dictionary = new ReadOnlyDictionary<TKey, TValue>(actualValue);
```
jnyrup: Again without testing, I think creating a new object breaks code like checking for reference equality.
```csharp
dictionary.Should().BeSameAs(dictionary)
```

dennisdoomen:
Are you planning to resolve the merge conflicts?

jnyrup:
I just tried out the branch, and here are some observations:

This fails to compile as `BeSameAs` expects an `IReadOnlyDictionary` and `dict` cannot implicitly be converted from `IDictionary` to `IReadOnlyDictionary`
```csharp
IDictionary<int, int> dict = new Dictionary<int, int>();
 // dict.Should().BeSameAs(dict); //< -- doesn't compile
dict.Should().BeSameAs(dict as IReadOnlyDictionary<int, int>);
```

Here `GenericDictionaryNotImplementingIDictionary` doesn't implement `IReadOnlyDictionary`, so it also not explicitly be converted.
```csharp
var dict = new GenericDictionaryNotImplementingIDictionary<int, int>();
// dict.Should().BeSameAs(dict); // doesn't compile
dict.As<object>().Should().BeSameAs(dict);
```

Because an `IDictionary` not implementing `IReadOnlyDictionary` is converted to a `ReadOnlyDictionary` this fails
```csharp
var dict = new GenericDictionaryNotImplementingIDictionary<int, int>();
dict.Should().BeOfType<GenericDictionaryNotImplementingIDictionary<int, int>>();
```

In believe all assertions from `ReferenceTypeAssertions` involving `TSubject` or `Type` are fragile to this approach.

shatl:
I'll resolve conflicts in couple days. However, better solution is need i order to support both IDictionary and IReadOnlyDictionary.
I'm going to try second option from my first post.

Issue #664, opened 2017-10-07
Configuration is not thread-safe


jnyrup:
When executing all tests in VS with "Run tests in parallel" enabled, sometimes some tests fails.

The stack trace:
```
Expected a <Xunit.Sdk.XunitException> to be thrown, but found a <System.InvalidOperationException>: System.InvalidOperationException with message "Nullable object must have a value."
     at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
     at System.Nullable`1.get_Value()
     at FluentAssertions.Common.Configuration.get_ValueFormatterDetectionMode() in C:\git\fluentassertions\Src\FluentAssertions\Common\Configuration.cs:line 43
     at FluentAssertions.Formatting.AttributeBasedFormatter.get_IsScanningEnabled() in C:\git\fluentassertions\Src\FluentAssertions\Formatting\AttributeBasedFormatter.cs:line 34
     at FluentAssertions.Formatting.AttributeBasedFormatter.CanHandle(Object value) in C:\git\fluentassertions\Src\FluentAssertions\Formatting\AttributeBasedFormatter.cs:line 29
     at FluentAssertions.Formatting.Formatter.<>c__DisplayClass8_0.<Format>b__0(IValueFormatter f) in C:\git\fluentassertions\Src\FluentAssertions\Formatting\Formatter.cs:line 138
     at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found)
     at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
     at FluentAssertions.Formatting.Formatter.Format(Object value, FormattingContext context, FormatChild formatChild) in C:\git\fluentassertions\Src\FluentAssertions\Formatting\Formatter.cs:line 138
     at FluentAssertions.Formatting.Formatter.ToString(Object value, Boolean useLineBreaks) in C:\git\fluentassertions\Src\FluentAssertions\Formatting\Formatter.cs:line 91
     at FluentAssertions.Execution.MessageBuilder.<FormatArgumentPlaceholders>b__7_0(Object a) in C:\git\fluentassertions\Src\FluentAssertions\Execution\MessageBuilder.cs:line 62
     at System.Linq.Enumerable.SelectArrayIterator`2.ToArray()
     at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
[...]

PR #714, opened 2017-12-26
Some threadsafety for Configuration

melchiork:
Configuration.Current.ValueFormatterAssembly and Configuration.Current.ValueFormatterDetectionMode can be accessed in concurrent execution safely. Spotted in #664.

### DIFF
--- a/Src/FluentAssertions/Common/Configuration.cs
+++ b/Src/FluentAssertions/Common/Configuration.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using FluentAssertions.Formatting;
+using System.Threading;
 
 namespace FluentAssertions.Common
 {
@@ -11,6 +12,7 @@ namespace FluentAssertions.Common
         private string valueFormatterAssembly;
         private ValueFormatterDetectionMode? valueFormatterDetectionMode;
         private string testFrameworkName;
+        private object propertiesAccessLock = new object();
 
         #endregion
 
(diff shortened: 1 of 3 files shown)


Src/FluentAssertions/Common/Configuration.cs
```diff
@@ -35,12 +37,17 @@ public ValueFormatterDetectionMode ValueFormatterDetectionMode
         {
             get
             {
+                Monitor.Enter(propertiesAccessLock);
```
dennisdoomen: Why not just use `lock`?

jnyrup:
@melchiork Thanks for having a look as this.

I would prefer using the `lock` statement instead.
Currently if `DetermineFormatterDetectionMode()` throws an exception `Monitor.Exit()` won't be called and causing a deadlock.

On the other hand the `lock` statement is [translated](https://blogs.msdn.microsoft.com/ericlippert/2009/03/06/locks-and-exceptions-do-not-mix/) into the following to ensure that `Monitor.Exit` is called.
```
bool lockWasTaken = false;
var temp = obj;
try { Monitor.Enter(temp, ref lockWasTaken); { body } }
finally { if (lockWasTaken) Monitor.Exit(temp); }
```

Out of pure curiosity, do you have any numbers on how taking an exclusive locks influences running all tests in parallel?

melchiork:
I've changed to use `lock` statement instead of Monitor. I was not aware of the thing that @jnyrup pointed out - thanks!

dennisdoomen:
Does this solve #664?

melchiork:
@dennisdoomen As far as I know.

PR #715, opened 2017-12-28
Do not replace double braces with context

### LINKED ISSUE #586: FormatException when asserting on the content of an exception message (using WithMessage(string)) when the expected exception message contains braces ({})
I'm using the latest (at time of writing) NuGet package distribution in VS2013: Id FluentAssertions, Version 4.19.3. 

I'm writing an extension method for the `GenericCollectionAssertions<T>` class and tests to go along with my new extension.   The extension is partially implemented, like so:

```C#
public static AndConstraint<GenericCollectionAssertions<T>> ContainSingleItemOfType<T>(this GenericCollectionAssertions<T> assertions, Type type, string reason = "", params object[] reasonArgs)
{
	var itemsOfType = assertions.Subject.Where(i => type.IsAssignableFrom(i.GetType()));

	Execute.Assertion
		.ForCondition(itemsOfType.Any())
		.BecauseOf(reason, reasonArgs)
		.FailWith("Expected collection {0} to contain a single item of type {1}{reason}, but it was empty.", assertions.Subject, type);

	throw new NotImplementedException();
}
```

The test looks like this:

```C#
[Test]
public void ContainSingleItemOfType_EmptyCollection_Throws()
{
	var collection = Enumerable.Empty<object>();

	Action assert = () => collection.Should().ContainSingleItemOfType(typeof(ICloneable));

	assert.ShouldThrow<AssertionException>()
		.WithMessage(@"Expected collection {empty} to contain a single item of type ICloneable, but it was empty.");
}
```

Debugging the test, the assertion does throw the `AssertionException` with the expected message.(`"Expected collection {empty} to contain a single item of type System.ICloneable, but it was empty."`)   However, `WithMessage` subsequently throws a `FormatException`, which fails my test. (This `FormatException` comes from line 66 of `MessageBuilder.cs`, according to the debugger.)

Am I missing some level of formatting that I should be doing in my test or in my assertion?   If I change my assertion message and test to look for a message that *doesn't* contain braces, then it all works fine:

```C#
public static AndConstraint<GenericCollectionAssertions<T>> ContainSingleItemOfType<T>(this GenericCollectionAssertions<T> assertions, Type type, string reason = "", params object[] reasonArgs)
{
	var itemsOfType = assertions.Subject.Where(i => type.IsAssignableFrom(i.GetType()));

	Execute.Assertion
		.ForCondition(itemsOfType.Any())
		.BecauseOf(reason, reasonArgs)
		.FailWith("Expected the collection to contain a single item of type {0}{reason}, but it was empty.", type);

	throw new NotImplementedException();
}
```
```C#
[Test]
public void ContainSingleItemOfType_EmptyCollection_Throws()
{
	var collection = Enumerable.Empty<object>();

	Action assert = () => collection.Should().ContainSingleItemOfType(typeof(ICloneable));

	assert.ShouldThrow<AssertionException>()
		.WithMessage(@"Expected the collection to contain a single item of type System.ICloneable, but it was empty.");
}
```

I'm thinking it could be an issue with `WithMessage` itself?

jnyrup:
This PR tries to resolve #586.
I haven't touched the formatting logic that much before, so I hope I got it right.

The Regex is changed to include both negative lookbehind and lookahead, such that `{{...}}` isn't matched.
It will still fail if one passes in strings such as e.g. `{ {empty}}`, but don't believe we can (or should) try to guard against all possible input.


dennisdoomen:
I'll need to delay this PR a bit since I've also changed the `MessageBuilder` significantly to support #588. See https://github.com/dennisdoomen/fluentassertions/blob/CallerContext/Src/FluentAssertions/Execution/MessageBuilder.cs

PR #716, opened 2018-01-01
Report missing and additional keys on dictionaries non-equivalency

### LINKED ISSUE #557: When comparing dictionary objects of different items count, final result does not show which items are missing.
Lets say I have 2 dictionary objects: Dictionary<string, object> dictA, Dictionary<string, object> dictB. I'm doing the following when comparing them:

`dictB.ShouldBeEquivalentTo<T>(dictA);`

When the total items do not match, I just get a message " Expected member dictB to be a dictionary with 26 item(s), but found 23 item(s).". How can I make it show what items are missing between the two?

krajek:
Fixes #557. When dictionaries differ in count of items, message includes keys that are missing or those that were unexpected(called "additional keys").


krajek:
It was suggested to show just first 3 items, but for simplicity and consistency, I used default formatter.

Src/FluentAssertions/Equivalency/GenericDictionaryEquivalencyStep.cs
```diff
@@ -154,15 +154,81 @@ private static Type GetIDictionaryInterface(Type expectedType)
...
+                IEnumerable<TSubjectKey> additionalKeys = subject.Keys.Except(presentKeys);
+
+                bool thereAreSomeMissingKeys = missingKeys.Count > 0;
+                bool thereAreSomeAdditionalKeys = additionalKeys.Any();
+
+                // Just missing keys
+                if(thereAreSomeMissingKeys && !thereAreSomeAdditionalKeys)
+                {
+                    return AssertionScope.Current
+                        .ForCondition(subject.Count == expectation.Count)
```
jnyrup: You can leave out the `ForCondition`. 
The condition is always false is it was checked on line 163

Src/FluentAssertions/Equivalency/GenericDictionaryEquivalencyStep.cs
```diff
@@ -154,15 +154,81 @@ private static Type GetIDictionaryInterface(Type expectedType)
...
+                            expectation.Count,
+                            subject.Count,
+                            missingKeys);
+                }
+
+                // Just additional keys
+                else if(!thereAreSomeMissingKeys && thereAreSomeAdditionalKeys)
+                {
+                    return AssertionScope.Current
+                        .ForCondition(subject.Count == expectation.Count)
```
jnyrup: Leave out `ForCondition`

Src/FluentAssertions/Equivalency/GenericDictionaryEquivalencyStep.cs
```diff
@@ -154,15 +154,81 @@ private static Type GetIDictionaryInterface(Type expectedType)
...
+                            expectation.Count,
+                            subject.Count,
+                            additionalKeys);
+                }
+
+                // Both missing and additional keys
+                else if(thereAreSomeMissingKeys && thereAreSomeAdditionalKeys)
+                {
+                    return AssertionScope.Current
+                        .ForCondition(subject.Count == expectation.Count)
```
jnyrup: Leave out `ForCondition`

Src/FluentAssertions/Equivalency/GenericDictionaryEquivalencyStep.cs
```diff
@@ -154,15 +154,81 @@ private static Type GetIDictionaryInterface(Type expectedType)
             return GetIDictionaryInterfaces(expectedType).Single();
         }
 
-        private static bool AssertSameLength<TSubjectKey, TSubjectValue, TExpectKey, TExpectedValue>(
+        private static bool AssertSameLength<TSubjectKey, TSubjectValue, TExpectedKey, TExpectedValue>(
             IDictionary<TSubjectKey, TSubjectValue> subject,
-            IDictionary<TExpectKey, TExpectedValue> expectation)
+            IDictionary<TExpectedKey, TExpectedValue> expectation) where TExpectedKey : TSubjectKey
```
jnyrup: Doesn't this introduce "asymmetry"?
E.g. if `TSubjectKey` derives from `TExpectedKey` instead, then `AssertSameLength(dict1, dict2)` would work, but not `AssertSameLength(dict2, dict1)`.


How about when `TSubjectKey` and `TExpectedKey` are completely different types?
Is this a too strict constraint?

I hope my thoughts make any sense.

Tests/Shared.Specs/DictionaryEquivalencySpecs.cs
```diff
@@ -786,10 +786,35 @@ public void When_two_dictionaries_asserted_to_be_equivalent_have_different_lengt
...
-                .WithMessage("Expected subject to be a dictionary with 2 item(s), but found 1 item(s)*");
+                .WithMessage("Expected subject to be a dictionary with 2 item(s), but found 1 item(s).*Missing key(s): {\"farewell\"}*");
             
             act2.Should().Throw<XunitException>()
-                .WithMessage("Expected subject to be a dictionary with 1 item(s), but found 2 item(s)*");
+                .WithMessage("Expected subject to be a dictionary with 1 item(s), but found 2 item(s).*Additional key(s): {\"farewell\"}*");
+        }
+
+        [Fact]
+        public void When_two_dictionaries_asserted_to_be_equivalent_have_different_lengths_with_both_missing_and_additional_items_it_should_fail_descriptively()
```
jnyrup: This might be personal preference, but I would split this into two separate tests.

Src/FluentAssertions/Equivalency/GenericDictionaryEquivalencyStep.cs
krajek: Thanks, I thought that it was still necessary, but you are right, it worked without ForCondition.

Src/FluentAssertions/Equivalency/GenericDictionaryEquivalencyStep.cs
krajek: Thanks, removed

Src/FluentAssertions/Equivalency/GenericDictionaryEquivalencyStep.cs
krajek: Thanks, removed.

dennisdoomen:
It would be nice if you could actually use a `git rebase` to keep the history clean. I never use a normal merge but always rebase my branch on the target branch. See also https://www.slideshare.net/dennisdoomen/git-like-a-pro-how-to-use-it-as-it-was-meant-to

Tests/Shared.Specs/DictionaryEquivalencySpecs.cs
krajek: Thanks, good idea, I did that.

Src/FluentAssertions/Equivalency/GenericDictionaryEquivalencyStep.cs
krajek: Yes, it does introduce asymmetry but it is the same asymmetry that is already enforced in AssertIsCompatiblyTypedDictionary which is executed just before AssertSameLength. I added two tests that document the asymmetric behaviour. Additionally, I explained the asymmetry issue just below the `where TExpectedKey : TSubjectKey`

Thanks for your input. Please, let me know if my explanation makes sense.

krajek:
I think that the PR is ready.

Regarding rebasing: sure, I will do that in next PRs. Nevertheless, I would suggest squashing the feature branch. My commits are sometimes too fine-grained for the main history.

dennisdoomen:
I'll clean-up the code a bit and merge it directly.

dennisdoomen:
Replaced by #723

PR #717, opened 2018-01-01
Explicitly mention that "because" formatting can throw FormatException

### LINKED ISSUE #551: When because message contains “{someText}”, then exception will thrown
Example:

public void Foo(string info)
{
    object someObject = null;
    someObject.Should().BeNull($“because {info}”);
}

public void Bar()
{
    //this will throw exception:
    Foo(“{a}”);
}

krajek:
Closes #551.
Does not change the behaviour.
Only documentation for "because" formatting is improved.

### DIFF
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -103,12 +103,17 @@ namespace FluentAssertions.Execution
         /// Specify the reason why you expect the condition to be <c>true</c>.
         /// </summary>
         /// <param name="because">
-        /// A formatted phrase explaining why the condition should be satisfied. If the phrase does not 
-        /// start with the word <i>because</i>, it is prepended to the message.
+        /// A formatted phrase compatible with <see cref="string.Format(string,object[])"/> explaining why 
+        /// the condition should be satisfied. If the phrase does not start with the word <i>because</i>, 
+        /// it is prepended to the message. 
         /// </param>
         /// <param name="becauseArgs">
         /// Zero or more values to use for filling in any <see cref="string.Format(string,object[])"/> compatible placeholders.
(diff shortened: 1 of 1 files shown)


dennisdoomen:
Let's first see if we can fix this properly in 5.0 rather than solve it using documentation.

krajek:
Could you please describe how it would even theoretically be possible to "fix" the problem "properly"? The scenario in the #551 is that caller provides incorrectly formatter "because" argument. What is the correct behaviour for the FA in that case?

dennisdoomen:
We could escape anything that doesn't look like a typical `{0]` structure and deescape after the formatting was done.

krajek:
Ok, I think I see your point. It would allow something like `BeNull("values {must} be null")` instead of current `BeNull("value {{must}} be null")`. I acknowledge that advantage, but I think added complexity is absolutely not worth it.

Moreover, in some cases, it would be even harmful. Consider following code: 
`someObject.Should().BeNull("because {O}", 1);`
I intentionally used letter `O` instead of digit `0`. Your solution would silently skip the obvious error and user would be surprised at runtime.

Another issue is that even if you will implement proposed solution perfectly, you will still need to throw formatting exception in cases like 
`someObject.Should().NotBeNull("because {0,abc}", 1);`
There is just no way around it, so documentation improvements from the PR would still be correct and useful.

Last, but not least, changing the allowed formats for `because` message would mean resigning from consistency with string.Format. I consider this consistency a value in itself. This format is well understood and documented. Any variation would introduce some confusion.

If I am missing something, I would be happy to learn more. 
Currently, it seems to me that messing with standard formatting rules is opening little Pandora box for very little gain. 

Hope that helps.

dennisdoomen:
Alright. I was looking into solutions too much. You have a good point here.

PR #719, opened 2018-01-03
Better message for empty collections

### LINKED ISSUE #718: Bug when comparing collection to empty collection
```C#
[Fact]
public void EmptyCollectionComparison()
{
    var collection1 = new[] { 1, 2, 3 };
    var collection2 = new int[] { };
    collection1.Should().Equal(collection2);
}
```

Fails with the following message:

`Expected collection to be equal to {empty}, but found empty collection.`

I think "but found empty collection" should read "but found {1, 2, 3}" or something that at least indicates that collection1 was not in fact empty.

I'm using version (4.19.4).

jnyrup:
When comparing collections `EitherIsEmpty` discarded information about which collection was the empty and which was the non-empty.

I've split that check into two separate assertions, such that we can include the subject in the failure message if the expected collection is empty, but the subject isn't.

This fixes #718 

This also fixes asserting `StartsWith` and `EndsWith` with empty collections, as the empty collection naturally is an prefix and suffix on any collection.


dennisdoomen:
Before I had time to read the original issue, you've already fixed it. Chapeau.

PR #727, opened 2018-01-08
Added BeDecoratedWithOrInherit and NotBeDecoratedWithOrInherit for type assertions

### LINKED ISSUE #706: BeDecoratedWith should be able to specify wether or not the attribute should be inherited
type.BeDecoratedWith() currently only checks for direct implementation. It should also be possible to check if the type, or any of its parent types, are decorated. This could be easily achieved since the build-in .NET method Type.GetCustomAttributes already has an overload that accepts a boolean wether or not to take inherited attributes into account.

Version 5.0.0-beta0004

MJLHThomassen-Eurocom:
Fixes #706


jnyrup:
Thanks for your contribution!

It all looks good to me 👍

PR #728, opened 2018-01-08
Change ...(3 dots) to …(elipsis character)

krajek:


dennisdoomen:
Is this just about styling?

krajek:
Yep, mainly good style. It seems to me like there is no good reason not to use it. 
One could argue though, that it takes one character instead of three in memory or that it takes less visual space when printed.

PR #735, opened 2018-01-21
Sanity checks

jnyrup:
Some sanity checks that should still pass if one tries to add support for `IReadOnlyDictionary`.


dennisdoomen:
I don't understand. What is this PR supposed to do?

jnyrup:
It is the three assertions from https://github.com/fluentassertions/fluentassertions/pull/710#issuecomment-354542171 that wouldn't work with that PR in its current state.

So the purpose is to help anyone who tries to solve the missing support for `IReadOnlyDictionary`.
Or put in another way, I'm a bit sceptical and I think any attempt shouldn't break these three assertions.

Issue #711, opened 2017-12-21
Too easy to accidentally use Should().NotThrow() against async void lambda


zarenner:
With the current async support in FA, it's too easy in my opinion to accidentally use .Should().NotThrow() against an async void lambda. This will result in a test that looks valid but doesn't actually test the continuation.

Example:
```csharp
Action asyncFunc = async () => { await Task.Delay(5000); throw new Exception("fail"); };
asyncFunc.Should().NotThrow();
// Test (incorrectly) passes
```

...instead of the correct test implementation:
```csharp
Func<Task> asyncFunc = async () => { await Task.Delay(5000); throw new Exception("fail"); };
asyncFunc.Should().NotThrow();
// Test (correctly) fails
```

I see a couple of separate but complementary possible improvements:
- Detect that the Action is an async method (e.g. action.Method.GetCustomAttribute(typeof(AsyncStateMachineAttribute)) != null), and throw. Would break anyone intentionally sending in an async void, but that seems unlikely. I don't believe it's possible to "fix" the async void method; if it was that would be another alternative.

- Add actual async [Not]ThrowAsync methods that only work on Func\<Task\>. Wouldn't directly prevent this mistake if NotThrow was still used, but the added Async (and await) would make it more obvious when the async assertions aren't being used. I don't know if there was a reason AsyncFunctionAssertions was designed to block instead of using async/await (possibly because it would be a bit less fluent), but personally I'd much prefer actual async assertions.

dennisdoomen:
Option 1 sounds like a decent solution.

jnyrup:
Can this be closed with #737 merged?

Issue #463, opened 2016-08-18
Ground work for Fluent Assertions 5.0


dennisdoomen:
**Ground Work**
- [x] Switch to XUnit (for Devices)?
- [x] Switch to Cake
- [x] Move JSON support to a separate repo to support different release cycles
- [x] Drops support for .NET 4.0, Silverlight, Windows 8.0/8.1, WPA and Windows Phone for Silverlight.
- [x] Redirect contributors to the master branch for bug-fixes.
- [x] Move to dedicated organization
- [x] Start logo contest
- [x] Cross-compiles .NET Standard 1.4 (so we can support XmlDocument and UWP), .NET Standard 1.6 (to support .NET Core)  and .NET Framework 4.5 (#558)

dennisdoomen:
@onovotny any thoughts on this?

onovotny:
LGTM :)

dennisdoomen:
@onovotny will this allow you to get rid of FA.Xamarin?

MannusEtten:
maybe also support for UWP?

dennisdoomen:
> maybe also support for UWP?

Hmm, that would mean I have to target 1.4. @onovotny can you confirm?

onovotny:
If you target 1.4, then that includes UWP.

SaroTasciyan:
@dennisdoomen Regarding #257 planning to remove `ShouldBeEquivalentTo()` in v5 to prevent confusion?

dennisdoomen:
I was planning to give them the same behavior.

PR #737, opened 2018-01-22
Fail when Should().[Not]Throw used on async void Action

zarenner:
## IMPORTANT 
SINCE FLUENT ASSERTIONS IS CURRENTLY UNDER HEAVY REFACTORING FOR [VERSION 5.0](https://github.com/fluentassertions/fluentassertions/issues/463), WE CURRENTLY ONLY ACCEPT BUGFIX REQUESTS. SORRY FOR THE INCONVENIENCE

* [ ] The code complies with the [Coding Guidelines for C# 3.0, 4.0 and 5.0](http://www.csharpcodingguidelines.com/)/.

#711

### DIFF
--- a/Src/FluentAssertions/Specialized/ActionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/ActionAssertions.cs
@@ -2,7 +2,10 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
 
+using FluentAssertions.Common;
 using FluentAssertions.Execution;
 using FluentAssertions.Primitives;
 
@@ -35,6 +38,8 @@ namespace FluentAssertions.Specialized
         public ExceptionAssertions<TException> Throw<TException>(string because = "", params object[] becauseArgs)
             where TException : Exception
         {
+            FailIfSubjectIsAsyncVoid();
+
             Exception actualException = InvokeSubjectWithInterception();
             IEnumerable<TException> expectedExceptions = extractor.OfType<TException>(actualException);
 
@@ -67,6 +72,8 @@ namespace FluentAssertions.Specialized
         /// </param>
         public void NotThrow<TException>(string because = "", params object[] becauseArgs) where TException : Exception
         {
+            FailIfSubjectIsAsyncVoid();
+
             Exception actualException = InvokeSubjectWithInterception();
 
             if (actualException != null)
@@ -92,6 +99,8 @@ namespace FluentAssertions.Specialized
         /// </param>
         public void NotThrow(string because = "", params object[] becauseArgs)
         {
+            FailIfSubjectIsAsyncVoid();
+
             try
             {
                 Subject();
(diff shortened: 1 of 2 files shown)


Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -119,6 +128,14 @@ private Exception InvokeSubjectWithInterception()
             return actualException;
         }
 
+        private void FailIfSubjectAsyncVoid(string assertionMethodName)
+        {
+            if (Subject.GetMethodInfo().HasAttribute<AsyncStateMachineAttribute>())
+            {
+                throw new InvalidOperationException($"Cannot use {assertionMethodName} on an Action representing an async void method.");
```
zarenner: Let me know if you prefer a more descriptive message indicating that Action very likely should have been Func<Task>

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -35,6 +38,8 @@ public ActionAssertions(Action subject, IExtractExceptions extractor)
         public ExceptionAssertions<TException> Throw<TException>(string because = "", params object[] becauseArgs)
             where TException : Exception
         {
+            FailIfSubjectAsyncVoid("Throw");
```
zarenner: Let me know if you disagree with adding the check on Should().Throw() as well. It doesn't need it as much as NotThrow IMO, but could help users avoid confusion while writing tests.

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -119,6 +128,14 @@ private Exception InvokeSubjectWithInterception()
             return actualException;
         }
 
+        private void FailIfSubjectAsyncVoid(string assertionMethodName)
+        {
+            if (Subject.GetMethodInfo().HasAttribute<AsyncStateMachineAttribute>())
```
zarenner: Not sure what FA's stance on reflection is - this is the only way I'm aware of to do this check though.

Src/FluentAssertions/Specialized/ActionAssertions.cs
dennisdoomen: Yes, I think emphasizing the possible solution does help the user understanding his mistake.

jnyrup:
If `Throw` gets it, please also add it to`ThrowExactly`, otherwise the exception message would mention `Throw` when calling `ThrowExactly`.

Shouldn't `FailIfSubjectAsyncVoid` be `FailIfSubjectIsAsyncVoid`?

Src/FluentAssertions/Specialized/ActionAssertions.cs
Meir017: Maybe use `nameof`

zarenner: afaik nameof doesn't let you capture the correct (generic or non-generic) NotThrow method. Not that I expect one name to change and not the other, but still seemed wrong.

zarenner:
Updated exception message to use term "action assertions" and suggest fix. Renamed FailIfSubjectAsyncVoid to FailIfSubjectIsAsyncVoid. Added test for ThrowExactly.

Issue #739, opened 2018-01-22
Treat types that override `Equals` as having value semantics


dennisdoomen:
Dogfooding 5.0 has revealed some inconsistent behavior of how `Should().BeEquivalenTo` should deal with value vs structural comparisons. This is currently governed by https://github.com/fluentassertions/fluentassertions/blob/release-5.0/Src/FluentAssertions/AssertionOptions.cs#L36 and https://github.com/fluentassertions/fluentassertions/blob/release-5.0/Src/FluentAssertions/Equivalency/ValueTypeEquivalencyStep.cs. I can think of two options that could fix this:
1. Detect that an object overrides `Equals` and assume value semantics for that type. Add a method `ComparingByMembers<T>` (to counter `ComparingByValue<T>`) to override this behavior. Since we want to use anonymous types as an expectation, we should not treat those as value types.
2. Allow objects to be annotated with an attribute which name and namespace should match `[FluentAssertions.HasValueSemantics]` (by name, not by type)

Any opinions?

mpopdam:
I don't like the annotation option.

ronaldkroon:
Agree, I also don't like to add annotations to production code, just for the specs.
I really feel that if you implement Equals yourself that we should trust that it is indeed equal.

dennisdoomen:
Solved in #740

PR #740, opened 2018-01-24
Assuming types that override Equals to have value semantics

dennisdoomen:
Replaces all mechanisms and options to control whether or not an object has value semantics by a single algorithm: Any object that overrides object.Equals and is not an anonymous type is treated as a value type, unless ComparingByMembers<T>() is used. Also removes the global AssertionOptions.IsValueType. For those cases where this algorithm misbehaves, ComparingByValue<T> can be used.

#739 

* [x] Initial implementation
* [x] Also treat tuples specially
* [x] Update extensibility docs
* [x] Update main docs on value equality

### DIFF
--- a/Src/FluentAssertions/Common/TypeExtensions.cs
+++ b/Src/FluentAssertions/Common/TypeExtensions.cs
@@ -3,7 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
-
+using System.Runtime.CompilerServices;
 using FluentAssertions.Equivalency;
 
 namespace FluentAssertions.Common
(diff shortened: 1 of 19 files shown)


Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -136,14 +113,10 @@ internal static Type[] GetClosedGenericInterfaces(Type type, Type openGenericTyp
...
-        {
-            return HasProperties(type) && !AssertionOptions.IsValueType(type);
-        }
-
-        private static bool HasProperties(Type type)
+        public static bool OverridesEquals(this Type type)
         {
-            return type.GetProperties(PublicMembersFlag).Any();
+            MethodInfo methodInfo = type.GetMethod(nameof(object.Equals), new[]{ typeof(object)});
+            return (methodInfo != null) && (methodInfo.GetBaseDefinition().DeclaringType != methodInfo.DeclaringType);
```
jnyrup: Does this check that `Equals(object)` also overrides `object.Equals(object)`? 
I.e. it is neither of
```
public bool Equals(object obj);
public new bool Equals(object obj);
```

dennisdoomen: Only the 2nd is possible (I think)

jnyrup: It fails for this corner case

```csharp
public class VirtualClass
{
	public virtual bool Equals(object obj) { ... }
}

public class VirtualClassOverride : VirtualClass
{
	public override bool Equals(object obj) { ... }
}
```

https://dotnetfiddle.net/qKKLbq

dennisdoomen: I'll fix it.

jnyrup: It should be enough to add an additional check that 
`methodInfo.GetBaseDefinition().DeclaringType == typeof(System.Object)`

dennisdoomen: Took a bit more than that:

```
            IEnumerable<MethodInfo> methods = type
                .GetMethods(BindingFlags.Public | BindingFlags.Instance);

            methods = methods
                .Where(m => m.Name == "Equals" &&
                            m.GetParameters().Select(p => p.ParameterType).SequenceEqual(new[] {typeof(object)}));

            methods = methods
                .Where(m => m.GetBaseDefinition().DeclaringType != m.DeclaringType)
                .ToArray();

            return methods.Any();
```

PR #750, opened 2018-01-27
Add support for unsigned types and sbyte

### LINKED ISSUE #435: Very unexpected omission: uint?.Should().Be(<value>) fails
```
            uint? u = 5;
            u.Should().Be(5);
```

Result: Expected object to be 5, but found 5.

It works fine for uint (not nullable) and int (both nullable and not nullable)

Used version: 3.2.2.0

jnyrup:
This PR provides overloads of `Should` for:

```
sbyte
sbyte?
ushort
ushort?
uint
uint?
ulong
ulong?
```

This fixes #435


Tests/Shared.Specs/NumericAssertionSpecs.cs
```diff
@@ -876,6 +876,25 @@ public void When_asserting_a_byte_value_it_should_treat_is_any_numeric_value()
             act.Should().NotThrow();
         }
 
+        [Fact]
+        public void When_asserting_a_sbyte_value_it_should_treat_is_any_numeric_value()
```
dennisdoomen: In general, I'd like to avoid these repetitive `When_asserting`. It adds superfluous noise.

PR #758, opened 2018-02-08
add a target for .net 4.7 that doesn't include ValueTuple fix #757

### LINKED ISSUE #757: When installing the package in .net 4.7 it forces me to install ValueTuple

m3zercat:
I hope this is the right way to do this. I did build using the script and it changed all the assembly info files - not sure if I was supposed to include those in this pull request - can do if necessary

### DIFF
--- a/Src/FluentAssertions.nuspec
+++ b/Src/FluentAssertions.nuspec
@@ -29,6 +29,8 @@
       <frameworkAssembly assemblyName="System.Xml.Linq" targetFramework="net45" />
     </frameworkAssemblies>
     <dependencies>
+      <group targetFramework="net47">
+      </group>
       <group targetFramework="net45">
         <dependency id="System.ValueTuple" version="4.3.0" />
       </group>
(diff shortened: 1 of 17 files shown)


dennisdoomen:
Looks like what I was expecting although I suspected only an update to the .nuspec would be enough. Did you try to build the nuget package locally and then see if you could add it to a .NET 4.7 unit test project?

m3zercat:
I'll admit I have no idea how to install a nuget package locally to a project. I did build the package and inspect it with Nuget Package Explorer to confirm that it looked ok.

dennisdoomen:
Create a nuget package source that points to a local folder and then try to add that package to a .NET 4.7 project ;-)

m3zercat:
Yep. done that. It installed just fine with no requirement for ValueTuple. Though I'm not sure - which bit of the fluent assertions code base required ValueTuple - I just want to confirm it works still. Though I'm sure it does.

dennisdoomen:
It tries to support value tuples.

m3zercat:
Awesome. Thanks

dennisdoomen:
No, thank you!

PR #762, opened 2018-02-11
Nullable struct assertions

jnyrup:
This PR add a new type of Assertions for nullable structs that do not fall into the current categories `{bool, DateTime, DateTimeOffset, Guid, TimeSpan}`.

Unlike the other nullable assertions, this one returns an `AndWhichConstraint` on `HaveValue/NotBeNull` for chaining assertions on the struct type.


dennisdoomen:
Crap. This caused a breaking change. You used to be able to do `nullableEnum.Should().Be`, but a nullable enum now ends up in the `NullableStructAssertions<T>`. Was this your intention?

jnyrup:
Damn,
It wasn't my intention that nullable enums should end up in `NullableStructAssertions<T>`.
I did intentionally leave out `Be(T expected)` on `NullableStructAssertions<T>`.

I'm not aware of any ways in the type system to exclude enums.
So in order for this to not be a breaking change, `NullableStructAssertions` would need to provide all methods from `ObjectAssertions`.

Any feelings about that.

I just stumbled upon a single place in our unit tests where assertions on nullable structs would be convenient. But I would rather leave it out, if it messes to much with the rest of FA.

dennisdoomen:
Yeah, that's how I found this myself. I was trying to upgrade a project with 19K unit tests ;-)

But there are quite a few methods on `ObjectAssertions` and `ReferenceTypeAssertions` that should be available for structs as well.

jnyrup:
I had a closer look and I'm leaning towards dropping this PR again.
Currently `nullableEnum.Should().HaveFlag(MyEnum.Flag)` returns an `AndConstraint<ObjectAssertions>`.

If I added `HaveFlag` to `NullableStructAssertions` and returned a `AndConstraint<NullableStructAssertions<TSource>>` instead of a `AndConstraint<ObjectAssertions>`
it would be a breaking change.
But returning a AndConstraint<ObjectAssertions> from a `NullableStructAssertions<TSource>` feels like an confusing API.

Even if that could be resovled, a lot of the assertions in `ReferenceTypeAssertions` don't make sense for structs.

dimkanewtown:
This was an unexpected curve-ball after the upgrade since this
```
            target.NullableEnumProperty.Should().Be(expectedNullOrNotNullVal, because);
```
had to become this for the logic to work properly :(
```
            if(expectedNullOrNotNullVal== null)
                target.NullableEnumProperty.Should().BeNull(because);
            else
                target.NullableEnumProperty.Value.Should().Be(expectedNullOrNotNullVal, because);

```

jnyrup:
@dimkanewtown I'm really sorry about the havoc this caused 😞.

dennisdoomen:
Can you create a revert PR for that? Then I can release 5.1.1

PR #764, opened 2018-02-13
Fixes a FormatException when a bad string is included in the 'because' clause

### LINKED ISSUE #756: FormatException when a bad string is included in the 'because' clause of a failed assertion
When a particular string is included in the because clause of a failed assertion, it causes the assertion to throw a FormatException instead of an AssertionFailedException. This string comes from the big list of naughty strings (https://github.com/minimaxir/big-list-of-naughty-strings) and corresponds to "<>?:\"{}|_+".

This problem can be reproduced with the following .NET Core 2.0 console app with a reference to the 5.0 version of FluentAssertions:

```
using FluentAssertions;

namespace BadStringForFluentAssertions
{
    class Program
    {
        static void Main(string[] args)
        {
            // When decoded, this string is "<>?:\"{}|_+"
            const string base64Encoded = "PD4/OiJ7fXxfKw==";
            var troublesomeString = System.Text.Encoding.UTF8.GetString(System.Convert.FromBase64String(base64Encoded));

            // This assertion throws a format exception rather than the correct AssertionFailedException
            System.Action a = () =>
                troublesomeString.Should().Be("foo", $"{troublesomeString} causes trouble in the 'because' clause");

            a.Should().Throw<AssertionFailedException>();
        }
    }
}

```

The actual error I get is:

```
FluentAssertions.Execution.AssertionFailedException
  HResult=0x80131500
  Message=Expected a <FluentAssertions.Execution.AssertionFailedException> to be thrown, but found a <System.FormatException>: System.FormatException with message "Input string was not in a correct format."
     at System.Text.StringBuilder.FormatError()
     at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)
     at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
     at System.String.Format(String format, Object[] args)
     at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\AssertionScope.cs:line 208
     at FluentAssertions.Primitives.StringEqualityValidator.ValidateAgainstLengthDifferences() in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Primitives\StringEqualityValidator.cs:line 30
     at FluentAssertions.Primitives.StringValidator.Validate() in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Primitives\StringValidator.cs:line 41
     at FluentAssertions.Primitives.StringAssertions.Be(String expected, String because, Object[] becauseArgs) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Primitives\StringAssertions.cs:line 41
     at BadStringForFluentAssertions.Program.<>c__DisplayClass0_0.<Main>b__0() in d:\dev\BadStringForFluentAssertions\BadStringForFluentAssertions\Program.cs:line 16
     at FluentAssertions.Specialized.ActionAssertions.InvokeSubjectWithInterception() in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Specialized\ActionAssertions.cs:line 118
.
  Source=FluentAssertions
  StackTrace:
   at FluentAssertions.Execution.FallbackTestFramework.Throw(String message) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\FallbackTestFramework.cs:line 21
   at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\AssertionScope.cs:line 215
   at FluentAssertions.Specialized.ActionAssertions.Throw[TException](String because, Object[] becauseArgs) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Specialized\ActionAssertions.cs:line 60
   at BadStringForFluentAssertions.Program.Main(String[] args) in d:\dev\BadStringForFluentAssertions\BadStringForFluentAssertions\Program.cs:line 18

```

bernard-chen:
## IMPORTANT 
SINCE FLUENT ASSERTIONS IS CURRENTLY UNDER HEAVY REFACTORING FOR [VERSION 5.0](https://github.com/fluentassertions/fluentassertions/issues/463), WE CURRENTLY ONLY ACCEPT BUGFIX REQUESTS. SORRY FOR THE INCONVENIENCE

* [ ] The code complies with the [Coding Guidelines for C# 3.0, 4.0 and 5.0](http://www.csharpcodingguidelines.com/)/.

Fixes #756.

### DIFF
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -115,18 +115,26 @@ namespace FluentAssertions.Execution
         /// <param name="because">
         /// A formatted phrase compatible with <see cref="string.Format(string,object[])"/> explaining why 
         /// the condition should be satisfied. If the phrase does not start with the word <i>because</i>, 
-        /// it is prepended to the message. 
+        /// it is prepended to the message. If the format of <paramref name="because"/> or 
+        /// <paramref name="becauseArgs"/> is not compatible with <see cref="string.Format(string,object[])"/>,
+        /// then a warning message is returned instead.
         /// </param>
         /// <param name="becauseArgs">
         /// Zero or more values to use for filling in any <see cref="string.Format(string,object[])"/> compatible placeholders.
         /// </param>
-        /// <exception cref="System.FormatException">
-        /// Thrown if the format of <paramref name="because"/> or <paramref name="becauseArgs"/> is not 
-        /// compatible with <see cref="string.Format(string,object[])"/>.
-        /// </exception>
         public AssertionScope BecauseOf(string because, params object[] becauseArgs)
         {
-            reason = () => string.Format(because ?? "", becauseArgs ?? new object[0]);
+            reason = () =>
+            {
+                try
+                {
+                    return string.Format(because ?? "", becauseArgs ?? new object[0]);
+                }
+                catch (FormatException formatException)
+                {
+                    return $"**WARNING** because message '{because}' could not be formatted with string.Format{Environment.NewLine}{formatException.StackTrace}";
+                }
+            };
             return this;
         }
 
--- a/Tests/Shared.Specs/AssertionScopeSpecs.cs
+++ b/Tests/Shared.Specs/AssertionScopeSpecs.cs
@@ -193,6 +193,21 @@ namespace FluentAssertions.Specs
         }
 
         [Fact]
+        public void When_invalid_format_is_used_in_because_parameter_it_should_render_default_text()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => 1.Should().Be(2, "use of {} is considered invalid in because parameter");
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act.Should().Throw<XunitException>()
+                .WithMessage("*because message 'use of {} is considered invalid in because parameter' could not be formatted with string.Format*");
+        }
+
+        [Fact]
         public void When_an_assertion_fails_in_a_scope_with_braces_it_should_use_the_name_as_the_assertion_context()
         {
             //-----------------------------------------------------------------------------------------------------------


jnyrup:
Thanks for having a look into this.

I'm a bit hesitant in trying to fix stuff, that throws a `System.FormatException`.
E.g. `1.Should().Be(2, "{{{{}}}}");` still fails with a `FormatException`.

In #717 the summary notes for `BecauseOf` was changed to mention, that it could throw a `FormatException` if the string is not compatible with `string.Format`.

jnyrup:
For the sake of insight, we changed `reason` to only evaluated on test failure, to avoid a potentially expensive `string.Format`, e.g. when formatting large collections.

But I agree that a test shouldn't fail, because the ´because` throws an exception.
I would prefer if we changed the `Func` in `BecauseOf` to wrap `string.Format` in a try/catch and provided a meaningful fallback message if `string.Format` throws a `FormatException` including the stack trace.

bernard-chen:
I can add a try/catch to the `Func`, but are you saying that the result of evaluating the `reason` would be a string that says something like "because of a reason that could not be formatted with string.Format"? I would think we would still want to throw a FormatException so that the behavior matches the summary notes of #717, but maybe rethrow it with another FormatException that says that it was the reason that could not be formatted.

dennisdoomen:
Something like that could work

jnyrup:
To reformulate myself. 
(because what I wrote earlier was not specific enough)

A test will only fail from bad `because` string, if:
* the assertion was wrapped inside an `Action`, 
* that assertion fails, 
* `BecauseOf` throws a `FormatException`,
* and you asserted an exception different from `FormatException` on the `Action`.

The notes in #717 was written before #725 was merged in.
Maybe @krajek has a different view on this topic now?

If we go with a fallback reason, I was thinking about something like `**WARNING** because message could not be formatted with string.Format\r\n[StackTrace]`.

krajek:
It still seems to me that changing anything in formatting grammar is harmful. Standard formatting grammar is well documented and understood, let's stick to it.

On the other hand, I really like the idea of enriching assertion message instead of throwing FormatException to the caller. Even if the reason part is flawed, part of the assertion message would still be rendered,  that's desirable property. 
Just repharsing, for clarity. I suggest:
 * try/catch the reason evaluation
 * even if reason throws, never let the exception out
 * display some simple message in the reason part, like it was suggested
 * remove the note about possibiliy of throwing `FormatException`  from documentation(basically revert #717)

@jnyrup what do you think? If I understood you correctly you suggested still throwing `FormatException`. How is the idea of completely containing the exception and using only assertion message works for you?

jnyrup:
@krajek I agree with everything you just suggested.

bernard-chen:
Let me take a pass at this. I'll undo the changes I proposed in this PR in favor of what I think we're now talking about.

bernard-chen:
I made the change proposed by @krajek, except that I also included the original because parameter in the warning message.

bernard-chen:
If `becauseArgs` is empty, should we skip the `string.Format` call?

Tests/Shared.Specs/AssertionScopeSpecs.cs
```diff
@@ -192,6 +192,22 @@ public void When_parentheses_are_used_in_the_because_arguments_it_should_render_
...
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            const string because = "use of {} is considered invalid in because parameter";
+            Action act = () => 1.Should().Be(2, because);
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act.Should().Throw<XunitException>()
+                .WithMessage($"*because message '{because}' could not be formatted with string.Format*");
```
krajek: Tiny issue: you are reproducing the logic from production code in tests. I suggest avoiding $ formatting and just prepare expected message manually.

krajek:
Good job, nice improvement.

Regarding empty `becauseArgs` message: that is an interesting question but best answered in other PR or issue.

Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -115,18 +115,26 @@ public AssertionScope UsingLineBreaks
...
-            reason = () => string.Format(because ?? "", becauseArgs ?? new object[0]);
+            reason = () =>
+            {
+                try
+                {
+                    return string.Format(because ?? "", becauseArgs ?? new object[0]);
+                }
+                catch (FormatException formatException)
+                {
+                    return $"**WARNING** because message '{because}' could not be formatted with string.Format\r\n{formatException.StackTrace}";
```
jnyrup: For the sake of platform independence, please use `{Environment.NewLine}` instead of `\r\n.

bernard-chen: I just pushed a commit with this change.

Tests/Shared.Specs/AssertionScopeSpecs.cs
bernard-chen: I embedded the text directly in the expected message with the most recent commit.

jnyrup:
Looks good to me.

dennisdoomen:
Looks good to me too. Now, to really help us keep the history clean, I would like you to update the PR so it only includes the final implementation. You can do that by using an interactive rebase and skipping the first commits or by creating a new branch, cherry picking the right commits and force pushing it into the one that is linked to the current PR. Would that work for you?

bernard-chen:
I did the rebase. Please let me know if the commit history and the file changes look like you expect.

dennisdoomen:
@bernard-chen congrats with your first contribution to Fluent Assertions

PR #765, opened 2018-02-13
Revert "Nullable struct assertions"

jnyrup:
Reverts fluentassertions/fluentassertions#762

This caused a breaking change for e.g. nullable enums.


jnyrup:
I'm not sure why "Add BeEquivalentTo to ComparableAssertions" is also noted to be reverted.
The files changed looks fine?

PR #766, opened 2018-02-14
Improve message for unequal length collections in BeEquivalentTo

jnyrup:
This PR improves the failure message when collections fail in `BeEquivalentTo` due to unequal length.
Before the failure message would only state the number of found items.
The failure message now includes:
* subject collection,
* expected collection,
* the difference in length, 
* and whether the subject had too many or too few items.

### DIFF
--- a/Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidatorExtensions.cs
+++ b/Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidatorExtensions.cs
@@ -1,0 +1,55 @@
+using System;
+
+using FluentAssertions.Execution;
+
+namespace FluentAssertions.Equivalency
+{
+    internal static class EnumerableEquivalencyValidatorExtensions
+    {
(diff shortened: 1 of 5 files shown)


Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidatorExtensions.cs
```diff
@@ -0,0 +1,55 @@
+using System;
+
+using FluentAssertions.Execution;
+
+namespace FluentAssertions.Equivalency
+{
+    internal static class EnumerableEquivalencyValidatorExtensions
```
dennisdoomen: Why are these extension methods if they are only used by the `EnumerableEquivalencyValidator`?

jnyrup: Purely make them extensions methods on `AssertionScope`,  so `AssertCollectionsHaveSameCount` can call them in a more fluent way instead of passing `AssertionScope` as a parameter to each.

PR #768, opened 2018-02-15
Cleanup

jnyrup:
I'm giving cleanup a shot again :)

* Add default access modifiers
* Use conditional access + coalesce instead of (x != null && x)
* Convert verbatim strings to regular strings, when verbatim is unused
* Remove empty lines
* Simplify lambdas, e.g. "() => {throw ex;}" to "() => throw ex"
* Merge declaration and assignment
* Add parentheses according to precedence
* Replace new EventArgs() with EventArgs.Empty
* Format whitespace
* C#6 dictionary initializer


dennisdoomen:
I trust you blindly with this ;-)

PR #769, opened 2018-02-16
Only attempt to format in BecauseOf if format parameters were supplied

bernard-chen:
This is admittedly a minor case, but we should only trigger the `FormatException`/warning message from attempting to use the `because` parameter in `string.Format` if we also include `becauseArgs`.

### DIFF
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -128,7 +128,8 @@ namespace FluentAssertions.Execution
             {
                 try
                 {
-                    return string.Format(because ?? "", becauseArgs ?? new object[0]);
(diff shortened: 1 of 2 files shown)


jnyrup:
Again, I'm hesitant in trying to "fix" this.

Take the case of `1.Should().Be(2, "{0}")`
By returning "{0}" we hide that the user either provider an invalid format string, or didn't include `becauseArgs`.

We simply don't know whether the user wanted to do:
* `1.Should().Be(2, "{{0}}")` or
* `1.Should().Be(2, "{0}", "foo")`.

dennisdoomen:
Well, we don't have to hold the hands of our users. Not running the format string through `string.format` when there are no because args is perfectly fine. Most people will probably use `$` anyway.

PR #770, opened 2018-02-16
becauseArgs can be null

jnyrup:


Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -129,7 +129,7 @@ public AssertionScope BecauseOf(string because, params object[] becauseArgs)
                 try
                 {
                     string becauseOrEmpty = because ?? "";
-                    return becauseArgs.Any() ? string.Format(becauseOrEmpty, becauseArgs) : becauseOrEmpty;
+                    return (becauseArgs?.Any() == true) ? string.Format(becauseOrEmpty, becauseArgs) : becauseOrEmpty;
```
bernard-chen: Thanks for fixing this. Can we remove the `== true` part of this code?

dennisdoomen: It's the result of using the `?` operator .

jnyrup: No.
`becauseArgs?` is a nullable bool, so `== true` cannot be removed.
In other words, it checks that the nullable bool is `true`, and neither `null` or `false`.

bernard-chen: Ah. It's because the `?` makes the expression return `bool?`, which can't be directly used in the ternary statement. It jumped out as odd to me because we don't typically want to compare explicitly with true or false. A different option would be something like `becauseArgs?.Any() ?? false`, but it's not much of an improvement.

PR #786, opened 2018-03-01
Adding Assertion Scope Documentation

michal-ciechan:
Adding documentation for a hidden gem of a feature (`AssertionScope`). Been wanting this for months, and finally thought I'd give it a go to implement, but was pleasantly surprised to find it in the Unit Tests and exposed already!

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -1268,3 +1268,30 @@ someAction.ExecutionTime().Should().BeGreaterThan(100.Milliseconds());
 someAction.ExecutionTime().Should().BeGreaterOrEqualTo(100.Milliseconds());
 someAction.ExecutionTime().Should().BeCloseTo(150.Milliseconds(), 50.Milliseconds());
 ```
+
+
+## Assertion Scope ##
+
+You can batch multiple Assertions, into an `AssertionScope` so that FluentAssertions throws one exception at the end of the scope with all failures.
+
+E.g.
+
+```csharp
+using (new AssertionScope())
+{
+    5.Should().Be(10);
+    "Actual".Should().Be("Expected");
+}
+```
+
+The above will batch the 2 failures, and throw an exception at the point of disposing the `AssertionScope` displaying both errors.
+
+E.g. Exception thrown at point of dispose contains:
+
+    Expected value to be 10, but found 5.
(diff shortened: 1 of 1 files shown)


docs/documentation.md
```diff
@@ -1268,3 +1268,30 @@ someAction.ExecutionTime().Should().BeGreaterThan(100.Milliseconds());
 someAction.ExecutionTime().Should().BeGreaterOrEqualTo(100.Milliseconds());
 someAction.ExecutionTime().Should().BeCloseTo(150.Milliseconds(), 50.Milliseconds());
 ```
+
+
+## Assertion Scope ##
+
+You can batch multiple Assertions, into an `AssertionScope` so that FluentAssertions throws one exception at the end of the scope with all failures.
```
dennisdoomen: Why is _assertions_ written with a capital? And why is there a comma?

docs/documentation.md
```diff
@@ -1268,3 +1268,30 @@ someAction.ExecutionTime().Should().BeGreaterThan(100.Milliseconds());
...
+
+```csharp
+using (new AssertionScope())
+{
+    5.Should().Be(10);
+    "Actual".Should().Be("Expected");
+}
+```
+
+The above will batch the 2 failures, and throw an exception at the point of disposing the `AssertionScope` displaying both errors.
```
dennisdoomen: It's good practice to write numbers under 10 as one, two, three, etc.

docs/documentation.md
```diff
@@ -1268,3 +1268,30 @@ someAction.ExecutionTime().Should().BeGreaterThan(100.Milliseconds());
...
+
+E.g. Exception thrown at point of dispose contains:
+
+    Expected value to be 10, but found 5.
+    Expected string to be "Expected" with a length of 8, but "Actual" has a length of 6.
+    
+        at........
+        
+For more information take a look at the `AssertionScopeSpecs.cs` in Unit Tests.
```
dennisdoomen: I would make it a link instead?

michal-ciechan:
As per review comments:

 - Lowercased "Assertions" + removed comma
 - Rewrote '2' as 'two'
 - Made `AssertionScopeSpecs.cs` a link to the actual file

Issue #645, opened 2017-09-29
More Assertions on types


jnyrup:
We do have an _impressive_ amount of type assertions.
I've tried to lookup the .net type hierarchy and keywords to see what we might be missing.
See the bottom for types and keywords.

Here's a rough sketch of what such APIs could look like, but please notice that I haven't checked if they are:
* even possible
* desirable.

- [ ] ClassAssertions : TypeAssertions<ClassAssertions>
- [ ] ClassSelector : TypeSelector
- [ ] EventInfoAssertions : MemberInfoAssertions<EventInfo, EventInfoAssertions>
- [ ] ~FieldInfoAssertions : MemberInfoAssertions<FieldInfo, FieldInfoAssertions>~
  - [ ] ~[Not]BeVolatile~
  - [ ] ~[Not]BeConst~
- [ ] ~FieldInfoSelector~
  - [ ] ~ThatAre[Not]Volatile~
  - [ ] ~ThatAre[Not]Const~
- [ ] InterfaceAssertions : TypeAssertions<InterfaceAssertions>
- [ ] InterfaceSelector : TypeSelector
- [ ] MemberInfoAssertions
  - [ ] [Not]HaveAccessModifier
  - [ ] ~[Not]BeOverridable~
  - [ ] [Not]BeStatic
  - [ ] [Not]BeAbstract
- [x] MethodInfoSelector
  - [x] ThatAre[Not]Async #1725
  - [ ] ~ThatAre[Not]Overridable~
  - [x] ThatAre[Not]Virtual #1744 
  - [x] ThatAre[Not]Static #1740
  - [x] ThatAre[Not]Abstract #2060
  - [ ] ThatAre[Not]Sealed
- [ ] PropertyInfoSelector
  - [x] NotBeWritable #1256
  - [ ] ~ThatAre[Not]Overridable~
  - [x] ThatAre[Not]Virtual #2054
  - [x] ThatAre[Not]Static #2054
  - [x] ThatAre[Not]Abstract #2054
  - [ ] ThatAre[Not]Sealed
- [ ] ~StructAssertions : TypeAssertions<StructAssertions>~
- [ ] ~StructSelector : TypeSelector~
- [x] TypeAssertions
  - [x] [Not]BeStatic #796
  - [x] [Not]BeAbstract #796
  - [x] [Not]BeSealed #796
- [ ] TypeSelector
  - [ ] ThatAre[Not]Virtual
  - [x] ThatAre[Not]Static #1306
[...]

PR #796, opened 2018-03-08
(Not)Be(Abstract|Sealed|Static) class assertions implemented

melchiork:
For  #645
Assertions for types (abstract, sealed, static) added. They are based on C# understanding of those keywords, as in CLR they differ.

### DIFF
--- a/Src/FluentAssertions/Types/TypeAssertions.cs
+++ b/Src/FluentAssertions/Types/TypeAssertions.cs
@@ -531,6 +531,8 @@ namespace FluentAssertions.Types
         /// <returns></returns>
         public AndConstraint<TypeAssertions> BeSealed(string because = "", params object[] becauseArgs)
         {
+            AssertThatSubjectIsClass();
+
             Execute.Assertion
                 .ForCondition(Subject.IsCSharpSealed())
                 .BecauseOf(because, becauseArgs)
@@ -548,6 +550,8 @@ namespace FluentAssertions.Types
         /// <returns></returns>
         public AndConstraint<TypeAssertions> NotBeSealed(string because = "", params object[] becauseArgs)
         {
+            AssertThatSubjectIsClass();
+
             Execute.Assertion
                 .ForCondition(!Subject.IsCSharpSealed())
                 .BecauseOf(because, becauseArgs)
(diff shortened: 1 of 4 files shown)


melchiork:
@jnyrup Thanks, fixed. Also I've added two examples to the docs.

jnyrup:
It just struck me that from a C# perspective the keywords `sealed`, `abstract` and `static` are only applicable to class like types, and not interfaces or structs.

Several assertions in `TypeAssertions` checks if the type is an interface.
One example is `BeDerivedFrom` as you cannot derive from an interface type, but implement it.

dennisdoomen:
So? You want to have some changea made to this PR?

jnyrup:
Yes, I would like to have a check that the `Subject` type is not a valuetype or an interface or throw an `InvalidOperationException` with a descriptive message.

This is similar to what we do in `ActionAssertions.FailIfSubjectIsAsyncVoid`, `XDocumentAssertions.HaveRoot` and `XDocumentAssertions.HaveElement`

Calling `typeof(StructType).IsSealed` returns `true`, but that is from a CLR perspective.
The purpose of the type assertions seems to be from a C# perspective, i.e. if the type under assertion has the `sealed` keyword.

melchiork:
@jnyrup @dennisdoomen 
Done, I've also check if the type is a Delegate.
I went with ArgumentException as similar was done in TypeAssertions.Implement and TypeAssertions.NotImplement

jnyrup:
`BeDerivedFrom` and the alike uses `ArgumentException` as they're checking the argument. 
As the `Subject` isn't an argument to the assertion method I prefer `InvalidOperationException` like in `HaveRoot`. 

I really like your use of `Theory` tests. 
Excellent job!

melchiork:
@jnyrup Done. Thanks for the reviews!

Src/FluentAssertions/Types/TypeAssertions.cs
```diff
@@ -1102,5 +1216,14 @@ protected override string Identifier
         {
             get { return "type"; }
         }
+
+        private void AssertThatSubjectIsClass()
+        {
+            var typeInfo = Subject.GetTypeInfo();
+            if (typeInfo.IsInterface || typeInfo.IsValueType || typeof(Delegate).IsAssignableFrom(typeInfo.BaseType))
+            {
+               throw new InvalidOperationException($"{Subject} must be a class.");
```
jnyrup: Last thing I can come up with is a missing space to keep four space indentation.

PR #803, opened 2018-03-17
[Not]BeCloseTo for integral types

### LINKED ISSUE #801: BeApproximately variant for integral types
I found myself writing code similar to this today

```csharp
int subject = 1;
int expected = 2

subject.Should().BeInRange(expected - 1, expected + 1);
```

While if I casted `subject` to a floating point type I could write the in my opinion much prettier
`subject.Should().BeApproximately(expected, 1)`.

I'm not sure the name for integral types should be `BeApproximately`, but a function with the same functionality could be a nice shorthand.

jnyrup:
This PR adds `[Not]BeCloseTo` for: byte, sbyte, ushort, short, uint, int, ulong and long.

I could use some feedback on the type of the `delta` parameter.
* for the signed cases they take a signed `delta`, but throw an `ArgumentOfOutRangeException` if the `delta` is negative.
* for the unsigned cases they take a unsigned `delta` which can never be negative.

So I'm unsure what the most consistent/unsurprising/user-friendly API:
* The signedness of `delta` follows the others, 
* They take a signed `delta` and conditionally throws, or
* They take an unsigned `delta`

This fixes #801

### DIFF
--- a/Src/FluentAssertions/NumericAssertionsExtensions.cs
+++ b/Src/FluentAssertions/NumericAssertionsExtensions.cs
@@ -10,6 +10,676 @@ namespace FluentAssertions
     public static class NumericAssertionsExtensions
     {
         /// <summary>
+        /// Asserts an integral value is close to another value within a specified value.
+        /// </summary>
+        /// <param name="parent">The <see cref="NumericAssertions{T}"/> object that is being extended.</param>
+        /// <param name="nearbyValue">
+        /// The nearby value to compare the actual value with.
+        /// </param>
+        /// <param name="delta">
+        /// The maximum amount of which the two values may differ.
+        /// </param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])"/> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because"/>.
+        /// </param>
+        public static AndConstraint<NumericAssertions<sbyte>> BeCloseTo(this NumericAssertions<sbyte> parent,
+            sbyte nearbyValue, byte delta, string because = "",
+            params object[] becauseArgs)
+        {
+            sbyte actualValue = (sbyte)parent.Subject;
+            sbyte minValue = (sbyte)(nearbyValue - delta);
+            if (minValue > nearbyValue)
+            {
+                minValue = sbyte.MinValue;
+            }
+
+            sbyte maxValue = (sbyte)(nearbyValue + delta);
+            if (maxValue < nearbyValue)
+            {
+                maxValue = sbyte.MaxValue;
+            }
+
+            Execute.Assertion
+                .ForCondition(minValue <= actualValue && actualValue <= maxValue)
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected {context:value} to be within {0} from {1}{reason}, but found {2}.",
+                    delta, nearbyValue, actualValue);
+
+            return new AndConstraint<NumericAssertions<sbyte>>(parent);
+        }
+
+        /// <summary>
+        /// Asserts an integral value is close to another value within a specified value.
+        /// </summary>
+        /// <param name="parent">The <see cref="NumericAssertions{T}"/> object that is being extended.</param>
+        /// <param name="nearbyValue">
+        /// The nearby value to compare the actual value with.
+        /// </param>
+        /// <param name="delta">
+        /// The maximum amount of which the two values may differ.
+        /// </param>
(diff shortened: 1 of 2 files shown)


melchiork:
@jnyrup As the delta can never be negative I believe it's better to enforce this constraint using the type system.

dennisdoomen:
> They take an unsigned delta

I vote for that

jnyrup:
I've updated the PR with unsigned types for `delta`. for working with `long` values and `ulong` delta this required two helper functions `GetMinValue` and `GetMaxValue` as `-` and `+` apparently are ambiguous operators for `long` and `ulong`.

Src/FluentAssertions/NumericAssertionsExtensions.cs
```diff
@@ -9,6 +9,676 @@ namespace FluentAssertions
     /// </summary>
     public static class NumericAssertionsExtensions
     {
+        /// <summary>
+        /// Asserts an integral value is within another value by a specified value.
```
dennisdoomen: _is close to_?

Src/FluentAssertions/NumericAssertionsExtensions.cs
```diff
@@ -9,6 +9,676 @@ namespace FluentAssertions
...
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because"/>.
+        /// </param>
+        public static AndConstraint<NumericAssertions<sbyte>> BeCloseTo(this NumericAssertions<sbyte> parent,
+            sbyte nearbyValue, byte delta, string because = "",
+            params object[] becauseArgs)
+        {
+            sbyte actualValue = (sbyte)parent.Subject;
+            sbyte minValue = (sbyte)(nearbyValue - delta);
```
dennisdoomen: Is `Math.Max` not clearer?

Src/FluentAssertions/NumericAssertionsExtensions.cs
```diff
@@ -9,6 +9,676 @@ namespace FluentAssertions
...
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because"/>.
+        /// </param>
+        public static AndConstraint<NumericAssertions<byte>> BeCloseTo(this NumericAssertions<byte> parent,
+            byte nearbyValue, byte delta, string because = "",
+            params object[] becauseArgs)
+        {
+            byte actualValue = (byte)parent.Subject;
```
dennisdoomen: I wonder if you can somehow replace this all this repetition with a mathod that takes an `IComparable`

Tests/Shared.Specs/NumericAssertionSpecs.cs
```diff
@@ -2069,6 +2069,1265 @@ public void When_a_nullable_decimal_has_no_value_and_should_not_approximate_it_s
...
+        [InlineData(sbyte.MaxValue, sbyte.MaxValue, sbyte.MaxValue)]
+        [InlineData(sbyte.MaxValue, sbyte.MaxValue - 1, 1)]
+        [InlineData(sbyte.MaxValue, sbyte.MaxValue - 1, sbyte.MaxValue)]
+        [Theory]
+        public void When_a_sbyte_value_is_close_to_expected_value_it_should_succeed(sbyte actual, sbyte nearbyValue, byte delta)
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Act / Assert
+            //-----------------------------------------------------------------------------------------------------------
+            actual.Should().BeCloseTo(nearbyValue, delta);
```
dennisdoomen: Generally I use a `Should().NotThrow` here.

Tests/Shared.Specs/NumericAssertionSpecs.cs
```diff
@@ -2069,6 +2069,1265 @@ public void When_a_nullable_decimal_has_no_value_and_should_not_approximate_it_s
...
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => actual.Should().BeCloseTo(nearbyValue, delta);
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act.Should().Throw<XunitException>();
```
dennisdoomen: Shouldn't we at least verify some parts of the message?

Src/FluentAssertions/NumericAssertionsExtensions.cs
jnyrup: The problem here is handling overflows. 
Another way would be to wrap the `nearbyValue +/- delta` inside `checked{}` and catch the `OverflowException`.

Issue #781, opened 2018-02-25
MethodImplAttribute doesn't work with BeDecoratedWith


melchiork:
Version: 5.1.2

When a method is decorated with `MethodImplAttribute` the assertion `Should().BeDecoratedWith<MethodImplAttribute>()` will not detect that and therefore produce a false negative. It's because `GetCustomAttributes` method that is used behind the scenes will not get this attribute as per [docs](https://msdn.microsoft.com/en-US/library/system.runtime.compilerservices.methodimplattribute(v=vs.110).aspx). The `GetMethodImplementationFlags` method has to be used.
I can implement a patch for this, but I'm unsure which of two approaches would be better:
1. [breaking] add capacity of finding the Implementation Flags to current `BeDecoratedWith` method
2. [non-braking] add a new assertion like `BeDecoratedWithImplementationFlags`

dennisdoomen:
Why would (1) be breaking?

melchiork:
Example "reconstruction" of `MethodImplAttribute`

```
private static int ImplementationOptionsMask = 
Enum.GetValues(typeof(MethodImplOptions)).Cast<int>().Sum(x => x);

                private static Attribute GetMethodImplAttribute(MemberInfo memberInfo)
        {
#if NETSTANDARD1_3
            return null;
#else
            var methodInfo = memberInfo.MemberType == MemberTypes.Method
                ? memberInfo as MethodInfo
                : null;

            var implementationAttributes = methodInfo.GetMethodImplementationFlags();

            var methodImplementationOptionsWithoutNotMatchingImplementationAttributes =
                (int)implementationAttributes & ImplementationOptionsMask;

            var implementationOptions =
                (MethodImplOptions)(short)
                methodImplementationOptionsWithoutNotMatchingImplementationAttributes;

            if (implementationOptions != 0)
            {
                return new MethodImplAttribute(implementationOptions);
            }

            return null;
#endif
        }
```

jnyrup:
The other type assertions take a C# perspective rather than a CLR perspective.
Seen from that point of view, I'm not against `[Not]BeDecoratedWith` getting capabilities to detect `MethodImplAttribute`.

If it is at any help for the person who might implement this, `MethodImplAttribute` is marked with `Inherited = false`. So overriden methods won't inherit this attribute.

jnyrup:
Thanks for another contribution @melchiork .
They are very appreciated.

PR #804, opened 2018-03-18
Handle MethodImplAttribute

melchiork:
PoC for #781
The code recreates the `MethodImplAttribute` using `MethodImplementationFlags`.


dennisdoomen:
Why do you call it a POC?

Src/FluentAssertions/Common/MethodInfoExtensions.cs
```diff
@@ -3,27 +3,64 @@
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 
 namespace FluentAssertions.Common
 {
     public static class MethodInfoExtensions
     {
+        private static Lazy<int> ImplementationOptionsMask = new Lazy<int>(() => Enum.GetValues(typeof(MethodImplOptions)).Cast<int>().Sum(x => x));
```
jnyrup: Why lazy and not just `static readonly`?

Src/FluentAssertions/Common/MethodInfoExtensions.cs
```diff
@@ -3,27 +3,64 @@
...
 
         internal static IEnumerable<TAttribute> GetMatchingAttributes<TAttribute>(this MemberInfo memberInfo, Expression<Func<TAttribute, bool>> isMatchingAttributePredicate) where TAttribute : Attribute
         {
-            return memberInfo.GetCustomAttributes(
+            var customAttributes = memberInfo.GetCustomAttributes(
                 typeof(TAttribute), false)
                 .Cast<TAttribute>()
+                .ToList();
+
+            if(typeof(TAttribute) == typeof(MethodImplAttribute))
```
jnyrup: I would add a check `memberInfo` is an `MethodInfo` here and send the casted value to `RecreateMethodImplAttribute`.

And then tests for `[Not]BeDecoratedWith` when calling on a non-`MethodInfo` member types.

Src/FluentAssertions/Common/MethodInfoExtensions.cs
```diff
@@ -3,27 +3,64 @@
...
+            return customAttributes
                 .Where(isMatchingAttributePredicate.Compile());
         }
 
         internal static bool IsNonVirtual(this MethodInfo method)
         {
             return !method.IsVirtual || method.IsFinal;
         }
+
+        private static (bool success, MethodImplAttribute attribute) RecreateMethodImplAttribute(MemberInfo memberInfo)
```
jnyrup: Then this function can take a `MethodInfo` as argument.

Src/FluentAssertions/Common/MethodInfoExtensions.cs
```diff
@@ -3,27 +3,64 @@
...
+        {
+            var methodInfo = memberInfo as MethodInfo ?? null;
+
+            var implementationFlags = methodInfo.MethodImplementationFlags;
```
jnyrup: and avoid NRE here if someone tried to check on a non-method member.

Src/FluentAssertions/Common/MethodInfoExtensions.cs
```diff
@@ -3,27 +3,64 @@
...
         }
 
         internal static bool IsNonVirtual(this MethodInfo method)
         {
             return !method.IsVirtual || method.IsFinal;
         }
+
+        private static (bool success, MethodImplAttribute attribute) RecreateMethodImplAttribute(MemberInfo memberInfo)
+        {
+            var methodInfo = memberInfo as MethodInfo ?? null;
```
jnyrup: Then we could remove the cast from here

Src/FluentAssertions/Common/MethodInfoExtensions.cs
```diff
@@ -3,27 +3,64 @@
...
+            var implementationFlags = methodInfo.MethodImplementationFlags;
+
+            var implementationFlagsMatchingImplementationOptions =
+                (int)implementationFlags & ImplementationOptionsMask.Value;
+
+            var implementationOptions =
+                (MethodImplOptions)
+                implementationFlagsMatchingImplementationOptions;
+
+            if (implementationOptions != 0)
```
jnyrup: A reason other than preference  to return `bool, MethodImplAttribute` and not just `MethodImplAttribute`?

Tests/Shared.Specs/MethodInfoAssertionSpecs.cs
```diff
@@ -125,6 +126,49 @@ public void When_asserting_a_method_is_decorated_with_attribute_and_it_is_it_suc
             act.Should().NotThrow();
         }
 
+        [Fact]
+        public void When_asserting_a_method_is_decorated_with_MethodImpl_attribute_and_it_is_it_succeeds()
```
jnyrup: If I know @dennisdoomen right, he prefers leaving out the `_asserting_` part of the test name as it doesn't add value.

Tests/Shared.Specs/MethodInfoAssertionSpecs.cs
```diff
@@ -125,6 +126,49 @@ public void When_asserting_a_method_is_decorated_with_attribute_and_it_is_it_suc
...
+            //-------------------------------------------------------------------------------------------------------------------
+            // Act
+            //-------------------------------------------------------------------------------------------------------------------
+            Action act = () =>
+                methodInfo.Should().BeDecoratedWith<MethodImplAttribute>();
+
+            //-------------------------------------------------------------------------------------------------------------------
+            // Assert
+            //-------------------------------------------------------------------------------------------------------------------
+            act.Should().Throw<XunitException>()
```
jnyrup: By replacing `WithMessage(...)` with something like `WithMessage("*PublicDoNothing*MethodImplAttribute*")` the test becomes more robust.
E.g if someone wanted to reword the failure message to end with ", but could not find that attribute.".

Tests/Shared.Specs/MethodInfoAssertionSpecs.cs
```diff
@@ -241,7 +328,7 @@ public void When_asserting_a_method_is_decorated_with_an_attribute_and_multiple_
         #region NotBeDecoratedWithOfT
 
         [Fact]
-        public void When_asserting_a_method_is_not_decorated_with_attribute_and_it_is_not_it_succeeds()
+        public void When_asserting_a_method_is_not_decorated_with__attribute_and_it_is_not_it_succeeds()
```
jnyrup: double underscore?

Src/FluentAssertions/Common/MethodInfoExtensions.cs
melchiork: I wanted to avoid computing that if it won't be used in a check. And if it were static it would have computed every time somebody uses the assembly. It's probably a nanosecond win anyway ;).

Src/FluentAssertions/Common/MethodInfoExtensions.cs
melchiork: Good point, will do that.

Src/FluentAssertions/Common/MethodInfoExtensions.cs
melchiork: True, I'll adjust.

Src/FluentAssertions/Common/MethodInfoExtensions.cs
melchiork: Just spotted that myself, won;t be a problem after the parameter type change.

Src/FluentAssertions/Common/MethodInfoExtensions.cs
melchiork: The idea is not to check if the returned value is null and the add it, but to rather have something like Result<T>. I don't like checking for null of the result a of a method, I considered TryGet... pattern with out keyword but this seems more readable.

Tests/Shared.Specs/MethodInfoAssertionSpecs.cs
melchiork: Sure, I just copied from previous tests, assuming it's the standard around.

Tests/Shared.Specs/MethodInfoAssertionSpecs.cs
melchiork: Omission, already corrected.

melchiork:
@jnyrup Thanks for the review, I don't find it rude nor harsh. In fact I really appreciate the feedback, as it gives me chance to improve. I'll try to fix and commit tomorrow.

Tests/Shared.Specs/MethodInfoAssertionSpecs.cs
jnyrup: Yeah, we do have _a lot_ of superfluous  `_asserting_` is test names.
It is on my todo, but quite far down.

dennisdoomen:
> I was thinking of some more ways to test this, but I can't quite nail it down. Anyway now I have some comments from @jnyrup to fix.

Sounds like this is going exactly as it should 😉

melchiork:
Post review changes are in place, I've also added a possibility to assert on constructor, as constructors too can be marked with `MethodImp` attribute.

Src/FluentAssertions/Common/MethodInfoExtensions.cs
```diff
@@ -3,27 +3,64 @@
...
         internal static IEnumerable<TAttribute> GetMatchingAttributes<TAttribute>(this MemberInfo memberInfo, Expression<Func<TAttribute, bool>> isMatchingAttributePredicate) where TAttribute : Attribute
         {
-            return memberInfo.GetCustomAttributes(
+            var customAttributes = memberInfo.GetCustomAttributes(
                 typeof(TAttribute), false)
                 .Cast<TAttribute>()
+                .ToList();
+
+            var methodBase = memberInfo as MethodBase;
+            if (typeof(TAttribute) == typeof(MethodImplAttribute) && methodBase != null)
```
jnyrup: If you'd like to avoid comparing with `null`, you can use the new(er) [is pattern matching](https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching#the-is-type-pattern-expression).

Src/FluentAssertions/Common/MethodInfoExtensions.cs
```diff
@@ -3,27 +3,64 @@
...
         }
 
         internal static bool IsNonVirtual(this MethodInfo method)
         {
             return !method.IsVirtual || method.IsFinal;
         }
+
+        private static (bool success, MethodImplAttribute attribute) RecreateMethodImplAttribute(MethodBase methodBase)
+        {
+            var implementationFlags = methodBase?.MethodImplementationFlags ?? 0;
```
jnyrup: You can remove the null conditional operator now after moving the type cast. 
 If something sends a null in here, let them be punished.

Tests/Shared.Specs/MethodInfoAssertionSpecs.cs
jnyrup: By "everywhere", do you mean everywhere in Fluent Assertions?
A quick look yesterday showed that at least 30% of `WithMessage` uses wildcards. 

By including the entire message in the assertion, one essentially asserts that the exception under test must have a very specific message. 

I'll put cleaning up exception message assertions on my todo. It is not very fair to contributors that the existing code is written in one way, but we tell them to write it another way.

jnyrup:
Nice catch that constructors can also be decorated, I didn't know that. 

I'm missing tests that verify the expected behavior when asserting on a non `MethodBase` type.

Src/FluentAssertions/Common/MethodInfoExtensions.cs
melchiork: Sure.

Src/FluentAssertions/Common/MethodInfoExtensions.cs
melchiork: Handy, I haven't used that before.

Tests/Shared.Specs/MethodInfoAssertionSpecs.cs
melchiork: I meant everywhere in the parts I was changing. Turns out it's not a very good sample for the population then.

melchiork:
Fixed.
I've added one test that assert on class type before `When_asserting_a_class_is_decorated_with_MethodImpl_attribute_and_it_is_not_it_throws`. It's in the `BeDecoratedWith` region. As `MethodImplAttribute` is marked to handle only methods and constructors I don't see a reasonable test for `NotBeDecoratedWith`.

dennisdoomen:
@jnyrup make sure you approve the changes so that I know you're happy with the changes.

jnyrup:
@melchiork Roger, `When_asserting_a_class_is_decorated_with_MethodImpl_attribute_and_it_is_not_it_throws` serves that purpose.

As `RecreateMethodImplAttribute` can't detect `[MethodImpl]` or `[MethodImpl(0)]` could you add tests for that too? That would also serve as documentation for the known limitations.

melchiork:
@jnyrup `MethodImpl` without any options specified seems pointless to apply on a class, .NET Core 1.1 even does not allow it. Anyway I've added some more tests.

Src/FluentAssertions/Common/MethodInfoExtensions.cs
```diff
@@ -3,27 +3,63 @@
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 
 namespace FluentAssertions.Common
 {
     public static class MethodInfoExtensions
     {
+        private static readonly Lazy<int> ImplementationOptionsMask =
```
dennisdoomen: I'm not a big fan of in-line comments, but this deserves some some.

Src/FluentAssertions/Common/MethodInfoExtensions.cs
```diff
@@ -3,27 +3,63 @@
...
         }
 
         internal static bool IsNonVirtual(this MethodInfo method)
         {
             return !method.IsVirtual || method.IsFinal;
         }
+
+        private static (bool success, MethodImplAttribute attribute) RecreateMethodImplAttribute(MethodBase methodBase)
+        {
+            var implementationFlags = methodBase.MethodImplementationFlags;
```
dennisdoomen: (Minor) I prefer explicit types for primitive types.

jnyrup:
@melchiork would you be willing to address the two review comments? 
Then I'll push the big fat merge button.

melchiork:
@jnyrup  @dennisdoomen Done.

PR #833, opened 2018-04-22
Update packages

jnyrup:
This PR has two parts:

* Port the remaining old csproj style projects to csproj2017 style. 
* Update NuGet package dependencies in test projects.

I intentionally didn't touch:
* The test framework dependencies as I wasn't sure about that.
* `Microsoft.NET.Test.Sdk` as I'm not fully aware of its usage or potential negative consequences by updating it.


dennisdoomen:
Did you compare the NuGet package that this PR builds against the one that is on Nuget? Just to make sure all dependencies are fine.

jnyrup:
I'm not sure what you're referring to? 

I thought about adding a [`cake.config` file](https://cakebuild.net/docs/fundamentals/configuration) to ensure that `nuget.exe` doesn't read any custom nuget feeds from `%AppData%\NuGet\NuGet.config`.
Would that be good enough?

dennisdoomen:
I just meant that the NuGet package and the assemblies are the same in terms of frameworks targeted.

jnyrup:
Ah, now I see.
I have compared the artifact from AppVeyor build with the one on nuget.org using Nuget Package Explorer and their dependencies are identical.

PR #836, opened 2018-04-24
Port test projects to NetCore

jnyrup:
This PR ports some tests projects to Net Core.
Besides making them cross-platform, I've added Test Runners to the projects.
This makes them testable with `dotnet test` and runnable from Visual Studio.


dennisdoomen:
Doesn't this mean we don't test with .NET 4.5 anymore? I can't completely foresee what we're missing here.

PR #842, opened 2018-05-02
ExecutionTimeAssertions works on cases where the code deadlocks, test…

peterekepeter:
ExecutionTimeAssertions works on cases where the code deadlocks, tests finish as soon as conditions are met or failed instead of running indefinitely.

I implemented the feature where you can have test cases can run indefinitely or deadlock. The assertions will only wait until necessary for them to determine if test passes or fails, and they continue on. The only side effect is that if the thread runs indefinitely or deadlocks then the thread will stay hanging, but then again in these cases your test should fail. 

## IMPORTANT 

* [x] The code complies with the [Coding Guidelines for C# 3.0, 4.0 and 5.0](http://www.csharpcodingguidelines.com/).


peterekepeter:
Looks like I have more work to do...

jnyrup:
I like the idea to handle deadlocks and return early, when the criteria has been meet.
Maybe I'm missing out on something, but the code looks quite complicated and involving low-level stuff like `Thread`.

How about this approach:
In `ExecutionTime` wrap the `Action` inside a `Task` and expose it in an internal property.
```csharp
internal Task Task { get; }

protected ExecutionTime(Action action, string actionDescription)
{
	ActionDescription = actionDescription;
	Task = Task.Run(action);
}
```

Add the following method to `ExecutionTime`;
```csharp
internal TimeSpan Execute(TimeSpan timeout)
{
    var stopwatch = Stopwatch.StartNew();
    Task.Wait(timeout);
    stopwatch.Stop();

    return stopwatch.Elapsed;
}
```

Change `ExecutionTimeAssertions` to save the `ExecutionTime` object as a field instead of unwrapping it;
```csharp
public ExecutionTimeAssertions(ExecutionTime executionTime)
{
    this.executionTime = executionTime;
}
```

In the assertion method like e.g. `BeLessOrEqualTo` we know for how long to run the task
run and wait the `Task` with a maximum timeout
```csharp
public void BeLessOrEqualTo(TimeSpan maxDuration, string because = "", params object[] becauseArgs)
{
    TimeSpan duration = executionTime.Execute(maxDuration);

    Execute.Assertion
        .ForCondition(duration.CompareTo(maxDuration) <= 0)
    ...
}
```

peterekepeter:
Also you are assuming that there is only one condition can be applied to the ExectionTime expression.
But I will have to look for a Task based solution, the build failed because Thread cannot be used in one of the target platforms. I'll see if I can works something out based on your suggestion.

peterekepeter:
I think I can make this work with Task I like most of your suggestion but assertion and execution are separate concerns, I don't want to specialize the execution for each type of assertion, it would unnecessarily complicate the code. And we don't just know for how long we need to run for `BeLessOrEqualTo`, we actually know it for all conditions.

peterekepeter:
Ended up spending more time on this than expected, but build succeeded, verify the PR again and let me know if I should make other changes.

Tests/Shared.Specs/ExecutionTimeAssertionsSpecs.cs
```diff
@@ -94,6 +94,30 @@ public void When_the_execution_time_of_an_action_is_less_or_equal_to_a_limit_it_
...
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => someAction.ExecutionTime().Should().BeLessThan(100.Milliseconds());
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act.Should().Throw<XunitException>();
```
dennisdoomen: Shouldn't you at least verify part of the message. Because if your code doesn't work, you'll still get an XUnit exception if I'm not mistaken. Same for the other tests.

Src/FluentAssertions/Specialized/ExecutionTimeAssertions.cs
```diff
@@ -10,18 +12,41 @@ namespace FluentAssertions.Specialized
...
+        private void PollUntil(Func<TimeSpan, bool> condition, bool result, TimeSpan rate)
+        {
+            var forcedStop = false;
+            while (execution.IsRunning)
+            {
+                if (condition(execution.ElapsedTime) == result)
+                {
+                    break;
+                }
+                Task.Delay(rate).Wait();
```
dennisdoomen: I'm not sure this is safe in all environments. I've seen lockups on AppVeyor while waiting for `Wait` call. I would just use a `Thread.Sleep` instead.

Tests/Shared.Specs/ExecutionTimeAssertionsSpecs.cs
peterekepeter: well, if my code doesnt work, the test never finishes to run, but fair enough, better to be safe

Src/FluentAssertions/Specialized/ExecutionTimeAssertions.cs
peterekepeter: yes, wanted to use that, originally and yes, .Result probably blocks a thread here, ill see if i can improve

peterekepeter:
So I think there is an even more elegant solution, the execution is already wrapped inside a Task.Run(), I just wait on it to finish with the polling rate as the timeout. If execution finishes earlier than that then the test finishes earlier as well, if not then the assertion passes or fail without needing for the execution to actually finish.

Src/FluentAssertions/Specialized/ExecutionTimeAssertions.cs
```diff
@@ -10,18 +12,41 @@ namespace FluentAssertions.Specialized
...
+        /// Checks the executing action if it satisfies a condition.
+        /// If the execution runs into an exeption, then this will rethrow it.
+        /// </summary>
+        /// <param name="condition">Condition to check un the current elapsed time.</param>
+        /// <param name="result">The expected result when polling is stopped.</param>
+        /// <param name="stopExecution">If polling is stopped then task execution is also stopped.</param>
+        /// <param name="rate">The rate at which the conition is re-checked.</param>
+        private void PollUntil(Func<TimeSpan, bool> condition, bool result, TimeSpan rate)
+        {
+            var forcedStop = false;
```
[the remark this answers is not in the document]
peterekepeter: remnants from a feature that's not possible, I originally planned to use thread and kill it if not needed anymore :)

Src/FluentAssertions/Specialized/ExecutionTimeAssertions.cs
```diff
@@ -10,18 +12,41 @@ namespace FluentAssertions.Specialized
...
+            while (execution.IsRunning)
+            {
+                if (condition(execution.ElapsedTime) == result)
+                {
+                    break;
+                }
+                execution.Task.Wait(rate);
```
jnyrup: If the return value of `Wait` is intentionally ignored, yo could use `_ = execution.Task.Wait(rate);` to explicitly state the intention.

peterekepeter: hah, this is a strange thing to do :)

jnyrup: I like to be explicit :)
I haven't tested it out, but this should help silence static analysers reporting unused return values, when I intentionally didn't use it.

peterekepeter:
So it looks like everything is fixed! Is there anything more I should add?

dennisdoomen:
As soon as @jnyrup approves the PR, it'll get merged.

Tests/Shared.Specs/ExecutionTimeAssertionsSpecs.cs
```diff
@@ -94,6 +94,33 @@ public void When_the_execution_time_of_an_action_is_less_or_equal_to_a_limit_it_
...
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => someAction.ExecutionTime().Should().BeLessThan(100.Milliseconds());
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act
+                .Should().Throw<XunitException>()
+                .And.Message.Should().StartWith(
```
jnyrup: Suggestion: use `WithMessage`.
E.g. `WithMessage("*action should be less than 0.100s, but it required*")`.

Src/FluentAssertions/Specialized/ExecutionTimeAssertions.cs
```diff
@@ -10,18 +12,39 @@ namespace FluentAssertions.Specialized
...
-            actionDescription = executionTime.ActionDescription;
+            this.execution = executionTime;
+        }
+
+        /// <summary>
+        /// Checks the executing action if it satisfies a condition.
+        /// If the execution runs into an exeption, then this will rethrow it.
+        /// </summary>
+        /// <param name="condition">Condition to check on the current elapsed time.</param>
+        /// <param name="result">The expected result when polling is stopped.</param>
```
jnyrup: Minor:
Maybe it's just me, but `result` indicates a computed value to me, whereas this is a stop condition.
So maybe rename to something like `stopCondition`, `expectedResult` or?

Tests/Shared.Specs/ExecutionTimeAssertionsSpecs.cs
peterekepeter: nice, did not know about this one

Tests/Shared.Specs/ExecutionTimeAssertionsSpecs.cs
```diff
@@ -94,6 +94,33 @@ public void When_the_execution_time_of_an_action_is_less_or_equal_to_a_limit_it_
...
+            {
+                // lets cause a deadlock
+                var semaphore = new Semaphore(0, 1); // my weapon of choice is a semaphore
+                semaphore.WaitOne(); // oops
+            };
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => someAction.ExecutionTime().Should().BeLessThan(100.Milliseconds());
```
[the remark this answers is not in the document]
peterekepeter: done

peterekepeter:
updated, also changed all assertions from that file to use WithMessage(), it looks readable that way, and it's consistent

jnyrup:
@peterekepeter I tried to poke around with the PR and noticed two things:

* If we break out of `while` before the `Task` has completed, the `stopWatch` is not stopped, is that intentional?
As `executionTime.ElapsedTime` is called several times afterwards I guess we could hit a case, where the failure *condition* and failure *message* are inconsistent.

* `Task` already has a `IsCompleted` property, it seems we can use that instead of `IsRunning`?

No tests failed when changing these two things, so I conclude that they are either compatible or we are missing a test.
Any thoughts?

peterekepeter:
- Yes it's intentional, I've added a test. Nothing stops the client code from calling multiple time assertions, so it should give consistent results.
 - It might seem the same, but I get tests failing. The way Tasks are handled might not always be how one might expect, so I think it's better to take the safe side with try finally and a boolean variable.

dennisdoomen:
@jnyrup I assume you're taking care of this PR, right?

jnyrup:
@dennisdoomen I will. I've just been short on spare time, but hope to review the last part during the weekend.

jnyrup:
As we can now break early from a long running action, the failure message is now a bit misleading if we broke out early.

E.g.
```csharp
Action act = () => Thread.Sleep(10.Seconds());
act.ExecutionTime().Should().BeLessThan(50.Milliseconds());
```
> Execution of the action should be less than 0.050s, but it required 0.052s and 30.7µs.

But we don't know how long it *will* take, we just know that it required more than 50 ms.

Maybe adjust parts of the failure message based on `execution.IsRunning`.
If task completed, keep the current failure message,
otherwise: adjust it to e.g. "Execution of the action should be less than 0.050s, but it did not finish after 0.052s and 30.7µs.".
I'm not sure whether to prefer "did not finish after" or "had not finished after", hopefully someone with better English skills than me can answer that.

peterekepeter:
Okay, I've added a "more than" in case there is an early exit, and the message will contain "exactly" if it's not an early exit. This way it does not break existing tests. So you get messages like:
`Execution of the action should be less than 0.100s, but it required more than 0.200s.`
and in case it's not an early exit:
`Execution of the action should be less than 0.100s, but it required exactly 0.200s.`

peterekepeter:
I don't understand AppVeyor's error message. I did not touch the test which failed.

jnyrup:
@peterekepeter, thanks for the contribution to Fluent Assertions!
We appreciate the time you've put into reworking this PR to its merged shape.

peterekepeter:
@jnyrup It was a long process, but finally!

dennisdoomen:
So? Time for another release?

PR #813, opened 2018-03-25
New site based on Minimal Mistakes.

dennisdoomen:
image

image

PR #843, opened 2018-05-10
Re-add lost examples

jnyrup:
The tips and examples somehow got lost in #813 .
Now they're back.


dennisdoomen:
Is that the reason that https://fluentassertions.com/tips/ is empy?

jnyrup:
Yes

PR #850, opened 2018-05-21
Add issue template

jnyrup:
We way to often ask issue openers to provide more information.
Let's help them by providing the questions up front.


dennisdoomen:
Good idea!

PR #853, opened 2018-05-29
Removed two GC tests that are unstable on AppVeyor

dennisdoomen:
Proposal to remove the two tests that keep failing on AppVeyor.


jnyrup:
They also fails for me, when running locally now.
I tried comparing the logs from AppVeyor of the last successful  and first failing runs, but I couldn't find anything there.

The only thing that comes to my mind, is that I recently installed .net 4.7.2 on my machine.

PR #856, opened 2018-06-06
updated exceptions tips

Meir017:
The exception tips are outdated and use the old syntax
```
act.ShouldThrowExactly()
```
instead of
```cs
act.Should().ThrowExactly()
```

I was wondering if about adding more tips for the `WithInnerException` and `WithInnerExceptionExactly` methods as well

something like this:
old:
```cs
act.Should().Throw<Exception>()
    .Which.InnerException.Should().BeOfType<ArgumentException>();
```
new:
```cs
act.Should().Throw<Exception>()
    .WithInnerException<ArgumentException>();
```

@jnyrup

### DIFF
--- a/docs/_data/tips/exceptions.yml
+++ b/docs/_data/tips/exceptions.yml
@@ -1,15 +1,15 @@
 - old: |
-    Action act = () => func();
+    Action act = () => throw new InvalidOperationException("Problems, errorCode2 and more Problems");
     
-    act.ShouldThrowExactly<InvalidOperationException>()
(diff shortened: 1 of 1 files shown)


jnyrup:
Thanks Meir.
Just saw your edits after merging, but tips for `WithInnerException` and `WithInnerExceptionExactly` sounds good.

I noticed that `mstest-migration/exceptions.yml` as well suffers from the old `ShouldThrowExactly`.
If you'd like to fix those as well while you're at it.

PR #861, opened 2018-06-09
Adds support for .NET Core 2.0 and restricts event assertion API.

### LINKED ISSUE #859: Please consider to add NetCoreApp2.0 targetting
Hello,

Since Microsoft unlists a month ago _System.Reflection.Emit_ and _System.Reflection.Emit.Lightweight_ both 4.3.0 version (https://github.com/dotnet/corefx/issues/29365) you should consider to remove the target **netstandard2.0** and specify **netcoreapp2.0**... or at least add **netcoreapp2.0** to the list of all supported targets. (And maybe **netcoreapp2.1** ?)

In fact, I can no longer uses FluentAssertions on a project which performs dynamic assemblies loading because of the following error I had during assembly discovering in units tests:

>An assembly specified in the application dependencies manifest (CKSetup.Runner.deps.json) was not found:
    package: 'System.Security.Cryptography.ProtectedData', version: '4.4.0'
    path: 'runtimes/win/lib/netstandard2.0/System.Security.Cryptography.ProtectedData.dll'

Of course, the previous error disappeared with a **dotnet publish** which will finaly resolves correctly the runtimes, but doing **dotnet publish** before running unit tests is not suitable...

Thank you !

dennisdoomen:
Some features such as ILEmit didn't work in .NET Standard 2.0, but where available up until recently when microsoft pulled System.Reflection.Emit from NuGet. So it appeared the API worked, but only really worked under the runtime platforms. See also https://github.com/dotnet/corefx/issues/29365

* {New} Added support for .NET Core 2.0 in addition to .NET Standard 2.0
* {Fix) Restricted the event assertion API to the run-time .NET platforms only 

Fixes #859

### DIFF
--- a/Src/FluentAssertions.nuspec
+++ b/Src/FluentAssertions.nuspec
@@ -7,12 +7,12 @@
     <owners>Dennis Doomen</owners>
     <authors>Dennis Doomen</authors>
     <summary>
-      Fluent API for asserting the results of unit tests that targets .NET Framework 4.5 and 4.7, as well as .NET Standard 1.3, 1.6 and 2.0.
+      Fluent API for asserting the results of unit tests that targets .NET Framework 4.5 and 4.7, as well as .NET Core 2.0, .NET Standard 1.3, 1.6 and 2.0.
       Supports the unit test frameworks MSTest, MSTest2, Gallio, NUnit, XUnit, MBunit, MSpec, and NSpec.
     </summary>
     <description>
       A very extensive set of extension methods that allow you to more naturally specify the expected outcome of a TDD or
-      BDD-style unit tests. Targets .NET Framework 4.5 and 4.7, as well as .NET Standard 1.3, 1.6 and 2.0.
+      BDD-style unit tests. Targets .NET Framework 4.5 and 4.7, as well as .NET Core 2.0, .NET Standard 1.3, 1.6 and 2.0.
       Supports the unit test frameworks MSTest, MSTest2, Gallio, NUnit, XUnit, MBUnit, MSpec, and NSpec.
     </description>
     <language>en-US</language>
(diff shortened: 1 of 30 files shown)


jnyrup:
I'm not familiar with the IL emitting part of the codebase, but here are some observations.

Places where .NET Core 2.0 is not mentioned:
```
docs/index.html:     BDD-style unit tests. Targets .NET Framework 4.5 and 4.7, as well as .NET Standard 1.3, 1.6 and 2.0.'
docs/index.html:**Targets** .NET 4.5, .NET 4.7, .NET Standard 1.3, 1.6 and 2.0.
docs/_config.yml:description              : "A very extensive set of extension methods that allow you to more naturally specify the expected outcome of a TDD or BDD-style unit tests. Targets .NET Framework 4.5 and 4.7, as well as .NET Standard 1.3, 1.6 and 2.0."
docs/_pages/about.md:Fluent Assertions cross-compiles to .NET Framework 4.5 and 4.7, as well as .NET Standard 1.3, 1.6 and 2.0. Because of that it supports the following minimum platforms.
Src/FluentAssertions/AssemblyInfo.cs:[assembly: AssemblyDescription("Fluent API for asserting the results of unit tests that targets .NET Framework 4.5 and 4.7, as well as .NET Standard 1.3, 1.6 and 2.0. "
```

Places that defines `NETSTANDARD2_0`, but not `NETCOREAPP2_0`
```
Src/FluentAssertions/CallerIdentifier.cs:#if NET45 || NET47 || NETSTANDARD2_0
Src/FluentAssertions/Common/AppSettingsConfigurationStore.cs:﻿#if NET45 || NET47 || NETSTANDARD2_0
Src/FluentAssertions/Common/FullFrameworkReflector.cs:﻿#if NET45 || NET47 || NETSTANDARD2_0
Src/FluentAssertions/Common/Services.cs:#if NET45 || NET47 || NETSTANDARD2_0
Src/FluentAssertions/Execution/AssertionFailedException.cs:#if NET45 || NET47 || NETSTANDARD2_0
Src/FluentAssertions/Execution/AssertionFailedException.cs:#if NET45 || NET47 || NETSTANDARD2_0
Src/FluentAssertions/Execution/AssertionFailedException.cs:#if NET45 || NET47 || NETSTANDARD2_0
Src/FluentAssertions/Execution/GallioTestFramework.cs:#if !NET45 && !NET47 && !NETSTANDARD2_0
Src/FluentAssertions/Execution/LateBoundTestFramework.cs:#if !NET45 && !NET47 && !NETSTANDARD2_0
Src/FluentAssertions/Formatting/Formatter.cs:#if NET45 || NET47 || NETSTANDARD2_0
```

Remove from `EnumAssertionSpecs.cs`
```
using System.Reflection;
using System.Reflection.Emit;
```

`FullFrameworkReflector.cs` 
has `using System.Reflection.Emit`
but preprocessor directives are
`#if NET45 || NET47 || NETSTANDARD2_0`
shouldn't this be 
`#if NET45 || NET47 || NETCOREAPP2_0`?

dennisdoomen:
Thanks for the thorough review @jnyrup. Processed all remarks.

olivier-spinelli:
If I may add a point regarding the multi framework support in general and the one of NetCoreApp 2.1 in particular, the System.Configuration.ConfigurationManager reference should be upgraded from 4.4.0 to the new (netcoreapp2.1 compliant) 4.5.0.  This is from what I can see in https://github.com/fluentassertions/fluentassertions/blob/master/Src/FluentAssertions/FluentAssertions.csproj

Actually, all other current references I can see in master.cs should use their latest 4.5.0 release: System.ValueTuple, System.Reflection.TypeExtensions.

And Microsoft.Extensions.DependencyModel should be upgraded to the current 2.1.0 version.

With those upgrades I think everything should work fine, even with dynamic dll loads...

dennisdoomen:
I kind of kept the versions on the low side to prevent conflicts with other libraries that a consumer using FA would use.

olivier-spinelli:
I perfectly understand this concern and  you are right for netcoreapp2.0 but netcoreapp2.1 should use the 4.5.0 runtime versions... 
And I would like to add that the others are fine with 4.5.0 (since ascending compatibility is guaranteed).

Issue #866, opened 2018-06-21
fluentassertions website tips


Meir017:
I'm working on the `fluentassertions.analyzers` project and I would like to have the ability to reference the actual website to show the improvement

Right not each test case is rendered in the following html
```html
<table class="code-table">
    ...
</table>
```
![image](https://user-images.githubusercontent.com/9786571/41727486-20d12b06-757d-11e8-8192-1c6caf9d24d5.png)

would it be possible to add an `id` attribute to each `table` element?

I'm not sure how the html is generated 😕

jnyrup:
The html is generated by [Jekyll](https://jekyllrb.com/) from this [template](
https://github.com/fluentassertions/fluentassertions/blob/master/docs/_includes/assertion-comparison.html) written in [Liquid](https://shopify.github.io/liquid/).

From the [Liquid documentation](https://help.shopify.com/themes/liquid/objects/for-loops#forloop-index) it seems easy by using the `forloop.index`

Something like `id="example{{forloop.index}}"`.

Meir017:
Looks to me I would need to add an `id` to each of test cases in the `tips` directory, am I right?

jnyrup:
That is not necessary.
Are you working on it? 
Otherwise I almost got it working.

Meir017:
I didn't start working on it yet, I'd be happy to see your solution 👏

dennisdoomen:
Now #867 is merged, can we close this?

PR #867, opened 2018-06-21
Add id to all examples

jnyrup:
From #866 
This PR adds an `id` to all examples on the website, such that one can link directly to them.

Pros:
* No need to add individual ids to each example

Cons:
* If the order of examples is mutated or new examples are not appended to the end, then the links become invalid.


Meir017:
so an exception id will be `Exceptions1` and `Exceptions2` and so on... ?

would it be possible to make the id's `kebab-case` so instead of `ComparableandNumerics1` it will be `comparable-and-numerics-1`?

jnyrup:
@Meir017 changed it to kebab-case.

PR #873, opened 2018-07-05
[docs] strings.yml

Meir017:
@jnyrup  added 
```cs
// old:
actual.Should().BeNull().And.BeEmpty();
// new
actual.Should().BeNullOrEmpty();
```

### DIFF
--- a/docs/_data/tips/strings.yml
+++ b/docs/_data/tips/strings.yml
@@ -95,3 +95,15 @@
 
   new-message: |
     Expected string with length 4, but found string "Actual" with length 6.
+
+- old: |
+    actual.Should().BeNull().And.BeEmpty();
+
+  new: |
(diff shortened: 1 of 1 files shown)


Meir017:
This is the opposite of https://github.com/fluentassertions/fluentassertions/blob/master/docs/_data/tips/strings.yml#L28

jnyrup:
@Meir017 I wouldn't say so.
The one you link to suggests transforming "not null and not empty" into "neither null nor empty" which is logically equivalent since `!A && !B == !(A || B)`.
This PR suggests transforming "null *and* empty" (which cannot happen) into "null *or* empty", which is not the same as `A && B != A || B`

Meir017:
hmm, I hadn't thought of that...

PR #879, opened 2018-07-26
Added support for .NET Core 2.1.

satano:


dennisdoomen:
Thanks for the PR. What does this PR add in addition to just another target? Does it unlock specific .NET Core 2.1 functionality?

satano:
No it does not add/ulock anyting new.

But today we found some behavior in our Team Foundation Server (on premise version). We have a .NET Core 2.1 test library for one of our projects. Recently we upgraded TFS to the latest version. And today we tried to setup an automatic build. But `nuget restore` was failing there with the message that it cannot restore FluentAssertions, because it targets only .NET Core 2.0. It seems this is some new behavior with latest version of `nuget.exe`. Because older version (in our local computer) was OK with this. Just restore using TFS build was failing. We had to change and target this test project to .NET Core 2.0.

So I hope this will make that `nuget.exe` happy.

I can post here exact message tomorrow when I will be at work.

dennisdoomen:
> I can post here exact message tomorrow when I will be at work.

That would be useful, because that doesn't make sense to me.

satano:
So. This is what I found out.

Everything is OK in Visual Studio (VS 2017 15.7.5). VS restores packages, builds the whole solution and tests are passing.

But it is not working from command line. When I run: `nuget.exe restore .\our-solution.sln -PackagesDirectory .\packages -Verbosity Detailed`, it fails with the error (`nuget.exe` version 4.1.0.2450):

> Errors in our-netcoreapp2.1-unittest-project.csproj
>     Package System.Configuration.ConfigurationManager 4.4.0 is not compatible with netcoreapp2.1 (.NETCoreApp,Version=v2.1). Package System.Configuration.ConfigurationManager 4.4.0 supports:
>       - net461 (.NETFramework,Version=v4.6.1)
>       - netstandard2.0 (.NETStandard,Version=v2.0)
>     One or more packages are incompatible with .NETCoreApp,Version=v2.1.

We do not reference `System.Configuration.ConfigurationManager`. It is referenced indirectly by FluentAssertions.

The problem is, that automatic build in TFS uses nuget from command line and so it fails in this step - restoring packages. So it will not continue to build the project.

I built FluentAssertions from this branch, published it in our local NuGet feed and referenced this version in our project. It works this way - at least locally on my computer.

dennisdoomen:
On second thought, it seems to be a VSTS Build-specific thing. See https://github.com/dotnet/core/issues/796#issuecomment-323298924

This is the main reason why I use PSake or Cake to completely encapsulate the build process in a source-controlled script in ALL my projects.

satano:
Hmmm. Thank you for pointing out nuget version. With the latest one (4.7.0) it is not failing. I used the older  one just because I got it from the TFS build agent to have the same version as it has and did not think about it.

So probably the best solution is to update nuget.exe on our TFS if it is possible - we will investigate this.

And so closs this PR?

satano:
The problem is because our build agent uses old version of `nuget.exe`. Solution is tu update our build agent/environment and so this PR is not needed.

PR #886, opened 2018-08-07
Custom rules were not executed on dictionary equivalency assertions

### LINKED ISSUE #530: Dictionary of DateTime comparison. Am I doing something wrong, or is this an issue?
Two dateTimes put in identical Dictionaries, just 1 millisecond difference...

var date1 = new DateTime(2017, 3, 1);
var date2 = date1.AddMilliseconds(1);

var myDict1 = new Dictionary<string, DateTime>() { { "dateField", date1 } };
var myDict2 = new Dictionary<string, DateTime>() { { "dateField", date2 } };

myDict1.ShouldAllBeEquivalentTo(myDict2, 
  options => options.Using< DateTime >
      (item =>   item.Subject.Should().BeCloseTo(item.Expectation, 1000))
   .WhenTypeIs< DateTime >());

Why does the BeCloseTo rule not work? The test fails (Also tried with <KeyValuePair<sting, DateTime>> as the generic type? Confused. Can anybody help. thanks so much!

dennisdoomen:
Fixes #530

### DIFF
--- a/Tests/Shared.Specs/DictionaryEquivalencySpecs.cs
+++ b/Tests/Shared.Specs/DictionaryEquivalencySpecs.cs
@@ -1221,5 +1221,37 @@ namespace FluentAssertions.Specs
             act.Should().Throw<XunitException>()
                 .WithMessage("Expected*String*JValue*");
         }
+
+        [Fact]
+        public void When_a_custom_rule_is_applied_on_a_dictionary_it_should_apply_it_on_the_values()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            var dictOne = new Dictionary<string, double>
+            {
+                { "a", 1.2345 },
+                { "b", 2.4567 },
+                { "c", 5.6789 },
+                { "s", 3.333 }
+            };
+
+            var dictTwo = new Dictionary<string, double>
+            {
+                { "a", 1.2348 },
+                { "b", 2.4561 },
+                { "c", 5.679 },
+                { "s", 3.333 }
+            };
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act / Assert
+            //-----------------------------------------------------------------------------------------------------------
+            dictOne.Should().BeEquivalentTo(dictTwo, options => options
+                .Using<double>(ctx => ctx.Subject.Should().BeApproximately(ctx.Expectation, 0.1))
+                .WhenTypeIs<double>()
+            );
+        }
+
     }
 }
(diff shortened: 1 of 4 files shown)


Tests/Shared.Specs/BasicEquivalencySpecs.cs
```diff
@@ -1085,6 +1085,7 @@ public void When_a_property_is_write_only_it_should_be_ignored()
             action.Should().NotThrow();
         }
 
+
```
jnyrup: Leave this change out.

Tests/Shared.Specs/CollectionEquivalencySpecs.cs
```diff
@@ -6,6 +6,7 @@
 using System.Linq;
 using FluentAssertions.Common;
 using FluentAssertions.Equivalency;
+using FluentAssertions.Execution;
```
jnyrup: I guess this change can be left out as well?

Tests/Shared.Specs/DictionaryEquivalencySpecs.cs
```diff
@@ -1221,5 +1221,37 @@ public void When_a_nested_dictionary_value_doesnt_match_it_should_throw()
             act.Should().Throw<XunitException>()
                 .WithMessage("Expected*String*JValue*");
         }
+
+        [Fact]
+        public void SomeOtherTest()
```
jnyrup: Needs a better test name.

Tests/Shared.Specs/DictionaryEquivalencySpecs.cs
```diff
@@ -1221,5 +1221,37 @@ public void When_a_nested_dictionary_value_doesnt_match_it_should_throw()
...
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            var dictOne = new Dictionary<string, double>
+            {
+                { "a", 1.2345 },
```
jnyrup: Very nitpicking, but I would use C# 6 dictionary initializers.

Tests/Shared.Specs/ExtensibilityRelatedEquivalencySpecs.cs
```diff
@@ -553,7 +553,7 @@ public void When_an_assertion_rule_is_added_it_appear_in_the_exception_message()
         }
 
         [Fact]
-        public void When_multiple_asertion_rules_are_added_they_should_be_evaluated_last_to_first()
+        public void When_multiple_assertion_rules_are_added_they_should_be_evaluated_last_to_first()
```
jnyrup: Depending on how strict/isolated you want to be on commits, this should be moved to it's own commit.

PR #891, opened 2018-08-11
Disabled parallel execution of unit tests

dennisdoomen:
When the `AssertionOptionSpecs` run in parallel with other tests, they can affect the outcome.

### DIFF
--- a/Tests/Shared.Specs/AssertionOptionSpecs.cs
+++ b/Tests/Shared.Specs/AssertionOptionSpecs.cs
@@ -17,7 +17,6 @@ namespace FluentAssertions.Specs
     namespace AssertionOptionsSpecs
     {
 #if NET45 || NET47 || NETCOREAPP2_0
-        [Collection("Equivalency")]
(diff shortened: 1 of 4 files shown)


dennisdoomen:
The problem was that those `AssertionOptionsSpecs` caused the other tests to fail if they ran in parallel. You're saying that this will allow the `AssertionOptionSpecs` to go first without parallelization, followed by the rest in parallel?

PS. I already had bad feelings about me pushing directly without requesting a review.... 😉

jnyrup:
I agree that all tests mutating `AssertionOptions` tests should be run sequentially.
They are mutating static state and should preferably be put in some sort of setup method in user test projects.

Except that it will run parallel test before sequential tests, yes.
https://github.com/xunit/xunit/issues/1179

dennisdoomen:
Well, after having merged this PR, it will now run the assemblies in parallel, but not the test inside an assembly. Unless you disagree, I think that's good enough for now.

Issue #888, opened 2018-08-10
AsyncFunctionAssertions.NotThrow() showing only exception.Message


theprateik:
### Description

With the current implementation of `FluentAssertions.Specialized.AsyncFunctionAssertions.NotThrow()` method, when the test fails, only the `exception.Message` is displayed from the original exception. I think it would be great to show `exception.ToString()` as a whole. Most of the time the `exception.Message` is not enough to understand the error; stack trace and others from the original exception would really help here.

Following is an extraction from current implementation of the method:
```
Execute.Assertion
    .BecauseOf(because, becauseArgs)
    .FailWith("Did not expect any exception{reason}, but found a {0} with message {1}.",
        exception.GetType(), exception.Message);
```

Following is my requested modification:
```
Execute.Assertion
    .BecauseOf(because, becauseArgs)
    .FailWith("Did not expect any exception{reason}, but found a {0} with error {1}.",
        exception.GetType(), exception.ToString());
```

Also,  I can gladly submit a PR if required.

dennisdoomen:
It's a reasonable request. I remember making that decision based on this feeling that the `ToString` implementation usually returns a lot of noise. 

What do you think @jnyrup?

dennisdoomen:
OK. Let's do this.

jnyrup:
This has been fixed in #892.

PR #892, opened 2018-08-13
show error instead of message for async throws

MNie:
## IMPORTANT 

* [x] The code complies with the [Coding Guidelines for C# 3.0, 4.0 and 5.0](http://www.csharpcodingguidelines.com/).

Related to #888 
CC: @jnyrup 
Maybe it would be sensible to create some extension method to check if correct exception (or type of exception) is included in `FailWith` continuation? What do you think?

### DIFF
--- a/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
@@ -73,12 +73,12 @@ namespace FluentAssertions.Specialized
             }
             catch (Exception exception)
             {
-                var deepestException = GetDeepestException(exception);
+                Exception nonAggregateException = GetFirstNonAggregateException(exception);
 
                 Execute.Assertion
                     .BecauseOf(because, becauseArgs)
                     .FailWith("Did not expect any exception{reason}, but found a {0} with message {1}.",
-                        deepestException.GetType(), deepestException);
+                        nonAggregateException.GetType(), nonAggregateException.ToString());
             }
         }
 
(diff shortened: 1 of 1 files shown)


Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -104,22 +101,30 @@ public void NotThrow<TException>(string because = "", params object[] becauseArg
...
+                        .ForCondition(!(deepestException is TException))
                         .BecauseOf(because, becauseArgs)
                         .FailWith("Did not expect {0}{reason}, but found one with message {1}.",
-                            typeof(TException), exception.Message);
+                            typeof(TException), deepestException);
                 }
             }
         }
 
+        private static Exception GetDeepestException(Exception exception)
```
dennisdoomen: 🔧 Isn't this returning the first non-aggregate exception?

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -104,22 +101,30 @@ public void NotThrow<TException>(string because = "", params object[] becauseArg
             }
             catch (Exception exception)
             {
-                while (exception is AggregateException)
-                {
-                    exception = exception.InnerException;
-                }
+                var deepestException = GetDeepestException(exception);
```
dennisdoomen: 🙃 The conventions avoid the use of `var`, unless the type is explicitly stated as a generic parameter.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -73,15 +73,12 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
 
                 Execute.Assertion
                     .BecauseOf(because, becauseArgs)
                     .FailWith("Did not expect any exception{reason}, but found a {0} with message {1}.",
-                        exception.GetType(), exception.Message);
+                        deepestException.GetType(), deepestException);
```
dennisdoomen: ❓ Can we safely assume the `ToString` of an exception always includes the `Message`?

jnyrup: The default override of [`ToString()`](https://referencesource.microsoft.com/#mscorlib/system/badimageformatexception.cs,84ce150b21b0bc42,references) includes `Message` and I think we can rely on that.

For subclasses of `Exception` most seem not to override `ToString()`.

For subclasses overriding `Exception` I would rely on that they know what information is most relevant.
E.g. [`FileNotFoundException`](https://referencesource.microsoft.com/#mscorlib/system/io/filenotfoundexception.cs,e292457f6ecd914d,references).

MNie:
@dennisdoomen all things fix'd :)

dennisdoomen:
Waiting for @jnyrup

jnyrup:
@MNie Thanks for the contribution to Fluent Assertions.

I'm not sure I understand your question about exception type check?

MNie:
@jnyrup I see that there are extension methods like `WithMessage` maybe it would be sensible to add some extension method to check if exact exception was included inside of `FailWith` method or maybe there is some kind of function like that right now?

jnyrup:
@MNie Like a `NotThrowExactly<TException>()`? 
That does not currently exist.

If that's not the case, please open a new issue with an example.
That makes it easier to track discussions, thanks.

PR #895, opened 2018-08-17
Split nested function calls into variables

jnyrup:
When debugging a statement such as `new X().Method()`, I can't easily, as far as I know, step over `new X()` and into `.Method()`.

So this PR makes it easier (for me) to debug as it splits nested function calls into separate variables.

It also has some nice side effects:
* Makes intermediate values visible in debugging
* Gives more informative stack traces if FA throws an Exception, as we do fewer operations per line.

### DIFF
--- a/Src/FluentAssertions/Collections/CollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/CollectionAssertions.cs
@@ -389,7 +389,8 @@ namespace FluentAssertions.Collections
                 Tracer = options.TraceWriter
             };
 
-            new EquivalencyValidator(options).AssertEquality(context);
+            var equivalencyValidator = new EquivalencyValidator(options);
+            equivalencyValidator.AssertEquality(context);
 
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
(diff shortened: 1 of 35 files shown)


dennisdoomen:
Interesting. Since I'm using Rider, I have not run into this specific problem anymore.

jnyrup:
Everything about it, or something particular?
I was aware that it might be style controversial.

How does Rider avoid this problem?
Maybe I'm just missing out on a very useful debugger feature.

dennisdoomen:
Well, if we're going to avoid chaining at all, then we're going too far IMO. I remember that Visual Studio 2017 started showing all the intermediate values in a chained call in the debugger. 

I'm 100% sure Rider does this as well. It suspect it's this setting that allows that (enabled by default):

image

dennisdoomen:
This is what I get in Visual Studio 2017 if I try to debug https://github.com/fluentassertions/fluentassertions/blob/13c7c04be55f2ba74af0497bb61ac2a62b15283f/Src/FluentAssertions/AssertionExtensions.Actions.cs#L37

image

PR #898, opened 2018-08-19
Limits the depth of rendering an object to a max of 5

### LINKED ISSUE #889: OutOfMemory exception when using assertions on HtmlNodeCollection
### Description

When trying to check an HtmlNodeCollection (HtmlAgilityPack) for NullOrEmpty we are getting an OutOfMemory exception. Upon quick investigation there seems to be an infinite loop in Formater.FormatChild method.

```
var col = new HtmlNodeCollection(new HtmlNode(HtmlNodeType.Element, new HtmlDocument(), 0));
col.Add(new HtmlNode(HtmlNodeType.Element, new HtmlDocument(), 0));
col.Should().BeNullOrEmpty();
```

### Expected behavior:

Assertion should fail with error message

### Actual behavior:

OutOfMemoryException

### Versions

FluentAssertions 5.4.1.0
.NET framework 4.6.1

dennisdoomen:
I see no value in rendering a graph 15 levels deep. This should help prevent OOM exceptions while rendering an object graph. 

Fixes #889

### DIFF
--- a/Src/FluentAssertions/Formatting/Formatter.cs
+++ b/Src/FluentAssertions/Formatting/Formatter.cs
@@ -13,8 +13,6 @@ namespace FluentAssertions.Formatting
     {
         #region Private Definitions
 
-        private const int MaxDepth = 15;
-
         private static readonly List<IValueFormatter> customFormatters = new List<IValueFormatter>();
 
         private static readonly List<IValueFormatter> defaultFormatters = new List<IValueFormatter>
@@ -115,7 +113,7 @@ namespace FluentAssertions.Formatting
                 {
                     return $"{{Cyclic reference to type {childValue.GetType()} detected}}";
                 }
-                else if (graph.Depth > MaxDepth)
+                else if (graph.Depth > 5)
                 {
                     return "{Maximum recursion depth was reached…}";
                 }
@@ -198,6 +196,11 @@ namespace FluentAssertions.Formatting
             }
 
             public int Depth => (pathStack.Count - 1);
+
+            public override string ToString()
+            {
+                return String.Join(".", pathStack.Reverse().ToArray());
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Src/FluentAssertions/Formatting/Formatter.cs
```diff
@@ -115,7 +113,7 @@ private static string FormatChild(string path, object childValue, bool useLineBr
                 {
                     return $"{{Cyclic reference to type {childValue.GetType()} detected}}";
                 }
-                else if (graph.Depth > MaxDepth)
+                else if (graph.Depth > 5)
```
Meir017: Maybe make this configurable instead of constant?

dennisdoomen: Personally I don't see why anybody would ever see an object graph in their unit test output window that goes beyond 5 levels deep.

Meir017: Maybe for comparing json or xml objects?

jnyrup:
A lot of the problem seems to be that `HtmlNode` declare it's backing fields as `internal` which we also print along the public property. 

Couldn't we detect that e.g. `_ownerDocument` and `OwnerDocument` references the same object in memory and only print out `OwnerDocument`?

Src/FluentAssertions/Formatting/Formatter.cs
dennisdoomen: For that we have specialized APIs that don't require cyclic reference detection (which is the real cause of the problem mentioned here)

dennisdoomen:
I believe we already skip it because of the recursive detection. The real issue is that we somehow include private and internal fields and properties. I think that's something we should change (in a separate PR).

dennisdoomen:
> When I run the test, the failure message both contains the internal field _ownerdocument and the public property OwnerDocument.
> OwnerDocument is just returning _ownerdocument.
> https://github.com/zzzprojects/html-agility-pack/blob/master/src/HtmlAgilityPack.Shared/HtmlNode.cs
> It's the same case for _parentnode and ParentNode
> 
> Could it be that we also select internal members that are not marked as VisibleTo the running test?
> It seems we also pick internal members of Encoding

We should fix both of them.

PR #904, opened 2018-08-27
Add generic overloads of BeEquivalentTo

### LINKED ISSUE #903: Adding to Should().BeEquivalentTo() options.WithTracing() changes outcome of the test
### Description

I got an error message which I think should not have happened, since the only difference is the order of the item and by default order shouldn't break equivalence. So I tried adding options.WithTracing() to understand what's going on, but then the test passes. This is the output without WithTracing():

> Message: Expected item[0] to be 
> 
> Deal.DealComputer+Deal
> {
>    Items = {Deal.DealComputer+SingleItem
>       {
>          Name = "Croissant"
>          Price = 12
>          Tags = {Pastry}
>       }, Deal.DealComputer+SingleItem
>       {
>          Name = "Cappuccino"
>          Price = 11
>          Tags = {HotDrink}
>       }}
>    Name = "Coffee + Pastry"
>    Price = 19
>    Tags = {HotDrink, Pastry}
> }, but found 
> 
> Deal.DealComputer+Deal
> {
>    Items = {Deal.DealComputer+SingleItem
>       {
>          Name = "Cappuccino"
>          Price = 11
>          Tags = {HotDrink}
>       }, Deal.DealComputer+SingleItem
>       {
>          Name = "Croissant"
>          Price = 12
>          Tags = {Pastry}
>       }}
>    Name = "Coffee + Pastry"
>    Price = 19
>    Tags = {HotDrink, Pastry}
> }.
> 
> With configuration:
> - Use declared types and members
> - Compare enums by value
> - Include all non-private properties
> - Include all non-private fields
> - Match member by name (or throw)
> - Without automatic conversion.
> - Without automatic conversion.
> - Be strict about the order of items in byte arrays

```csharp

public enum Tag
    {
        HotDrink,
        Pastry,
        Breakfast
    }

public interface Dish
        {
            string Name { get; set; }
            int Price { get; set; }
            IEnumerable<Tag> Tags { get; set; }
        }

        public class SingleItem : Dish, IEquatable<SingleItem>
        {
            public int Price { get; set; }
            public IEnumerable<Tag> Tags { get; set; }

            public override bool Equals(object obj)
                => Equals(obj as SingleItem);

            public bool Equals(SingleItem other) 
                => !(other is null) &&
                   Name == other.Name &&
                   Price == other.Price &&
                   Tags.SequenceEqual(other.Tags);

                => HashCode.Combine(Name, Price, Tags);

            public static bool operator ==(SingleItem item1, SingleItem item2)
            {
                if (item1 is null)
                {
                    return item2 is null;
                }

                return item1.Equals(item2);
            }

            public static bool operator !=(SingleItem item1, SingleItem item2)
                => !(item1 == item2);
        }

        public class Deal : Dish, IEquatable<Deal>
        {
            public int Price { get; set; }
            public IEnumerable<Tag> Tags { get; set; }
            public IEnumerable<SingleItem> Items { get; set; }

            public override bool Equals(object obj) => Equals(obj as Deal);

            public bool Equals(Deal other)
            {
                return !(other is null) &&
                       Name == other.Name &&
                       Price == other.Price &&
                       Tags.SequenceEqual(other.Tags) &&
                       Items.SequenceEqual(other.Items);
            }

            public override int GetHashCode() => HashCode.Combine(Name, Price, Tags, Items);

            public static bool operator ==(Deal deal1, Deal deal2)
            {
                if (deal1 is null)
                {
                    return deal2 is null;
                }

                return deal1.Equals(deal2);
            }

            public static bool operator !=(Deal deal1, Deal deal2) => !(deal1 == deal2);
        }

        readonly SingleItem cappuccino = new SingleItem { Name = "Cappuccino", Price = 11, Tags = new[] { Tag.HotDrink }};
        readonly SingleItem croissant = new SingleItem { Name = "Croissant", Price = 12, Tags = new[] { Tag.Pastry } };
        readonly SingleItem tea = new SingleItem { Name = "Tea", Price = 8, Tags = new[] { Tag.HotDrink } };
        readonly SingleItem datesPastry = new SingleItem { Name = "Dates Pastry", Price = 10, Tags = new[] { Tag.Pastry } };
        readonly Deal CoffeeAndPastryDeal = new Deal { Name = "Coffee + Pastry", Price = 19, Tags = new [] { Tag.HotDrink, Tag.Pastry } };

        [Fact]
        public void FourItemsSingleDeal()
{
            var items = new[] {
                tea,
                datesPastry,
                cappuccino,
                croissant,
            };

            var deals = new[] { CoffeeAndPastryDeal };

            var completedDeal1 = CoffeeAndPastryDeal.DeepClone();
            completedDeal1.Items = new[] { croissant, cappuccino };
            var expected = new Dish[]
            {
                completedDeal1,
                tea,
                datesPastry
            };

            var completedDeal2 = CoffeeAndPastryDeal.DeepClone();
            completedDeal2.Items = new[] { cappuccino, croissant };

            var actual =
                new Dish[]
            {
                completedDeal2,
                tea,
                datesPastry
            };

            actual.Should().BeEquivalentTo(expected, o => o.WithTracing()); // passes
            actual.Should().BeEquivalentTo(); //fails
}
```

### Expected behavior:

I expect options.WithTracing() to not affect the outcome of tests

### Actual behavior:

options.WithTracing() affects the outcome of tests

### Versions

* Which version of Fluent Assertions are you using? 5.4.1
* Which .NET runtime and version are you targeting? .net core 2.1
* For deep cloning I'm using nuget DeepCloner 0.10.2

jnyrup:
When using an array as the expectation for `BeEquivalentTo`, the behavior varies whether you provide e.g. the identity config `opt => opt`, as it changes the picked overload from the non-generic 
```c#
BeEquivalentTo(params object[] expectations)
``` 
to the generic 
```c#
BeEquivalentTo<TExpectation>(IEnumerable<TExpectation> expectation, ...)
```

This PR adds two generic overloads to give a more consistent / less surprising experience.

This fixes #903

### DIFF
--- a/Src/FluentAssertions/Collections/CollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/CollectionAssertions.cs
@@ -323,6 +323,89 @@ namespace FluentAssertions.Collections
         /// <param name="becauseArgs">
         /// Zero or more objects to format using the placeholders in <see cref="because" />.
         /// </param>
+        public AndConstraint<TAssertions> BeEquivalentTo<TExpectation>(TExpectation[] expectations, string because = "",
+            params object[] becauseArgs)
+        {
+            BeEquivalentTo(expectations, config => config, because, becauseArgs);
+
+            return new AndConstraint<TAssertions>((TAssertions)this);
+        }
+
+        /// <summary>
+        /// Asserts that a collection of objects is equivalent to another collection of objects.
+        /// </summary>
+        /// <remarks>
+        /// Objects within the collections are equivalent when both object graphs have equally named properties with the same
+        /// value,  irrespective of the type of those objects. Two properties are also equal if one type can be converted to another
+        /// and the result is equal.
+        /// The type of a collection property is ignored as long as the collection implements <see cref="IEnumerable"/> and all
+        /// items in the collection are structurally equal.
+        /// </remarks>
+        /// <param name="config">
+        /// A reference to the <see cref="EquivalencyAssertionOptions{TSubject}"/> configuration object that can be used
+        /// to influence the way the object graphs are compared. You can also provide an alternative instance of the
+        /// <see cref="EquivalencyAssertionOptions{TSubject}"/> class. The global defaults are determined by the
+        /// <see cref="AssertionOptions"/> class.
+        /// </param>
+        /// <param name="because">
+        /// An optional formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the
+        /// assertion is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> BeEquivalentTo<TExpectation>(TExpectation[] expectations,
+            Func<EquivalencyAssertionOptions<TExpectation>, EquivalencyAssertionOptions<TExpectation>> config, string because = "",
+            params object[] becauseArgs)
+        {
+            BeEquivalentTo((IEnumerable<TExpectation>)expectations, config, because, becauseArgs);
+
+            return new AndConstraint<TAssertions>((TAssertions)this);
+        }
+
+        /// <summary>
+        /// Asserts that a collection of objects is equivalent to another collection of objects.
+        /// </summary>
+        /// <remarks>
+        /// Objects within the collections are equivalent when both object graphs have equally named properties with the same
+        /// value, irrespective of the type of those objects. Two properties are also equal if one type can be converted to another
+        /// and the result is equal.
+        /// The type of a collection property is ignored as long as the collection implements <see cref="IEnumerable"/> and all
+        /// items in the collection are structurally equal.
+        /// Notice that actual behavior is determined by the global defaults managed by <see cref="AssertionOptions"/>.
+        /// </remarks>
+        /// <param name="because">
+        /// An optional formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the
+        /// assertion is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> BeEquivalentTo(IEnumerable expectations, string because = "", params object[] becauseArgs)
+        {
+            BeEquivalentTo(expectations.Cast<object>(), because, becauseArgs);
+
+            return new AndConstraint<TAssertions>((TAssertions)this);
+        }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


jnyrup:
One caveat is potential breaking change at runtime.

If one previously used the non-generic overload and asserted the following to throw, it would pass because it discarded the `IInterface` type of `expected` and used the runtime types of `MyClass`.

Then upgrading to a version of FA containing this PR would break the test,  because it would now use the defined properties of `IInterface` and not throw as asserted. 

```c#
public interface IInterface
{
    int Value { get; set; }
}

public class MyClass : IInterface
{
    public int Value { get; set; }
    public int Tag { get; set; }
}

[Fact]
public void Test_breaking_change()
{
    var actual = new IInterface[] { new MyClass() { Value = 1, Tag = 1 } };
    var expected = new IInterface[] { new MyClass() { Value = 1, Tag = 2 } };

    Action act = () => actual.Should().BeEquivalentTo(expected);

}
```

dennisdoomen:
> If one previously used the non-generic overload and asserted the following to throw, it would pass because it discarded the IInterface type of expected and used the runtime types of MyClass.

But that would have been a false positive, right?

jnyrup:
If one expected the assertion to take the runtime types into account, the test was correct.
This change would require you to speccify `RespectingRuntimeTypes()`.

Actually, I'll like to fiddle a bit more with this before merging. 
Just to get a better feeling that I didn't miss something.

jnyrup:
I think I got it now.

The difficult part is to determine when `params object[] expectations` is multiple items or an array of multiple items.
E.g. should `BeEquivalentTo(new[] { 1, 2 })` be treated as two items, or a single item which *then* is an array of two items?
That depends whether subject to compare against is
* `new[] { new[] { 1, 2 } }`, or
* `new[] {1, 2 }`.

I don't think this can be solved, as it depends on the subject how to interpret to expectation.

What we can do is add more overloads of `BeEquivalentTo` such that explicit array creation will pick another overload.
E.g.
* `BeEquivalentTo(1, 2)` still chooses `BeEquivalentTo(params object[] expectations)`
* `BeEquivalentTo(new[] { 1, 2})` now chooses `BeEquivalentTo<T>(params T[])`

dennisdoomen:
The scary part of these kinds of changes is that it might cause unexpected side-effects in existing code-bases.

jnyrup:
> Versioning is hard and makes my head hurt.
https://codeblog.jonskeet.uk/2018/04/13/backward-compatibility-and-overloading/

I guess this PR is equivalent to the Generics section in the blog post above.
https://semver.org/ is not very specific on what kind of "broken compatibility" requires increasing the MAJOR version.

dennisdoomen:
So when you added those overloads, did it break anything?

jnyrup:
It didn't break anything in the current tests, because we didn't have a test exercises this path.
`When_the_expectation_is_an_array_of_interface_type_it_should_respect_interface_members` is an example of a test that fails in master but succeeds with this PR.

jnyrup:
I'm closing this in favor of #908 .

PR #908, opened 2018-09-03
Add tests to prevent breaking changes

jnyrup:
The changes proposed in #904 can cause a breaking change at runtime if one uses an array of interface types and implicitly asserts that runtime properties are respected.
Even though the consistent way would be to respect the compile type and require `RespectRuntimeTypes()`, I'd like to keep my breaking change score at 1. #762 

This PR instead adds tests to help anyone who tries to fix the inconsistent behavior of `BeEquivalentTo(params object[])`, at least for 5.x.x, not to cause a breaking change.

### DIFF
--- a/Tests/Shared.Specs/CollectionEquivalencySpecs.cs
+++ b/Tests/Shared.Specs/CollectionEquivalencySpecs.cs
@@ -14,6 +14,17 @@ namespace FluentAssertions.Specs
 {
     public class CollectionEquivalencySpecs
     {
+        public interface IInterface
+        {
+            int InterfaceProperty { get; set; }
+        }
+
+        public class MyClass : IInterface
+        {
+            public int InterfaceProperty { get; set; }
+            public int ClassProperty { get; set; }
+        }
+
         public class SubDummy
         {
             public SubDummy(int id)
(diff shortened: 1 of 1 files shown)


Tests/Shared.Specs/CollectionEquivalencySpecs.cs
```diff
@@ -181,6 +192,117 @@ public void Add(Guid userId, params string[] roles)
...
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => actual.Should().BeEquivalentTo(expected);
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act.Should().NotThrow("Subject and expectation has same degree of nesting.");
```
dennisdoomen: 🤔 I'm not sure what your trying to say with the because. Same with other test.

Tests/Shared.Specs/CollectionEquivalencySpecs.cs
```diff
@@ -181,6 +192,117 @@ public void Add(Guid userId, params string[] roles)
...
+        }
+
+        [Fact]
+        public void When_the_expectation_is_less_nested_than_the_subject_it_should_fail()
```
dennisdoomen: 🔧 Maybe _has less dimensions_?

Issue #918, opened 2018-09-18
BeEquivalentTo with WithStrictOrdering produces messy failure message


shuebner:
### Description

Comparing collections with option "WithStrictOrdering" produces messy failure messages.

### Complete minimal example reproducing the issue
```csharp
new[] { 1, 2, 3 }.Should().BeEquivalentTo(new[] { 1, 3, 2 }, opts => opts.WithStrictOrdering());
```

### Expected behavior:

Clear failure message about the difference between the lists.

### Actual behavior:

Confusing and malformatted message with wrong statements:

Expected item[1] to be a collection with 3 item(s)Expected item[1] to be 3, but found 2.
Expected item[2] to be a collection with 3 item(s)Expected item[2] to be 2, but found 3.

### Versions

FluentAssertions 5.4.2.0
.NET Framework 4.6.1

dennisdoomen:
> I also found it as part of #920. It is already fixed there.

That would surprise me. I'm still fixing it by changing the way assertion scopes get chained.

PR #920, opened 2018-09-19
Improve performance of AllBeEquivalentTo

### LINKED ISSUE #914: Assert that a large enumerable contains only the very same item n times?
I have a collection of 100K sample numbers. In one test case I want to assert that all of them are of the same value. I tried:

    values.Should().AllBeEquivalentTo(5);

This takes quite some time (around five seconds). Which is WAY to long. Is this a bug in the implementation? Why is it taking so long? I think it's the wrong function, but I couldn't find a `.AllBe(5)` for which I was looking for originally.

My workaround now is `values.Distinct().Should().BeEquivalentTo(new[]{5});` but I don't like it because it doesn't express my intent.

Ideas on how to do that properly?

krajek:
Attempt to fix #914.

There are (at least) two bottlenecks:

- [x] `AllBeEquivalentTo` created a collection of items matching the subject's length, and then used loose ordering which results in O(n^2) complexity. Change to strict ordering is necessary to have hope for any kind of reasonable performance.
- [x] Building collection of messages. The more inequivalent items there is the more time it takes to create assert failure messages. I think that some kind of early break is needed in `EnumerableEquivalencyValidator` after say 10 failures.
- [x] Extract `OrderingRules.IsOrderingStrictFor(context)` in `EnumerableEquivalencyValidator.AssertElementGraphEquivalency` before the loop. During profiling, it stands out as an easy-fix with the execution time around 8-10% spent in this method.

I checked that after those three issues there are no more outstanding hot spots. Execution time is split among some `CanHandle` methods, creating objects like `EquivalencyValidationContext` and so on.

Any ideas are warmly welcomed.

### DIFF
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -331,12 +331,22 @@ namespace FluentAssertions.Collections
         /// Zero or more objects to format using the placeholders in <see cref="because" />.
         /// </param>
         public void AllBeEquivalentTo<TExpectation>(TExpectation expectation,
-            Func<EquivalencyAssertionOptions<TExpectation>, EquivalencyAssertionOptions<TExpectation>> config, string because = "",
+            Func<EquivalencyAssertionOptions<TExpectation>, EquivalencyAssertionOptions<TExpectation>> config,
+            string because = "",
             params object[] becauseArgs)
         {
             TExpectation[] repeatedExpectation = RepeatAsManyAs(expectation, Subject).ToArray();
 
-            BeEquivalentTo(repeatedExpectation, config, because, becauseArgs);
+            // Because we have just manually created the collection based on single element
+            // we are sure that we can force string ordering, because ordering does not matter in terms
+            // of correctness. On the other hand we do not want to change ordering rules for nested objects
+            // in case user needs to use them. Strict ordering improves algorithmic complexity
+            // from O(n^2) to O(n). For bigger tables it is necessary in order to achieve acceptable
+            // execution times.
+            Func<EquivalencyAssertionOptions<TExpectation>, EquivalencyAssertionOptions<TExpectation>> forceStringOrderingConfig =
+                x => config(x).WithStrictOrderingFor(s => s.SelectedMemberPath == "");
+
+            BeEquivalentTo(repeatedExpectation, forceStringOrderingConfig, because, becauseArgs);
         }
 
         private static IEnumerable<TExpectation> RepeatAsManyAs<TExpectation>(TExpectation value, IEnumerable<T> enumerable)
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -338,7 +338,7 @@ namespace FluentAssertions.Collections
             TExpectation[] repeatedExpectation = RepeatAsManyAs(expectation, Subject).ToArray();
 
             // Because we have just manually created the collection based on single element
-            // we are sure that we can force string ordering, because ordering does not matter in terms
+            // we are sure that we can force strict ordering, because ordering does not matter in terms
             // of correctness. On the other hand we do not want to change ordering rules for nested objects
             // in case user needs to use them. Strict ordering improves algorithmic complexity
             // from O(n^2) to O(n). For bigger tables it is necessary in order to achieve acceptable
--- a/Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidator.cs
+++ b/Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidator.cs
@@ -62,22 +62,27 @@ namespace FluentAssertions.Equivalency
             unmatchedSubjectIndexes = new List<int>(subjects.Length);
             unmatchedSubjectIndexes.AddRange(Enumerable.Range(0, subjects.Length));
 
-            foreach (int index in Enumerable.Range(0, expectations.Length))
+            if (OrderingRules.IsOrderingStrictFor(context))
             {
-                T expectation = expectations[index];
+                foreach (int index in Enumerable.Range(0, expectations.Length))
+                {
+                    T expectation = expectations[index];
 
-                if (!OrderingRules.IsOrderingStrictFor(context))
-                {
-                    using (context.TraceBlock(path => $"Finding the best match of {expectation} within all items in {subjects} at {path}[{index}]"))
-                    {
-                        LooselyMatchAgainst(subjects, expectation, index);
-                    }
-                }
-                else
-                {
                     using (context.TraceBlock(path => $"Strictly comparing expectation {expectation} at {path} to item with index {index} in {subjects}"))
                     {
                         StrictlyMatchAgainst(subjects, expectation, index);
+                    }
+                }
+            }
+            else
+            {
+                foreach (int index in Enumerable.Range(0, expectations.Length))
+                {
+                    T expectation = expectations[index];
+                    
+                    using (context.TraceBlock(path => $"Finding the best match of {expectation} within all items in {subjects} at {path}[{index}]"))
+                    {
+                        LooselyMatchAgainst(subjects, expectation, index);
                     }
                 }
             }
--- a/Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidator.cs
+++ b/Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidator.cs
@@ -74,10 +74,11 @@ namespace FluentAssertions.Equivalency
                         bool succeed = StrictlyMatchAgainst(subjects, expectation, index);
                         if(!succeed)
                         {
+                            const int limit = 10;
                             failedCount++;
                             if(failedCount >= 10)
                             {
-                                //trace;
+                                context.TraceSingle(path => $"Fail failing strict order comparison of collection after {limit} items failed at {path}");
                                 break;
                             }
                         }
--- a/Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidator.cs
+++ b/Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidator.cs
@@ -11,6 +11,8 @@ namespace FluentAssertions.Equivalency
     /// </summary>
     internal class EnumerableEquivalencyValidator
     {
+        private const int FailedItemsFastFailThreshold = 10;
+
         #region Private Definitions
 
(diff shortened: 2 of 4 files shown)


Tests/Shared.Specs/CollectionEquivalencySpecs.cs
```diff
@@ -900,14 +900,45 @@ public void When_all_subject_items_are_not_equivalent_to_expectation_object_it_s
...
                 "Expected item[1] to be 1, but found 2.*Expected item[2] to be 1, but found 3*");
         }
 
+        [Fact]
+        public void When_some_subject_items_are_not_equivalent_to_expectation_for_huge_table_execution_time_should_still_be_short()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            const int N = 100000;
```
jnyrup: 🔧 How about using `Enumerable.Range(1, 100_000).ToList()`?

Tests/Shared.Specs/CollectionEquivalencySpecs.cs
```diff
@@ -900,14 +900,45 @@ public void When_all_subject_items_are_not_equivalent_to_expectation_object_it_s
...
+            Action action = () =>
+            {
+                try { subject.Should().AllBeEquivalentTo(1); }
+                catch (Exception)
+                {
+                    // ignored, we only care about execution time
+                }
+            };
+
+            ////-----------------------------------------------------------------------------------------------------------
```
jnyrup: 🙃 `////`

jnyrup:
Out of curiosity, Did you measure the wall time improvement of this PR for the new test?

krajek:
What does "wall time" mean? :-)

jnyrup:
Wall time is just another word for the runtime measured in e.g. seconds instead of algorithmic complexity.
E.g. if applying this PR improves the test from 100s to 10s.

Tests/Shared.Specs/CollectionEquivalencySpecs.cs
krajek: Notice that first item is not equal to the rest. If the test will ultimately use single value I will refactor to `Enumerable.Range`.

krajek:
I tried to measure the time for 10000 items but gave up after a few minutes :-). Even for a few hundreds of items times were in the order of minutes. With strict ordering and 100 000 items, I have around 13s.

krajek:
FYI: Last commit(502977c) fixed issue unrelated to the topic of the PR. When I set strict ordering for the top-level collection exception message for existing test case changed. The reason was the discrepancy between `LooselyMatchAgainst` and `StrictlyMatchAgainst`.
`LooselyMatchAgainst` created scope for each item whereas `StrictlyMatchAgainst` didn't. 

Please review that particular scenario, I am not sure I understand the code deeply enough.

dennisdoomen:
I'm loosing track of what each PR is supposed to. Normally, if a PR consists of multiple commits, I assume you have carefully squashed individual changes into a couple of well-focused commits. We also merge these PRs by keeping all commits. So please either separate unrelated changes into separate PRs or make sure you squash commits that change stuff in previous commits into the same one. I regularly use [fixup](https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html) commits and/or amend existing commits.

Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidator.cs
```diff
@@ -62,35 +64,73 @@ private void AssertElementGraphEquivalency<T>(object[] subjects, T[] expectation
...
-                    using (context.TraceBlock(path => $"Finding the best match of {expectation} within all items in {subjects} at {path}[{index}]"))
+                    bool succeed = StrictlyMatchAgainst(subjects, expectation, index);
+                    if (!succeed)
                     {
-                        LooselyMatchAgainst(subjects, expectation, index);
+                        failedCount++;
+                        if (failedCount >= FailedItemsFastFailThreshold)
+                        {
+                            context.TraceSingle(path =>
+                                $"Fail failing strict order comparison of collection after {FailedItemsFastFailThreshold} items failed at {path}");
```
dennisdoomen: 🤔 I _think_ you mean that you already found 10 mismatches and decided to abort the comparison for performance reasons, right?

Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -284,6 +284,12 @@ public string[] Discard()
             return assertionStrategy.DiscardFailures().ToArray();
         }
 
+        public bool AnyFailures()
```
dennisdoomen: 🤔 Why is it a method and not a property? 
🙃 I would call it `HasFailures`

Tests/Shared.Specs/CollectionEquivalencySpecs.cs
```diff
@@ -900,14 +922,66 @@ public void When_all_subject_items_are_not_equivalent_to_expectation_object_it_s
...
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action action = () =>
+            {
+                try { subject.Should().AllBeEquivalentTo(1); }
+                catch (Exception)
+                {
+                    // ignored, we only care about execution time
```
dennisdoomen: 👍 That _is_ a very helpful comment.

Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidator.cs
```diff
@@ -137,9 +179,20 @@ private string[] TryToMatch<T>(object subject, T expectation, int expectationInd
...
 
-        private void StrictlyMatchAgainst<T>(object[] subjects, T expectation, int expectationIndex)
+        private bool StrictlyMatchAgainst<T>(object[] subjects, T expectation, int expectationIndex)
         {
-            parent.AssertEqualityUsing(context.CreateForCollectionItem(expectationIndex.ToString(), subjects[expectationIndex], expectation));
+            using (var scope = new AssertionScope())
+            {
+                object subject = subjects[expectationIndex];
+                string indexString = expectationIndex.ToString();
+                var equivalencyValidationContext = context
```
dennisdoomen: 🙃 In this code-base, we prefer explicit types, unless the type itself is clear already.

Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidator.cs
```diff
@@ -62,35 +64,73 @@ private void AssertElementGraphEquivalency<T>(object[] subjects, T[] expectation
...
+            int failedCount = 0;
+            foreach (int index in Enumerable.Range(0, expectations.Length))
+            {
+                T expectation = expectations[index];
+
+                using (context.TraceBlock(path =>
+                    $"Finding the best match of {expectation} within all items in {subjects} at {path}[{index}]"))
                 {
-                    using (context.TraceBlock(path => $"Strictly comparing expectation {expectation} at {path} to item with index {index} in {subjects}"))
+                    bool succeed = LooselyMatchAgainst(subjects, expectation, index);
```
dennisdoomen: 🙃 It's either `success` or `succeeded`

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -331,12 +331,22 @@ private bool IsValidProperty<TSelector>(Expression<Func<T, TSelector>> propertyE
...
-            Func<EquivalencyAssertionOptions<TExpectation>, EquivalencyAssertionOptions<TExpectation>> config, string because = "",
+            Func<EquivalencyAssertionOptions<TExpectation>, EquivalencyAssertionOptions<TExpectation>> config,
+            string because = "",
             params object[] becauseArgs)
         {
             TExpectation[] repeatedExpectation = RepeatAsManyAs(expectation, Subject).ToArray();
 
-            BeEquivalentTo(repeatedExpectation, config, because, becauseArgs);
+            // Because we have just manually created the collection based on single element
+            // we are sure that we can force string ordering, because ordering does not matter in terms
```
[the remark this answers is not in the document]
krajek: Good catch

Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidator.cs
krajek: Ok, fixed.

Tests/Shared.Specs/CollectionEquivalencySpecs.cs
```diff
@@ -324,6 +324,28 @@ public void When_a_collection_does_not_match_it_should_include_items_in_message(
...
         }
 
+        [Fact]
+        public void When_collection_of_same_count_does_not_match_it_should_include_at_most_10_items_in_message()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            const int commonLength = 11;
+            var subject = new int[commonLength] { 0, 1, 2, 2, 4, 5, 6, 7, 8, 9, 10 };
```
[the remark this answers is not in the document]
krajek: Repeated 2 fixed, I have added a comment explaining why I'm not using `Enumerable.Repeat` this time

Src/FluentAssertions/Equivalency/EnumerableEquivalencyValidator.cs
krajek: "Fail failing strict order comparison" - jeez what was I thinking? Fixed.

krajek:
First of all thanks for mentioning fixup commits, I had no idea such a thing exists.
Secondly, thanks for describing the merging policy. I am more used to no-constraints feature branches merged with squash to the main branch. I would be fine with such final for this PR as well.

Regarding the topic of this PR. Every commit but one is directly tackling performance problem as stated in the PR title. The one outstanding commit (502977c) was necessary (at least at the time of writing) because after I introduced the `WithStrictOrdering` usage I accidentally reproduced the #918 issue. The fix in my case what to introduce new `AssertionScope` for strict ordering path just like a loose ordering path was already doing.

Now we need to decide how to proceed. Possible scenarios:

A) After my explanation, you understand the scope of the feature and you merge it with squash.
B) I abandon this particular PR and split it into three, one for each "sub-task".
@dennisdoomen would you be ok with A) this time?

In either case from now on, for new PR's I will prepare my branches
1) to be as small as possible
2) to be prepared for a merge with all the commits using fixup/amend/rebase etc.

dennisdoomen:
> would you be ok with A) this time?

Absolutely. My intention was just to explain my preference for this repo.

Issue #897, opened 2018-08-17
BeEquivalentTo shows enums as decimal in error message


robvanuden:
### Description
BeEquivalentTo shows enums as decimal in error message
```csharp
public enum Color
{
    Red,Blue
}
```
`Expected member Color to be 1M, but found 0M.`

### Complete minimal example reproducing the issue

```csharp
public enum Color
{
    Red,Blue
}

public class Car
{
    public Color Color { get; set; }
}

[Fact]
public void Test()
{
    var car = new Car
    {
        Color = Color.Red
    };
    car.Should().BeEquivalentTo(new Car() { Color = Color.Blue });
}
```

### Expected behavior:
`Message: Expected member Color to be Blue, but found Red.`

### Actual behavior:
`Message: Expected member Color to be 1M, but found 0M.`

### Versions

* FluentAssertions 5.4.1
* .NET Core 2.0

### Additional Information

Running in Xunit, when relevant.

jnyrup:
Thanks for bringing this up.

The default behavior for comparing `enum` in `BeEquivalentTo` is to compare them by value.
That is, two different enums are considered equivalent if and only if there underlying values are equal.

If you want to compare the enums by name instead, you can use

```c#
car.Should().BeEquivalentTo(new Car() { Color = Color.Blue },
    opt => opt.ComparingEnumsByName());
```

which changes the failure message to

```
Message: Expected member Color to be "Blue" with a length of 4, but "Red" has a length of 3.

With configuration:
- Use declared types and members
- Compare enums by name
- Match member by name (or throw)
- Without automatic conversion.
- Be strict about the order of items in byte arrays
```

Getting back to your issue, I think the failure message can be improved for both `ComparingEnumsByName` and `ComparingEnumsByValue`.

Some ideas for the failure message:
* `Expected member Color to be Color.Blue, but found Color.Red.`
* `Expected member Color to be Blue (1), but found Red (0).`
* `Expected member Color to be Color.Blue (1), but found Color.Red (0).`

Including more details is useful if one e.g. compares `Color.Red = 0` with `AnotherColor.Red = 1` by value. 
Only displaying `Message: Expected member Color to be Blue, but found Red.` would be confusing.

The relevant class is [`EnumEqualityStep`](https://github.com/fluentassertions/fluentassertions/blob/master/Src/FluentAssertions/Equivalency/EnumEqualityStep.cs).

robvanuden:
[...]

PR #923, opened 2018-09-24
Improve enum comparison assert messages

krajek:
Applies changes discussed in #897.

Example of the new message:
`"Expected enum to be EnumFour.Three(3) by value, but found EnumOne.Two(3)`

Reviews are welcome.


Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
```diff
@@ -56,5 +52,59 @@ public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAsserti
 
             return true;
         }
+
+        private static void HandleByValue(IEquivalencyValidationContext context)
+        {
+            decimal? subjectsUnderlyingValue = ExtractDecimal(context.Subject);
+            decimal? expectationsUnderlyingValue = ExtractDecimal(context.Expectation);
+
+            string subjectsName = EnumDescription(context.Subject, subjectsUnderlyingValue);
```
jnyrup: If #921 get's merged before this one, the lazy `FailWith` could be used here.

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
```diff
@@ -56,5 +52,59 @@ public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAsserti
...
+            string expectationName = EnumDescription(context.Expectation, expectationsUnderlyingValue);
+
+            Execute.Assertion
+                .ForCondition(subjectsUnderlyingValue == expectationsUnderlyingValue)
+                .FailWith($"Expected enum to be {expectationName}{{reason}}, but found {subjectsName}.");
```
jnyrup: ❓ How about something like `Expected enum to equal {expectationName} by name{{reason}}, but found {subjectsName}.`

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
```diff
@@ -56,5 +52,59 @@ public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAsserti
...
+            string expectationName = EnumDescription(context.Expectation, expectationsUnderlyingValue);
+
+            Execute.Assertion
+                .ForCondition(subject == expected)
+                .FailWith($"Expected enum to be {expectationName}{{reason}}, but found {subjectsName}.");
+        }
+
+        private static string EnumDescription(object o, decimal? v)
+        {
+            string PrintDecimal(decimal? x) => x?.ToString(CultureInfo.InvariantCulture);
```
jnyrup: 🔧 As `v` is null checked before calling `PrintDecimal`, it could take a regular `decimal` as parameter.

krajek:
@jnyrup Thanks for the review. All suggestions are implemented and PR is my opinion PR is ready to go

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
```diff
@@ -56,5 +52,65 @@ public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAsserti
...
+            decimal? expectationsUnderlyingValue = ExtractDecimal(context.Expectation);
+
+            Execute.Assertion
+                .ForCondition(subjectsUnderlyingValue == expectationsUnderlyingValue)
+                .FailWith(() =>
+                {
+                    string subjectsName = EnumDescription(context.Subject, subjectsUnderlyingValue);
+                    string expectationName = EnumDescription(context.Expectation, expectationsUnderlyingValue);
+
+                    return new FailReason($"Expected enum to be {expectationName} by value{{reason}}, but found {subjectsName}.");
```
jnyrup: 🔧 I think "Expected enum to *equal* {expectationName} by value" sounds more natural?

krajek: Thanks, I missed that part. Marking difference with italics helps.

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
```diff
@@ -56,5 +52,65 @@ public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAsserti
...
+            decimal? expectationsUnderlyingValue = ExtractDecimal(context.Expectation);
+
+            Execute.Assertion
+                .ForCondition(subjectsUnderlyingValue == expectationsUnderlyingValue)
+                .FailWith(() =>
+                {
+                    string subjectsName = EnumDescription(context.Subject, subjectsUnderlyingValue);
+                    string expectationName = EnumDescription(context.Expectation, expectationsUnderlyingValue);
+
+                    return new FailReason($"Expected enum to equal {expectationName} by value{{reason}}, but found {subjectsName}.");
```
dennisdoomen: 🤡 Consider using `{context:enum}` so that the caller C# code can be used to get the real name of the variable that is being asserted.

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
```diff
@@ -56,5 +52,65 @@ public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAsserti
...
+                });
+        }
+
+        private static void HandleByName(IEquivalencyValidationContext context)
+        {
+            string subject = context.Subject.ToString();
```
dennisdoomen: 🤔 What if the subject or expectation is `null`?

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
```diff
@@ -56,5 +52,65 @@ public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAsserti
...
+            string expected = context.Expectation.ToString();
+
+            Execute.Assertion
+                .ForCondition(subject == expected)
+                .FailWith(() =>
+                {
+                    decimal? subjectsUnderlyingValue = ExtractDecimal(context.Subject);
+                    decimal? expectationsUnderlyingValue = ExtractDecimal(context.Expectation);
+
+                    string subjectsName = EnumDescription(context.Subject, subjectsUnderlyingValue);
```
dennisdoomen: 🙃 I would keep the subject-related lines together and then the expectation-related ones.

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
```diff
@@ -56,5 +52,65 @@ public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAsserti
...
+                    string expectationName = EnumDescription(context.Expectation, expectationsUnderlyingValue);
+                    return new FailReason(
+                            $"Expected enum to equal {expectationName} by name{{reason}}, but found {subjectsName}.");
+                });
+        }
+
+        private static string EnumDescription(object o, decimal? v)
+        {
+            string PrintDecimal(decimal x) => x.ToString(CultureInfo.InvariantCulture);
```
dennisdoomen: 🤔 Why would you use a local function here? What algorithm are you trying to encapsulate?

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
```diff
@@ -56,5 +52,65 @@ public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAsserti
...
+                    decimal? expectationsUnderlyingValue = ExtractDecimal(context.Expectation);
+
+                    string subjectsName = EnumDescription(context.Subject, subjectsUnderlyingValue);
+                    string expectationName = EnumDescription(context.Expectation, expectationsUnderlyingValue);
+                    return new FailReason(
+                            $"Expected enum to equal {expectationName} by name{{reason}}, but found {subjectsName}.");
+                });
+        }
+
+        private static string EnumDescription(object o, decimal? v)
```
dennisdoomen: 🙃 Methods are normally named by a verb and a noun.

Tests/Shared.Specs/BasicEquivalencySpecs.cs
```diff
@@ -3599,6 +3595,26 @@ public void When_asserting_members_from_different_enum_types_are_equivalent_by_s
             act.Should().NotThrow();
         }
 
+        [Fact]
+        public void When_asserting_members_from_different_enum_types_are_equivalent_by_value_but_comparing_by_name_it_should_throw()
```
dennisdoomen: 🤔 What are _asserting members_?

Tests/Shared.Specs/EnumAssertionSpecs.cs
```diff
@@ -51,14 +51,15 @@ public void When_both_enums_are_equal_and_of_different_underlying_types_it_shoul
         public void When_both_enums_are_large_and_not_equal_it_should_throw()
         {
             // Arrange
```
dennisdoomen: 🤡 Consider updating the AAA style to match the rest.

krajek: Please confirm if the comment is valid, this particular test seems to follow AAA style.

Tests/Shared.Specs/BasicEquivalencySpecs.cs
krajek: Would you accept the change to "[...]asserting_that_members_from[...]"? BTW there are many places like this is this particular file and tens in other files :-).

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
krajek: Got it, fixed.

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
krajek: Extracting function made sense at the beginning, not so much after it was simplified. Inlined.

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
krajek: If that's not a problem I will keep it like it is. I tried the other version but it looked more chaotic.

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
krajek: Could you point me to some places in the code where I should look to understand that feature better? I guess `{context:enum}` would replace `"enum"`, wouldn't it? Any other considerations after that change, maybe some more tests?

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
jnyrup: The enums can only be `null` if they are nullable enums.
`CanHandle` checks that the types of `subject` and `expectation` has `IsEnum` which nullable enums do not.

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
```diff
@@ -56,5 +52,65 @@ public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAsserti
...
+            if (o == null || v == null)
+            {
+                return "null";
+            }
+
+            if (o.GetType().GetTypeInfo().IsEnum)
+            {
+                string typePart = o.GetType().Name;
+                string namePart = Enum.GetName(o.GetType(), o);
+                string valuePart = PrintDecimal(v.Value) ?? "?";
```
jnyrup: The `?? "?"` should not be necessary since the redefinition of `PrintDecimal`

Tests/Shared.Specs/EnumAssertionSpecs.cs
jnyrup: I could imagine Dennis refers to surrounding the three section header comments with `//------------`

krajek: Oh, right. Fixed.

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
krajek: Actually, it was a bug. In that case NRE was thrown. @jnyrup you missed the fact that `CanHandle` checks if `subject` **or** `expectaction` is enum, which makes it possible for one side of the comparison to be null. I have added tests for both cases.

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
krajek: Removed, thanks.

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
krajek: Nevermind, I took a look at existing tests and `MessageBuilder.SubstituteIdentifier`. Fixed.

PR #927, opened 2018-09-25
Document catching trace for successful assertions

### LINKED ISSUE #846: Traces of successful runs
Traces should be available for successful runs as well to understand why assertions are not failing.

krajek:
I think this documentation extension closes #846.

### DIFF
--- a/docs/_pages/documentation.md
+++ b/docs/_pages/documentation.md
@@ -980,6 +980,19 @@ With trace:
   }
   ```
 
+  By default, the trace is displayed only when an assertion fails as part of the message.
+  If the trace is needed for the successful assertion you can catch it like this:
+
+  ```csharp
+object object1 = [...];
+object object2 = [...];
+var traceWriter = new StringBuilderTraceWriter();            
+object1.Should().BeEquivalentTo(object2, opts => opts.WithTracing(traceWriter));             
+string trace = traceWriter.ToString();
+  ```
+
+Alternatively, you could write your own implementation of `ITraceWriter` for special purposes e.g. writing to a file.
+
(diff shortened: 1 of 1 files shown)


docs/_pages/documentation.md
```diff
@@ -980,6 +980,18 @@ With trace:
...
+
+  ```csharp
+object object1 = [...];
+object object2 = [...];
+var traceWriter = new StringBuilderTraceWriter();            
+object1.Should().BeEquivalentTo(object2, opts => opts.WithTracing(traceWriter));             
+string trace = traceWriter.ToString();
+  ```
+
+Alternatively, you could write your own implementation of ITraceWriter for special purposes e.g. writing to a file.
```
dennisdoomen: 🙃 I would use back-tickets for the reference to `ITraceWriter` so it looks like code.

docs/_pages/documentation.md
```diff
@@ -980,6 +980,18 @@ With trace:
   }
   ```
 
+  By default, trace is displayed only when assertion fails as part of the message. If the trace is needed for successful assertion you can catch it like this:
```
dennisdoomen: 🙃 It's _the trace_
🙃 _successful assertions_

krajek: Thanks, I passed the test through Grammarly. Applied your first comment, but not the other.

docs/_pages/documentation.md
krajek: Good catch, thanks

krajek:
I have just fixed styling issues. PR is ready to merge.

PR #932, opened 2018-09-30
Add support for private protected access modifier

### LINKED ISSUE #878: Add support for "private protected"
C# 7.2 added support for the new access modifier [`private protected`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/private-protected).

This is used for the following public methods: 
* `HaveAccessModifier` 
* `NotHaveAccessModifier`
* `WhichSetterHas`
* `WhichSetterDoesNotHave`
* `WhichGetterHas`
* `WhichGetterDoesNotHave`
* `BeWritable`
* `BeReadable`

It would be nice to have support for `private protected`.

This should be accomplishable by:

* Adding another enum to [CSharpAccessModifier.cs](https://github.com/fluentassertions/fluentassertions/blob/master/Src/FluentAssertions/Common/CSharpAccessModifier.cs), 
* add a check for `IsFamilyAndAssembly` for the methods in
[CSharpAccessModifierExtensions.cs](https://github.com/fluentassertions/fluentassertions/blob/master/Src/FluentAssertions/Common/CSharpAccessModifierExtensions.cs), and
* write some tests

krajek:
I changed Spec assemblies configuration to `<LangVersion>7.2</LangVersion>`. `private protected` access modifier is not available in earlier versions.

Closes #878.

### DIFF
--- a/Tests/Shared.Specs/BasicEquivalencySpecs.cs
+++ b/Tests/Shared.Specs/BasicEquivalencySpecs.cs
@@ -1442,21 +1442,24 @@ namespace FluentAssertions.Specs
                 "protected",
                 "internal",
                 "protected-internal",
-                "private");
+                "private",
+                "private-protected");
 
             var expected = new ClassWithAllAccessModifiersForMembers(
                 "public",
                 "protected",
                 "ignored-internal",
                 "ignored-protected-internal",
-                "private");
+                "private",
+                "ignore-private-protected");
 
             //-----------------------------------------------------------------------------------------------------------
             // Act
             //-----------------------------------------------------------------------------------------------------------
             Action act = () => subject.Should().BeEquivalentTo(expected, config =>
                 config.Excluding(ctx => ctx.WhichGetterHas(CSharpAccessModifier.Internal) ||
-                                        ctx.WhichGetterHas(CSharpAccessModifier.ProtectedInternal)));
+                                        ctx.WhichGetterHas(CSharpAccessModifier.ProtectedInternal) ||
+                                        ctx.WhichGetterHas(CSharpAccessModifier.PrivateProtected)));
 
             //-----------------------------------------------------------------------------------------------------------
             // Assert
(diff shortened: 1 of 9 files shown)


Tests/Shared.Specs/MethodBaseAssertionSpecs.cs
```diff
@@ -642,6 +682,27 @@ public void When_asserting_a_public_member_is_not_private_it_succeeds()
...
         }
 
+        [Fact]
+        public void When_asserting_a_private_protected_member_is_not_private_it_succeeds()
+        {
+            //-------------------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-------------------------------------------------------------------------------------------------------------------
+            PropertyInfo propertyInfo = typeof(TestClass).GetPropertyByName("PublicGetPrivateProtectedSet");
+            MethodInfo getMethod = propertyInfo.SetMethod;
```
jnyrup: Should be named `setMethod`

Tests/Shared.Specs/BasicEquivalencySpecs.cs
```diff
@@ -4193,14 +4200,17 @@ public class ClassWithAllAccessModifiersForMembers
         protected internal string ProtectedInternalProperty { get; set; }
         private string PrivateProperty { get; set; }
 
+        private protected string PrivateProtectedProperty { get; set; }
+
         public ClassWithAllAccessModifiersForMembers(string publicValue, string protectedValue, string internalValue,
-            string protectedInternalValue, string privateValue)
+            string protectedInternalValue, string privateValue, string privateProtected)
```
jnyrup: 🙃 Rename the parameter to `privateProtectedValue`

Tests/Shared.Specs/MethodBaseAssertionSpecs.cs
krajek: Good catch, thanks.

Tests/Shared.Specs/BasicEquivalencySpecs.cs
krajek: Good catch, thanks.

PR #934, opened 2018-10-01
Add `BeApproximately` version for both subject and expected nullable

### LINKED ISSUE #922: double?.Should().BeApproximately is looking for a float?
I am using 5.4.2 and trying to work with both double and double? type data elements

```
double? d1 = 1.0123;
double expected = 1.01;

d1.Should().BeApproximately(expected, 1.0); <-- this extension doesn't exist
d1.Value.Should().BeApproximately(expected, 1.0) <-- complains that "expected" should be float
```

Shouldn't both work?

krajek:
Closes #922.

For now, I have tackled only the case of `double?`. 
Please review. Upon approval, I will copy the logic and tests for `float?` and `decimal?`.


krajek:
PR is ready.
Let me know what you think about `NotBeApproximately(nullable, nullable[...])`.
For me it makes sense, but it was not discussed not requested in the #922. 
I will create another PR if we agree to add it.

dennisdoomen:
Yeah, I wonder where @krajek is getting the time from.... ;-)

PR #936, opened 2018-10-03
Nullable assertion extenions cleanup

krajek:
Two main topics of the PR:

- Introduce regions for `NumericAssertionsExtensions`, one for each method group
- Extract repeated code from each method group

I would like that to be merged before I will tackle extension of `NotBeEquivalentTo` for nullable subject and expected value (I mean folloup of #934).

### DIFF
--- a/Src/FluentAssertions/NumericAssertionsExtensions.cs
+++ b/Src/FluentAssertions/NumericAssertionsExtensions.cs
@@ -45,11 +45,7 @@ namespace FluentAssertions
                 maxValue = sbyte.MaxValue;
             }
 
-            Execute.Assertion
-                .ForCondition(minValue <= actualValue && actualValue <= maxValue)
-                .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context:value} to be within {0} from {1}{reason}, but found {2}.",
-                    delta, nearbyValue, actualValue);
+            FailIfValueOutsideBounds(minValue <= actualValue && actualValue <= maxValue, nearbyValue, delta, actualValue, because, becauseArgs);
 
             return new AndConstraint<NumericAssertions<sbyte>>(parent);
         }
@@ -88,11 +84,7 @@ namespace FluentAssertions
                 maxValue = byte.MaxValue;
             }
 
-            Execute.Assertion
-                .ForCondition(minValue <= actualValue && actualValue <= maxValue)
-                .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context:value} to be within {0} from {1}{reason}, but found {2}.",
-                    delta, nearbyValue, actualValue);
+            FailIfValueOutsideBounds(minValue <= actualValue && actualValue <= maxValue, nearbyValue, delta, actualValue, because, becauseArgs);
 
             return new AndConstraint<NumericAssertions<byte>>(parent);
         }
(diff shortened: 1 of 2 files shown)


Src/FluentAssertions/NumericAssertionsExtensions.cs
```diff
@@ -43,11 +45,7 @@ public static class NumericAssertionsExtensions
                 maxValue = sbyte.MaxValue;
             }
 
-            Execute.Assertion
-                .ForCondition(minValue <= actualValue && actualValue <= maxValue)
-                .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context:value} to be within {0} from {1}{reason}, but found {2}.",
-                    delta, nearbyValue, actualValue);
+            BeCloseToFailIfValueOutsideBounds(minValue <= actualValue && actualValue <= maxValue, nearbyValue, delta, actualValue, because, becauseArgs);
```
dennisdoomen: 🙃 `BeCloseToFailIfValueOutsideBounds` is a bit of a weird name. I'm not sure how to read it.

Src/FluentAssertions/NumericAssertionsExtensions.cs
```diff
@@ -507,11 +480,7 @@ public static class NumericAssertionsExtensions
                 maxValue = ushort.MaxValue;
             }
 
-            Execute.Assertion
-                .ForCondition(!(minValue <= actualValue && actualValue <= maxValue))
-                .BecauseOf(because, becauseArgs)
-                .FailWith("Did not expect {context:value} to be within {0} from {1}{reason}, but found {2}.",
-                    delta, distantValue, actualValue);
+            NoBeCloseToFailIfValueInsideBounds(!(minValue <= actualValue && actualValue <= maxValue), distantValue, delta, actualValue, because, becauseArgs);
```
dennisdoomen: 🤔 I suppose you meant _not be_? 
🙃 See earlier remark

Src/FluentAssertions/NumericAssertionsExtensions.cs
```diff
@@ -9,6 +9,8 @@ namespace FluentAssertions
     /// </summary>
     public static class NumericAssertionsExtensions
     {
+
+        #region BeCloseTo
```
jnyrup: 🔧 I've become more fond of using partial classes instead of `#region`s for cases such as this.
We could extract this into four partial classes `NumericAssertionsExtensions.BeCloseTo.cs` etc.

krajek: I agree that partial classes would be even more maintainable.
@dennisdoomen since such an approach is not yet implemented anywhere in the codebase, could you comment on that?

Src/FluentAssertions/NumericAssertionsExtensions.cs
krajek: Name simplified.

Src/FluentAssertions/NumericAssertionsExtensions.cs
krajek: Name simplified.

Src/FluentAssertions/NumericAssertionsExtensions.cs
dennisdoomen: You don't even need `partial` classes here.

krajek: Do you mean simply defining few classes, one for each method group(`BeCloseNumericAssertionExtension`, `NotBeCloseNumericAssertionExtensions` and so on)?

dennisdoomen: Yes, you could do that. Although I personally prefer regions for this. Being able to quickly find all the assertions from Github is something I often do.

krajek: Let's stick with regions for now then.

Issue #738, opened 2018-01-22
Conversion rules end up in the failure message multiple times


dennisdoomen:
```
With configuration:
- Use declared types and members
- Compare enums by value
- Match member by name (or throw)
- Try conversion of all members. Try conversion of member x.SelectedMemberPath.Contains("Birthdate"). Try conversion of all members. 
- Be strict about the order of items in byte arrays
" does not.
```

dennisdoomen:
Can't reproduce

dennisdoomen:
The `TryConversionStep` is added every time a new instance of the `SelfReferenceEquivalencyAssertionOptions` is created. First when it copies over the `UserEquivalencySteps` from the defaults (which always adds the `TryConversionStep` as part of it), and possibly later

krajek:
I will tackle the problem. Any comments before I start? Is there any test that reproduces the behaviour? Is it only about duplication in the description or the behaviour itself is also affected?

dennisdoomen:
It's purely a cosmetic issue.

jnyrup:
@krajek I guess this can be closed with #941 ?

PR #941, opened 2018-10-08
Fix multiplied ConversionSelector's descriptions appearing in assertion message

krajek:
Refers to #738.
For sure it is a step forward, but I am not certain if it fixes the root cause. I will investigate more, and possibly follow up with another PR.  But for the time being...

the problem was the `ConversionSelector.Clone` method that treated `description` of type `StringBuilder` like it was immutable `string`. Obviously, `StringBuilder` is mutable and needs to be cloned.

Two main changes:

- Create new `StringBuilder` in `ConversionSelector.Clone`

- Rename `description` field to `descriptionBuilder` because `description` suggests it is a `string` indeed.

### DIFF
--- a/Src/FluentAssertions/Equivalency/IEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/IEquivalencyAssertionOptions.cs
@@ -53,7 +53,7 @@ namespace FluentAssertions.Equivalency
         /// <summary>
         /// Gets an ordered collection of Equivalency steps how a subject is compared with the expectation.
         /// </summary>
-        IEnumerable<IEquivalencyStep> UserEquivalencySteps { get; }
+        IEnumerable<IEquivalencyStep> GetUserEquivalencySteps(ConversionSelector conversionSelector);
 
         /// <summary>
         /// Gets a value indicating whether the runtime type of the expectation should be used rather than the declared type.
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
@@ -77,11 +77,12 @@ namespace FluentAssertions.Equivalency
             includeProperties = defaults.IncludeProperties;
             includeFields = defaults.IncludeFields;
 
+            ConversionSelector = defaults.ConversionSelector.Clone();
+
             selectionRules.AddRange(defaults.SelectionRules);
-            userEquivalencySteps.AddRange(defaults.UserEquivalencySteps);
+            userEquivalencySteps.AddRange(defaults.GetUserEquivalencySteps(ConversionSelector));
             matchingRules.AddRange(defaults.MatchingRules);
             orderingRules = new OrderingRuleCollection(defaults.OrderingRules);
-            ConversionSelector = defaults.ConversionSelector.Clone();
 
             getDefaultEqualityStrategy = defaults.GetEqualityStrategy;
             TraceWriter = defaults.TraceWriter;
@@ -125,8 +126,8 @@ namespace FluentAssertions.Equivalency
         /// <summary>
         ///     Gets an ordered collection of Equivalency steps how a subject is compared with the expectation.
         /// </summary>
-        IEnumerable<IEquivalencyStep> IEquivalencyAssertionOptions.UserEquivalencySteps =>
-            userEquivalencySteps.Concat(new[] { new TryConversionStep(ConversionSelector) });
+        IEnumerable<IEquivalencyStep> IEquivalencyAssertionOptions.GetUserEquivalencySteps(ConversionSelector convertionSelector) =>
+            userEquivalencySteps.Concat(new[] {new TryConversionStep(convertionSelector) });
 
         public ConversionSelector ConversionSelector { get; } = new ConversionSelector();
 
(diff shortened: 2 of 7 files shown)


krajek:
I have just run all the tests and haven't found a place when duplicated description would appear.
I have used a conditional breakpoint and debugged all the tests.

@dennisdoomen are you aware of MCVE for the #738 that this fix does not cover?

Tests/Shared.Specs/BasicEquivalencySpecs.cs
```diff
@@ -2581,7 +2581,10 @@ public void When_a_specific_mismatching_property_is_excluded_from_conversion_it_
             //-----------------------------------------------------------------------------------------------------------
             // Assert
             //-----------------------------------------------------------------------------------------------------------
-            act.Should().Throw<XunitException>().WithMessage("Expected*<1973-09-20>*\"1973-09-20\"*");
+            string exceptionMessage = act.Should().Throw<XunitException>().Which.Message;
```
jnyrup: 🙃 Can't the assertions be chained with `And`?

krajek: We should eat our own dog food right? ;-).

krajek:
After @jnyrup comment, I found that I created misspelt the assertion in test:
```.Should().NotMatch("*Try conversion of all members*", "conversion description should be present");```
should be 
```.Should().**Match**("*Try conversion of all members*", "conversion description should be present");```
hence I have a problem to fix.

**Comment by krajek** [comment]
Thanks. 
While debugging I start to see that problem with `ConversionSelector.Clone` is just the tip of the iceberg ;-).
Work in progress to fixing the root causes.

**Comment by krajek** [comment]
Ok, by now I think I have gotten to the bottom of the issue. The more facts I gathered, the more clear it was that until this PR, the feature was working almost by accident :-).

The most important parts of the code to understand it are:

- `ConversionSelector.Clone()` copied the `StringBuilder` reference instead of cloning. @dennisdoomen found a case when the description was duplicated, but because default static `StringBuilder` was copied all over the place, the description could be repeated without upped bound. However, cloning would also be incorrect, because of the conditional logic in `ConvertionSelector.ToString()`. One of the incorrect scenarios would be when you clone `ConvertionSelector` without rules, and you get `"Without automatic conversion."`. Next, you add some conversion rule and try to add the description the builder, but  `"Without automatic conversion."` is not erased! **The only solution was to make `ConvertionSelector`  delay building the description string until `ToString` is called.**

- `SelfReferenceEquivalencyAssertionOptions` constructor that copies from `defaults` had following line: `userEquivalencySteps.AddRange(defaults.UserEquivalencySteps);`. Seems logical, but when we go to `UserEquivalencySteps` implementation we can see that its not a single retrieval of the list but ` userEquivalencySteps.Concat(new[] { new TryConversionStep(ConversionSelector) });`. This code generates convertion step with a current selector. So when it copied from `defaults` it used `defaults.ConvertionSelector` instead of using cloned selector. **The only solution I found yet is to inject `ConvertionSelector`**

I plan to add some more tests tomorrow, but I could use some review. I hope I did not miss anything important.

File: Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs [hunk 2]
```diff
@@ -125,8 +126,8 @@ IEnumerable<IMemberSelectionRule> IEquivalencyAssertionOptions.SelectionRules
         /// <summary>
         ///     Gets an ordered collection of Equivalency steps how a subject is compared with the expectation.
         /// </summary>
-        IEnumerable<IEquivalencyStep> IEquivalencyAssertionOptions.UserEquivalencySteps =>
-            userEquivalencySteps.Concat(new[] { new TryConversionStep(ConversionSelector) });
+        IEnumerable<IEquivalencyStep> IEquivalencyAssertionOptions.GetUserEquivalencySteps(ConversionSelector convertionSelector) =>
```
jnyrup: typo: `convertionSelector` => `conversionSelector`

krajek: 👍 done.

Src/FluentAssertions/Equivalency/IEquivalencyAssertionOptions.cs
```diff
@@ -53,7 +53,7 @@ public interface IEquivalencyAssertionOptions
         /// <summary>
         /// Gets an ordered collection of Equivalency steps how a subject is compared with the expectation.
         /// </summary>
-        IEnumerable<IEquivalencyStep> UserEquivalencySteps { get; }
+        IEnumerable<IEquivalencyStep> GetUserEquivalencySteps(ConversionSelector conversionSelector);
```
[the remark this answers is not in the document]
krajek: Yes, now I think that we need to change the public API because it's design was flawed.
On the other hand, the change is of the kind I like the most:

- impact on a tiny part of the user base
- compile-time
- trivial to fix

Nonetheless, I will look into it over the upcoming days, maybe so idea will emerge. I would not count on that though.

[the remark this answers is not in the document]
dennisdoomen: Just do it. That could might be public, but it's really just internal stuff and never mentioned in any of the extensibility docs.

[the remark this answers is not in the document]
krajek: Thanks, I won't worry about it then.

krajek:
Ready for merge(with squash).

I have added one test checking the exception message exactly, I hope you won't crucify me for that :-). My intent was to have at least one test that verifies the whole mechanism of gathering descriptions from the equivalency steps. 

As it was discussed, the PR introduces minor breaking change in public API.

Issue #940, opened 2018-10-06
NotThrow with wait time


frederik-h:
As far as I know, fliuentassertions contains no equivalent to NUnit's [DelayedConstraints](https://github.com/nunit/docs/wiki/DelayedConstraint) which allows to specifiy that a condition should be satisfied after a certain wait time.
I am using an overload of ActionAssertions.NotThrow which takes additional waitTime and pollingInterval parameters to achieve a similar effect. That is, it repeatedly executes the Action every pollingInterval milliseconds, catching any exceptions, until the Action either stops throwing exceptions or the waitTime has passed. In the latter case, it fails with an exception.

I was wondering if there is some interest to include this functionality in fluentassertions?

dennisdoomen:
Could you give us an example of what this would look like?

frederik-h:
@dennisdoomen Sure. Let's consider the following artificial example:

```         var watch = Stopwatch.StartNew();
            Action throwForSomeTime = delegate {
                    if(watch.ElapsedMilliseconds < 1000)
                            throw new Exception("Not yet done.");
            };

            throwForSomeTime.Should().NotThrow(1100, 50);
            Console.WriteLine("Action stopped throwing as expected.");
```

Here the assertion would succeed and return after about 1000ms since the Action stops throwing an exception after 1000ms. The second argument of NotThrow is the polling interval, i.e. the Action will be executed every 50ms until it succeeds or the maximum wait time of 1100ms is exceeded.

If we change the Action to throw for, say, 1200ms, the assertion would fail with 
an exception:  `Unhandled Exception: FluentAssertions.Execution.AssertionFailedException: Did not expect any exception after 1100 milliseconds, but found System.Exception with message "Not yet done."
`

If you have tests which have to wait, for instance, for some network service to become available, the easiest thing to do is to insert a sleep into your code. This is not ideal since your tests will always have to wait for a fixed time, even if your service becomes available much more quickly. Implementing the polling in each test is quite cumbersome.

frederik-h:
Sure, it has to be decided what should be used to achieve the delay in the implementation. I currently use `Task.Delay` for this. Another question is what would be the best name for the method? Should it be an overload of `NotThrow` or would `NotThrowAfter` be clearer?

dennisdoomen:
I like `Should().NotThrowAfter`

eajhnsn1:
[...]

PR #942, opened 2018-10-08
Add Should().NotThrowAfter assertion for actions

frederik-h:
This adds a `Should().NotThrowAfter` assertion for Actions which is a variant of `Should().NotThrow`
 which asserts that the Action should stop throwing exceptions after a given wait time. See issue #940 for a discussion of this feature.

If this gets merged, it should probably also be considered to add corresponding methods for the other action assertions such as `ThrowAfter` etc.

## IMPORTANT 

- Regarding compliance to the guidelines: I was unable to run the CSharpGuidelinesAnalyzer, since I do not have access to Windows/VisualStudio at home.
-  Regarding the documentation: I am not sure where to put this right now; I think we can add it later.


Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,54 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+        /// </param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        /// <exception cref="ArgumentOutOfRangeException">Throws if waitTime or pollInterval are negative.</exception>
+        public void NotThrowAfter(int waitTime, int pollInterval, string because = "", params object[] becauseArgs)
```
jnyrup: 🔧 I would prefer to use `TimeSpan` for `waitTime` and `pollInterval`.

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,54 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+        {
+            FailIfSubjectIsAsyncVoid();
+
+            if(waitTime < 0 || pollInterval < 0)
```
jnyrup: 🔧 Split into two `if`s to give a more precise exception message.

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -895,6 +894,62 @@ public void When_no_exception_should_be_thrown_and_none_was_it_should_not_throw(
             //-----------------------------------------------------------------------------------------------------------
             foo.Invoking(f => f.Do()).Should().NotThrow();
         }
+
+        private static void ThrowUntil(Stopwatch watch, int throwTime)
```
jnyrup: 🔧 Take two `TimeSpan` parameters instead and pass in `StopWatch.Elapsed` as the first argument.

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -895,6 +894,62 @@ public void When_no_exception_should_be_thrown_and_none_was_it_should_not_throw(
...
+            //-----------------------------------------------------------------------------------------------------------
+            Action action = () => foo.Invoking(f => f.Do())
+                                     .Should().NotThrowAfter(waitTime, pollInterval, "we passed valid arguments");
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            action
+                .Should().Throw<XunitException>()
+                         .Where(e => e.Message.Contains(
```
jnyrup: 🔧 Use `WithMessage` with wildcards as the test should not depend on how FakeItEasy name their exceptions.
E.g. 
```csharp
action.Should().Throw<XunitException>()
    .WithMessage("*we passed valid arguments*");
```

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -895,6 +894,62 @@ public void When_no_exception_should_be_thrown_and_none_was_it_should_not_throw(
...
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            var foo = A.Fake<IFoo>();
+            var watch = Stopwatch.StartNew();
+            var waitTime = 100;
+            var pollInterval = 10;
+
+            void ThrowLongerThanWaitTime() => ThrowUntil(watch, waitTime + waitTime / 2);
+
+            A.CallTo(() => foo.Do()).Invokes(ThrowLongerThanWaitTime);
```
jnyrup: 🙃 Can this be simplified to the following?

```csharp
Action throwLongerThanWaitTime = () => ThrowUntil(watch, waitTime + waitTime / 2);

// Act
Action act = () => throwLongerThanWaitTime.Should().NotThrowAfter(waitTime, pollInterval, "we passed valid arguments");

// Assert
    .WithMessage("*we passed valid arguments*");
```

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -895,6 +894,62 @@ public void When_no_exception_should_be_thrown_and_none_was_it_should_not_throw(
...
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            var foo = A.Fake<IFoo>();
+            var watch = Stopwatch.StartNew();
+            var waitTime = 100;
+            var pollInterval = 10;
+
+            void ThrowShorterThanWaitTime() => ThrowUntil(watch, waitTime / 2);
```
jnyrup: 🙃 The same simplification as above

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,54 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        /// <exception cref="ArgumentOutOfRangeException">Throws if waitTime or pollInterval are negative.</exception>
+        public void NotThrowAfter(int waitTime, int pollInterval, string because = "", params object[] becauseArgs)
+        {
+            FailIfSubjectIsAsyncVoid();
```
jnyrup: Please add a test that makes `FailIfSubjectIsAsyncVoid` fail.

Src/FluentAssertions/Specialized/ActionAssertions.cs
jnyrup: Please add two tests that makes the conditional fail.

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,54 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+                exception = InvokeSubjectWithInterception();
+                if (exception is null)
+                    return;
+
+                Task.Delay(pollInterval).Wait();
+                invocationEndTime = watch.ElapsedMilliseconds;
+            }
+            Execute.Assertion
+                        .BecauseOf(because, becauseArgs)
+                        .FailWith("Did not expect any exception after {0} milliseconds{reason}, but found {1}.", waitTime, exception);
```
jnyrup: After changing `waitTime` to a `TimeSpan` the `TimeSpanFormatter` will handle units, so `milliseconds` can be removed.

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,54 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+        /// Asserts that the current <see cref="Action"/> stops throwing any exception
+        /// after a specified amount of time.
+        /// </summary>
+        /// <remarks>
+        /// The <see cref="Action"/> is invoked. If it raises an exception,
+        /// the invocation is repeated until it either stops raising any exceptions
+        /// or the specified wait time is exceeded.
+        /// </remarks>
+        /// <param name="waitTime">
+        /// The time in milliseconds after which the <see cref="Action"/> should have stopped throwing any exception.
```
jnyrup: Remove `in milliseconds` as `waitTime` should be a `TimeSpan` per next comment.

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,54 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+            long invocationEndTime = -1;
+            Exception exception = null;
+
+            while (invocationEndTime < waitTime)
+            {
+                exception = InvokeSubjectWithInterception();
+                if (exception is null)
+                    return;
+
+                Task.Delay(pollInterval).Wait();
```
dennisdoomen: ❌ Since this code is synchronous, I would just use `Thread.Sleep`here. Calling `Wait`on a `Task` can be very dangerous.

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
dennisdoomen: 🤔 I prefer not to use helper methods in test. They tend to obscure the test.

Src/FluentAssertions/Specialized/ActionAssertions.cs
dennisdoomen: 🔧 Maybe `NotThrowAnymoreAfter`

dennisdoomen: 🤔 Do we also need to be able to specify an exception type here?
🤔 Do we also need this assertion for `Func<Task>` for consistency with the other exception APIs?

frederik-h: > Maybe `NotThrowAnymoreAfter`

Are you sure? I think that this would be a tad too long. It might also be misleading. Doesn't "anymore" suggest that the Action must have been throwing exceptions at some point

frederik-h: > I would prefer to use `TimeSpan` for `waitTime` and `pollInterval`.
Sounds good. Would you keep an overload with the current signature or would you expect the user to convert from that representation to the TimeSpan?

dennisdoomen: True. Very true. Alright. Let's stick with `NotThrowAfter`

dennisdoomen: > Would you keep an overload with the current signature or would you expect the user to convert from that representation to the TimeSpan?

They can use the `TimeSpan` extension methods like `12.Seconds()`.

frederik-h: > Do we also need to be able to specify an exception type here?

I have not encountered a use case for this so far. This would mean that the Action could be still throwing exceptions but not the specified one ... :confused. We could add it for consistency's sake. You could also wait until someone requests it. I don't know your policy regarding these kind of questions :smirk:
> Do we also need this assertion for `Func<Task>` for consistency with the other exception APIs?
Probably.

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
frederik-h: > 🙃 Can this be simplified to the following?
Sure, but the same would apply to some of the surrounding tests? I also do not really get why the mocking should be used here. @dennisdoomen Should the Arrange-Act-Assert structure of the other tests by strictly imitated?

jnyrup: You should always divide the tests into comment separated Arrange, Act & Assert sections.
There are a few exceptions to this, where e.g. Arrange and Act sections are collapsed into one section.

If existing tests has room for improvements, you do not need to copy their bad behavior.
If you have specific tests in mind, you can write me and I'll include them in occasional cleanup PRs.

FYI: make a blank line after quoting someone to separate your comment in Github Markdown.

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -889,12 +888,132 @@ public void When_no_exception_should_be_thrown_and_none_was_it_should_not_throw(
...
+        public void NotThrowAfter_when_no_exception_should_be_thrown_after_wait_time_but_it_was_it_should_throw()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            var watch = Stopwatch.StartNew();
+            var waitTime = 100.Milliseconds();
+            var pollInterval = 10.Milliseconds();
+
+            Action ThrowLongerThanWaitTime = () =>
```
jnyrup: 🙃 variables should be lower camel case

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -889,12 +888,132 @@ public void When_no_exception_should_be_thrown_and_none_was_it_should_not_throw(
...
+                {
+                    if (watch.Elapsed <= waitTime + (waitTime.Milliseconds / 2).Milliseconds())
+                        throw new ArgumentException("An exception was forced");
+                };
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action action = () =>
```
jnyrup: 🙃 variables should be lower camel case

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -889,12 +888,132 @@ public void When_no_exception_should_be_thrown_and_none_was_it_should_not_throw(
...
+        public void NotThrowAfter_when_no_exception_should_be_thrown_after_wait_time_and_none_was_it_should_not_throw()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            var watch = Stopwatch.StartNew();
+            var waitTime = 100.Milliseconds();
+            var pollInterval = 10.Milliseconds();
+
+            Action ThrowShorterThanWaitTime = () =>
```
jnyrup: 🙃 variables should be lower camel case

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -889,12 +888,132 @@ public void When_no_exception_should_be_thrown_and_none_was_it_should_not_throw(
...
+                {
+                    if (watch.Elapsed <= (waitTime.Milliseconds / 2).Milliseconds())
+                        throw new ArgumentException("An exception was forced");
+                };
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act / Assert
+            //-----------------------------------------------------------------------------------------------------------
+            ThrowShorterThanWaitTime.Should().NotThrowAfter(waitTime, pollInterval);
```
jnyrup: 🔧 The entire test suite might not be consistent about it, but to adhere to the AAA-style you could write

```
// Act
Action act = () => ThrowShorterThanWaitTime.Should().NotThrowAfter(waitTime, pollInterval);

// Assert
act.Should().NotThrow();
```

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,57 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+        /// </param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        /// <exception cref="ArgumentOutOfRangeException">Throws if waitTime or pollInterval are negative.</exception>
+        public void NotThrowAfter(TimeSpan waitTime, TimeSpan pollInterval, string because = "", params object[] becauseArgs)
```
jnyrup: 🤔 Would `pollingInterval` be better than `pollInterval` ?

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -889,12 +888,132 @@ public void When_no_exception_should_be_thrown_and_none_was_it_should_not_throw(
...
-
             //-----------------------------------------------------------------------------------------------------------
             // Act / Assert
             //-----------------------------------------------------------------------------------------------------------
             foo.Invoking(f => f.Do()).Should().NotThrow();
         }
+
+        #pragma warning disable CS1998
+        [Fact]
+        public void NotThrowAfter_when_subject_is_async_it_should_not_throw()
```
jnyrup: @dennisdoomen usually has some opinions on naming tests

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,57 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+        {
+            FailIfSubjectIsAsyncVoid();
+
+            if(waitTime < TimeSpan.Zero)
+                throw new ArgumentOutOfRangeException(nameof(waitTime), $"The value of {nameof(waitTime)} must be positive.");
+
+            if(pollInterval < TimeSpan.Zero)
+                throw new ArgumentOutOfRangeException(nameof(pollInterval), $"The value of {nameof(pollInterval)} must be positive.");
+
+            var watch = Stopwatch.StartNew();
```
jnyrup: @dennisdoomen can you shed some light on when we like explicit types?

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,57 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+                exception = InvokeSubjectWithInterception();
+                if (exception is null)
+                    return;
+
+                Thread.Sleep(pollInterval);
+                invocationEndTime = watch.Elapsed;
+            }
+            Execute.Assertion
+                        .BecauseOf(because, becauseArgs)
+                        .FailWith("Did not expect any exception after {0}{reason}, but found {1}.", waitTime, exception);
```
jnyrup: 🤔 Shouldn't this either be "any exception**s**" or "**an** exception"?

jnyrup:
The failing build should have been fixed with #944 .

Src/FluentAssertions/Specialized/ActionAssertions.cs
frederik-h: I am not a native english speaker. But I am sure that I have seen `pollInterval` used quite often in various APIs. I've also checked it with a dictionary and I think it is ok.

Src/FluentAssertions/Specialized/ActionAssertions.cs
frederik-h: Oh, that topic ... :wink: I think its fine here since the type is plain to infer from the right-hand side - without any ambiguity.

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
frederik-h: I tried to stick to the naming pattern found in the tests. If a test covers a single method, I think its nice to reflect this in the name. This is also done in this test suite, e.g. in the tests for `WithInnerExceptionExactly`

jnyrup:
The build fails as [`Thread.Sleep`](https://docs.microsoft.com/da-dk/dotnet/api/system.threading.thread.sleep?view=netframework-4.7.2#applies-to) is not available in .net standard earlier than 2.0.
In https://github.com/fluentassertions/fluentassertions/blob/master/Src/FluentAssertions/Specialized/ExecutionTimeAssertions.cs this is handled by wrapping the action inside a `Task`.

frederik-h:
> The build fails as [`Thread.Sleep`](https://docs.microsoft.com/da-dk/dotnet/api/system.threading.thread.sleep?view=netframework-4.7.2#applies-to) is not available in .net standard earlier than 2.0.
> In https://github.com/fluentassertions/fluentassertions/blob/master/Src/FluentAssertions/Specialized/ExecutionTimeAssertions.cs this is handled by wrapping the action inside a `Task`. Note that this to enable interrupting the `Action`.

@jnyrup Did you have a chance to take a look at my attempt to solve this?

jnyrup:
@frederik-h Sorry for the delayed response.
I did have a look at it and my initial gut feeling was that it felt like a hack to use a synchronization mechanism to implement a wait.

What about if we used something similar to https://github.com/fluentassertions/fluentassertions/blob/master/Src/FluentAssertions/Specialized/ExecutionTimeAssertions.cs and wrap the synchronous function inside a `Task` which can then be polled or re-run.
That should also have the benefit that for a long-running method, the test method will not run longer than `waitTime`.

frederik-h:
> @frederik-h Sorry for the delayed response.
> I did have a look at it and my initial gut feeling was that it felt like a hack to use a synchronization mechanism to implement a wait.

Yes, I also hesitated at first, but since the waiting is part of the specified behavior of the method and not some implementation specific side effect, I decided that there is no substantial reason not to use it.  

> What about if we used something similar to https://github.com/fluentassertions/fluentassertions/blob/master/Src/FluentAssertions/Specialized/ExecutionTimeAssertions.cs and wrap the synchronous function inside a `Task` which can then be polled or re-run.
> That should also have the benefit that for a long-running method, the test method will not run longer than `waitTime`.

I am not sure that this is a benefit, but I don't have a strong opinion on this matter. This would implement a different functionality than what I was trying to implement, but it could probably be used for more or less the same purpose. What would you do if the Task is not done after the `waitTime`? Would you throw an exception? Even if the subject `Action` did not throw any exceptions?

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,60 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+        /// </param>
+        /// <exception cref="ArgumentOutOfRangeException">Throws if waitTime or pollInterval are negative.</exception>
+        public void NotThrowAfter(TimeSpan waitTime, TimeSpan pollInterval, string because = "", params object[] becauseArgs)
+        {
+            FailIfSubjectIsAsyncVoid();
+
+            if(waitTime < TimeSpan.Zero)
+                throw new ArgumentOutOfRangeException(nameof(waitTime), $"The value of {nameof(waitTime)} must be positive.");
+
+            if(pollInterval < TimeSpan.Zero)
```
jnyrup: ❌ See comment above about positive vs non-negative.

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -889,12 +888,139 @@ public void When_no_exception_should_be_thrown_and_none_was_it_should_not_throw(
...
```
jnyrup: The method name says that it _should not throw_, but the code asserts that `action` _does throw_

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,60 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        /// <exception cref="ArgumentOutOfRangeException">Throws if waitTime or pollInterval are negative.</exception>
+        public void NotThrowAfter(TimeSpan waitTime, TimeSpan pollInterval, string because = "", params object[] becauseArgs)
+        {
+            FailIfSubjectIsAsyncVoid();
+
+            if(waitTime < TimeSpan.Zero)
```
jnyrup: x The opposite of negative is not positive, but non-negative as is includes zero.
Either the condition should be `<=` or the exception message should be must be non-negative".

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,60 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+
+            if(pollInterval < TimeSpan.Zero)
+                throw new ArgumentOutOfRangeException(nameof(pollInterval), $"The value of {nameof(pollInterval)} must be positive.");
+
+            TimeSpan? invocationEndTime = null;
+            Exception exception = null;
+
+            var watch = Stopwatch.StartNew();
+
+            using(var waiter = new ManualResetEvent(false)) {
```
jnyrup: 🙃 brace should be on new line.
If your editor supports [EditorConfig](https://editorconfig.org/), formatting the entire document should automatically use the project preferences. 
I've recently updated the `.editorconfig` file which defines this, so rebase on master to get the best auto format experience.

jnyrup:
@dennisdoomen do you have an opinion on the usage of `ManualResetEvent` for waiting?

dennisdoomen:
> do you have an opinion on the usage of `ManualResetEvent` for waiting?

He's not doing concurrent calls, is he? So a simple Thread.Sleep would suffice.

jnyrup:
@dennisdoomen `Thread.Sleep` is not available in netstandard until 2.0, see earlier comment.

dennisdoomen:
> `Thread.Sleep` is not available in netstandard until 2.0, see earlier comment.

So maybe we should not support this method except for .NET Standard 2.0 and higher, and the full .NET Framework.

Src/FluentAssertions/Specialized/ActionAssertions.cs
frederik-h: 👍 good spot ;-) The message should  say non-negative

frederik-h:
> > `Thread.Sleep` is not available in netstandard until 2.0, see earlier comment.
> 
> So maybe we should not support this method except for .NET Standard 2.0 and higher, and the full .NET Framework.

Sounds like the best option if you are opposed to using a workaround for the missing Thread.sleep.
But I don't see a hard technical reason for not using the `ManuelResetEvent`. Do I miss something?

Src/FluentAssertions/Specialized/ActionAssertions.cs
frederik-h: I will check it out.

frederik-h:
> > > `Thread.Sleep` is not available in netstandard until 2.0, see earlier comment.
> > 
> > 
> > So maybe we should not support this method except for .NET Standard 2.0 and higher, and the full .NET Framework.
> 
> Sounds like the best option.

I have now implemented this.

dennisdoomen:
> I don't think that my changes have caused the test failure in the test from

No, that one is unstable.

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,60 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+
+            
+            TimeSpan? invocationEndTime = null;
+            Exception exception = null;
+            var watch = Stopwatch.StartNew();
+
+            while (invocationEndTime is null || invocationEndTime < waitTime)
+            {
+                exception = InvokeSubjectWithInterception();
+                if (exception is null)
```
dennisdoomen: 🙃 We prefer explicit braces for all blocks

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -863,6 +866,136 @@ public void When_no_exception_should_be_thrown_and_none_was_it_should_not_throw(
             //-----------------------------------------------------------------------------------------------------------
             foo.Invoking(f => f.Do()).Should().NotThrow();
         }
+
+#if NET45 || NET47 || NETSTANDARD2_0 || NETCOREAPP2_0
+        #pragma warning disable CS1998
+        [Fact]
+        public void NotThrowAfter_when_subject_is_async_it_should_throw()
```
dennisdoomen: 🙃 We try to avoid using the literal names of code elements in unit test names

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,60 @@ public void NotThrow(string because = "", params object[] becauseArgs)
             }
         }
 
+#if NET45 || NET47 || NETSTANDARD2_0 || NETCOREAPP2_0
```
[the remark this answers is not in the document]
frederik-h: I have changed it. I hope the new condition is equivalent, I have checked the list of target frameworks in the project file.

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -863,6 +866,136 @@ public void When_no_exception_should_be_thrown_and_none_was_it_should_not_throw(
...
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            var watch = Stopwatch.StartNew();
+            var waitTime = 100.Milliseconds();
+            var pollInterval = 10.Milliseconds();
+
+            Action throwLongerThanWaitTime = () =>
+                {
+                    if (watch.Elapsed <= waitTime + (waitTime.Milliseconds / 2).Milliseconds())
```
jnyrup: Explicit braces

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -114,6 +114,61 @@ public void NotThrow(string because = "", params object[] becauseArgs)
...
+            TimeSpan? invocationEndTime = null;
+            Exception exception = null;
+            var watch = Stopwatch.StartNew();
+
+            while (invocationEndTime is null || invocationEndTime < waitTime)
+            {
+                exception = InvokeSubjectWithInterception();
+                if (exception is null) {
+                    return;
+		}
```
jnyrup: Please align the braces. Running a document format (at least in Visual Studio or Rider), should format the document with respect to our EditorConfig style definitions.

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -863,6 +866,136 @@ public void When_no_exception_should_be_thrown_and_none_was_it_should_not_throw(
...
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            var watch = Stopwatch.StartNew();
+            var waitTime = 100.Milliseconds();
+            var pollInterval = 10.Milliseconds();
+
+            Action throwShorterThanWaitTime = () =>
+                {
+                    if (watch.Elapsed <= (waitTime.Milliseconds / 2).Milliseconds())
```
jnyrup: Explicit braces

Src/FluentAssertions/Specialized/ActionAssertions.cs
frederik-h: Sorry for the incoveniences. I usually don't work with with dotnet at home, but I use FluentAssertions at work. I don't have a proper setup for dotnet development on my old Linux laptop (Rider is completely unusable on this machine). But I should at least install some EditorConfig plugin for my editor ;-).

jnyrup:
For a consistent API, we should probably also expose `NotThrowAfter` for `AsyncFunctionAssertions` and `FunctionAssertions` as well.
@dennisdoomen Could those be implemented in a separate PR or would that block for the master branch for potential minor releases in between?

dennisdoomen:
> Could those be implemented in a separate PR or would that block for the master branch for potential minor releases in between?

I think we need to be a bit more critical.

frederik-h:
> 
> I think we need to be a bit more critical.

Regarding the `FunctionAssertions` etc. I suppose? Is there anything left to do regarding this PR? :smile:

dennisdoomen:
> Regarding the `FunctionAssertions` etc. I suppose? Is there anything left to do regarding this PR? 😄

Would you be willing to pick up the suggestions from @jnyrup? Then we would have a single PR that adds this new API in a consistent API.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -174,6 +174,119 @@ private static void NotThrow(Exception exception, string because, object[] becau
...
+        /// </param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        /// <exception cref="ArgumentOutOfRangeException">Throws if waitTime or pollInterval are negative.</exception>
+        public async Task NotThrowAfterAsync(TimeSpan waitTime, TimeSpan pollInterval, string because = "", params object[] becauseArgs)
```
jnyrup: To have the desired effect of eager argument validation, `async` needs to be removed from `NotThrowAfterAsync` and just `return assertionTask();`.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -174,6 +174,119 @@ private static void NotThrow(Exception exception, string because, object[] becau
...
+            }
+
+            if (pollInterval < TimeSpan.Zero)
+            {
+                throw new ArgumentOutOfRangeException(nameof(pollInterval), $"The value of {nameof(pollInterval)} must be non-negative.");
+            }
+
+	    await assertionTask();
+
+	    async Task assertionTask() {
```
jnyrup: Please put the `{` on next line and indent the body of the local function.

jnyrup:
@frederik-h regarding the two failing tests, you might want to relax the time limits.
Timing tests easily fluctuate on slower machines, such as build servers.

jnyrup:
Regarding the documentation, a description of `NotThrowAfter` could be added just after
`act.Should().NotThrow<InvalidOperationException>();`

and just add an example to the block of async examples.

dennisdoomen:
So? Is this one ready to be merged?

Tests/Shared.Specs/FunctionExceptionAssertionSpecs.cs
```diff
@@ -328,6 +331,151 @@ public void When_function_throw_aggregate_in_aggregate_exception_and_that_was_no
...
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            AndWhichConstraint<FunctionAssertions<int>, int> act =
+                throwShorterThanWaitTime.Should().NotThrowAfter(waitTime, pollInterval);
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act.Subject.Should().Be(42);
```
jnyrup: Just to double check.
Could this be written as
```
throwShorterThanWaitTime.Should().NotThrowAfter(waitTime, pollInterval)
    .Which.Should().Be(42);
```

frederik-h: Yes!

jnyrup:
@frederik-h Thanks for implementing this feature and contributing to Fluent Assertions.
It takes some endurance as a contributor to go through three months of reviewing (and silence), but I'm really satisfied with the merged result.

dennisdoomen:
Completely agree with @jnyrup here. Because of the maturity of FA, it becomes much more difficult to get new functionality in. So thanks for that. 

@jnyrup I think this deserved a release, right?

jnyrup:
@dennisdoomen Push the button.

frederik-h:
Thank you for the constructive review!

PR #949, opened 2018-10-16
Ensure specific nuget.config is used

jnyrup:
Specifies that both `DotNetCoreRestore` and `NuGetRestore` should use `nuget.config` to ensure that `nuget.org` is used.


dennisdoomen:
Hmm, normally the `.nuget.config` that is next to the solution file should be picked up automatically. What problem were you trying to solve?

Issue #954, opened 2018-10-30
Type.Should().Be(Type) doesn't support open generics


thepinkmile:
I would expect this:
```
var response = result.GetType(); // returns ICustomInterface<int>
response.Should().Be(typeof(ICustomInterface<>));
```
to succeed.

Instead it fails with message:

Expected type to be ICustomInterface`1, but found ICustomInterface`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]].

I believe this is related to the change made for: #458

P.S. Workaround for now:
```
result.GetType().GetGenericTypeDefinition()
        .Should().Be(typeof(ICustomInterface<>));
```

I am also using this in a few "And" constraints which forces me to do:

```
 result.Should().NotBeNull()
    .And.Subject.GetGenericTypeDefinition().Should().Be(typeof(ICustomInterface<>));
```
Which is ugly and stops me being able to add more assertions as the Subject is now the GenericTypeDefinition and not the actual Type I am validating.

jnyrup:
Thanks for raising this issue.
First of all to assert on the type of an object you should probably use one of 
```c#
response.Should().BeAssignableTo<TType>()
response.Should().BeAssignableTo(typeof(TType))
``` 
to test that `response` is assignable to `TType`.

The changes related to #458 added support for open generic when using `BeOfType()`.

Here are my findings:

First let's see how Fluent assertions behaves for non-generic types to establish a baseline.
```c#
interface IInterface { }

class Base : IInterface { }

class Derived : Base { }

Derived derived = new Derived();
derived.Should().BeOfType<Derived>(); // success
derived.Should().BeAssignableTo<Base>(); // success
derived.Should().BeAssignableTo<IInterface>(); // success
derived.Should().BeOfType<Base>(); // fail
derived.Should().BeOfType<IInterface>(); // fail

Base baseClass = new Derived();
baseClass.Should().BeOfType<Derived>(); // success
baseClass.Should().BeAssignableTo<Base>(); // success
baseClass.Should().BeAssignableTo<IInterface>(); // success
baseClass.Should().BeOfType<Base>(); // fail
baseClass.Should().BeOfType<IInterface>(); // fail

IInterface interfaced = new Derived();
interfaced.Should().BeOfType<Derived>(); // success
interfaced.Should().BeAssignableTo<Base>(); // success
interfaced.Should().BeAssignableTo<IInterface>(); // success
[...]

PR #955, opened 2018-10-31
Consider types assignable to open generics (#954)

mdonoughe:
I think I did this right, at least for BeAssignableTo. Base types and interfaces are searched for matching generic types.

Should BeDerivedFrom do the same thing but disallowing interfaces?

#954

## IMPORTANT 


Src/FluentAssertions/Types/TypeAssertions.cs
```diff
@@ -72,6 +72,45 @@ public new AndConstraint<TypeAssertions> BeAssignableTo<T>(string because = "",
             return BeAssignableTo(typeof(T), because, becauseArgs);
         }
 
+        bool CheckIfAssignableToOpenGeneric(Type definition)
```
dennisdoomen: 🔧 Splitting the method in two, one for interfaces and one for classes might help read this code
🔧 Please move the method below the first use. We follow the order of execution ordering rule.

Src/FluentAssertions/Types/TypeAssertions.cs
```diff
@@ -72,6 +72,45 @@ public new AndConstraint<TypeAssertions> BeAssignableTo<T>(string because = "",
...
+            // assignable to any generic type based on the given generic type definition.
+
+            if (definition.GetTypeInfo().IsInterface)
+            {
+                // check subject and its interfaces against definition
+                TypeInfo subjectInfo = Subject.GetTypeInfo();
+                if (subjectInfo.IsInterface && subjectInfo.IsGenericType &&
+                    subjectInfo.GetGenericTypeDefinition().IsSameOrEqualTo(definition))
+                {
+                    return true;
```
dennisdoomen: 🙃 I personally find all these intermediate `return` statements a more difficult to read.

Src/FluentAssertions/Types/TypeAssertions.cs
```diff
@@ -113,8 +162,18 @@ public new AndConstraint<TypeAssertions> NotBeAssignableTo<T>(string because = "
         /// <returns>An <see cref="AndConstraint{T}"/> which can be used to chain assertions.</returns>
         public new AndConstraint<TypeAssertions> NotBeAssignableTo(Type type, string because = "", params object[] becauseArgs)
         {
+            bool condition;
```
dennisdoomen: 🔧 Maybe rename it to `isAssignable`?

Tests/Shared.Specs/TypeAssertionSpecs.cs
```diff
@@ -484,6 +484,65 @@ public void When_an_unrelated_type_instance_it_fails_with_a_useful_message()
                 .WithMessage($"*{typeof(DummyImplementingClass)} to be assignable to {typeof(DateTime)}*failure message*");
         }
 
+        [Fact]
+        public void When_constructed_of_open_generic_it_succeeds()
```
dennisdoomen: 🙃  We normally use the `When_...._it_should_....` format for all our tests. 
🤔 I don't get the _when constructed_. We're not constructing anything here, right?

Tests/Shared.Specs/TypeAssertionSpecs.cs
```diff
@@ -484,6 +484,65 @@ public void When_an_unrelated_type_instance_it_fails_with_a_useful_message()
...
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange / Act / Assert
+            //-----------------------------------------------------------------------------------------------------------
+            typeof(IDummyInterface<IDummyInterface>).Should().BeAssignableTo(typeof(IDummyInterface<>));
+        }
+
+        [Fact]
+        public void When_implementation_of_open_generic_interface_it_succeeds()
```
dennisdoomen: 🙃 It feels like the name is missing a verb or something.

mdonoughe: I was trying to emulate the "When_its_own_type_instance_it_succeeds" pattern the other BeAssignableTo methods were following without making the names super long.

Src/FluentAssertions/Types/TypeAssertions.cs
mdonoughe: Is this better now?

mdonoughe:
I've added in support for BeDerivedFrom too.

mdonoughe:
BeOfType already had open generics.

Is there a convention for both `BeAssignableTo`s to use the same logic?

jnyrup:
Yes, but `NotBeOfType` does not.

I'm not sure I understand your question.
If you're asking where to put shared logic, I would put it in `TypeExtensions.cs`.

jnyrup:
Don't know if this is WIP, but `BeAssignableTo<T>` should probably call `BeAssignableTo(Type)` instead?

mdonoughe:
It probably should, but it doesn't seem to make any difference since you can't pass open generics to methods like that. I assumed it did already.

mdonoughe:
Actually should it? `BeAssignableTo<T>` casts the result to `T`, so for cases where it wouldn't match `BeAssignableTo` (shouldn't be any?) it would fail to cast even if it relied on `BeAssignableTo` to do the assertion.

jnyrup:
It looks very good now and feature wise could be merge.
But I would like to have both a positive and a negative test for `BeAssignableTo`, `NotBeAssignableTo` and `NotBeOfType` from `ReferenceTypeAssertions` when the expectation is an open generic.

dennisdoomen:
Maybe it's better to request changes then. It makes the state of the PR more visual.

PR #960, opened 2018-11-05
Including and/or excluding properties in graph comparison should honor the declared type

### LINKED ISSUE #956: IsEquivalentTo and BeEquivalentTo do not handle member hiding
### Description

I noticed the implementation for IsEquivarentTo in TypeExtensions considers two members to be equivalent if they have the same name and the declaring type of either property is the same _or inherits_ the type of the other property. This is incorrect due to member hiding.

E.g.

```csharp
void Main()
{
	var aProperty = SelectedMemberInfo.Create(typeof(A).GetProperty("Property"));
	var bProperty = SelectedMemberInfo.Create(typeof(B).GetProperties().First(p => p.DeclaringType == typeof(B)));
	aProperty.IsEquivalentTo(bProperty).Should().BeFalse();
	
	B b1 = new B(), b2 = new B();
	b1.Should().BeEquivalentTo(b2, config => config.Including(b => b.Property));
}

class A
{
	public string Property { get; set; } = Guid.NewGuid().ToString();
}

class B: A
{
	public new string[] Property { get; set; }
}
```

### Expected behavior:

`Main` should execute successfully, because the two properties are different and the two instances of `B` are equivalent when comparing only the null arrays.

### Actual behavior:

Both assertions throw because of the logic error in `IsEquivalentTo`. The second assertion throws because the logic for `BeEquivalentTo` gets the property from the expression and then uses `IsEquivalentTo` to compare it against other properties later.

### Versions

* Which version of Fluent Assertions are you using? 5.4.2
* Which .NET runtime and version are you targeting? .NET Framework 4.7.2

### Additional Information

Using `Excluding(b => ((A)b).Property)` instead of `Including(b => b.Property)` does not work either because the code for extracting the member from the expression does not support casts.

dennisdoomen:
Excluding or including a property with a certain name would include or exclude all properties with that name in the entire hierarchy, even those that are hidden by the `new` keyword.

Fixes #956


Src/FluentAssertions/Common/MemberPath.cs
```diff
@@ -21,23 +26,37 @@ public bool IsParentOrChildOf(string candidate)
             return IsParent(candidate) || IsChild(candidate);
         }
 
+        public bool IsSameAs(string candidate, Type memberDeclaringType)
+        {
+            string[] candidateSegments = Segmentize(candidate);
```
jnyrup: 🙃 fail fast on `memberDeclaringType == declaringType`
I.e. `if (memberDeclaringType != declaringType) return false;`

Src/FluentAssertions/Common/MemberPath.cs
```diff
@@ -21,23 +26,37 @@ public bool IsParentOrChildOf(string candidate)
...
+
+            return candidateSegments.SequenceEqual(segments) && memberDeclaringType == declaringType;
+        }
+
         private bool IsChild(string candidate)
         {
             string[] candidateSegments = Segmentize(candidate);
 
-            return candidateSegments.Take(segments.Count).SequenceEqual(segments);
+            return candidateSegments.Take(segments.Count).SequenceEqual(segments) &&
```
jnyrup: 🙃 `(candidateSegments.Length > segments.Count) && candidateSegments.Take(segments.Count).SequenceEqual(segments)` instead.

Src/FluentAssertions/Common/MemberPath.cs
```diff
@@ -21,23 +26,37 @@ public bool IsParentOrChildOf(string candidate)
...
+                   candidateSegments.Length > segments.Count;
         }
 
         private bool IsParent(string candidate)
         {
             string[] candidateSegments = Segmentize(candidate);
 
-            return candidateSegments.SequenceEqual(segments.Take(candidateSegments.Length));
+            return candidateSegments.SequenceEqual(segments.Take(candidateSegments.Length))
```
jnyrup: 🙃 `candidateSegments.Length < segments.Count` first

Issue #818, opened 2018-03-30
Add version of StringAssertions.Contain that allows specifying the number of times a substring should occur.


basbossink:
Currently StringAssertions has no easy way of asserting that a substring should occur a particular number of times within a string. Imagine the syntax could be something like

    actual.Shoud().Contain("Fred", Twice, "Fred is important");

or

    actual.Should().Contain("Fred").Twice("Fred is important");

or

    actual.Should().Contain("Fred", 2, "Fred is important");

jnyrup:
How about the `Match` assertion?

This asserts that `actual` has at least two *distinct* `Fred`s in it.
```csharp
actual.Should().Match("*Fred*Fred*", "Fred is important");
```

I emphasize distinct because I feel it could be difficult to get such an API right.
E.g. does `aaa` contain `aa` one or two times?

Would `actual.Should().Contain(expectedSubstring, k)` mean that `actual` should contain `expectedSubstring` *exactly* `k` times or *at least* `k` times?

Just for the record a very unreadable way to assert that `actual` contains exactly two distinct `Fred`s.
```csharp
actual.Should().MatchRegex("^(?:(?!Fred).)*Fred(?:(?!Fred).)*Fred(?:(?!Fred).)*$", "Fred is important");
```

basbossink:
Thanks for the explanation, in this case the `Match` assertion is a good alternative. The second suggestion is imho a little bit more readable. In my example I would expect `Twice` to mean exactly two times. I could also image something like:

    actual.Should().Contain("Fred").AtLeast.Twice("Fred is important");
    actual.Should().Contain("Fred").AtMost.Twice("Fred is important");
    actual.Should().Contain("Fred").Exactly.Twice("Fred is important"); //which would be equivalent to:
    actual.Should().Contain("Fred").Twice("Fred is important");

Given that my request can by implemented using the `Match`, I would treat this request as a _light suggestion_ and handle it as you see fit. Thanks for your efforts.

dennisdoomen:
It could work, provided that `Fred` does indeed exists in that `actual`. The successive checks would just apply additional checks. But then we should also add those extensions to all related assertions.

mkolumb:
Hi,

I would like to do this.
But I think we should do it using the following api
`actual.Should().ContainExactly("Fred", 2, "Fred is important");`

API proposed above is more readable, but I think it will be a breaking change.

mkolumb:
[...]

PR #962, opened 2018-11-12
[WIP] Add more specific variants of contain (#818)

mkolumb:
Hi,

I have implemented this enhancement.
You can find the API in documentation.md
Because I didn't receive a response, I made the following assumptions:
- `Contain` should throw always if expected string wouldn't be found (as in the previous version)
- `aaa` contains `aa` once, `aaaaa` contains `aa` twice


mkolumb:
If I should change something in assumptions please comment and I will fix it.

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -618,7 +618,7 @@ public AndConstraint<StringAssertions> NotEndWithEquivalentOf(string unexpected,
         /// <param name="becauseArgs">
         /// Zero or more objects to format using the placeholders in <see cref="because" />.
         /// </param>
-        public AndConstraint<StringAssertions> Contain(string expected, string because = "", params object[] becauseArgs)
+        public StringContainConstraint Contain(string expected, string because = "", params object[] becauseArgs)
```
dennisdoomen: 🔧 Would `StringContainmentConstraints` be a better name?

Src/FluentAssertions/StringContainConstraint.cs
```diff
@@ -0,0 +1,82 @@
...
+    public class StringContainConstraint : AndConstraint<StringAssertions>
+    {
+        public StringContainConstraint(StringAssertions parentConstraint, string expected) : base(parentConstraint)
+        {
+            ParentConstraint = parentConstraint;
+            Expected = expected;
+            Occurrences = CountOccurrences(parentConstraint.Subject, expected, StringComparison.Ordinal);
+        }
+
+        private string Expected { get; }
```
dennisdoomen: 🔧 I would change them into private fields.

Src/FluentAssertions/StringContainConstraint.cs
```diff
@@ -0,0 +1,82 @@
...
+        public AndConstraint<StringAssertions> Twice(string because = "", params object[] becauseArgs) =>
+            Exactly.Twice(because, becauseArgs);
+
+        public AndConstraint<StringAssertions> Thrice(string because = "", params object[] becauseArgs) =>
+            Exactly.Thrice(because, becauseArgs);
+
+        public AndConstraint<StringAssertions> Times(int times, string because = "", params object[] becauseArgs) =>
+            Exactly.Times(times, because, becauseArgs);
+
+        private static int CountOccurrences(string actual, string expected, StringComparison comparison)
```
dennisdoomen: 🔧 We normally order the members of a method in order of execution so you can read the file as a book (from Clean Code).

Src/FluentAssertions/StringWithOccurenceConstraint.cs
```diff
@@ -0,0 +1,56 @@
...
+            Times(times, $"{times} times", because, becauseArgs);
+
+        private AndConstraint<StringAssertions> Times(int times, string displayValue, string because = "",
+            params object[] becauseArgs)
+        {
+            Execute.Assertion
+                .ForCondition(Comparator(Occurrences, times))
+                .BecauseOf(because, becauseArgs)
+                .FailWith(
+                    $"Expected {{context:string}} {{0}} to contain {{1}} {Mode} {displayValue}{{reason}}, but found {Occurrences} times.",
```
dennisdoomen: ❓ What if `Occurrences` is 1? Should it not say _time_ instead of _times_?

Src/FluentAssertions/StringWithOccurenceConstraint.cs
```diff
@@ -0,0 +1,56 @@
...
+            StringOccurrenceComparer comparator)
+        {
+            Assertion = parentConstraint;
+            Expected = expected;
+            Occurrences = occurrences;
+            Mode = mode;
+            Comparator = comparator;
+        }
+
+        private StringAssertions Assertion { get; }
```
dennisdoomen: 🔧 Why not use private fields?

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -1867,6 +1869,816 @@ public void When_string_containment_is_asserted_and_actual_value_is_null_then_it
 #endif
         }
 
+        #endregion
+		
+		#region Default with times
+
+        [InlineData("ABCDEF", "BCD", 1)]
+        [InlineData("ABCDEBCDF", "BCD", 2)]
+        [InlineData("ABXYZCDXYZEBCDFXYZ", "XYZ", 3)]
```
dennisdoomen: 🤔 Hmm, aren't you duplicating the number of scenarios? We do gray box testing. So you don't have to academically prove that it works. Just be smart about the scenarios. Same with the other tests.

Src/FluentAssertions/Primitives/StringAssertions.cs
mkolumb: Maybe `StringContainmentConstraint` (without 's' at the end)?

Src/FluentAssertions/StringWithOccurenceConstraint.cs
mkolumb: Good point, I will fix it

Src/FluentAssertions/StringContainConstraint.cs
mkolumb: Ok, my mistake, I used resharper to order the methods.
So please confirm the order:

1. CountOccurrences
2. Once
3. Twice
4. Thrice
5. Times
6. Exactly
7. AtLeast
8. MoreThan
9. AtMost
10. LessThan

Src/FluentAssertions/StringContainConstraint.cs
mkolumb: Ok

Src/FluentAssertions/StringWithOccurenceConstraint.cs
mkolumb: Ok

Tests/Shared.Specs/StringAssertionSpecs.cs
mkolumb: Ok, I have heard about gray box testing, but I'm not sure what you exactly mean (in this case).
So please confirm.

Should I remove tests from `Default with times` region?
These tests are exactly the same as in `Exactly` region (but for different API).

Should I have less input data?

Src/FluentAssertions/StringContainConstraint.cs
dennisdoomen: Correct.

Tests/Shared.Specs/StringAssertionSpecs.cs
dennisdoomen: Just the `InlineData`.

mkolumb: Ok, thanks

Src/FluentAssertions/Primitives/StringAssertions.cs
mkolumb: I changed the name to `StringContainmentConstraint` because I think it better fits the convention e.g. `AndWhichConstraint`. If you really want to have `StringContainmentConstraints` I can change this.

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -1785,43 +1785,904 @@ public void When_asserting_string_does_not_end_with_equivalent_of_a_value_and_ac
...
             //-----------------------------------------------------------------------------------------------------------
             // Arrange
             //-----------------------------------------------------------------------------------------------------------
             string actual = "ABCDEF";
+            string expectedSubstring = "XYS";
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => actual.Should().Contain(expectedSubstring).AtMost.Once("that is {0}", "required");
```
jnyrup: ❌ This is a prime example of why I don't like this augmentation of `Contain`. 
`ABCDEF` contains `XYS` 0 times, which is less than 1, but the test fails.

Src/FluentAssertions/Primitives/StringAssertions.cs
dennisdoomen: No, not at all. A 🔧 is a suggestion anyway.

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -3032,6 +3906,28 @@ public void Should_fail_with_descriptive_message_when_asserting_string_length_to
...
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            string actual = "ABCDEFGHI";
+            string substring = "EF";
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act / Assert
+            //-----------------------------------------------------------------------------------------------------------
+            actual.Should()
```
jnyrup: 🔧 I'd like to be able to write
```c#
actual.Should().Contain(substring).AtLeast.Once().And.AtMost.Twice();
```
Right now `CountOccurrences is re-computed` for every call to `Contain()`?

Src/FluentAssertions/StringWithOccurenceConstraint.cs
```diff
@@ -0,0 +1,56 @@
...
+        private readonly StringOccurrenceComparer comparator;
+        private readonly string expected;
+        private readonly string mode;
+        private readonly int occurrences;
+
+        public StringWithOccurenceConstraint(StringAssertions parentConstraint,
+            string expected,
+            int occurrences,
+            string mode,
+            StringOccurrenceComparer comparator)
```
jnyrup: ❓ Why a `delegate` instead of `Func<int, int, bool>`?

Src/FluentAssertions/Primitives/StringAssertions.cs
mkolumb: True, I know it is just a suggestion.
I didn't find any class with name ending with `Constraints`, so I thought it is convection.
Now I see I was wrong.
I changed class name to your suggestion :)

Src/FluentAssertions/StringContainmentConstraints.cs
```diff
@@ -0,0 +1,82 @@
...
+        private readonly StringAssertions parentConstraint;
+
+        public StringContainmentConstraints(StringAssertions parentConstraint, string expected) : base(parentConstraint)
+        {
+            this.parentConstraint = parentConstraint;
+            this.expected = expected;
+            occurrences = CountOccurrences(parentConstraint.Subject, expected, StringComparison.Ordinal);
+        }
+
+        private static int CountOccurrences(string actual, string expected, StringComparison comparison)
```
jnyrup: 🔧 How about inlining comparison as it's always StringComparison.Ordinal?

Src/FluentAssertions/StringContainmentConstraints.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            actual = actual ?? "";
+            expected = expected ?? "";
+
+            int occurrences = 0;
+
+            int index;
+
+            while ((index = actual.IndexOf(expected, comparison)) >= 0)
```
jnyrup: x Use [`public int IndexOf (string value, int startIndex, int count, StringComparison comparisonType)`](https://docs.microsoft.com/en-us/dotnet/api/system.string.indexof?view=netframework-4.7.2#System_String_IndexOf_System_String_System_Int32_System_Int32_System_StringComparison_) to avoid implementing a variant of [Schlemiel the Painter's Algorithm](https://en.wikichip.org/wiki/schlemiel_the_painter%27s_algorithm). This will also avoid any calls Substring.
Here's an [implementation](https://www.dotnetperls.com/string-occurrence)

Src/FluentAssertions/StringContainmentConstraints.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        private readonly string expected;
+        private readonly int occurrences;
+        private readonly StringAssertions parentConstraint;
+
+        public StringContainmentConstraints(StringAssertions parentConstraint, string expected) : base(parentConstraint)
+        {
+            this.parentConstraint = parentConstraint;
+            this.expected = expected;
+            occurrences = CountOccurrences(parentConstraint.Subject, expected, StringComparison.Ordinal);
```
jnyrup: x I think this be removed from the constructor and only computed if any methods are called.
Right now `CountOccurrences` is called even for simple `actual.Should().Contain(expectedSubstring)`.

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -1785,43 +1785,904 @@ public void When_asserting_string_does_not_end_with_equivalent_of_a_value_and_ac
...
+            Action act = () => actual.Should().Contain(expectedSubstring).AtMost.Times(expectedTimes);
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_string_containment_at_most_is_asserted_and_actual_value_contains_the_expected_string_but_not_at_most_expected_times_it_should_throw()
```
jnyrup: x This is a prime example of why I don't like this augmentation of `Contain`.
"ABCDEF" contains "XYS" 0 times, which is less than 1, but the test fails.

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -1785,43 +1785,904 @@ public void When_asserting_string_does_not_end_with_equivalent_of_a_value_and_ac
...
+            //-----------------------------------------------------------------------------------------------------------
+            string someString = null;
+            Action act = () => someString.Should().Contain("XYZ", "that is {0}", "required");
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act.Should().Throw<XunitException>().WithMessage(
+#if NETCOREAPP1_1
+                 "Expected string <null> to contain \"XYZ\" because that is required.");
```
jnyrup: 🙃 As we're not testing the context replacement, I would remove the preprocessor directive and replace `string`/`someString` with `*`.

Tests/Shared.Specs/StringAssertionSpecs.cs
mkolumb: You're right, that was my assumption.
I think that you as maintainers should decide about API.

I see following solutions:
1. Leave it as is
2. Replace the current `Contain` by something like that `actual.Should().Contain("Fred").Any`
3. Expand `Contain` in different way, e.g. as you suggested in #818 
4. Introduce new method, but I don't know how to named it.

Tests/Shared.Specs/StringAssertionSpecs.cs
mkolumb: Ok, I can change it.

Src/FluentAssertions/StringWithOccurenceConstraint.cs
mkolumb: Personally, I think that if you have more than one parameter of the same type then delegate is more readable.

Src/FluentAssertions/StringContainmentConstraints.cs
mkolumb: Good point. I will fix it

Src/FluentAssertions/StringContainmentConstraints.cs
mkolumb: Sure, I forgot to remove this parameter.
I will add this when I will implement similar API for `ContainEquivalentOf`.

Src/FluentAssertions/StringContainmentConstraints.cs
mkolumb: Yes, you're right. I will change this.

Tests/Shared.Specs/StringAssertionSpecs.cs
jnyrup: @dennisdoomen Here's my beef with this API.

Tests/Shared.Specs/StringAssertionSpecs.cs
mkolumb: The same as in https://github.com/fluentassertions/fluentassertions/pull/962#discussion_r232810763

Tests/Shared.Specs/StringAssertionSpecs.cs
mkolumb: Hi,

I think that I have found the way to fix this without breaking changes, by using `new AssertionScope()` instead of `Execute.Assertion`.
Please give me some additional time, I will update the PR probably on Wednesday.

dennisdoomen: The problem they were facing had to do with the word _repeated_. We don't have this issue with this API. It reads very fluently IMHO.

Src/FluentAssertions/StringContainmentConstraints.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        public StringWithOccurenceConstraint AtMost => new StringWithOccurenceConstraint(this,
+            "at most",
+            (occuredTimes, expectedTimes) => occuredTimes <= expectedTimes);
+
+        public StringWithOccurenceConstraint LessThan => new StringWithOccurenceConstraint(this,
+            "less than",
+            (occuredTimes, expectedTimes) => occuredTimes < expectedTimes);
+
+        internal int Occurrences
```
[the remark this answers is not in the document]
mkolumb: Done

Tests/Shared.Specs/StringAssertionSpecs.cs
jnyrup: @dennisdoomen I agree that it reads fluently, but the result of
```c#
"A".Should().Contain("B").AtMost.Once();
```
would fail and be logically inconsistent with the written assertion.

dennisdoomen: Yeah, either we make sure our API does not support that, or we reconsider the API

dennisdoomen:
> With the lessons learned from #925 this PR would break [ABI compatibility](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md#signatures) by changing the return of a public member.

What if the return type inherits from the original return type?

dennisdoomen:
So what do we want to do with this PR?

jnyrup:
I would prefer if this was reusable, such that other `Contain*` assertions could benefit from it.

E.g.
* `CollectionAssertions.ContainEquivalentOf(TExpectation expectation, ...)`
* `NonGenericCollectionAssertions.Contain(object expected, ...)`
* `SelfReferencingCollectionAssertions.Contain(T expected, ...)`
* `SelfReferencingCollectionAssertions.Contain(Expression<Func<T, bool>> predicate, ...)`
* `StringAssertions.Contain(string expected, ...)`
* `StringAssertions.ContainEquivalentOf(string expected, ...)`
* (`GenericDictionaryAssertions.ContainValue(TValue expected, ...)`)

dennisdoomen:
Hmm. Looking through the entire discussion, I feel (1) is still the best API but will cause breaking changes. (2) doesn't have that problem, but may cause naming conflicts with people that are using FakeitEasy as well. I definitely don't like (3) and (4).

jnyrup:
> > I would prefer if this was reusable, such that other `Contain*` assertions could benefit from it.
> 
> That's a different concern, right? As far as I remember, the discussion ended with the breaking nature of this PR. IMO, this is something we should avoid.

It's an additional concern. 
It would feel weird to me if one overload of `Contain` got support but not the other.
E.g.
```c#
"aaa".Should().Contain("a") // <-- would get support from this PR
new int[] { 1, 1, 1 }.Should().Contain(1) // <-- wouldn't
```

jnyrup:
@mkolumb: @dennisdoomen and I have discussed the API and we found that your earlier [API proposal](https://github.com/fluentassertions/fluentassertions/issues/818#issuecomment-437426673) ended up being the best non-breaking logically consistent way to implement this feature.
At least from our perspectives.

```c#
public delegate bool MatchesOccurenceCount(int actual);

public AndConstraint<StringAssertions> Contain(string expected, MatchesOccurenceCount predicate, string because = "", params object[] becauseArgs)
```

```c#
// Usage
actual.Should().Contain("Fred", AtLeast.Twice, "Fred is important");

// Predefined delegates
{AtLeast, AtMost, MoreThan, LessThan, Exactly}.{Once, Twice, Times(int count)}
(Maybe leave out weird ones such at LessThan.Once?)

// Custom predicates
actual.Should().Contain("Fred", count => count % 2 == 0, "Fred must appear an equal number of times");
```

mkolumb:
> @mkolumb: @dennisdoomen and I have discussed the API and we found that your earlier [API proposal](https://github.com/fluentassertions/fluentassertions/issues/818#issuecomment-437426673) ended up being the best non-breaking logically consistent way to implement this feature.
> At least from our perspectives.
> 
> ```cs
> public delegate bool MatchesOccurenceCount(int actual);
> 
> public AndConstraint<StringAssertions> Contain(string expected, MatchesOccurenceCount predicate, string because = "", params object[] becauseArgs)
> ```
> 
> ```cs
> // Usage
> actual.Should().Contain("Fred", AtLeast.Twice, "Fred is important");
> 
> // Predefined delegates
> {AtLeast, AtMost, MoreThan, LessThan, Exactly}.{Once, Twice, Times(int count)}
> (Maybe leave out weird ones such at LessThan.Once?)
> 
> // Custom predicates
> actual.Should().Contain("Fred", count => count % 2 == 0, "Fred must appear an equal number of times");
> ```

Ok, I will change it, probably next weekend.

dennisdoomen:
> Ok, I will change it, probably next weekend.

Sorry for the confusion. But deciding on the trade-off between a nice fluent and intuitive API and backwards compatibility proved to be less than trivial.

dennisdoomen:
Is this still on your radar?

mkolumb:
> Is this still on your radar?

Yes, sorry, I didn't have a time to finish it. I will do this this week

jnyrup:
I'm very interested in how good failure messages we can get.

dennisdoomen:
@mkolumb can we still motivate you to finish this?

dennisdoomen:
Since there has been no activity for 6 months, let's close it.

PR #963, opened 2018-11-12
Add support for IReadOnlyDictionary assertions

### LINKED ISSUE #568: Support for IReadOnlyDictionary, IReadOnlyCollection

krajek:
Closes #568.

The main idea is that due to limitations of method overload resolution we cannot have `[...]Should(IReadOnlyDictionary<TKey,TValue>[...]`. It will always conflict with overload accepting `IDictionary`. 

Therefore I propose using `ShouldReadOnlyDictionary` name. It seems that it would be beneficial to cut this Gordian knot.

Please let me know what you think. Upon acceptance, I will complete documentation etc.


dennisdoomen:
> The main idea is that due to limitations of method overload resolution we cannot have `[...]Should(IReadOnlyDictionary<TKey,TValue>[...]`. It will always conflict with overload accepting `IDictionary`.

You mean that the compiler doesn't accept a `Should(this IDictionary)` and ` Should(this IReadOnlyDicitonary)`?

jnyrup:
@dennisdoomen the compiler does support having both overloads.
`IDictionary` doesn't inherit from `IReadOnlyDictionary`, but `Dictionary` implements both interfaces, so the conflict is that `Dictionary.Shoul()` is ambiguous between `Should(this IDictionary)` and `Should(this IReadOnlyDictionary)`.

dennisdoomen:
Frankly, I don't like a `ShouldReadOnlyDictionary` extension method. With version 5, I've went to great lengths to make sure all APIs are consistent.

krajek:
I think I understand your point. It is in a sense a scratch on the glass.

However, I dislike the fact that I do not have all dictionary assertions on `IReadOnlyDictionary` more. I stumbled upon it today yet again when testing some `Serilog` properties that are exposed using that interface.

I think that there is a decision to make. (A) We can go practical on this one and agree to some form of impurity (whether we will choose `ShouldReadOnlyDictionary` or some other name does not matter that much). (B) We could also decide to abandon assertions on `IReadOnlyDictionary` and just close related issues. 

@jnyrup @dennisdoomen I would appreciate a clear decision on that matter, thanks.

dennisdoomen:
What about replacing the overload for `IDictionary` with `IReadOnlyDictionary`? Both `Dictionary` and `ReadOnlyDictionary` implement that interface?

krajek:
> hat about replacing the overload for `IDictionary` with `IReadOnlyDictionary`? Both `Dictionary` and `ReadOnlyDictionary` implement that interface?

Clever idea, I will try it out in the evening.

krajek:
The only test failing is `When_the_execution_time_of_a_member_is_close_to_a_limit_it_should_not_throw`. I guess it needs to be dealt with on master.

Moreover, I have just cleaned up classes from initial `ShouldReadOnlyDictionary` implementation.

jnyrup:
From a unit test perspective I like the change, as the assertions shouldn't use any methods not available from `IReadOnlyDictionary`.

But this would be both an ABI and API breaking change.
* Own custom dictionaries only (currently) implementing `IDictionary<TKey, TValue>`
* Third party libraries not compiled for .NET 4.5+ (The first edition of the `IReadOnly` interfaces), where it's harder/impossible to add `IReadOnlyDictionary<TKey, TValue>`.

While the potential problems mentioned above are most likely only the case for a minority, my gut feeling is still that such a breaking change should be postponed to a major release.

dennisdoomen:
I have to agree with @jnyrup. Unless we can think of a way to prevent the breaking change, we'll have to postpone this to a future major release.

krajek:
Planning the feature for the next major release is not a problem at all.  I will have time to polish the implementation and think about downsides.

Thanks for mentioning the fact that `IReadOnlyDictionary` is not present before 4.5. I was not aware.

samcragg:
I know there's been lots of talk how to support this, however, I think one way to solve it would be to have the dictionary methods on an `IEnumerable<KeyValyePair<TKey, TValue>>`:
```C#
public static class DictionaryExtension
{
    public static void Example<TKey, TValue>(this IEnumerable<KeyValuePair<TKey, TValue>> value)
    {
        if (value is IDictionary<TKey, TValue> dictionary)
        {
            UseDictionary(dictionary);
        }
        else if (value is IReadOnlyDictionary<TKey, TValue> readOnlyDictionary)
        {
            UseReadOnlyDictionary(readOnlyDictionary);
        }
        else
        {
            throw new NotSupportedException("value must be a dictionary type");
        }
    }

    private static void UseDictionary<TKey, TValue>(IDictionary<TKey, TValue> dictionary)
    {
    }

    private static void UseReadOnlyDictionary<TKey, TValue>(IReadOnlyDictionary<TKey, TValue> readOnlyDictionary)
    {
    }
}
```
The following works fine with the above:
```C#
new Dictionary<string, int>().Example();
```
The above works as both interfaces (`IDictionary` and `IReadOnlyDictionary`) implement `IEnumerable<KeyValuePair<TKey, TValue>>`. Since we've limited it to `KeyValuePair`, a normal `List<string>` won't be applicable for the extension.

However, the above would match a `List<KeyValuePair<string, int>>` - I personally have never seen that used in production code though (and especially now we have named tuples in C# 7, I think it's useage would be extremely rare). To solve that we could perhaps provide another extension method on the diciontary assertions to cater for the rare chance someone is using it (maybe, `AsCollection`?) that would only be applicable to `IEnumerable<KeyValuePair<TKey,TValue>` to help move them back to the collection assertions. It might be a case though of waiting for the error to be reported, as I think it will be far more rarer than the use of `IReadOnlyDictionary`

Happy to have a stab at another PR if you prefer to avoid confusion with this one.

aalmada:
I'm late for this but here's my two cents.

The conflict happens not only for `IDictionary`/`IReadOnlyDictionary` but also for `ICollection`/`IReadOnlyCollection` and `IList`/`IReadOnlyList`. Although FluentAssertions implements assertions for `ICollection`, `IList` and `IDictionary`, it only uses the enumeration part of these interfaces, which is read-only. The assertions should take the read-only interfaces as parameter and not the read-write counterparts.

Still, these assertions are incomplete. They test only one form of enumerating the collection. For example, a collection that implements `IReadOnlyList<T>` can be enumerated using the indexer, using `IEnumerable<T>.GetEnumerator()`, using `IEnumerable.GetEnumerator()`. There's no guarantee that they are all correctly implemented. The `Count` property can also return the wrong value. 

The collection [may not even implement any interface](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAgrgOwD4AEBMAjALABQZuAzAAQJwQAmA9pgDYCe1sGAxjNQFl2AJQiYA5nDIBvMtXnUqteszacAlpn4BhJln7TqkmAG5qAXzkKr8pUNES4ACk38eerQEob1WaQUB1Lr61AC81O76Jj6W/tZxtjR0jCwc1ACimOhgiBAwTAjUAOJwMJnZufkITp5hAHzUmHAA7hlZOQh5BU7BXtEJ1D5K3Oh8bRWdVT5+gQrJqmmuER6mPgFLPOgIdFr9s/JrCkrlHV3VGyveAwEz+wEwABbqUAB0kVphy1GHs5vbcB9wgBaQh7faxO4/RRJFSpDQfbRbHb8UINP7IsGzKFKYBMJisQRMABucAAcnAAB4wGr1agAajp6IB/AAPF9djEyLEyBR8DQZEN8PhBgNbrMAGYFeg8B5OIkQQry1joODUTSNFqCERiSROQgABk8VzuAEhcIQAJxyiDKuCeTEWLlAA) or have [a public `GetEnumerator()` that is not an overloaded interface](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAgrgOwD4AEBMAjALABQuADAAS6EB0AwgPYA2LcAxjAJZOZQBuMpRqEALENLCAzFQRwIAEz4sAnlVgYuVALKqAShEwBzOFRCjpAHm6YYAPjIBvMlTc1Z8pSvW2YVZix/JypTGAEqAF9Xdxi3XFk9QxM4AAo/Kg4mIIBKOKoXUndigOy7KgBeTLLw/Oii2Ib4zwVlTDUqAFFMdDBECBgmBCoAcTgYbt7+wYRUnMr7Kkw4AHcunr6EAaHUwLscyWKvNo6ASUnN7YQbO0Xaaz97OjGJjemd+YrF5bWAEW4oAAHJhQCDAdgXd6zPYwA75Y4+KjnN5bGaWZ7jSGoj4LJarKj/IEgsEQlFXXY1OFNKj5BIaGBafxYq75QoldwI9q+cpZIKHdlUDIcdAIeR2fns/LFOnMmbpHmUqXuNkC4owAAWALovPKVR1tWpquForguqoAFpCBKSvVVUrmnJWoiMgwRWL/F9Mm7TQa7Yb3HTgExWLomAA3OAAOTgAA8YHNcQBqRPG91UKzVPl1fK0/AEgHA0HguCyobmJGl66PVn2x3eLmChVZ/1uIXe8W12t0wmFkklslyoWKlsFWtqzVQbU1SqZjsj4qpn0zy3W4q2gVd2Qu9sexaLueqpjAABWnH8yKm2IQjB3uP3vo3I8DwZYoYj0bjCc9yfv6dnD8lJ9ZFwMQqH0OAoHGL891vKoV07ICaFAnsQTSeYnHXNx6nqMhhHwWRnFzPMaxHAAzIYFA4dVUjDCBhlolh0DMWw8TWJIjFMVJCAoHI8hHABIWgAE4aIgRi4CpNcyEiIA=).

This is fine if you assume the iterators are well implemented but not if you're implementing an iterator. That's my case so I implemented a `BeExactlyAs()` extension to FluentAssertions: https://github.com/NetFabric/NetFabric.Hyperlinq/blob/419762f44f54f67f2ff312585cea3db8cabeb987/NetFabric.Hyperlinq.UnitTests/Utils/FluentAssertions/Collections/ReadOnlyListAssertions.cs

It would be great is this feature was added to the API.

dennisdoomen:
Now that we're accepting breaking changes again, I think it's time to reconsider this PR.

krajek:
Hi again, I managed to merge with `master` and fix all functional tests successfully but some approval tests are failing. I have yet to learn what they are and how to fix them :-).

krajek:
Ok, I merged with develop and fixed the approval tests. I changed `*.approved.txt` files since the API really does change for this PR.

What is important is that I had to disable one test which did not compile with the new API. It was the test that used a dictionary that did not implement `IReadOnlyDictionary`. It would be a breaking change. The tradeoff is:
* we don't support dictionary assertions of dictionaries not implementing `IReadOnlyDictionary`
* instead we do support dictionary assertions on `IReadOnlyDictionary` objects, which is arguably more important and much more common

Do any more discussion/decisions/doubts come to your mind?

dennisdoomen:
What we could do is to provide an `AsReadOnlyDictionary` extension method that converts the `IDictionary` into a `IReadOnlyDictionary` for those folks that really need to work with an `IDictionary` and for which the `GenericCollectionAssertions` don't provide the value they want.

Src/FluentAssertions/AssertionExtensions.cs
```diff
@@ -294,7 +294,7 @@ public static StringCollectionAssertions Should(this IEnumerable<string> @this)
         /// current <see cref="IDictionary{TKey, TValue}"/>.
```
dennisdoomen: Most of the XML docs still refer to `IDictionary`

Tests/Shared.Specs/GenericReadOnlyDictionaryAssertionSpecs.cs
```diff
@@ -0,0 +1,3031 @@
+﻿using System;
```
jnyrup: This file seems to duplicate nearly all tests from `GenericDictionaryAssertions`.
I'm wondering what the purpose is?

Src/FluentAssertions/AssertionExtensions.cs
krajek: Fixed

Tests/Shared.Specs/GenericReadOnlyDictionaryAssertionSpecs.cs
krajek: I was also not sure whether to "duplicate" all the tests. In general the older file tests dictionaries while the newer one tests read only dictionaries. 

I wanted full coverage of both scenarios, so I basically copied all the tests. They were extremely useful during the development of the PR, now a little bit less I guess. I would err on the side on leaving them, but will not protest if your consensus would go the other way.

Src/FluentAssertions/Extensions/DictionaryExtensions.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+
+namespace FluentAssertions.Extensions
```
dennisdoomen: I would keep this in the root namespace so that it becomes discoverable.

Src/FluentAssertions/Extensions/DictionaryExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+
+            public bool TryGetValue(TKey key, out TValue value) => Dictionary.TryGetValue(key, out value);
+
+            public TValue this[TKey key] => Dictionary[key];
+
+            public IEnumerable<TKey> Keys => Dictionary.Keys;
+            public IEnumerable<TValue> Values => Dictionary.Values;
+        }
+
+        public static IReadOnlyDictionary<TKey, TValue>
```
dennisdoomen: I would add some XML docs to explain why this is useful. 
And move the helper class below its usage please. We prefer to read our code like a book.

krajek: Thanks, documentation added.

Src/FluentAssertions/Extensions/DictionaryExtensions.cs
krajek: Moved to root namespace

Src/FluentAssertions/Extensions/DictionaryExtensions.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+
+namespace FluentAssertions.Extensions
+{
+    public static class DictionaryExtensions
+    {
+        internal class ReadOnlyDictionaryAdapter<TKey, TValue>
```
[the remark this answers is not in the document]
krajek: Thanks.

Src/FluentAssertions/Extensions/DictionaryExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+            : IReadOnlyDictionary<TKey, TValue>
+        {
+            private readonly IDictionary<TKey, TValue> Dictionary;
```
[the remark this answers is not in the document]
krajek: Changed to camelCase.

krajek:
I think I adjusted the PR according to all your comments. 
Tomorrow, for completeness, I will write some tests for the extension method. 

Anything else to complete the task?
Maybe changelog or docs?

dennisdoomen:
> Maybe changelog or docs?

Yeah, good thinking. We wanted to start updating https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/releases.md for anything on `develop`. And updating the docs needs to happen as well. 

> BTW: I wrote simple Powershell script approving new API, maybe you will find it useful:

Yep. Please add it to the root of the repo. Something like `AcceptApiChanges.ps1`. Maybe good to add some remarks at the top of the file to explain what the script is supposed to do for contributors.

krajek:
I have just added an entry in the `releases.md` and some clarification in `dictionaries.md`.
LGTM I guess.

krajek:
I merged with the `develop` branch but test `When_the_execution_time_of_a_member_is_close_to_a_limit_it_should_not_throw ` failed to time related reason. Is there any way to trigger re-build?

dennisdoomen:
@jnyrup is this one still on your radar?

PR #968, opened 2018-11-15
Add support for async assertion on Func<Task<T>>

### LINKED ISSUE #964: Using Should().Throw() with Func<T> fails after upgrading to version 5.5.0
### Description

After upgrading FluentAssertions from version 5.4.2 to 5.5.0, usage of Should().Throw<T> with async method does not yield expected behavior. The following code example functions as expected with version 5.4.2.

```
public class FluentAssertionsExceptionTests
{
    [Fact]
    public void Test1()
    {
        Func<Task<int>> func1 = () => TestAsync(false);
        func1.Should().NotThrow<Exception>();

        // Does not assert thrown exception
        Func<Task<int>> func2 = () => TestAsync(true);
        func2.Should().Throw<Exception>();
    }

    private async Task<int> TestAsync(bool throwException)
    {
        if (throwException)
        {
            throw new Exception("Error");
        }

        return await Task.FromResult(123);
    }
}
```

### Expected behavior:

The assertion for the thrown exception passes.

### Actual behavior:

The assertion for the thrown exception does not pass.  The assertion does not detect the thrown exception.

```
Xunit.Sdk.XunitException
  HResult=0x80131500
  Message=Expected System.Exception, but no exception was thrown.
  Source=FluentAssertions
  StackTrace:
   at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message)
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
   at FluentAssertions.Specialized.FunctionAssertions`1.Throw[TException](Exception exception, String because, Object[] becauseArgs)
   at FluentAssertions.Specialized.FunctionAssertions`1.Throw[TException](String because, Object[] becauseArgs)
```

### Versions

Visual Studio 2017, .NET Core 2.1, XUnit 2.4.1, FluentAssertions 5.5.0

### Additional Information

Removal of the 'TResult' argument from the Func<Task<T>> yields expected behavior and the assertion functions as expected with version 5.5.0.

```
Func<Task> func3 = () => TestAsync(true);
func3.Should().Throw<Exception>();
```

krajek:
Closes #964.

In hindsight, missing support for `Func<Task<T>>` seems so obvious :-).

### DIFF
--- a/Src/FluentAssertions/AssertionExtensions.cs
+++ b/Src/FluentAssertions/AssertionExtensions.cs
@@ -643,6 +643,16 @@ namespace FluentAssertions
         }
 
         /// <summary>
+        /// Returns a <see cref="AsyncFunctionAssertions"/> object that can be used to assert the
+        /// current <see><cref>System.Func{Task{T}}</cref></see>.
+        /// </summary>
+        [Pure]
+        public static AsyncFunctionAssertions Should<T>(this Func<Task<T>> action)
+        {
+            return new AsyncFunctionAssertions(action, extractor);
+        }
+
+        /// <summary>
         /// Returns a <see cref="FunctionAssertions{T}"/> object that can be used to assert the
         /// current <see cref="System.Func{T}"/> .
         /// </summary>
@@ -651,6 +661,7 @@ namespace FluentAssertions
         {
             return new FunctionAssertions<T>(func, extractor);
         }
+        
 
 #if NET45 || NET47 || NETCOREAPP2_0
 
--- a/Tests/Shared.Specs/AsyncFunctionExceptionAssertionSpecs.cs
+++ b/Tests/Shared.Specs/AsyncFunctionExceptionAssertionSpecs.cs
@@ -190,6 +190,36 @@ namespace FluentAssertions.Specs
         }
 
         [Fact]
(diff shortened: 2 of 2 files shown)


Src/FluentAssertions/AssertionExtensions.cs
```diff
@@ -642,6 +642,16 @@ public static AsyncFunctionAssertions Should(this Func<Task> action)
             return new AsyncFunctionAssertions(action, extractor);
         }
 
+        /// <summary>
+        /// Returns a <see cref="AsyncFunctionAssertions"/> object that can be used to assert the
+        /// current <see><cref>System.Func{Task{T}}</cref></see>.
+        /// </summary>
+        [Pure]
+        public static AsyncFunctionAssertions Should<T>(this Func<Task<T>> action)
```
dennisdoomen: So the overload that takes a `Func<T>` does not override this one?

krajek: No, it does not.

For example all tests for `Should<T>(this Func<T> func)` and `Should(this Func<Task> action)` were not affected and still call respective overload. New overload is called only for two new tests I 've added.

Another important bit is `public class Task<TResult> : System.Threading.Tasks.Task`. Generic version inherits non-generic one, that allows overload resolution to choose `Task<T>` overload as one that is more specific.

PR #969, opened 2018-11-16
Optimize performance of IncludeMemberByPathSelectionRule 

jnyrup:
This PR has three parts:
* Lazy-loading `Segments` instead of eagerly in constructor
* Caching segments for candidates, by wrapping them as a `MemberPath`
* Use type check as fail-fast in `IsSameAs`

### DIFF
--- a/Src/FluentAssertions/Common/MemberPath.cs
+++ b/Src/FluentAssertions/Common/MemberPath.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Linq;
 
 namespace FluentAssertions.Common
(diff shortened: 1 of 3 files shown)


dennisdoomen:
What are you trying to accomplish with this PR?

jnyrup:
It triggered me that:
* the candidate `memberpath` was segmentized 1-3 times in `IncludeMemberByPathSelectionRule`, while we can do with atmost 1
* `segments` were eagerly computed, where I prefer to minimize the work in the constructor
* the segments were (to me) unnecessarily copied into a List.
* the candidate memberpath was segmentized before the `declaringTypes` are compared.

I think `IncludeMemberByPathSelectionRule` becomes prettier

Issue #966, opened 2018-11-15
Provide opt-out to `AssertionOptions(o => o.WithStrictOrdering())`


BrunoJuchli:
### Breaking change in 5.5 (coming from 5.4.2)
 
We globally set the default assertion equivalency options using:

	AssertionOptions.AssertEquivalencyUsing(options => options
		.WithStrictOrdering());

Because we believe it's safer to require strict ordering by default (I can provide a reasoning if anyone is interested).

Now, in special cases where the order did not matter, we used:

    collection.Should().BeEquivalentTo(
        ...,
        options => options.WithoutStrictOrderingFor(info => true));

to counteract the effect of `WithStrictOrdering()`. For certain scenarios this has stopped working with 5.5.0 (we now have a test fail which didn't fail before).

### Proposal

`WithoutStrictOrderingFor(info => true))` was a bit of an ugly workaround in the first place.
Instead of "reverting" FluentAssertions to work as before it seems more sensible to actually support `WithoutStrictOrdering` as a proper counterpart to `WithStrictOrdering`.

[`SelfReferenceEquivalencyAssertionOptions<TSelf>.WithStrictOrdering()`](https://github.com/fluentassertions/fluentassertions/blob/7738a652886af2e2ca0918ce71b197f7e06d8441/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs#L436)

performs

    orderingRules.Add(new MatchAllOrderingRule());

The suggested `SelfReferenceEquivalencyAssertionOptions<TSelf>.WithoutStrictOrdering()` should perform something along the lines of:

    var strictOrderingRules = orderingRules.OfType<MatchAllOrderingRule>().ToList();
    foreach(var strictOrderingRule in strictOrderingRules)
    {
        orderingRules.Remove(strictOrderingRule)
    }

### Complete minimal example reproducing the issue

I have taken the liberty of not providing a repro - for now. If necessary I can produce one, but since I don't really consider the 5.5 behavior a bug It seems more appropriate to focus on the feature request - which, I believe, does resolve the issue appropriately.

Note: I suspect this to be connected to https://github.com/fluentassertions/fluentassertions/issues/965
the reason is that this problem only surfaces in a very few tests even though we apply `WithoutStrictOrderingFor(info => true)` 166 times.

Please also note that I have tested the following cases:

#### Default `WithStrictOrdering`, override with `WithoutStrictOrderingFor(info => true)`
[...]

PR #974, opened 2018-11-18
Provide opt-out to `AssertionOptions(o => o.WithStrictOrdering())`

BrunoJuchli:
This handles #966 

Added the following configuration options:
- `WithoutStrictOrdering()`
- `WithoutStrictOrderingForBytes()`

I'm a bit unsure about `WithoutStrictOrderingForBytes()`. Reasoning:

Should `WithoutStrictOrdering()` remove `ByteArrayOrderingRule`? It seems likely that strict byte array ordering is almost always demanded. This poses a design-problem though. Since it's possible it was removed from the default it's not clear whether it should be present after performing:

     .WithStrictOrdering()
     .WithoutStrictOrdering()

Personally, it seems risky to have `WithoutStrictOrdering` remove the `ByteArrayOrderingRule`. One might expect bytes still to be compared in order. And you'll only find out your expectation is wrong when a bug occurs which you believed covered by your tests...

Having `WithoutStrictOrdering` always clearing the `OrderingRuleCollection` and then adding a `ByteArrayOrderingRule` might not be expected - especially if the `ByteArrayOrderingRule` was not present on the default. The issue will make itself visible by throwing an exception, though.
To remove the `ByteArrayOrderingRule` I could add another method to the interface (`WithoutStrictOrderingForBytes`). Removing all ordering rules would then become:

    .WithoutStrictOrdering()
    .WithoutStrictOrderingForBytes()

from a usage perspective I'd prefer this.


dennisdoomen:
I think we can live with the situation that byte arrays are _always_ compared in order, irrespective to the options you provide. Do you agree @jnyrup

jnyrup:
I agree that `ByteArrayOrderingRule` should still be present after calling `WithoutStrictOrdering`.

I'm not sure about `WithoutStrictOrderingForBytes`. 
How would you express that functionality today?

`WithoutStrictOrdering` should be documentated in https://github.com/fluentassertions/fluentassertions/blob/master/docs/_pages/documentation.md.

dennisdoomen:
> I'm not sure about `WithoutStrictOrderingForBytes`.

I don't think we need this, do we really?

BrunoJuchli:
> I'm not sure about `WithoutStrictOrderingForBytes`.
> How would you express that functionality today?

IMHO you can't.
Originally I had another impression because `IEquivalencyAssertionOptions.OrderingRules` is public gettable.

---------------

Hence I agree that `WithoutStrictOrderingForBytes` should not be introduced. Strict ordering for bytes should always be kept (as is the current behavior).
Furthermore `OrderingRuleCollection.Clear` must be internal, otherwise it would be possible to remove strict ordering for bytes.

> `WithoutStrictOrdering` should be documentated in https://github.com/fluentassertions/fluentassertions/blob/master/docs/_pages/documentation.md.

Will do, thanks for the hint.

jnyrup:
@BrunoJuchli Sounds good and thanks for thinking about making `Clear()` internal. No need to expose it.

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
```diff
@@ -435,6 +435,7 @@ public TSelf Using(IEquivalencyStep equivalencyStep)
         /// </summary>
         public TSelf WithStrictOrdering()
         {
+            orderingRules.Clear();
```
jnyrup: ❓ Was `orderingRules` ever non-empty before this PR? Does this affect existing tests in any way?

BrunoJuchli: @jnyrup 
In case of `SelfReferenceEquivalencyAssertionOptions`, `OrderingRules` always contained the Byte-Ordering rule (see [here](https://github.com/fluentassertions/fluentassertions/blob/7738a652886af2e2ca0918ce71b197f7e06d8441/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs#L64)).

This change did not make any existing tests fail.

As I've now just found out is that determining whether ordering should be `Strict` or `NotStrict` works differently than I thought. I thought the last rule takes priority, but that's not the case. Ordering remains `NotStrict` if the `OrderingRuleCollection` a rule which determines the member to be `NotStrict`, even if there's a newer rule that says "everything should be strict".
Hence calling `OrderingRuleCollection.Clear()` does affect the behavior. `WithStrictOrdering` calling `Clear()` overrides pre-existing `WithoutStrictOrderingFor` rules. Otherwise (not calling `Clear()`) `WithoutStrictOrderingFor` will remain effective.

Hence I've smuggled in a breaking change through the backdoor...

For reference, my design choice was based on the following understand (which I have found to be wrong):

It's my believe (i.E. i don't *know" for sure) that this should work. There's only two types of rules which can be contained in the list: those which require strict ordering and does which require non-strict ordering. Furthermore the last added rule wins. So if there's multiple contradictory rules for a member the last added one wins. Since `WithStrictOrdering()` and `WithoutStrictOrdering()` affect *all* members and - at this point - is the latest addition it will always win. All previously added rules become superfluous at this point.

Of course it would also have been *correct* (but not that nice to debug and not so performant) to just skip `orderingRules.Clear()`. `WithoutStrictOrdering()` would have, however, required a `MatchAllOrderingRule` counterpart which returns `OrderStrictness.NonStrict`.

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
```diff
@@ -449,6 +450,17 @@ public TSelf WithStrictOrderingFor(Expression<Func<IMemberInfo, bool>> predicate
             return (TSelf)this;
         }
 
+        /// <summary>
+        ///     Causes all collections - except bytes - to be compared ignoring the order in which the items appear in the expectation.
+        ///     To ignore the order of bytes use <seealso cref="WithoutStrictOrderingForBytes"/>.
```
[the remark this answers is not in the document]
BrunoJuchli: Thanks!

BrunoJuchli:
As per discussion with @jnyrup I've introduced a breaking change.

Previously:

    o => o.WithoutStrictOrderingFor(x => x.Foo)
               .WithStrictOrdering();

Would result in strict ordering for everything **but** `.Foo`.

Now the same configuration results in strict ordering for **everything**.

dennisdoomen:
I think option 1 is the best. The order of the methods is the important part.

jnyrup:
I agree that option 1 is the best.
A potential runtime breaking change is fine with me, as the current behavior is unexpected to me.

I would expect the following two equivalences:
```c#
.WithoutStrictOrderingFor(x => x.Foo)
.WithStrictOrdering();

// to be equivalent to

.WithStrictOrdering();
```

```c#
.WithStrictOrderingFor(x => x.Foo)
.WithoutStrictOrdering();

// to be equivalent to

.WithoutStrictOrdering();
```

We apparently only have a single test for each of `WithStrictOrderingFor` and `WithStrictOrderingFor`.
It would make me less worried with tests for both the `Expression<Func<IMemberInfo, bool>>` and the `Expression<Func<TExpectation, object>>` overloads of `WithStrictOrderingFor`/`WithoutStrictOrderingFor`.
This also acts as documentation for the intended behavior.

BrunoJuchli:
Thanks.

@jnyrup 
> We apparently only have a single test for each of `WithStrictOrderingFor` and `WithStrictOrderingFor`.
> It would make me less worried with tests for both the `Expression<Func<IMemberInfo, bool>>` and the `Expression<Func<TExpectation, object>>` overloads of `WithStrictOrderingFor`/`WithoutStrictOrderingFor`.
> This also acts as documentation for the intended behavior.

Will add the test for `WithStrictOrderingFor(Expression<Func<TExpectation, object>>)`.
AFAIR there's no `WithoutStrictOrderingFor(Expression<Func<TExpectation, object>>)`. I remember to have expected it but I didn't find it.

BrunoJuchli:
Thanks

@jnyrup 
> We apparently only have a single test for each of `WithStrictOrderingFor` and `WithStrictOrderingFor`.
> It would make me less worried with tests for both the `Expression<Func<IMemberInfo, bool>>` and the `Expression<Func<TExpectation, object>>` overloads of `WithStrictOrderingFor`/`WithoutStrictOrderingFor`.
> This also acts as documentation for the intended behavior.

I'm not really sure I understand what you mean.
As i gather there's no `WithoutStrictOrderingFor(Expression<Func<TExpectation, object>>)`. Only `WithStrictOrderingFor(...)` has two overloads.
For the existing overloads there's a test checking that performing `WithStrictOrdering` / `WithoutStrictOrdering` resets them. (=> only in the "resetting" direction, i.E. `WithStrictOrderingFor(..)` is reset by `WithoutStrictOrdering()`).

jnyrup:
To my surprise you're right.
I just assumed both overloads to exist for both `WithStrictOrderingFor` and `WithoutStrictOrderingFor`.
Seems like it should have been a part of #307.

My comment about a single test was regarding the existing test coverage for variants of `StrictOrderingFor`. As the fourth overload is non-existing, the test coverage you provide looks great.

PR #975, opened 2018-11-18
Allows BeEquivalentTo to handle a non-generic collection as the SUT

### LINKED ISSUE #973: ShouldBeEquivalentTo flags two seemingly equivalent objects
### Description

Version 5.5.1 fails existing test - used to to work in 5.4.2

Checkout the following repo: https://github.com/LogoFX/logofx-client-mvvm-viewmodel
Run the Unit tests - all tests are green
Upgrade to 5.5.1
Run the Unit tests - now some are red.

### Expected behavior:

Tests remain green

### Actual behavior:

Some tests are red

### Versions

- [x] FA 5.5.1
- [x] .NET Framework 4.6.1

dennisdoomen:
The `GenericEnumerableEquivalencyStep` did not properly support the subject to be a non-generic collection. 

Fixes #973

### DIFF
--- a/Tests/Shared.Specs/CollectionEquivalencySpecs.cs
+++ b/Tests/Shared.Specs/CollectionEquivalencySpecs.cs
@@ -649,6 +649,24 @@ namespace FluentAssertions.Specs
                 .WithMessage("*Customers[0].Name*John*Jane*");
         }
 
+
+        [Fact]
+        public void When_the_subject_is_a_non_generic_collection_it_should_still_work()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            object item = new object();
+            object[] array = new[] { item };
+            IList readOnlyList = ArrayList.ReadOnly(array);
+
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act / Assert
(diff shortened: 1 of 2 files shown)


Tests/Shared.Specs/CollectionEquivalencySpecs.cs
```diff
@@ -1598,30 +1616,6 @@ public void When_subject_is_null_and_expectation_is_enumerable_it_should_throw()
                 "Expected subject not to be <null>*");
         }
 
-        [Fact]
-        public void When_the_expectation_is_not_a_multi_dimensional_array_it_should_throw()
```
jnyrup: ❓ Why is this test no longer valid?

dennisdoomen: Because since v5, the expectation drives the type of comparison. Since it's a string, it'll never trigger the `MultiDimensionalArrayEquivalencyStep`

jnyrup: Makes sense. 
Couldn't see how that test should be affected by the other changes in this PR.

dennisdoomen: Don't even understand how this entire bug only surfaced now.

PR #977, opened 2018-11-20
Reverted changes to AssertionScope to maintain binary compatibility

dennisdoomen:
In #925, some of the methods of `AssertionScope` were changed to return an `IAssertionScope` instead of the full class. However, this causes binary compatibility issues for libraries extending FluentAssertions. If such a library depended on a version between 5.0 and 5.5, upgrading to 5.5 would cause missing method exception. I've reverted those changes by exposing the `IAssertionScope` interface as explicit interface implementations. 

This should fix https://github.com/fluentassertions/fluentassertions.json/issues/27

### DIFF
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -78,7 +78,7 @@ namespace FluentAssertions.Execution
             private set => current = value;
         }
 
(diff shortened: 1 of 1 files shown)


dennisdoomen:
Note: I decided not to change ``Continuation.Then` since it would not be possible and it is very unlikely that a lot of people are using that one.

PR #987, opened 2018-12-30
Fix (un)escaping when building message

### LINKED ISSUE #986: Backslashes in subject or expected result are not correctly shown in the message
### Description
It seems that when running a test on a string variable backslashes are escaped incorrectly. Especially trailing backslashes are removed from the message.

See multiple unit tests in this gist: https://gist.github.com/rubenrorije/2caf183ffb7e598546d220797252795c

One of the examples of the unit tests:
```csharp
[TestMethod]
public void Trailing_Backslash_In_Expected_Result_Should_Not_Be_Removed()
{
    "A".Should().Be("A\\");
    /*
     * Output:
     * Expected string to be "A" with a length of 2, but "A" has a length of 1, differs near "A" (index 0).
     * Expected Output:
     * Expected string to be "A\" with a length of 2, but "A" has a length of 1, differs near "A" (index 0).
     */
}
```

### Versions
Fluent Assertions 5.5.3
.Net Framework 4.6.1

### Remarks
I would have looked for myself whether this was something that I could solve myself and include a pull-request. But at the moment (travelling in Iran) I do not have reliable internet connection to do this. 
Either someone else can take a look, or when I have the opportunity I will look into it more.

jnyrup:
`Unescape` is the inverse of `Escape`, such that `str == Unescape(Escape(str))`.
But it is problematic if the escaped pattern already exists in the non-escaped string, as calls to `Escape` and `Unescape` doesn't seem to be balanced(?).

It's difficult to fix this properly, as `Unescape` would need to know if some substring was escaped or not.
The approach here is based on the assumption that no one ever uses the ASCII character `0x1B` (arbitrarily chosen).
Even though that is not universally true, I believe that character is less used than `\`.
By surrounding the escaped string with `0x1B` we can now see when a substring was escaped, and also guard against double escaping.

Two tests now fails, as `Unescape` is not called on the escaped string, so it still contains `0x1B`.
`When_element_has_attributes_it_should_include_them_in_the_output` `When_element_has_child_element_it_should_not_include_them_in_the_output`
I'm not sure if it's a problem though.

Fixes #986


dennisdoomen:
One alternative, albeit more difficult to do, would require us to introduce an more specialized string class that tracks whether something is escaped or not.

jnyrup:
Just tried to remove the replacement of `\r`, `\n` and `\"` from both `Escape` and `Unescape` and only the same two tests as above failed.

`Unescape` was introduced in b444a390 with no tests
and `Escape` was introduced in 7662fe3b.
So now I'm getting curious about what problem the replacement actually solves, as no tests fails when removing it.

dennisdoomen:
Actually, it was introduced as part of #705, so I guess it had to do that stuff didn't get formatted correctly. I also remember plenty of cases where FA crashed when somebody used a `{` or `}` in their message.

jnyrup:
The two failing tests asserts that the output is escaped.
I'm uncertain if that is a strict requirement or just a too strict matching assertion.

My test above with (un)escaping still escaped `{` and `}` as there are multiple tests exercising that.
I only removed (un)escaping of `\r`, `\n` and `\"` and that only failed to two tests that currently fails.
I've tried searching the issue/pull history but I couldn't find issues describing escaping those three characters.

dennisdoomen:
So the real problem you're trying to fix here is that we don't have a proper way of escaping something that contains escaped characters?

jnyrup:
@dennisdoomen I've updated the PR.
The failing test is a timing related test.

dennisdoomen:
Any reason not to merge it?

PR #989, opened 2019-01-06
[NRTM] Issue 296: Generic Collections NotBeInAscendingOrder and NotBeInDescendingOrder

### LINKED ISSUE #296: Collection Assertion NotBeInAscendingOrder with lambda not available but documentation says it is
Hi

The documentation says there is an overload for NotBeInAscendingOrder with a lambda but I can't find this one into the 3.4.1 version.  The latest 4.0.0 version does not seem to have this overload neither.

I'm looking for
`collection.Should().NotBeInAscendingOrder(x => x.SomeProperty);`

but only found 
`collection.Should().NotBeAscendingInOrder()`

Thanks
Phil

david-a-jetter:
[This ](https://github.com/fluentassertions/fluentassertions/issues/296) seemed like a simple enough issue to pick up as my first contribution.

- Add properly named non-generic assertion methods
-- Add Obsolete attribute to improperly named methods, to be removed in a major release
- Add methods to Generic Collection Assertions for NotBeInAscendingOrder and NotBeInDescendingOrder
- Add unit tests for above
- Add documentation for above

Resolves #296


docs/_pages/documentation.md
```diff
@@ -480,8 +480,8 @@ collection.Should().NotIntersectWith(anotherCollection);
 
 collection.Should().BeInAscendingOrder();
 collection.Should().BeInDescendingOrder();
-collection.Should().NotBeAscendingInOrder();
-collection.Should().NotBeDescendingInOrder();
+collection.Should().NotInBeAscendingOrder();
```
jnyrup: ❌ typo: `NotInBeAscendingOrder` => `NotBeInAscendingOrder`

docs/_pages/documentation.md
```diff
@@ -480,8 +480,8 @@ collection.Should().NotIntersectWith(anotherCollection);
...
+collection.Should().NotInBeDescendingOrder();
```
jnyrup: ❌ typo: `NotInBeDescendingOrder` => `NotBeInDescendingOrder`

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -267,6 +396,46 @@ public AndConstraint<GenericCollectionAssertions<T>> OnlyHaveUniqueItems<TKey>(E
...
+            {
+                throw new ArgumentNullException(nameof(comparer),
+                    "Cannot assert collection ordering without specifying a comparer.");
+            }
+
+            if (IsValidProperty(propertyExpression, because, args))
+            {
+                ICollection<T> unordered = Subject.ConvertOrCastToCollection();
+
+                if(! unordered.Any())
```
jnyrup: 🙃 Move the space in front of the opening parentheses. 
`if (!unordered.Any())`

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -267,6 +396,46 @@ public AndConstraint<GenericCollectionAssertions<T>> OnlyHaveUniqueItems<TKey>(E
...
+                IOrderedEnumerable<T> expectation = ConvertToOrderedEnumerable(
+                    propertyExpression,
+                    comparer,
+                    direction,
+                    unordered);
+
+                string orderString = GetExpressionOrderString(propertyExpression);
+
+                Execute.Assertion
+                    .ForCondition(! unordered.SequenceEqual(expectation))
```
jnyrup: 🙃 Remove the space from `! unordered`.

CONTRIBUTING.md
```diff
@@ -5,4 +5,4 @@ No open-source project is going to be successful without contributions. After we
 * The [Pull Request](https://help.github.com/articles/using-pull-requests) is targeted at the `master` branch.
 * The code complies with the [Coding Guidelines for C# 3.0, 4.0 and 5.0](https://csharpcodingguidelines.com/)/.
 * The changes are covered by a new or existing set of unit tests which follow the Arrange-Act-Assert syntax such as is used [in this example](https://github.com/fluentassertions/fluentassertions/blob/daaf35b9b59b622c96d0c034e8972a020b2bee55/Tests/FluentAssertions.Shared.Specs/BasicEquivalencySpecs.cs#L33).
-* If the contribution affects the documentation, please update [**the documentation**](https://github.com/fluentassertions/fluentassertions/tree/master/docs), which is published on the [website](https://fluentassertions.com/documentation/).
+* If the contribution affects the documentation, please update [**the documentation**](https://github.com/fluentassertions/fluentassertions/tree/master/docs), which is published on the [website](https://fluentassertions.com/documentation/).
```
jnyrup: ❓ Is this just a change of line ending?

Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -931,11 +931,28 @@ public AndConstraint<TAssertions> BeInDescendingOrder(IComparer<object> comparer
         /// <param name="becauseArgs">
         /// Zero or more objects to format using the placeholders in <see cref="because" />.
         /// </param>
+        [Obsolete("Use NotBeInAscendingOrder instead")]
         public AndConstraint<TAssertions> NotBeAscendingInOrder(string because = "", params object[] becauseArgs)
         {
             return NotBeAscendingInOrder(Comparer<object>.Default, because, becauseArgs);
```
jnyrup: 🔧 Use the new non-obsolete `NotBeInAscendingOrder`

Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -966,11 +1003,28 @@ public AndConstraint<TAssertions> NotBeAscendingInOrder(IComparer<object> compar
         /// <param name="becauseArgs">
         /// Zero or more objects to format using the placeholders in <see cref="because" />.
         /// </param>
+        [Obsolete("Use NotBeInDescendingOrder instead")]
         public AndConstraint<TAssertions> NotBeDescendingInOrder(string because = "", params object[] becauseArgs)
         {
             return NotBeDescendingInOrder(Comparer<object>.Default, because, becauseArgs);
```
jnyrup: 🔧 Use the new non-obsolete `NotBeInDescendingOrder`

CONTRIBUTING.md
david-a-jetter: Yeah, I don't know... I first tried to start recording which SDKs I had to install to get the solution to build and test, but I realized that there's no way I could provide an exhaustive list since of course I already had several installed that were presumably being used. I also figured... maybe that's better left to a separate PR.

I will try to eliminate.

david-a-jetter:
@jnyrup Addressed comments.

I agree with moving to the non-obsolete methods since those will be covered by unit tests and the "redirect" makes sense.

I personally prefer the space after the `!` since I find it stands out - people I work with often comment it is ugly, and I say that it's supposed to be - otherwise it's too easy to miss, but I can acquiesce to other styles too.

Lastly - I use this library all over the place, in nearly every project I work on. It's kinda cool to be contributing to it now.

dennisdoomen:
cc @jnyrup

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -267,6 +396,46 @@ public AndConstraint<GenericCollectionAssertions<T>> OnlyHaveUniqueItems<TKey>(E
...
+            {
+                throw new ArgumentNullException(nameof(comparer),
+                    "Cannot assert collection ordering without specifying a comparer.");
+            }
+
+            if (IsValidProperty(propertyExpression, because, args))
+            {
+                ICollection<T> unordered = Subject.ConvertOrCastToCollection();
+
+                if (!unordered.Any())
```
jnyrup: ❓ How about sequences with a single item?
I guess they are, like empty lists,  also both ordered and unordered at the same time.

david-a-jetter: @jnyrup My initial thought was that such an assertion should `pass` / not throw. 

However, then I think about what problem this poses to the end user - and that would be that their test fails when really their test is invalid, and is either not using the appropriate assertion or does not have the appropriate arrange/act. And in that case, I would prefer the safety net of the failing assertion so I have an opportunity to correct my test.

The above may be confirmation bias because this is also the current behavior 😕

dennisdoomen:
@jnyrup can this me merged?

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
jnyrup: Here are some quick tests (if incorporated, please update them to FA style) on how FA currently behaves.
Would this PR break any of those?

```c#
[TestMethod]
public void EmptyCollection_BeInAscendingOrder()
{
    Action act = () => new int[] { }.Should().BeInAscendingOrder();
    act.Should().NotThrow();
}

[TestMethod]
public void EmptyCollection_BeInDescendingOrder()
{
    Action act = () => new int[] { }.Should().BeInDescendingOrder();
    act.Should().NotThrow();
}

[TestMethod]
public void SingleElement_BeInAscendingOrder()
{
    Action act = () => new int[] { 42 }.Should().BeInAscendingOrder();
    act.Should().NotThrow();
}

[TestMethod]
public void SingleElement_BeInDescendingOrder()
{
    Action act = () => new int[] { 42 }.Should().BeInDescendingOrder();
    act.Should().NotThrow();
}

[TestMethod]
public void EmptyCollection_NotBeInAscendingOrder()
{
    Action act = () => new int[] { }.Should().NotBeAscendingInOrder();
    act.Should().Throw<Exception>();
}

[TestMethod]
public void EmptyCollection_NotBeInDescendingOrder()
{
    Action act = () => new int[] { }.Should().NotBeDescendingInOrder();
    act.Should().Throw<Exception>();
}

[TestMethod]
public void SingleElement_NotBeInAscendingOrder()
{
    Action act = () => new int[] { 42 }.Should().NotBeAscendingInOrder();
    act.Should().Throw<Exception>();
}

[TestMethod]
public void SingleElement_NotBeInDescendingOrder()
{
    Action act = () => new int[] { 42 }.Should().NotBeDescendingInOrder();
    act.Should().Throw<Exception>();
}
```

jnyrup:
@dennisdoomen Not yet, see discussion above about the behavior of `NotBeOrderedBy`.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
jnyrup: @david-a-jetter Have you had a chance to look if the tests above still passes?

david-a-jetter: @jnyrup These tests pass - and I can work to rearrange them so they match in style of the other tests.

However, it seems like a similar test using an anonymous type and specifying the property results in a failure. Looks like more work to do still... different behavior of primitive collections vs. complex types, which is not okay. For example:

```csharp
        [Fact]
        public void When_asserting_single_item_collections_are_not_in_an_descendingly_ordered_collection_are_not_ordered_descending_it_should_succeed()
        {
            // Arrange
            var collection = new[]
            {
                new { Text = "b", Numeric = 1 }
            };

            // Act
            Action act = () => collection.Should().NotBeInDescendingOrder(o => o.Numeric, Comparer<int>.Default);

            // Assert
            act.Should().NotThrow();
        }
```

dennisdoomen: Hmm. If somebody uses these methods, aren't they already expecting a collection? Why else would you use those specific assertions? So if instead of a collection, the test results in an empty collection or a collection with just one item, I would expect it to fail.

dennisdoomen: > Hmm. If somebody uses these methods, aren't they already expecting a collection? Why else would you use those specific assertions? So if instead of a collection, the test results in an empty collection or a collection with just one item, I would expect it to fail.

@jnyrup did you see ⬆️ ?

jnyrup: @david-a-jetter I'm not sure what the problem is (or maybe I forgot).

I just tried asserting collection sortedness in NUnit:
empty and single element collections both:
* passes `Assert.That(list, Is.Ordered);`, and
* fails `Assert.That(list, Is.Not.Ordered);`

I came to think of two inherent properties of empty/single element collections.
They are:
* always ordered, and
* both in ascending *and* descending order at the same time.

dennisdoomen:
Can we come up with some guidelines on what behavior we expect, so @david-a-jetter or somebody else can finish this PR?

danielmpetrov:
Is this PR still of interest? @dennisdoomen @jnyrup 

What's the stopper here?

Maybe if @david-a-jetter is MIA I can try to finish it instead.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
dennisdoomen: > I came to think of two inherent properties of empty/single element collections.
> They are:
> 
> * always ordered, and
> * both in ascending _and_ descending order at the same time.

I think this is what we should implement.

Issue #1003, opened 2019-03-02
Tests are language dependent


lg2de:
I wondered why some tests fail while your build seams to complete successfully.

I found, that some tests are language dependend, e.g. `When_asserting_collections_not_to_be_equal_but_expected_collection_is_null_it_should_throw`:
```cs
act.Should().Throw<ArgumentNullException>().WithMessage(
    "Cannot compare collection with <null>.*Parameter name: unexpected");
```

The test verifies the text from exception, but the ArgumentNullException generates text depending on user/windows settings. On my (german) system the text is "Parametername" instead of "Parameter name".

I see three options:
1. remove (always) language dependend part of verified message ("Cannot compare collection with <null>.*: unexpected")
2. run all tests explicitely with english locale (either in build.ps1 or [better] within the test)
3. introduce and use UseCultureAttribute from [xUnit samples](https://github.com/xunit/samples.xunit/blob/master/UseCulture/UseCultureAttribute.cs)

What do you think?
For me, option 3 would be the best.

dennisdoomen:
I would go for option 2. I would always code in English. In the Netherlands, it's quite common to work with non-Dutch developers. 

Regardless, this issue is not related to Fluent Assertions.

lg2de:
Hi @dennisdoomen, I think this is a misunderstandig: I was talking about the Unit Tests of FluentAssertions. So, it is related to FA.
I'm coding in English, but other things are fine in German. So, I would like to have configured my Windows to German. But THAN "your" Fluent Assertions Unit Tests will fail.

I think, this should be fixed. The tests should NOT be related to the language selected for the OS they are running on!

PR #1005, opened 2019-03-03
Possible solution for culture invariant tests using UseCultureAttribute

lg2de:
In #1003 I listed several options to run the Fluent Assertions tests on all Windows locales.
In this PR I want to show a possible by running culture related test explicitely on the expected culture.


dennisdoomen:
> This PR modifies global static variables between parallel runs of unit tests.

For the .NET versions that support it, it will change it only for the length of a test. However, an alternative would be to set the culture for the entire test suite. 

We could be more liberal in what we expect, but we have no way to test if this works. AppVeyor agent always use an English version of Windows.

jnyrup:
@dennisdoomen I'm not sure I understand you regarding the length of a test?
When tests are run in parallel, setting a static variable has nothing to do with the lifetime of a single test.

I would rather set the culture at one single point, which is ensured to be run before any tests, than setting the culture on individual tests.
I don't know how to do that in xunit, or if it even possible without making each class inherit from some base class, which I'm not the biggest fan of.

I was thinking of trying to run all your tests in all locales, to check if we have more tests fragile to different locales.

dennisdoomen:
Yeah, you're right. Even when it only changes the culture of the current thread temporarily, thread jumping may happen in `async` tests. We could try to force the entire test to the invariant culture, but XUnit does not have a nice way for that other than using collection fixtures on each test class.

lg2de:
You are right. This is not the best solution. Do you have another idea to MAKE the tests culture invariant?
I tried setting the culture within build.ps1, but I did not succeed.

This solution helps ME to get successful completed tests. It will not affect the CI build environment because it is already english locale. The modified tests are NOT using async.
I would like to provide PR for another issue(s). Do I need to switch my PC to english locale to contribute to FluentAssertions?
There are other test failing due to timing issues. Maybe we can live with this too?

jnyrup:
It seems to be possible in xUnit.net: https://github.com/xunit/samples.xunit/blob/master/AssemblyFixtureExample/Samples.cs

Such culture fixing should not be be set in ps1/cake code.
Running tests from VS or via CLI should yield identical results.

lg2de:
Hi @jnyrup, I did not use AssemblyFixture, because this is not a build-in feature of Xunit.
Instead I introduced CollectionFixture on the classes with "localized" tests.

jnyrup:
@lg2de bummer...
I'm still no fan of this approach.
How do you decide when a class should be decorated with `[Collection("Invariant Culture Collection")]`?

I liked how #726 handled culture problems.

lg2de:
Mmh, it looke quite difficult to match your expectations...
The approach of mentioned PR does not match this problem because I cannot control how ArgumentNullException builds its text. It is just depending on the locale.

I think we need to change the tests to NOT check the locale dependent tests...

jnyrup:
@lg2de 
> I cannot control how ArgumentNullException builds its text

Exactly, and that's why the test shouldn't assert on that part of the exception message.
What we are interested in, is that the property `ParamName` should be `unexpected`.

That's what I proposed in the my first examples of rewriting the test.

jnyrup:
I'm not sure if you're trying to alter the tests as little as possible?
As the framework has evolved, some tests may have suffered from not being rewritten with newer features. 
I think the usage of `.Where(ex => ...)` in `ObjectAssertionSpecs` is such an example.
In my opinion those assertions are more clearly expression through `WithMessage()`.

Here are some examples on how I would probably write those tests.

`When_an_object_is_binary_serializable_but_not_deserializable_it_should_fail`
```c#
    .WithMessage("*to be serializable, but serialization failed with:**BinarySerializableClassMissingDeserializationConstructor*");
```

`When_an_object_is_not_data_contract_serializable_it_should_fail`
```c#
    .WithMessage("*we need to store it on disk*EnumMemberAttribute*");
```
I agree that the `ObjectAssertionSpecs` specs are not the easiest to write robust and clear tests for.


`When_asserting_collections_not_to_be_equal_but_expected_collection_is_null_it_should_throw`	
```c#
act.Should().Throw<ArgumentNullException>()
    .WithMessage("Cannot compare collection with <null>*")
    .And.ParamName.Should().Be("unexpected");
```
Instead of relying on what the message contains, it not relies on the type system.

lg2de:
Dear @jnyrup, this was my third attempt to fulfil your expectations. Step by step I loss motivation to continue...
For sure, I tried to change as less as possible. I replaced ONLY (!!!) the checks of german texts. These texts are comparing texts from .NET framework and should not be checked by FluentAssertions tests.

I've updated When_an_object_is_not_data_contract_serializable_it_should_fail according to your suggestion. The other test you may change later. I think, I was following existing patterns of other tests.

My main goal was to get the master branch teastable on a German PC (like my one) to contribute another feature(s).

lg2de:
... and your recommendation When_an_object_is_binary_serializable_but_not_deserializable_it_should_fail will not work: "to be serializable, but serialization failed with" is part of the localized text!

dennisdoomen:
@jnyrup  Let's just merge this one. If it fixes the problem @lg2de describes, it's fine by me. Feel free to apply a better fix as a reference example after that.

PR #1008, opened 2019-03-10
Improve robustness against localization

jnyrup:
This PR improves robustness when the tests are run on computers with non-us locales.

The first commit builds upon #1005 by asserting on `ParamName` instead of the message of `ArgumentException`.

The second commit improves conversion between `string` and `{DateTime, DateTimeOffset, TimeSpan}` by either specifying `Culture.InvariantCulture` or using the `FluentDateTimeExtensions`.

When using `{DateTime, DateTimeOffset, TimeSpan}.ToString(string format)` the current `CultureInfo` is used which might have other [`DateSeparator`](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.datetimeformatinfo.dateseparator?view=netframework-4.7.2) [`TimeSeparator`](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.datetimeformatinfo.timeseparator?view=netframework-4.7.2) than expected.
E.g. `When_a_full_date_and_time_is_specified_all_parts_should_be_included_in_the_output` failed in Danish, as the default Danish `TimeSeparator` is `.` and not `:`.


lg2de:
I can confirm, the tests are (still) running fine with German PC... :)

Issue #1001, opened 2019-02-28
Assertion that asynchronous operation completes within time span


lg2de:
Testing asynchronous operations is currently available like this:
```cs
this.myServer.Invoking(s => s.OpenAsync).Should().NotThrow();
```

I'm missing the option just check, that the operation will complete (and not throw) within specific time range. Therefore a new assertion could be created:
```cs
this.myServer.Invoking(s => s.OpenAsync).Should().Complete(TimeSpan.FromMilliseconds(100));
```

Such an idea I already mentioned in #990.

While thinking about the implementation I found a different approach.
Why to I need the indirection using "Invoking"? Why not asserting the Task directly:
```cs
this.myServer.OpenAsync().Should().Complete(TimeSpan.FromMilliseconds(100));
```

bitbonk:
I have come across this problem too. Sometimes those asynchronous operations that are being tested may never complete or may take an unusual long time to complete. As a result, the whole test run takes forever to complete. 

Sometimes I just want a test to fail if the asynchronous operation does not complete within a given amount of time. In most cases actually *is* something wrong if a task takes forever to complete.

bitbonk:
One more thing: What should happen to those tasks that did not complete in case the assertions failed? Is here a problem if tasks stay alive after the test is done? If there is a `CancellationToken`, can FA take care of cancelling those tasks?

lg2de:
So, @bitbonk, you prefer following the Invoking pattern?
Thinking about again, the timing may be more exact than working on the Task directly.
But, the approach without `Invoking` would be shorter.

Regarding the `CancellationToken` I do not see an option for FA.
But using e.g. xUnit the testing framework can take care, isn't it?

dennisdoomen:
> If there is a CancellationToken, can FA take care of cancelling those tasks?

The `CompleteWithin` could take the cancellation token as a parameter, but it won't be very intuitive.

lg2de:
@dennisdoomen, which approach you prefer? Based on Invoking or directly on the Task?

dennisdoomen:
Both ;-)

lg2de:
> Both ;-)

äähhmmm, oooh...
I'll try to do my best.

ArndtMichael:
I would prefer handling with the Task directly because:
The assertion is related to exactly that task and not that one Invoking returns. It's more clear what happens.

lg2de:
[...]

Issue #990, opened 2019-01-08
Assertions for Func<Task<T>>


jnyrup:
Currently both `Func<Task>` and `Func<Task<T>>` are handled by `AsyncFunctionAssertions` which has a `Subject` of type `Func<Task>` and hence ignores the return value for `Func<Task<T>>`.

If we split `AsyncFunctionAssertions` into:
* `AsyncFunctionAssertions` for `Func<Task<T>>`, and
* `AsyncActionAssertions` for `Func<Task>`

That should enable us to return the result of `Task<T>` for further assertions, similar to what was implement ed for `Func<T>` in #951, such as:

```c#
// Arrange/Act
Func<Task<int>> func = () => Task.FromResult(42);

// Assert
func.Should().NotThrow().Which.Should().Be(42);
(await func.Should().NotThrowAsync()).Which.Should().Be(42);
```

This was inspired by this [question](https://stackoverflow.com/questions/54092547/chceck-return-value-in-fluentassertion-exception-syntax)

dennisdoomen:
Yeah, sounds like a good idea.

lg2de:
I would like to combine this extension with an additional feature:
```c#
func.Should().Complete(TimeSpan.FromMilliseconds(100));
```

The test would fail if the Task does not complete within specified time span.

What do you think about such feature?

dennisdoomen:
That's a nice idea, and a very useful one as well.

lg2de:
While thinking about the implementation I would like to check whether my "Complete" extension should be based on `Func<Task<T>>` or maybe better on `Task<T>`.
The latter would result into easier usage:
```c#
this.myServer.OpenAsync().Should().Complete(TimeSpan.FromMilliseconds(100));
```

In this case my changes would not touch anymore the refactoring mentioned by @jnyrup.

jnyrup:
@lg2de I would prefer if you opened a new issue with an API proposal describing the purpose of the functions and their function signatures.

PR #1011, opened 2019-03-19
Task func extensions

lg2de:
This PR wants to implement the issues #990 and #1001 .
Maybe it is not yet completed, but I think it is time for a review.

Please note:
## Timing issues
I implemented new functionality which is related to timing. Unfortunately this cause failed tests from time to time because the real timing cannot be setup as exact as required (without creating real slow tests).

I already have seen some other tests which have the same problem. But I have no (easy) idea yet, how to get the problem fixed.

I think it will only work if the timing is fully mocked. We could introduce a new internal interface like that:
```cs
internal interface ITimer
{
    Task Delay(TimeSpan delay);
}
```
Maybe you know already a good abstraction for the timer.
Anyway, the full solution is a new issue, I guess.

## Unit test method naming
I added several unit tests for the new functions. But for me, your naming style for the test methods is completely new (never seen before).

I tried to build names according to the existing tests. Hopefully it is ok. Otherwise please tell me the correct names or even better: Please just rename the tests directly, if needed.

BTW: What do you think about the popular style [UnitOfWork_StateUnderTest_ExpectedBehavior]?


jnyrup:
Are you aware that #990 is marked with `breaking change` label?
As we follow [semantic versioning](https://semver.org/), we try to our very best *only* to introduce breaking changes in major releases.
The CoreFX repo has a nice [guide](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md#source-and-binary-compatibility-changes) about what counts as a source or binary breaking change.

The `master` branch should always be deployable, i.e. be ready to package as a nupkg and releasing on nuget.org.
As such the breaking changes in this PR blocks it from being merged to master.

dennisdoomen:
Indeed, @jnyrup is right. And I don't foresee any 6.0 version before the end of the year. Would it be possible to only include the changes needed for #1001?

lg2de:
Hi @jnyrup and @dennisdoomen!

Yes, I had seen the label "breaking change". You are absolutely right, the changes are not binary compatible.

But for my understanding, the backend classes like `AsyncFunctionAssertions` are not intended for direct use. The users of FA will only use the `Should()` method which returns the correct implementation. The user is not interested in the concrete class.

I found that FA is a strong named assembly. So, changing version to 5.7.X would require recompilation anyway. Do we really need to wait for version 6.0 for this update? What are your planes for 6.0 to wait until end of year?

With PR #1013 I provided subset of this PR for #1001, as requested.

lg2de:
abort

PR #1012, opened 2019-03-22
Trying to fix #821 by traversing interfaces

### LINKED ISSUE #821: Event Monitor Doesn't Detect Events on Interface
Given:

```c#
public Interface ITest : INotifyPropertyChanged {...}
public class Test : ITest {...}
```

I would expect:

```c#
var subject = new Test() as ITest;
using(var monitor = subject.Monitor()) {....}
```

would be a reasonable test. Currently, the `Monitor()` extension method throws an exception saying the object has no events. From looking at the code, it would appear this is because it depends on the `Type.GetEvents()` method, which does not consider interface inheritance in its examination. As far as I'm concerned, that would easily be argued as a bug in the .NET Framework, but given how long it's been there and the potentially breaking nature of a change, I don't imagine we will see it changed ever.

Replacing this code is likely non-trivial, but not overly hard either (and if I have time I will try to get around to it if nobody beats me to the punch), given that interfaces show up differently in the type hierarchy than base classes do.

Ignoring the contrived nature of cast to `ITest` in the above sample, the problem is showing up for me when trying to write tests to the interface (ie. the contract) when I don't know the object implementing it because I have the same tests for multiple objects implementing the interface (Liskov substitution).

Lette:
Ran into #821 when upgrading to FA 5.

Inspiration from: https://stackoverflow.com/a/26766221/11808

### DIFF
--- a/Src/FluentAssertions/Events/EventMonitor.cs
+++ b/Src/FluentAssertions/Events/EventMonitor.cs
@@ -82,7 +82,7 @@ namespace FluentAssertions.Events
                 throw new InvalidOperationException("Cannot monitor events on garbage-collected object");
             }
 
-            EventInfo[] events = typeDefiningEventsToMonitor.GetEvents();
+            EventInfo[] events = GetPublicEvents(typeDefiningEventsToMonitor);
             if (!events.Any())
             {
                 throw new InvalidOperationException($"Type {typeDefiningEventsToMonitor.Name} does not expose any events.");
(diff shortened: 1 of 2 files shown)


dennisdoomen:
Great work @Lette

lg2de:
With the PR this issue seems to be fixed. It should be closed.
I found event monitoring on interface to be working fine.

PR #1013, opened 2019-03-23
Assertions for Task and Task<T> with CompleteWithin checks

lg2de:
This PR implements #1001.
In contrast to PR #1011 this attempt provides only the extensions on `Task`/`Task<T>` and should not be a breaking change.

As already noted, I found timing issues in my and other tests. Today I added a prototype to abstract real timing in the tests. What do you think?

Unfortunately I could not make `ITimer` internal as planned. FluentAssertions is a strong named assembly. So, the test assembly must use strong named too, if it should refer to `ITimer`. (Otherwise the `InternalVisiblesTo` attribute will fail.) This is not possible because the Chill package is not strong named.
Please help to find solution to make `ITimer` internal.


dennisdoomen:
You can't make `ITimer` internal because it's a parameter of the `TaskAssertions` class. And we kept all those assertion classes (the ones returned by the `Should` methods) because this allows people to extend them.

lg2de:
Ok. Lets keep it public and start official review.

Src/FluentAssertions/Common/TaskTimer.cs
```diff
@@ -0,0 +1,21 @@
...
+    internal class TaskTimer : ITimer
+    {
+        public Task DelayAsync(TimeSpan delay)
+        {
+            return Task.Delay(delay);
+        }
+
+        public bool Wait(Task task, TimeSpan timeout)
+        {
+            return task.Wait(timeout);
```
dennisdoomen: This can cause deadlocks (which seem to happen more often on AppVeyor). Consider using a construct like https://github.com/ChillBDD/Chill/blob/master/Src/Chill/Common/TaskExtensions.cs#L16 or use a `Task.Run`.

Src/FluentAssertions/Specialized/TaskAssertions.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public void CompleteWithin(
+            TimeSpan timeSpan, string because = "", params object[] becauseArgs)
+        {
+            var completed = this.timer.Wait(Subject, timeSpan);
```
dennisdoomen: (minor) You don't need `this` here.

Src/FluentAssertions/Specialized/TaskAssertions.cs
```diff
@@ -0,0 +1,71 @@
...
+            this.timer = timer;
+        }
+
+        /// <summary>
+        /// Gets the <see cref="Task"/> that is being asserted.
+        /// </summary>
+        public Task Subject { get; private set; }
+
+        /// <summary>
+        /// Asserts that the current <see cref="Task"/> will complete within specified time range.
```
dennisdoomen: (nitpick) It's not really a time _range_, is it?

Src/FluentAssertions/Specialized/TaskOfTAssertions.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public async Task<AndWhichConstraint<TaskOfTAssertions<T>, T>> CompleteWithinAsync(
+            TimeSpan timeSpan, string because = "", params object[] becauseArgs)
+        {
+            var delayTask = this.timer.DelayAsync(timeSpan);
+            var completedTask = await Task.WhenAny(Subject, delayTask);
```
dennisdoomen: If the `Task` represented by `Subject` completes before the timeout, you must cancel the timer. Something like https://github.com/liquidprojections/LiquidProjections/blob/master/Src/LiquidProjections.Testing/TaskExtensions.cs#L9

Tests/Shared.Specs/TestingTimer.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.Threading.Tasks;
+using FluentAssertions.Common;
+
+namespace FluentAssertions.Specs
+{
+    /// <summary>
+    /// Implementation of <see cref="ITimer"/> for testing purposes only.
+    /// </summary>
+    public class TestingTimer : ITimer
```
dennisdoomen: Can be made `internal`, right?

Src/FluentAssertions/Common/TaskTimer.cs
lg2de: I have introduced solution provided by #1027.

Src/FluentAssertions/Specialized/TaskAssertions.cs
lg2de: I do not want to start such discussion here. I have different understanding for good coding style. I would like to identify members (and I do not line the "_member" aproach).
If you want I can remove "this.".

Src/FluentAssertions/Specialized/TaskAssertions.cs
lg2de: done

Src/FluentAssertions/Specialized/TaskOfTAssertions.cs
lg2de: Why do you think, that the delay task needs to cancelled. It will just complete in background without reasonable effort, I think.
Introducing new items like a `CancellationToken` will also create effort.
I think this is in balance.

Tests/Shared.Specs/TestingTimer.cs
lg2de: I have changed this.
But is it really useful? This class is defined in the assembly for tests an will never get public.

dennisdoomen: If you keep it public, it implictly becomes part of our "contract". I'd like to avoid that.

Src/FluentAssertions/Specialized/TaskOfTAssertions.cs
dennisdoomen: Just imagine a situation with thousands of tests running concurrently. If the timeout expires before the task completes, you might end up with a lot of unfinished background activity. This is a situation that you need to avoid.

Src/FluentAssertions/AssertionExtensions.cs
```diff
@@ -653,6 +653,26 @@ public static AsyncFunctionAssertions Should<T>(this Func<Task<T>> action)
...
         }
 
+        /// <summary>
+        /// Returns a <see cref="TaskAssertions"/> object that can be used to assert the
+        /// current <see cref="Task"/> .
+        /// </summary>
+        [Pure]
+        public static TaskAssertions Should(this Task task)
+        {
+            return new TaskAssertions(task, new TaskTimer());
```
jnyrup: 🤔 If we assert on a `Task` hasn't the execution already started? 
That would mean we aren't measuring the execution time from the beginning of the execution.

Tests/Shared.Specs/TestingTimer.cs
```diff
@@ -0,0 +1,35 @@
...
+        private TaskCompletionSource<bool> Signal { get; } = new TaskCompletionSource<bool>();
+
+        Task ITimer.DelayAsync(TimeSpan delay)
+        {
+            return this.Signal.Task;
+        }
+
+        bool ITimer.Wait(Task task, TimeSpan timeout)
+        {
+            Signal.Task.Wait();
```
jnyrup: 🔧 Use `GetAwaiter().GetResult()`? Related: #1038

Src/FluentAssertions/AssertionExtensions.cs
lg2de: Yes, that may happen.
Do you have any suggestion?

Tests/Shared.Specs/TestingTimer.cs
lg2de: done

Src/FluentAssertions/AssertionExtensions.cs
jnyrup: How about reusing the existing `Func<Task>` and `Func<Task<T>>` extension methods?

lg2de: Could you be a little more specific, please?
Further I think that such a problem cannot be solved completely precisely: The timing starts either before the start of the task and measures too much time. Or the measurement starts after the task and measures too much time.

jnyrup: Of course, I would reuse these two existing overloads of `Should` that takes `Func<Task>` and `Func<Task<T>>`
```c#
public static AsyncFunctionAssertions Should(this Func<Task> action)
public static AsyncFunctionAssertions Should<T>(this Func<Task<T>> action)
```

Measuring execution times is inherently difficult, for unit tests without warm-up, is downright unreliable.
That requires dedicated benchmark tools such as [BenchmarkDotNet](https://benchmarkdotnet.org/).
E.g. the order in which the unit tests are executed now matters, due to GC, JITting, threads, etc.
As we can only execute a single statement at a time, we cannot start measuring at the exact same time the execution starts.
All other places (I hope so) we start the clock before starting the execution.
See e.g. [ExecutionTimeAssertions.cs](https://github.com/fluentassertions/fluentassertions/blob/master/Src/FluentAssertions/Specialized/ExecutionTimeAssertions.cs)

lg2de: Within this kind of extension we have no chance to start timing before starting the `Task`. It is an extension for the task (instance) itself.
We can either start timing after task is running or we can stop here. Isn't it?

jnyrup: My thought is to avoid these two overloads of `Should` and use the existing ones.
The reason is to avoid fragmenting the API.
Then `CompleteWithin` could be added to `AsyncFunctionAssertions`?

dennisdoomen: I agree with @jnyrup. And it's not just about _when_ your timer starts running. It's purely to ensure we have a consistent API. After reading our response on Slack, I really hope you're willing to finalize this and the other comments. I'd love to have this in the next release.

dennisdoomen:
Replaced by #1048

PR #1019, opened 2019-03-31
Removed redundant multithreading (issue #1020)

davidomid:
This resolves issue [#1020](https://github.com/fluentassertions/fluentassertions/issues/1020)

Removed usages of Task.Run and Task.Factory.StartNew, in places where additional threads are redundant.


dennisdoomen:
Some of those were specifically introduced to prevent deadlocks. And your build is now deadlocking as well. Or am I missing something here?

davidomid:
> Some of those were specifically introduced to prevent deadlocks. And your build is now deadlocking as well. Or am I missing something here?

Hi Dennis, 

This branch's tests were running fine locally initially. I could be wrong but it now appears the tests which are failing right now are similar to the ones failing on the master branch currently.  I think it's best we fix the master branch, then I can re-evaluate the status of this branch. 

As for specifically introducing the additional threads to prevent deadlocks, are there any specific examples of where these deadlocks initially occurred? I believe there are more effective ways of handling deadlocks than using additional threads, although it may well be that there was a particular case where this wasn't possible in some other way. I'm more than happy to look into this, once we get the master branch build working again. 

Thanks very much :)

dennisdoomen:
In general, you're right. But we've noticed that the combination of AppVeyor, xUnit and task.Wait() often results in deadlocks.

dennisdoomen:
It looks like I'm right. Maybe you can undo only the ones where you call `Wait()` explicitly. xUnit has build their own scheduler and I think that's what's hurting us.

davidomid:
I'll investigate those deadlocks and create another pull request in the future with a different approach.

Issue #1020, opened 2019-03-31
AsyncFunctionAssertions methods contain redundant multithreading


davidomid:
Fixed in PR #1027

### Description

The AsyncFunctionAssertions class contains methods with redundant multithreading. 

These methods are:

- public void NotThrow(string because = "", params object[] becauseArgs)
- public async Task NotThrowAsync(string because = "", params object[] becauseArgs)
- public void NotThrow<TException>(string because = "", params object[] becauseArgs)
            where TException : Exception
- public async Task NotThrowAsync<TException>(string because = "", params object[] becauseArgs)
            where TException : Exception
- private Exception InvokeSubjectWithInterception()
- private async Task<Exception> InvokeSubjectWithInterceptionAsync()

### Complete minimal example reproducing the issue

E.g. this is how NotThrowAsync currently looks:

```csharp
public async Task NotThrowAsync(string because = "", params object[] becauseArgs)
{
     try
     {
          await Task.Run(Subject);
     }
     catch (Exception exception)
     {
          NotThrow(exception, because, becauseArgs);
     }
 }
```

This is how I would expect it to look:

```csharp
public async Task NotThrowAsync(string because = "", params object[] becauseArgs)
{
     try
     {
          await Subject();
     }
     catch (Exception exception)
     {
          NotThrow(exception, because, becauseArgs);
     }
 }
```

### Expected behavior:

The method should await the task returned by the Subject func. 

In the other async methods listed above, I expect similar behaviour. In the non-async methods, I expect the task to be awaited in a blocking way (i.e. using .Wait() instead of await). 

### Actual behavior:

The method awaits a new task returned by Task.Run, which is sent the Subject func. 
This schedules the work to be done on the thread pool, potentially using a different thread. 
This adds overhead and is otherwise redundant. 

This behaviour is similar for other methods listed above. 

### Versions

Any version

jnyrup:
Fixed in #1027

PR #1027, opened 2019-04-19
Removed redundant multithreading

davidomid:
Fix for issue #1020

Removed redundant multithreading which was implemented using Task.Run.

### DIFF
--- a/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
@@ -2,7 +2,6 @@
 using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
-
 using FluentAssertions.Execution;
 
 namespace FluentAssertions.Specialized
@@ -74,7 +73,7 @@ namespace FluentAssertions.Specialized
         {
             try
             {
-                Task.Run(Subject).Wait();
+                Subject().Wait();
             }
             catch (Exception exception)
             {
@@ -96,7 +95,7 @@ namespace FluentAssertions.Specialized
         {
             try
             {
-                await Task.Run(Subject);
+                await Subject();
             }
             catch (Exception exception)
             {
@@ -119,7 +118,7 @@ namespace FluentAssertions.Specialized
         {
             try
             {
-                Task.Run(Subject).Wait();
+                Subject().Wait();
             }
             catch (Exception exception)
             {
@@ -142,7 +141,7 @@ namespace FluentAssertions.Specialized
         {
             try
             {
-                await Task.Run(Subject);
+                await Subject();
             }
             catch (Exception exception)
             {
@@ -323,7 +322,7 @@ namespace FluentAssertions.Specialized
         {
             try
             {
-                Task.Run(Subject).Wait();
+                Subject().Wait();
                 return null;
             }
             catch (Exception exception)
@@ -336,7 +335,7 @@ namespace FluentAssertions.Specialized
         {
             try
             {
-                await Task.Run(Subject);
+                await Subject();
                 return null;
             }
             catch (Exception exception)
--- a/Src/FluentAssertions/AssertionExtensions.cs
+++ b/Src/FluentAssertions/AssertionExtensions.cs
@@ -7,6 +7,7 @@ using System.Reflection;
 using System.Threading.Tasks;
 using System.Xml.Linq;
 using FluentAssertions.Collections;
+using FluentAssertions.Common;
 using FluentAssertions.Events;
 using FluentAssertions.Numeric;
 using FluentAssertions.Primitives;
@@ -77,7 +78,7 @@ namespace FluentAssertions
         [MustUseReturnValue /* do not use Pure because this method executes the action before returning to the caller */]
         public static ExecutionTime ExecutionTime(this Func<Task> action)
         {
-            return new ExecutionTime(action);
+            return new ExecutionTime(action.ExecuteInDefaultSynchronizationContext);
         }
 
         /// <summary>
@@ -639,7 +640,7 @@ namespace FluentAssertions
         [Pure]
         public static AsyncFunctionAssertions Should(this Func<Task> action)
         {
-            return new AsyncFunctionAssertions(action, extractor);
+            return new AsyncFunctionAssertions(action.ExecuteInDefaultSynchronizationContext, extractor);
         }
 
         /// <summary>
@@ -649,7 +650,7 @@ namespace FluentAssertions
         [Pure]
         public static AsyncFunctionAssertions Should<T>(this Func<Task<T>> action)
         {
-            return new AsyncFunctionAssertions(action, extractor);
+            return new AsyncFunctionAssertions(action.ExecuteInDefaultSynchronizationContext, extractor);
         }
 
         /// <summary>
--- a/Src/FluentAssertions/Common/NoSynchronizationContextScope.cs
+++ b/Src/FluentAssertions/Common/NoSynchronizationContextScope.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Threading;
+
+namespace FluentAssertions.Common
+{
+    internal static class NoSynchronizationContextScope
+    {
+        public static DisposingAction Enter()
+        {
+            var context = SynchronizationContext.Current;
+            SynchronizationContext.SetSynchronizationContext(null);
+            return new DisposingAction(() => SynchronizationContext.SetSynchronizationContext(context));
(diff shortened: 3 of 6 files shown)


dennisdoomen:
It hang again.

dennisdoomen:
Maybe we try using what I've done in Chill: https://github.com/ChillBDD/Chill/pull/78/files

davidomid:
> Maybe we try using what I've done in Chill: https://github.com/ChillBDD/Chill/pull/78/files

Thanks very much for this link. Interesting solution! I'll see what I can do :)

davidomid:
@dennisdoomen the build is now passing. Please review this pull request and let me know yours thoughts. 

Thanks very much again for that ChillBDD suggestion.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -74,7 +74,7 @@ public void NotThrow(string because = "", params object[] becauseArgs)
         {
             try
             {
-                Task.Run(Subject).Wait();
+                Subject().Wait();
```
dennisdoomen: Shouldn't you use the `ExecuteInDefaultSynchronizationContext` method in all the below changes?

davidomid: I don't believe we need that, based on the results I'm seeing. 
According to AppVeyor, the only hanging tests on each of the my failing builds were:

- FluentAssertions.Specs.AsyncFunctionExceptionAssertionSpecs.When_async_method_throws_exception_expected_not_to_be_thrown_it_should_fail

- FluentAssertions.Specs.FluentActionsSpecs.Awaiting_works_with_func

What both tests have in common is that they both use Awaiting extension methods. 

Awaiting_works_with_func uses:

FluentActions.cs 
`public static Func<Task<T>> Awaiting<T>(Func<Task<T>> func)`

When_async_method_throws_exception_expected_not_to_be_thrown_it_should_fail uses:

AssertionExtensions.cs
`public static Func<Task> Awaiting<T>(this T subject, Func<T, Task> action)`

I believe these methods, combined with no longer using Task.Run in the NotThrow method, were causing the tests to hang. 
I've made a change to Awaiting in FluentActions.cs and the tests are passing fine. Interestingly, I haven't modified the Awaiting method in AssertionExtensions.cs, but the the related test is fine too. At this stage I'm not entirely sure why. It might be that this is a transient issue which only shows up on certain test runs. 

Maybe we can somehow re-run the tests a few times on AppVeyor and see if it manifests itself again?

dennisdoomen: I'm inclined to remove the synchronization context for all invocations to an `Action` and `Func<Task>`. See also this issue reported on Chill. https://github.com/ChillBDD/Chill/issues/77

davidomid: I'm inclined to agree with you on this. I want to dig around a bit first though and try to determine categorically why specific tests are hanging and which ones aren't. 
I've created a separate branch and PR just for diagnosing these hanging tests in AppVeyor. I'll reply here and update this PR when I find out more.

dennisdoomen: Kudos for helping us solve this long-lasting problem.

davidomid: @dennisdoomen No worries, I'm just happy to try contributing to this fantastic project :) 

I made the changes you suggested, by adding ExecuteInDefaultSynchronizationContext in various places. I've built these changes several times on AppVeyor and the tests seem stable. 

I'll leave you to resolve this if you're satisfied, otherwise let me know if you have other suggestions.

jnyrup:
If I understand the purpose of `NoSynchronizationContextScope`, it is meant to be called at the boundaries and then avoiding having to call `ConfigureAwait(false)` everywhere deeper in the stack.

If that is correct, shouldn't it be enough to *only* call `NoSynchronizationContextScope` in:

* `public static ExecutionTime ExecutionTime(this Func<Task> action)`
* `public static AsyncFunctionAssertions Should(this Func<Task> action)`
* `public static AsyncFunctionAssertions Should<T>(this Func<Task<T>> action)`

as they are, as far as I could see, exactly the boundaries of calling `Task` and `Task<T>`?

davidomid:
> If I understand the purpose of `NoSynchronizationContextScope`, it is meant to be called at the boundaries and then avoiding having to call `ConfigureAwait(false)` everywhere deeper in the stack.
> 
> If that is correct, shouldn't it be enough to _only_ call `NoSynchronizationContextScope` in:
> 
> * `public static ExecutionTime ExecutionTime(this Func<Task> action)`
> * `public static AsyncFunctionAssertions Should(this Func<Task> action)`
> * `public static AsyncFunctionAssertions Should<T>(this Func<Task<T>> action)`
> 
> as they are, as far as I could see, exactly the boundaries of calling `Task` and `Task<T>`?

Thanks for the feedback, you've raised an excellent point. I'll get right on it :)

davidomid:
> If I understand the purpose of `NoSynchronizationContextScope`, it is meant to be called at the boundaries and then avoiding having to call `ConfigureAwait(false)` everywhere deeper in the stack.
> 
> If that is correct, shouldn't it be enough to _only_ call `NoSynchronizationContextScope` in:
> 
> * `public static ExecutionTime ExecutionTime(this Func<Task> action)`
> * `public static AsyncFunctionAssertions Should(this Func<Task> action)`
> * `public static AsyncFunctionAssertions Should<T>(this Func<Task<T>> action)`
> 
> as they are, as far as I could see, exactly the boundaries of calling `Task` and `Task<T>`?

@jnyrup I've now pushed these changes. Please let me know if this is what you had in mind, or if you have other suggestions. 

Thanks very much

jnyrup:
Looks good!
(minor note: There are now a few superfluous namespaces and whitespace additions)

What happens if I execute one the following sync-over-async snippets, that calls into `ActionAssertions` and `FunctionAssertions`, respectively?

```c#
Action act = () => Task.Delay(0).Wait(0);
act.Should().NotThrow();
```

```c#
Func<bool> func = () => Task.Delay(0).Wait(0);
func.Should().NotThrow();
```

davidomid:
> Looks good!
> (minor note: There are now a few superfluous namespaces and whitespace additions)

My bad, I'll resolve this. 

> What happens if I execute one the following sync-over-async snippets, that calls into `ActionAssertions` and `FunctionAssertions`, respectively?
> 
> ```cs
> Action act = () => Task.Delay(0).Wait(0);
> act.Should().NotThrow();
> ```
> 
> ```cs
> Func<bool> func = () => Task.Delay(0).Wait(0);
> func.Should().NotThrow();
> ```

You've raised an excellent point. I could be mistaken but I don't think we have any test cases like this already. I'll create some and see if we have any potential issues. 

Thanks very much!

davidomid:
@jnyrup I've made the changes to remove the superfluous namespaces and whitespace. I've also added some new tests with the scenarios you described above. Those tests seem to pass fine but it looks like the current build is failing due to one of those other flaky tests which are being investigated separately. Do we need to retry the AppVeyor build somehow? 

Please let me know if you have any other issues or suggestions. 

Thanks very much :)

davidomid:
It just looked odd to me. When I looked into it, it looked a lot like it was only there to prevent deadlocks, so I thought it would be good to investigate and try to fix the underlying issue instead. 

In terms of retrying the build, I could always push a blank commit to force it to rebuild but I don't want to pollute the commit history and would rather ask if there's an alternative first :)

jnyrup:
@davidomid or, if you haven't tried before and would like to train your git-fu, you could trigger a build by rebasing on master into atomic commits, e.g. 
* Removed redundant multithreading
* Added additional test cases for ActionAssertions and FunctionAssertions

Not the case for this PR, but for large PR atomic commits greatly simplifies reviewing.

davidomid:
> @davidomid or, if you haven't tried before and would like to train your git-fu, you could trigger a build by rebasing on master into atomic commits, e.g.
> 
> * Removed redundant multithreading
> * Added additional test cases for ActionAssertions and FunctionAssertions
> 
> Not the case for this PR, but for large PR atomic commits greatly simplifies reviewing.

I'll do it for this one :) 

Thanks for the advice!

davidomid:
@jnyrup @dennisdoomen 

I've made the commits atomic and the latest build was successful :) 

Please let me know if there's anything else you'd like me to resolve.

dennisdoomen:
Well done @davidomid. Thanks for the hard work.

PR #1028, opened 2019-04-20
Attempt to fix a timing issue for the NotThrowAfter specs

dennisdoomen:
Attempts to fix a timing issue for the NotThrowAfter specs. Not entirely sure why this misbehaves on AppVeyor, but this seems to fix this.

### DIFF
--- a/Src/FluentAssertions/Specialized/ActionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/ActionAssertions.cs
@@ -166,8 +166,8 @@ namespace FluentAssertions.Specialized
                     return;
                 }
 
-                Thread.Sleep(pollInterval);
                 invocationEndTime = watch.Elapsed;
+                Thread.Sleep(pollInterval);
             }
 
             Execute.Assertion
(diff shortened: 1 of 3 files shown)


Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -166,8 +166,8 @@ public void NotThrowAfter(TimeSpan waitTime, TimeSpan pollInterval, string becau
                     return;
                 }
 
-                Thread.Sleep(pollInterval);
                 invocationEndTime = watch.Elapsed;
+                Thread.Sleep(pollInterval);
```
jnyrup: 🤔I don't think these `Thread.Sleep`s should have been moved.
When checking the loop condition the second time, `invocationEndTime` would be e.g. `x` while `Subject` was first invoked `x + pollInterval` ago. (If that makes any sense)

dennisdoomen: I don't h ave any clue what I was doing. But I can't think of any theoretical reason why it would fail on AppVeyor 😢

davidomid: I think it's failing on AppVeyor and not locally because presumably the build agent is less powerful than your machine. 
I think this a similar issue to other methods related to timing failing occasionally. We're relying on something being executed either within a precise amount of time, or to be executed after waiting a precise amount of time. I've found that these tests can be unreliable, especially when running other things on the same machine, or running it on a slower machine. 

I'm also looking into possible fixes for this. I'll let you know if I have anything.

Issue #1037, opened 2019-05-03
Throw<TException>, NotThrow, NotThrow<TException> and NotThrowAfter are throwing redundant AggregateExceptions


davidomid:
Resolved in PR #1038 

### Description

This is a minor issue but tasks being called synchronously with .Wait() will throw AggregateExceptions. This means that certain FluentAssertions methods are throwing AggregateExceptions, when they should only be throwing the inner exception. 

### Example

Take the following test case:

E.g.

```csharp
public void When_async_method_throws_the_expected_inner_exception_it_should_succeed()
        {
            //-----------------------------------------------------------------------------------------------------------
            // Arrange
            //-----------------------------------------------------------------------------------------------------------
            Func<Task> task = async () =>
            {
                await Task.Delay(100);
                throw new InvalidOperationException();
            };

            //-----------------------------------------------------------------------------------------------------------
            // Act
            //-----------------------------------------------------------------------------------------------------------
            Action action = () => task
                .Should().Throw<AggregateException>()
                .WithInnerException<InvalidOperationException>();

            //-----------------------------------------------------------------------------------------------------------
            // Assert
            //-----------------------------------------------------------------------------------------------------------
            action.Should().NotThrow();
        }
```

### Expected behavior:

I would expect this test to fail, given that the task doesn't throw an AggregateException and instead explicitly just throws an InvalidOperationException. 

### Actual behavior:

The test passes, because currently certain methods are executing the task synchronously using .Wait(), so if it throws an exception it will wrap it in an AggregateException, 

The methods which do this are:

- `public ExceptionAssertions<TException> Throw<TException>(string because = "", params object[] becauseArgs) where TException : Exception`

-  `public void NotThrowAfter(TimeSpan waitTime, TimeSpan pollInterval, string because = "", params object[] becauseArgs)`

- `public void NotThrow(string because = "", params object[] becauseArgs)`

- `public void NotThrow<TException>(string because = "", params object[] becauseArgs) where TException : Exception`

### Proposal

What I propose is replacing all calls to .Wait(), with .GetAwaiter().GetResult(). 
This should make sure only the top exception is thrown. If the given task does in fact throw an AggregateException, it will still work in the same way. The only difference is that Fluent Assertions itself does not throw AggregateExceptions internally, unnecessarily. 

This would require changing 2 tests which, like the one shown above, expect an AggregateException to be thrown with an inner exception. The tests would be modified so the given tasks explicitly throws an AggregateException.

[...]

PR #1038, opened 2019-05-03
Redundant AggregateException removal

davidomid:
Resolves issue #1037


jnyrup:
I agree that it is better to `.GetAwaiter().GetResult()` as that should remove the need to unwrap `AggregateExceptions`, as we then know they are not added.

I'm a bit suspicious/worried that we need to alter a test, that means the behavior has changed.
We seem to be inconsistent in how we have been unwrapping `AggregateExceptions` :(.

`NotThrow` and `NotThrow<TException>` call `GetFirstNonAggregateException`, which unwraps all `AggregateExceptions`.

`NotThrowAfter` and `Throw<TException>` call `InvokeSubjectWithInterception`, which only unwraps an `AggregateException` if its `InnerException` is also an `AggregateException`.

dennisdoomen:
I agree with your concern. In general, I think we should unwrap an `AggregateException` if it is FluentAssertions that causes it to be inserted. Only if invoking a `Task` directly would also raise an `AggregateException`, we should expose it. `Throw<TException>()` is special here. If you call `Should().Throw<InvalidOperationException>()` and the calling code raises this specific exception wrapped in an `AggregateException`, I believe it should succeed. This is the way it is currently implemented and documented. The same (should) apply to `ThrowExactly`, `ThrowAfter`, etc.

davidomid:
@jnyrup  @dennisdoomen  

I've added another passing unit test to make sure we can check for an exception type when an AggregateException contains an instance of it: 

```csharp
        [Fact]
        public void When_async_method_throws_aggregate_exception_containing_expected_exception_it_should_succeed()
        {
            // Arrange
            Func<Task> task = async () =>
            {
                await Task.Delay(100);
                throw new AggregateException(new InvalidOperationException());
            };

            // Act
            Action action = () => task
                .Should().Throw<InvalidOperationException>();

            // Assert
            action.Should().NotThrow();
        }
```

The latest build fails due to one of those flaky tests again. 

Aside from adding this test case, are there any other changes I should make at this stage? 

Thanks very much

Issue #696, opened 2017-11-29
Automatic extraction of aggregate exceptions makes it harder for me to test


chrischu:
I want to test a method that is synchronous but calls an asynchronous method and I want to ensure that I use `AsyncMethod().WaitAndUnwrapExceptions()` instead of `AsyncMethod.Result`, however this is currently not easily testable since FluentAssertions automatically unwraps the exception.

This is in my opinion a pretty bad idea as it might lead to a false feeling of security when in reality it might cover up a bug, because in reality no one unwraps the exception if I use `.Result` and then I have to catch AggregateException instead of the contained exceptions.

Therefore I think it would be best to completely remove the automatic exception unwrapping from assertions. However, if that is not desired for backwards compatibility I would at least like an option to disable the unwrapping in my solution (similiarly to how I can configure how ShouldBeEquivalentTo works).

dennisdoomen:
You're specifically talking about the `AggregateException` that FA is trying to intercept, right? If so, what would the API look like?

chrischu:
Something like `AssertionOptions.UnwrapAggregateExceptionsAutomatically()`?

dennisdoomen:
That would be a global setting. Did you try `Should().ThrowExact()`?

chrischu:
You probably mean `ShouldThrowExactly<TException>()`? 

Unfortunately that also unwraps (it just calls the normal `Throws` under the hood as seen here: https://github.com/fluentassertions/fluentassertions/blob/release-5.0/Src/FluentAssertions/AssertionExtensions.Actions.cs#L36)

davidomid:
@dennisdoomen I've taken a look at this and written some tests which confirm that current ShouldThrowExactly will work when an AggregateException is thrown. This is not the behaviour I would expect either, and would instead expect it to only work for the inner exception type. 

Is this how it's meant to work or is this an issue? I'm happy to write a fix for this if needed.

dennisdoomen:
The general guidelines are:

> I think we should unwrap an AggregateException if it is FluentAssertions that causes it to be inserted. Only if invoking a Task directly would also raise an AggregateException, we should expose it. Throw<TException>() is special here. If you call Should().Throw<InvalidOperationException>() and the calling code raises this specific exception wrapped in an AggregateException, I believe it should succeed. This is the way it is currently implemented and documented. The same (should) apply to ThrowExactly, ThrowAfter, etc.

[...]

PR #1041, opened 2019-05-05
Unwrapping of AggregateExceptions

jnyrup:
After #1038 got merged, the logic for unwrapping `AggregateException` becomes a lot easier, as we now know that FluentAssertions doesn't wrap exceptions in an `AggregateException`.

Calling `Throw<TException>` is designed to also succeed if an exception of type `TException` is wrapped inside an `AggregateException`.
The same behavior should be present for `NotThrow<TException>`

This PR solves these problems.
1. `NotThrow<TException>` didn't look inside `AggregateException`s to see if an exception of type `TException` was thrown, which made it non-symmetric to `Throw<TException>`.
2. `GetFirstNonAggregateException` only looked at the `InnerException` of an `AggregateException` instead of `InnerExceptions` (*all* wrapped exceptions). E.g. it failed to handle if `TException` was the second `InnerException`.
3. As we no longer unnecessarily wrap exceptions, `InterceptException` unwrapped to much.

### DIFF
--- a/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
@@ -151,25 +151,22 @@ namespace FluentAssertions.Specialized
 
         private static void NotThrow(Exception exception, string because, object[] becauseArgs)
         {
-            Exception nonAggregateException = GetFirstNonAggregateException(exception);
-
             Execute.Assertion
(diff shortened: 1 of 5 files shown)


jnyrup:
Found a related issue #696

dennisdoomen:
Tried to run this PR against the 20K tests of a project I'm involved with and everything seems to be fine.

Issue #1039, opened 2019-05-03
Missing null checks


jnyrup:
I'm playing around with [nullable reference types](https://devblogs.microsoft.com/dotnet/nullable-reference-types-in-csharp/) in [this branch](https://github.com/jnyrup/fluentassertions/tree/nullableReferenceTypes).
So far I've found several places that could benefit from an extra null check.

I'll try to keep the branch in sync with master as the null checks are merged.

```
ReferenceTypeAssertions
	BeAssignableTo
	BeAssignableTo<T>
		null Subject

ComparableTypeAssertions
	Be
		null Subject
	
NumericAssertions
	BeOfType
	NotBeOfType
		null Subject
		
ObjectAssertions
	HaveFlag
	NotHaveFlag
		null expectedFlag
		
ActionAssertions
	NotThrow
	InvokeSubjectWithInterception (public methods that calls it)
		null Subject
		
AsyncFunctionAssertions
	NotThrow
	NotThrowAsync
	NotThrow<TException>
	NotThrowAsync<TException>
	InvokeSubjectWithInterception (public methods that calls it)
	InvokeSubjectWithInterceptionAsync (public methods that calls it)
		null Subject
		
FunctionAssertions
	NotThrow
	NotThrow<TException>
	InvokeSubjectWithInterception (public methods that calls it)
	NotThrowAfter
		null Subject
		
AssemblyAssertions
	DefineType
		null Subject
	Reference
	NotReference
		null Subject
		null assembly
		
ExecutionTimeAssertions
	ctor
		null action
		null Func<T>
		
SimpleTimeSpanAssertions
	BePositive
	BeNegative
	Be
	NotBe
	BeLessThan
	BeLessOrEqualTo
	BeGreaterThan
	BeGreaterOrEqualTo
		null Subject
```

dennisdoomen:
You know C# 8 is still in preview, don't  you?

jnyrup:
I sure know that C# 8 is still in preview, that's why I keep the nullable annotation changes in a branch, which I have no intentions to create a PR on right now. 
That said, nullable reference types are already a great tool to let the compiler spot places with insufficient null checks.
My intention with this issue is to bring attention to those missing null checks, such that they can be added to master.

jnyrup:
I hacked up some [reflection code](https://github.com/jnyrup/fluentassertions/tree/nullReflection) in my search to eliminate `NullReferenceException`s thrown from FA.

[...]

PR #1042, opened 2019-05-05
Null checks for SimpleTimeSpanAssertions

jnyrup:
This PR adds missing null checks to `SimpleTimeSpanAssertions` that could cause `NullReferenceException`s when using a `null` `TimeSpan`.

Part of #1039

### DIFF
--- a/Tests/Shared.Specs/SimpleTimeSpanAssertionSpecs.cs
+++ b/Tests/Shared.Specs/SimpleTimeSpanAssertionSpecs.cs
@@ -41,6 +41,30 @@ namespace FluentAssertions.Specs
         }
 
         [Fact]
+        public void When_asserting_null_value_to_be_positive_it_should_fail()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Tests/Shared.Specs/SimpleTimeSpanAssertionSpecs.cs
```diff
@@ -40,6 +40,25 @@ public void When_asserting_negative_value_to_be_positive_it_should_fail()
...
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => nullTimeSpan.Should().BePositive();
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act.Should().Throw<XunitException>();
```
dennisdoomen: Usually we do verify that at least parts of the exception message make sense

jnyrup: Thanks
I've updated the assertions with separated `null` check statements and added assertions on the failure message in the tests.

PR #1046, opened 2019-05-18
Fix for ThrowExactly and ThrowExactlyAsync not working with AggregateExceptions

### LINKED ISSUE #696: Automatic extraction of aggregate exceptions makes it harder for me to test

### LINKED ISSUE #1045: ThrowExactly and ThrowExactlyAsync do not work properly with AggregateExceptions
This issue has been discussed as part of #696.

This is resolved by pull request #1046 

### Description

The ThrowExactly and ThrowExactlyAsync methods do not work in the expected way, when AggregateExceptions are thrown. 

The expectation is that asserting that a particular type of exception is thrown should not pass if that exception type is thrown within an AggregateException, instead of just throwing an exception of the expected type. 

Consider the following unit test: 

```csharp
[Fact]
public async Task When_subject_throws_aggregate_exception_instead_of_exact_exception_it_should_throw()
{
    // Arrange
    var asyncObject = new AsyncClass();

    // Act
    Func<Task> action = () => asyncObject.ThrowAggregateExceptionAsync<ArgumentException>();
    Func<Task> testAction = () => action.Should().ThrowExactlyAsync<ArgumentException>("ABCDE");

    // Assert
    (await testAction.Should().ThrowAsync<XunitException>())
        .WithMessage("*ArgumentException*ABCDE*AggregateException*");
}
```

### Expected behavior:

The test should pass, as we're asserting that an XunitException is thrown by ThrowExactlyAsync, indicating that the action throws an AggregateException instead of an ArgumentException. 

### Actual behavior:

The test fails, as ThrowExactlyAsync doesn't throw an exception as it treats the action as throwing an ArgumentException as it's an inner exception of the AggregateException.

davidomid:
This fixes #1045 and resolves #696. 

This pull request fixes ThrowExactly and ThrowExactlyAsync, when using AggregateExceptions. The documentation has also been updated to reflect the changes. 

These two methods previously would not throw an exception if an AggregateException was thrown containing an instance of the expected exception type. These methods will now throw an exception, unless you're asserting that an AggregateException was thrown.


davidomid:
@dennisdoomen @jnyrup 

I've updated the documentation and tested that the site still looks good locally.

In terms of the comment I made above about how I would like to reduce the amount of duplicated code, this actually doesn't seem entirely feasible at this point and appears to be a widespread issue in the solution, including the areas where I deleted old code. 
An example is that it seems difficult to share common functionality easily between FunctionAssertions and AsyncFunctionAssertions, (i.e. the ThrowExactly method) as they don't use a common type. Perhaps in a future item we can look at strategies for refactoring this logic so we have to worry less about duplicate implementations of common methods? 

In terms of this PR, do you have any other suggestions for other changes to be made at this stage, or are things at least good enough for this fix to be merged? 

Thanks very much :)

dennisdoomen:
> An example is that it seems difficult to share common functionality easily between FunctionAssertions and AsyncFunctionAssertions, (i.e. the ThrowExactly method) as they don't use a common type. P

Actually, this is exactly what I did in https://github.com/fluentassertions/fluentassertions/pull/1048.

davidomid:
> Actually, this is exactly what I did in #1048.

Thanks, I'll check out what you did in that PR and try and refactor things in a similar way.

dennisdoomen:
Or rebase on mine ;-)

davidomid:
> Or rebase on mine ;-)

I gave rebasing a go and had a look over all the changes but it turns out they weren't really in line with what I needed to reduce the duplicate code between AsyncFunctionAssertions and ActionAssertions. 

I took a different approach which you can see here #1054

davidomid:
Awaiting merge of #1054

davidomid:
@dennisdoomen that duplicate code has now been removed, due to merging in the latest refactorings from master. 

Please let me know if there's any other changes needed for this PR. 

Thanks very much

PR #1048, opened 2019-05-25
Assertions for Task and Task<T> with CompleteWithin checks

dennisdoomen:
Replaces #1013 by applying some more refactoring on #1013. 

Thanks to @lg2de for the original implementation.


Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -10,11 +10,11 @@ namespace FluentAssertions.Specialized
     /// Contains a number of methods to assert that an asynchronous method yields the expected result.
     /// </summary>
     [DebuggerNonUserCode]
-    public class AsyncFunctionAssertions
+    public abstract class AsyncFunctionAssertions
```
lg2de: Why making this `abstract` without having abstract members?
Please note https://rules.sonarsource.com/csharp/RSPEC-1694

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -10,11 +10,11 @@ namespace FluentAssertions.Specialized
...
     {
         private readonly IExtractExceptions extractor;
 
-        public AsyncFunctionAssertions(Func<Task> subject, IExtractExceptions extractor)
+        protected AsyncFunctionAssertions(Func<Task> subject, IExtractExceptions extractor)
```
lg2de: Making constructor protected is a breaking change, isn't it?

Src/FluentAssertions/Specialized/GenericAsyncFunctionAssertions.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public async Task<AndWhichConstraint<GenericAsyncFunctionAssertions<TResult>, Task<TResult>>> CompleteWithinAsync(
+            TimeSpan timeSpan, string because = "", params object[] becauseArgs)
+        {
+            var timeoutCancellationTokenSource = new CancellationTokenSource();
```
lg2de: CancellationTokenSource implements `IDisposable`.
You should dispose it (e.g. with `using`).

Src/FluentAssertions/Specialized/GenericAsyncFunctionAssertions.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            Task<TResult> task = subject();
+
+            Task completedTask =
+                await Task.WhenAny(task, timer.DelayAsync(timeSpan, timeoutCancellationTokenSource.Token));
+
+            if (completedTask == task)
+            {
+                timeoutCancellationTokenSource.Cancel();
+                await completedTask;
```
lg2de: Why awaiting the task which should be already completed?

Tests/Shared.Specs/TaskAssertionSpecs.cs
```diff
@@ -0,0 +1,97 @@
+﻿using System;
+using System.Threading.Tasks;
+using FluentAssertions.Extensions;
+using Xunit;
+using Xunit.Sdk;
+
+namespace FluentAssertions.Specs
+{
+    public class TaskAssertionSpecs
```
lg2de: You are testing here `Func<Task>` not `Task`.
So this tests should be placed in the file for the other Func tests.

Tests/Shared.Specs/TaskOfTAssertionSpecs.cs
```diff
@@ -0,0 +1,123 @@
+﻿using System;
+using System.Threading.Tasks;
+using FluentAssertions.Extensions;
+using Xunit;
+using Xunit.Sdk;
+
+namespace FluentAssertions.Specs
+{
+    public class TaskOfTAssertionSpecs
```
lg2de: So above!

lg2de:
Today I looked deeper in your PR. In addition to my review comments above, please note the following:

- You named the PR similar to #1013 but you implemented it similar to #1011.
- You closed #1013 as replaced. But my intention there was the extending `Task` directly. You extended `Func<Task>`.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
dennisdoomen: You're right. I've reverted my change.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
dennisdoomen: You're right as well.

Src/FluentAssertions/Specialized/GenericAsyncFunctionAssertions.cs
dennisdoomen: To make sure any exceptions are rethrown

Tests/Shared.Specs/TaskAssertionSpecs.cs
dennisdoomen: Which one do you mean?

dennisdoomen:
> I'm not in the position to "request" changes, but I recommend at least thinking about...

Of course you are. And you were even right.

> * You named the PR similar to #1013 but you implemented it similar to #1011.
> * You closed #1013 as replaced. But my intention there was the extending `Task` directly. You extended `Func<Task>`.

I took your PR and implemented it according to the review comment @jnyrup made [here](https://github.com/fluentassertions/fluentassertions/pull/1013#discussion_r281028341) and with which I agreed.

Src/FluentAssertions/AssertionExtensions.cs
```diff
@@ -662,19 +664,19 @@ public static ActionAssertions Should(this Action action)
         /// current <see cref="System.Func{Task}"/> .
         /// </summary>
         [Pure]
-        public static AsyncFunctionAssertions Should(this Func<Task> action)
+        public static NonGenericAsyncFunctionAssertions Should(this Func<Task> action)
```
jnyrup: Strictly speaking this is apparently a breaking change
https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md#signatures

but those rules are also ultra strict as they need to be reflection and serialization compatible.
Since we're returning a more derived type, we should be covered?

Src/FluentAssertions/AggregateExceptionExtractor.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions.Common;
+using FluentAssertions.Specialized;
+
+namespace FluentAssertions
+{
+    public class AggregateExceptionExtractor : IExtractExceptions
```
jnyrup: Should we make this `internal`?

Src/FluentAssertions/Common/TaskTimer.cs
```diff
@@ -0,0 +1,25 @@
...
+    internal class TaskTimer : ITimer
+    {
+        public Task DelayAsync(TimeSpan delay, CancellationToken cancellationToken)
+        {
+            return Task.Delay(delay, cancellationToken);
+        }
+
+        public bool Wait(Task task, TimeSpan timeout)
+        {
+            using (NoSynchronizationContextScope.Enter())
```
jnyrup: Is there any problem in Tasks being nested twice in a `NoSynchronizationContextScope`.
First in `Should` and then again here?

Tests/Shared.Specs/AssertionExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace FluentAssertions.Specs
+{
+    internal static class AssertionExtensions
+    {
+        private static readonly AggregateExceptionExtractor extractor = new AggregateExceptionExtractor();
+
+        public static NonGenericAsyncFunctionAssertions Should(this Func<Task> action, ITimer timer)
+        {
+            return new NonGenericAsyncFunctionAssertions(action, extractor, timer);
```
jnyrup: We're not wrapping `action` in `ExecuteInDefaultSynchronizationContext`.

Src/FluentAssertions/AssertionExtensions.cs
dennisdoomen: From a compile-time perspective, it'll be fine indeed. But technically, this could cause a binary incompatibility of some other library relied on this method. I'm not sure I'm willing to be that strict.

Src/FluentAssertions/AggregateExceptionExtractor.cs
dennisdoomen: We can only do that if we give the specs access to its internals using `[InternalsVisibleTo`

Src/FluentAssertions/Common/TaskTimer.cs
dennisdoomen: It's fine if you nest them. It'll just get the `null` context, set it to `null` and then restore the `null` ;-)

Tests/Shared.Specs/AssertionExtensions.cs
dennisdoomen: You mean here, in the constructor?

jnyrup: Yes, we currently use `ExecuteInDefaultSynchronizationContext` in three extensions methods in `AssertionExtensions`.
But I have been thinking about moving the wrapping closer to calls to `Wait()`.
E.g. to null check the action before invoking it.

dennisdoomen: This is the unit test project. That `ExecuteInDefaultSynchronizationContext` is not available there.

jnyrup: okay.

Src/FluentAssertions/AssertionExtensions.cs
dennisdoomen: Do you agree with this?

jnyrup: Wouldn't that other library be able to match it against the base type `AsyncFunctionAssertions`?
Maybe I'm just to tired to review any more code for today.

dennisdoomen:
I'm pretty sure it could be a binary compatibility issue. I'm not sure I care enough considering this is not an extension point.

Outlook voor Android downloaden<https://aka.ms/ghei36>

________________________________

Src/FluentAssertions/AssertionExtensions.cs
jnyrup: I was to apparently to tired to even answer your question.
I'm good with this change and the potential risk.

Src/FluentAssertions/Specialized/NonGenericAsyncFunctionAssertions.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public async Task<AndWhichConstraint<AsyncFunctionAssertions, Task>> CompleteWithinAsync(
+            TimeSpan timeSpan, string because = "", params object[] becauseArgs)
+        {
+            var timeoutCancellationTokenSource = new CancellationTokenSource();
```
jnyrup: I assume this `CancellationTokenSource` should also be disposed?

Src/FluentAssertions/Specialized/GenericAsyncFunctionAssertions.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// </summary>
+        /// <param name="timeSpan">The allowed time span for the operation.</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndWhichConstraint<GenericAsyncFunctionAssertions<TResult>, Task<TResult>> CompleteWithin(
```
jnyrup: Could we return `TResult` instead of `Task<TResult>`?
That would remove the need in tests to call `.Result`.

dennisdoomen:
Crap. I was too fast.

Tests/Shared.Specs/TaskAssertionSpecs.cs
lg2de: AsyncFunctionExceptionAssertionSpecs.cs

dennisdoomen: I suspected that one, but that one acts on exceptions, not tasks.

lg2de: Then I recommend `AsyncMethodAssertionSpecs.cs` here and `AsyncFunctionAssertionSpecs.cs` below.
Or - following your implementation classes `GenericAsyncFunctionAssertionsSpecs.cs` and `NonGenericAsyncFunctionAssertionsSpecs.cs`.

BTW: I myself prefer positive logic. This is why I do not that like "NonGeneric".

Src/FluentAssertions/Specialized/NonGenericAsyncFunctionAssertions.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// </summary>
+        /// <param name="timeSpan">The allowed time span for the operation.</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndWhichConstraint<AsyncFunctionAssertions, Task> CompleteWithin(
```
jnyrup: If the other `CompleteWithin` is changed to 
`AndWhichConstraint<GenericAsyncFunctionAssertions<TResult>, TResult>` 
then this should be changed to
AndConstraint<AsyncFunctionAssertions<TResult>>

dennisdoomen: I don't get it. They both return `AndWhichConstraint`.  One just wraps it in `Task`.

jnyrup:
I was thinking about using AndWhichConstraint<...,TResult> when the subject is e.g. Func<Task<TResult>> 
and AndConstraint<...> when the Subject is e.g. Func<Task>. 

I'm vacationing until Monday without a computer, so my responses might be delayed and inferior.

Issue #1053, opened 2019-05-27
AsyncFunctionAssertions has duplicate implementations of ActionAssertions methods


davidomid:
Resolved in PR #1054 

### Description

AsyncFunctionAssertions has non-async methods in common with ActionAssertions (i.e. Throw, NotThrow). However, these are implemented separately for AsyncFunctionAssertions and this has lead to:

- Redundant code from duplicate logic
- Inconsistent error messages between the similar methods. I.e. Throw for AsyncFunctionAssertions may output a slightly different error message to the same method contained within ActionAssertions. 
- Overhead of a developer when modifying the behaviour of one of these common methods, as the implementation has to be duplicated. 

My proposal is to make AsyncFunctionAssertions  a subclass of ActionAssertions. This ensures that any methods available for an `Action` are also available for a synchronous call to a `Func<Task>`.

jnyrup:
@davidomid FYI: if you include the text "This fixes #1053" in the PR, Github will close the issue once the PR is merged.

davidomid:
> @davidomid FYI: if you include the text "This fixes #1053" in the PR, Github will close the issue once the PR is merged.

Thanks for the tip :) 

I'm saying similar things in the PR but I guess I have to use the correct phrase.

PR #1054, opened 2019-05-27
Refactoring of AsyncFunctionAssertions and ActionAssertions

davidomid:
Resolves issue #1053 

This PR also contains the changes made in PR #1050

### DIFF
--- a/Src/FluentAssertions/Specialized/ActionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/ActionAssertions.cs
@@ -158,7 +158,7 @@ namespace FluentAssertions.Specialized
                 .ForCondition(expectedExceptions.Any())
                 .BecauseOf(because, becauseArgs)
                 .FailWith(
-                    "Expected a <{0}> to be thrown{reason}, but found <{1}>: {3}.",
+                    "Expected a <{0}> to be thrown{reason}, but found <{1}>: {2}{3}.",
                     typeof(TException), exception?.GetType(),
                     Environment.NewLine,
                     exception);
--- a/Src/FluentAssertions/Specialized/ActionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/ActionAssertions.cs
@@ -40,24 +40,8 @@ namespace FluentAssertions.Specialized
         {
             FailIfSubjectIsAsyncVoid();
 
-            Exception actualException = InvokeSubjectWithInterception();
-            IEnumerable<TException> expectedExceptions = extractor.OfType<TException>(actualException);
-
-            Execute.Assertion
-                .ForCondition(actualException != null)
-                .BecauseOf(because, becauseArgs)
-                .FailWith("Expected a <{0}> to be thrown{reason}, but no exception was thrown.", typeof(TException));
-
-            Execute.Assertion
-                .ForCondition(expectedExceptions.Any())
-                .BecauseOf(because, becauseArgs)
-                .FailWith(
-                    "Expected a <{0}> to be thrown{reason}, but found a <{1}>: {3}.",
-                    typeof(TException), actualException.GetType(),
-                    Environment.NewLine,
-                    actualException);
-
-            return new ExceptionAssertions<TException>(expectedExceptions);
+            Exception exception = InvokeSubjectWithInterception();
+            return Throw<TException>(exception, because, becauseArgs);
         }
 
         /// <summary>
@@ -74,18 +58,8 @@ namespace FluentAssertions.Specialized
             where TException : Exception
         {
             FailIfSubjectIsAsyncVoid();
-
-            Exception actualException = InvokeSubjectWithInterception();
-
-            if (actualException != null)
-            {
-                IEnumerable<TException> expectedExceptions = extractor.OfType<TException>(actualException);
-
-                Execute.Assertion
-                    .ForCondition(!expectedExceptions.Any())
-                    .BecauseOf(because, becauseArgs)
-                    .FailWith("Did not expect {0}{reason}, but found {1}.", typeof(TException), actualException);
-            }
+            Exception exception = InvokeSubjectWithInterception();
+            NotThrow<TException>(exception, because, becauseArgs);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


dennisdoomen:
Should we finish/merge #1050 first?

davidomid:
> Should we finish/merge #1050 first?

Yes please :) 

As far as I can tell, it should be fine for you to merge it when ready. 

Thanks very much

davidomid:
> Should we finish/merge #1050 first?

@dennisdoomen I resolved the merge conflicts. Looks good aside from a flaky time-related test now 👍

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -172,6 +144,45 @@ public void NotThrowAfter(TimeSpan waitTime, TimeSpan pollInterval, string becau
...
+            Execute.Assertion
+                .ForCondition(exception != null)
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected a <{0}> to be thrown{reason}, but no exception was thrown.", typeof(TException));
+
+            Execute.Assertion
+                .ForCondition(expectedExceptions.Any())
+                .BecauseOf(because, becauseArgs)
+                .FailWith(
+                    "Expected a <{0}> to be thrown{reason}, but found <{1}>: {3}.",
```
jnyrup: 🙃 The format string does not use `Environment.NewLine` and `{2}`.

dennisdoomen:
You need to rebase it on `master`, because I still see the changes from #1050.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -10,37 +9,20 @@ namespace FluentAssertions.Specialized
...
             Subject = subject;
         }
 
         /// <summary>
         /// Gets the <see cref="Func{Task}"/> that is being asserted.
         /// </summary>
-        public Func<Task> Subject { get; private set; }
-
-        /// <summary>
-        /// Asserts that the current <see cref="Func{Task}"/> throws an exception of type <typeparamref name="TException"/>.
```
jnyrup: ❓ This changes the xml summary from `Func{Task}` to `Action` in `ActionAssertions`.

Src/FluentAssertions/Specialized/ActionAssertions.cs
davidomid: Sorry this slipped through. I'm surprised the tests didn't pick this up. 
I'll look into this right now. Thanks!

davidomid: Sorry about this. I just looked into it and found that this was how things looked before, but in a different area of ActionAssertions.
I'll place it directly. before the {3}

jnyrup: No worries, they've snuck in during refactorings. 
I'm aware of at least one more format string in the repo that has similar "typo".

davidomid: It does seem that way. This pull request is a good opportunity to resolve all of those at least :)

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
davidomid: This is a good point. I guess we have a few solutions: 

1. We could add methods in AsyncFunctionAssertions which have the same signatures and just call the base methods, but with their own XML summaries. Downside to this would be having method signatures for the sake of the comments, which is fine but might be a bit ugly.
Another downside is that anyone adding functionality to ActionAssertions will need to remember to go to any subclasses and add a new method stub with the dedicated summary. 

2. We could remove all mentions of either `Action` or `Task<Func>` in the summaries and just refer to the `Subject` property. Might look like this:

```csharp
/// <summary>
/// Asserts that the current Subject throws an exception of type <typeparamref name="TException"/>.
```
Downside to this is that "Subject" doesn't mean much to the developer. 

3. We could change the summary to show multiple types, so it might look something look:
```csharp
/// <summary>
/// Asserts that the current <see cref="Action"/> or <see cref="Func{Task}"/> throws an exception of type <typeparamref name="TException"/>.
```
Although a downside to this is that ActionAssertions is therefore then aware of AsyncFunctionAssertions, which it probably shouldn't be. 

4. We could ignore it. Using a `Func<Task>` is obviously a different type to `Action`, but the developer might understand that methods like Throw, NotThrow, etc. are synchronous, thereby essentially making their asynchronous action a regular Action. 

@dennisdoomen  @jnyrup Which solution would you guys prefer? Any other solutions I haven't thought of? If need be then we can discuss different approaches to this refactoring altogether.

davidomid: @dennisdoomen @jnyrup  I can also replace the text so it refers to it as a "delegate". I mean both Action and Func<Task> are just delegates anyway. 

```csharp
/// <summary>
/// Asserts that the current delegate throws an exception of type <typeparamref name="TException"/>.
```

What are your thoughts on what to do about the comments?

jnyrup: I dislike 1 and 2 for the same reasons as you.

Referring simply to delegates would be fine with me.
That seems to be in line with `ReferenceTypes` which uses the similar general phrase `Asserts that the object`.

davidomid: Perfect, thanks very much for the feedback 👍

Src/FluentAssertions/Specialized/ActionAssertions.cs
davidomid: @jnyrup The other typo isn't in the same area that I changed for this feature, so I've resolved this conversation and I'd recommend we fix that typo in another PR

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
davidomid: @jnyrup I pushed the latest changes with "delegate". Resolved this conversation for now.

davidomid:
@jnyrup @dennisdoomen 

Please can this be merged? I frequently need to resolve conflicts due to the moving of methods which are edited in other merged pull requests. 

Failing tests are the flaky ones again. 

Thanks :)

Issue #1052, opened 2019-05-27
Add `async`-support for `WithMessage()`


mu88:
### Description
It would be great to use `WithMessage()` on `ThrowAsync<>()`.

### Complete minimal example reproducing the issue
```csharp
// Arrange
Func<Task> act = () => throw new InvalidOperationException("Hello World");

// Act & Assert
await act().Should().ThrowAsync<InvalidOperationException>().WithMessage("HelloWorld");
```

### Actual behavior:
Compile-time error. If have to await/block the Task to use `WithMessage()` (e. g. `await act().Should().ThrowAsync<InvalidOperationException>().Result.WithMessage("HelloWorld");`).

### Versions
5.6.0

jnyrup:
Currently you need to `await` the `ThrowAsync` before calling `WithMessage`.
```c#
(await act().Should().ThrowAsync<InvalidOperationException>()).WithMessage("HelloWorld");
```
note the extra parentheses.

To avoid those parentheses you could create an extension method like this
```c#
public static async Task<ExceptionAssertions<TException>> WithMessage<TException>(
    this Task<ExceptionAssertions<TException>> task,
    string expectedWildcardPattern,
    string because = "",
    params object[] becauseArgs)
    where TException : Exception
{
    return (await task).WithMessage(expectedWildcardPattern, because, becauseArgs);
}
```

and then call
```c#
await act().Should().ThrowAsync<InvalidOperationException>().WithMessage("HelloWorld");
```

mu88:
Works like a charm! Is it okay for you if I provide this extension method as a PR?

dennisdoomen:
Absolutely!

jnyrup:
Contributed in #1055

PR #1055, opened 2019-05-28
Provide extension method 'WithMessage()' for async situations (#1052)

mu88:
Resolves issue #1052


Src/FluentAssertions/AssertionExtensions.Async.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Threading.Tasks;
+using FluentAssertions.Specialized;
+
+namespace FluentAssertions
+{
+    public static partial class AssertionExtensions
```
dennisdoomen: I would keep it with the rest of the extension methods. No need for partial classes here.

Src/FluentAssertions/AssertionExtensions.Async.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public static async Task<ExceptionAssertions<TException>> WithMessage<TException>(
```
dennisdoomen: Can you please copy the original documentation?

Tests/Shared.Specs/AsyncFunctionExceptionAssertionSpecs.cs
```diff
@@ -48,6 +48,22 @@ public async Task When_async_method_throws_an_empty_AggregateException_it_should
             await act2.Should().ThrowAsync<XunitException>();
         }
 
+        [Fact]
+        public async Task When_async_method_throws_an_Exception_it_should_provide_the_message()
+        {
+            //-----------------------------------------------------------------------------------------------------------
```
jnyrup: As `act` and `act2` are mutually independent, please split this into two separate tests.

Issue #1047, opened 2019-05-23
System.BadImageFormatException: Bad IL Format when monitoring events on .NET Core 3.0-preview5


Wildenhaus:
### Description

FluentAssertions throws a `BadImageFormatException` when invoking a monitored event. This only occurs in the .NET Core 3.0 preview.

### Complete minimal example reproducing the issue

```csharp
[TestClass]
public class Repro
{
   public delegate void EventDelegate();

   public class Tester
   {
      public event EventDelegate TestEventDelegate;
      public event EventHandler TestEventHandler;

      public void RaiseEventDelegate()
      {
         TestEventDelegate?.Invoke();
      }

      public void RaiseEventHandler()
      {
         TestEventHandler?.Invoke( this, EventArgs.Empty );
      }
   }

   [TestMethod]
   public void TestRaiseEventDelegate()
   {
      var tester  = new Tester();
      var monitor = tester.Monitor();

      tester.RaiseEventDelegate();

      monitor.Should().Raise( nameof( Tester.TestEventDelegate ) );
   }

   [TestMethod]
   public void RaiseEventEventHandler()
   {
      var tester  = new Tester();
      var monitor = tester.Monitor();

      tester.RaiseEventHandler();

      monitor.Should().Raise( nameof( Tester.TestEventHandler ) );
   }
}
```

### Expected behavior:

`EventHandlerFactory` should emit a handler that does not cause a `BadImageFormatException` to be thrown.

### Actual behavior:

A `BadImageFormatException` is thrown when the event is invoked.

### Versions

* Which version of Fluent Assertions are you using?
FluentAssertions v5.6.0 via NuGet

* Which .NET runtime and version are you targeting?
.NET Core 3.0-preview5

### Additional Information

Have not looked into it. The exception is thrown immediately upon invoking the event. Bug does not occur in earlier versions of .NET Core (tested with 2.2).

jnyrup:
I can confirm the crash.
* `netcoreapp2.2` passes the tests
* `netcoreapp3.0` fails the tests

```
Failed   TestRaiseEventDelegate
Error Message:
 Test method FA_1047.Repro.TestRaiseEventDelegate threw exception:
System.BadImageFormatException: Bad IL format.
Stack Trace:
    at EventDelegateDynamicHandler(IEventRecorder )
[...]

PR #1057, opened 2019-05-29
Fix BadImageFormatException on .NET Core 3.0

lg2de:
This is a PR to provide testing environment for .NET Core for now. We would like to verify e.g. #1047.
Later this PR can be used to add the target for 3.0 in next release.

Regarding double format changes please note https://devblogs.microsoft.com/dotnet/floating-point-parsing-and-formatting-improvements-in-net-core-3-0/

I would recommend to adopt the expectation because this seems to be default behavior in .NET Core 3.


dennisdoomen:
I think this doesn't work because we still use `msbuild` (through a Cake API).

jnyrup:
Also from `build.cake`
```
settings.ToolVersion = MSBuildToolVersion.VS2017;
```

jnyrup:
And from https://github.com/cake-build/cake/releases there has been some updates from the 0.30.0 that we use to 0.33.0 about NugetRestore and msbuild 16.

lg2de:
With the recent comment the `build.ps1` was running fine.
I guess the build server need to have installed the .NET Core 3 SDK.

Lets retry when  appveyor has finished the VS2019 (16.1) image.

jnyrup:
Could it be that the build server do have 3.0 installed, but since it's preview it's not picked by default?

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -19,7 +19,7 @@ public AssemblyAssertions(Assembly assembly)
             Subject = assembly;
         }
 
-#if NET45 || NET47 || NETSTANDARD2_0 || NETCOREAPP2_0 // TODO :: Should be able to remove based on: https://github.com/dotnet/corefx/issues/1784#issuecomment-218803619
```
lg2de: I tried to remove this, as noted in the TODO.
But I got a lot of errors.
So I recommend to remove now this TODO, because it is not correct (anymore).

Src/FluentAssertions/AssertionExtensions.cs
```diff
@@ -690,7 +690,7 @@ public static FunctionAssertions<T> Should<T>(this Func<T> func)
         }
 
 
-#if NET45 || NET47 || NETCOREAPP2_0
+#if NET45 || NET47 || NETCOREAPP2_0 || NETCOREAPP3_0
```
lg2de: Should we exclude what is not supported instead include?

jnyrup:
Hmm... I can't really see from the build logs if:
* .net core 3.0 is not installed,
* Restore-NuGet-Packages can't find .net core 3.0, or
* Restore-NuGet-Packages actively chooses .net core 2.2 over .net core 3.0

Raising the logging level from `DotNetCoreVerbosity.Normal` to `DotNetCoreVerbosity.Diagnostic`
could maybe shed some light on those questions.

Maybe related links:
https://help.appveyor.com/discussions/problems/22448-dotnet-core-30-and-netstandard-21
https://github.com/cake-build/cake/issues/2519

Src/FluentAssertions/AssertionExtensions.cs
jnyrup: I agree with this.
Blacklisting instead of whitelisting makes adding new target frameworks easier.

dennisdoomen:
I pushed a commit that switched our Nuget restore and build steps to `dotnet`. The `Visual Studio 2019 Preview` agent has SDK `3.0.100-preview-009812` installed. It now complains about C# issues.

jnyrup:
I don't see any complaints about C# issues?
It fails before finishing restoring nuget packages.

Do we want to add a netcoreapp3.0 target of the Library?
It should suffice to add a netcoreapp3.0 test project.

dennisdoomen:
> I don't see any complaints about C# issues?

I got the entire script running on my local machine, and it even reports the IL errors reported in #1047.

> Do we want to add a netcoreapp3.0 target of the Library?

Not until we need to support assertions on specific types introduced by .NET Core 3.0

jnyrup:
Aha, the #1047 c# issues causing runtime problems.
I thought you were calling out some sort of c# compile time issues.

dennisdoomen:
> I thought you were calling out some sort of c# compile time issues.

As well. But that's probably because that agent is using an old preview of .NET Core 3.0

`C:\projects\fluentassertions\Tests\Shared.Specs\MethodBaseAssertionSpecs.cs(829,77): error CS8107: Feature 'private protected' is not available in C# 7.0. Please use language version 7.2 or greater. `

I'm trying to figure out a way to install the right dotnet SDK as part of our build process. We do this as well in our projects.

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
jnyrup: @dennisdoomen Do you have an opinion to this?

jnyrup:
@lg2de As #1060 has been merged it should now be possible to build .net core 3.0 targets.
As we're not utilizing any .net core 3.0 features (yet!), please remove that from the core library target frameworks.
When I tried locally I could reproduce #1047 with a netcoreapp2.0 core library and a netcoreapp3.0 test project.

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
dennisdoomen: If we (still) can't remove, don't

lg2de:
I have reworked as requested and added the fix mentioned by @jnyrup.
Please review my changes and comment the TODOs added.
The build failed now "only" due to timing issues.

Tests/Benchmarks/Benchmarks.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net471</TargetFramework>
+    <TargetFramework>net47</TargetFramework>
```
dennisdoomen: Why is this needed?

Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -14,7 +14,7 @@
     <Authors>Dennis Doomen;Jonas Nyrup</Authors>
     <PackageDescription>
       A very extensive set of extension methods that allow you to more naturally specify the expected outcome of a TDD or
-      BDD-style unit tests. Targets .NET Framework 4.5 and 4.7, as well as .NET Core 2.0, .NET Standard 1.3, 1.6 and 2.0.
+      BDD-style unit tests. Targets .NET Framework 4.5 and 4.7, as well as .NET Core 2.0, .NET Core 3.0, .NET Standard 1.3, 1.6 and 2.0.
```
jnyrup: We're no longer targeting net core 3.0

Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -56,6 +56,9 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
     <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.4.0" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.0'">
```
jnyrup: We're no longer targeting net core 3.0

build.cake
```diff
@@ -50,7 +50,6 @@ Task("Restore-NuGet-Packages")
 	DotNetCoreRestore(new DotNetCoreRestoreSettings
 	{
 		NoCache = true,
-        ConfigFile = "./nuget.config",
```
jnyrup: Please keep this

Tests/Shared.Specs/Utilities.cs
```diff
@@ -2,7 +2,10 @@
 
 namespace FluentAssertions.Specs
 {
-    public static class Utilities
+    /// <summary>
+    /// Implements extensions to <see cref="TimeSpan"/> available in .NET Core 2, but not in .NET Framework.
+    /// </summary>
+    public static class Utilities // TODO Rename to TimeSpanExtensions?
```
jnyrup: Good idea, that's a way better name than my `Utilities`

Tests/Shared.Specs/EventAssertionSpecs.cs
```diff
@@ -744,7 +744,7 @@ public void When_monitoring_an_object_through_an_interface_it_should_monitor_onl
             }
         }
 
-#if NET45 || NET47
+#if NET45 || NET47 // TODO Why this is not available in e.g. core2?
```
jnyrup: If you add `NETCOREAPP2_0` it does not compile, as `DefineDynamicAssembly` and `DefineDynamicModule` are not defined for that target.

Src/FluentAssertions/FluentAssertions.csproj
dennisdoomen: But we do support it officially. So I would keep it in.

jnyrup: Isn't this superfluous if we aren't targeting net core 3.0?

dennisdoomen: Correct.

Src/FluentAssertions/FluentAssertions.csproj
dennisdoomen: But we do support it officially. So I would keep it in.

Src/FluentAssertions/FluentAssertions.csproj
dennisdoomen: My response was meant for your comment on the nuspec file.

Src/FluentAssertions/FluentAssertions.csproj
dennisdoomen: Maybe the PR can also update the `docs` folder.

Tests/Benchmarks/Benchmarks.csproj
lg2de: This is not needed, but I would recommend this for consistency of the solution.
The target net47 is also build for 4.7.0.
Currently you need to install SDK for 4.7.0 **and** 4.7.1 to build the solution.

Src/FluentAssertions/FluentAssertions.csproj
lg2de: Done.
Please check.

build.cake
lg2de: This was a mistake due to merge from master.
Reverted.

PR #1060, opened 2019-05-30
Auto-install .NET Core version if it is missing

dennisdoomen:
To support #1057 on AppVeyor

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -50,16 +50,7 @@ Task("Restore-NuGet-Packages")
 	DotNetCoreRestore(new DotNetCoreRestoreSettings
 	{
 		NoCache = true,
-		ConfigFile = "./nuget.config",
 		Verbosity = DotNetCoreVerbosity.Normal
-	});
-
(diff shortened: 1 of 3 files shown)


build.cake
```diff
@@ -50,40 +50,20 @@ Task("Restore-NuGet-Packages")
...
 });
 
 Task("Build")
     .IsDependentOn("Restore-NuGet-Packages")
 	.IsDependentOn("GitVersion")
     .Does(() =>
 {
-    if(IsRunningOnWindows())
-    {
       // Use MSBuild
```
jnyrup: 🙃 Remove comment about MSBuild

build.cake
```diff
@@ -50,40 +50,20 @@ Task("Restore-NuGet-Packages")
 	DotNetCoreRestore(new DotNetCoreRestoreSettings
 	{
 		NoCache = true,
-		ConfigFile = "./nuget.config",
```
jnyrup: 🤔 I would prefer if we kept this, #949

PR #1061, opened 2019-05-31
Avoid casting throught object

jnyrup:
Use pattern matching to cast directly from `TSubject` to `T` instead of through `object`.
As `TSubject` is not constrained to be a class type, we cannot use ` as ` for the conversion.

### DIFF
--- a/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
+++ b/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
@@ -123,11 +123,9 @@ namespace FluentAssertions.Primitives
         {
             BeOfType(typeof(T), because, becauseArgs);
 
-            T typedSubject = default(T);
-            if (Subject is T)
-            {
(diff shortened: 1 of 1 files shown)


dennisdoomen:
> Is this an issue we should look into resolving by setting the language version, either by project or solution?

IMO, we should fix the version to a specific one.

davidomid:
@dennisdoomen  @jnyrup 

Resolved in this PR #1065

jnyrup:
@davidomid, sorry my bad.
When introducing minor version of C# Microsoft changed the way to resolve the language version when no `LangVersion` was set. 
https://devblogs.microsoft.com/dotnet/an-update-to-c-versions-and-c-tooling/

If I remember correctly VS2017 resolves `latest` to "latest major", while VS2019 resolves it to `to latest minor`.

eNeRGy164:
Having `latest` as LangVersion in VS2017 also resolves to latest Minor.
The _absence_ of LangVersion is VS2017 resolves to latest Major

Issue #1063, opened 2019-05-31
Solution requires C# 7.1 to build


davidomid:
Resolved in PR #1065 

The changes merged in PR #1061 make use of generic pattern matching, which is a language feature introduced in C# 7.1. 

When I attempt to compile locally, I get an error as the projects are using C# 7.0. I can't find any explicit setting of language version anywhere in the solution. It's possible it's using the latest major version by default, and I don't have C# 8 installed.

PR #1065, opened 2019-05-31
Set language version of FluentAssertions.csproj to C# 7.1

davidomid:
Resolves issue #1063


davidomid:
@dennisdoomen @jnyrup 

The flaky timing tests are failing, aside from that it seems fine. I'd recommend merging soon if possible, as getting latest from master causes compilation issues (for some of us anyway).

PR #1071, opened 2019-06-01
Implement NotBeEquivalent for objects (fixes #170)

### LINKED ISSUE #170: AreNotEqual
Is there anything that asserts two object not equivalent?

lg2de:
Is this the right way?


jnyrup:
The implementation looks good to me, but I think we may need to be careful here.

`BeEquivalentTo` is heavily overloaded with 21 overloads whereas `NotBeEquivalentTo` has 7 overloads.
So adding overloads of `NotBeEquivalentTo` dripwise could change the resolved overload between releases. 
Not saying that we shouldn't add these overloads, but I think it would be great if we could add them all in one release.

```c#
"AA".Should().BeEquivalentTo("BB"); // StringAssertions
"AA".Should().NotBeEquivalentTo("BB"); // does not exists

new object().Should().BeEquivalentTo(new object()); // Object Assertions
new object().Should().NotBeEquivalentTo(new object());  // does not exists, but would exist with this PR

new[] { "AA" }.Should().BeEquivalentTo(new[] { "BB" }); // StringCollectionAssertions
new[] { "AA" }.Should().NotBeEquivalentTo(new[] { "BB" });  // CollectionAssertions

new[] { 1 }.Should().BeEquivalentTo(new[] { 2 }); // GenericCollectionAssertions
new[] { 1 }.Should().NotBeEquivalentTo(new[] { 2 }); // CollectionAssertions

new Dictionary<int, int>().Should().BeEquivalentTo(1); // GenericDictionaryAssertions
new Dictionary<int, int>().Should().NotBeEquivalentTo(1); // does not exists

new ComparableOfInt(1).Should().BeEquivalentTo(1); // ComparableTypeAssertions
new ComparableOfInt(1).Should().NotBeEquivalentTo(1); // does not exists
```

lg2de:
Currently there are some overloads for NotEquivalent, some are missing. 
With adding this one the Situation is not changed, but #170 is fixed.

jnyrup:
Had a deeper look at the inheritance graphs.
Only `StringCollectionAssertions` and `GenericCollectionAssertions` inherits `NotBeEquivalentTo` from `CollectionAssertions`, otherwise it looks fine.

So no objections anyway 👍

lg2de:
I added small portion of documentation.
I did not find additional senseful references to `BeEquivalentTo`. So, I do not know where to have additional documentation for the new overload.

PR #1076, opened 2019-06-04
Fix timing issues with Should().NotThrowAfter()

dennisdoomen:
Extended the existing `ITimer` abstraction and renamed it to `IClock` so all tests that rely on delays, waits and stopwatches can be tested in a reliable fashion. This should resolve the flakiness we've been experiences.

## IMPORTANT 


FluentAssertions.sln
```diff
@@ -145,8 +145,8 @@ Global
 		{FCAFB0F1-79EA-4D49-813B-188D4BC4BE71}.Release|Any CPU.Build.0 = Release|Any CPU
 		{05727385-DA56-4FC1-B44A-79DFECC3198A}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
 		{05727385-DA56-4FC1-B44A-79DFECC3198A}.Debug|Any CPU.Build.0 = Debug|Any CPU
-		{05727385-DA56-4FC1-B44A-79DFECC3198A}.Release|Any CPU.ActiveCfg = Release|Any CPU
-		{05727385-DA56-4FC1-B44A-79DFECC3198A}.Release|Any CPU.Build.0 = Release|Any CPU
+		{05727385-DA56-4FC1-B44A-79DFECC3198A}.Release|Any CPU.ActiveCfg = Debug|Any CPU
```
lg2de: Why do you build "Debug" with Release configuration?

Src/FluentAssertions/Common/Clock.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace FluentAssertions.Common
+{
+    /// <summary>
+    /// Default implementation for <see cref="IClock"/> for production use.
+    /// </summary>
+    internal class Clock : IClock
```
lg2de: Why is it internal?
How to use it in extensions?

Src/FluentAssertions/Common/IClock.cs
```diff
@@ -7,8 +7,10 @@ namespace FluentAssertions.Common
     /// <summary>
     /// Represents an abstract timer that is used to make some of this library's timing dependent functionality better testable.
     /// </summary>
-    public interface ITimer
+    public interface IClock
```
lg2de: In a similar "challenge" it was named `ITimerFactory`.
I think, "clock" is a very generic name.
Maybe, it could also be named `IAssertionTiming` and `IAssertionTimer`.

Src/FluentAssertions/Common/IClock.cs
```diff
@@ -25,5 +27,12 @@ public interface ITimer
         /// <param name="timeout">The time span to wait.</param>
         /// <returns><c>true</c> if the task completes before specified timeout.</returns>
         bool Wait(Task task, TimeSpan timeout);
+
+        ITimer StartTimer();
+    }
+
+    public interface ITimer
```
lg2de: documentation is missing

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
```diff
@@ -1080,13 +1080,15 @@ public void When_no_exception_should_be_thrown_after_wait_time_but_it_was_it_sho
             //-----------------------------------------------------------------------------------------------------------
             // Arrange
             //-----------------------------------------------------------------------------------------------------------
-            var watch = Stopwatch.StartNew();
             var waitTime = 100.Milliseconds();
```
lg2de: With using the abstracted timer, concrete time value should not be used anymore.
The expected behavior should be fully controlled by the test.

Tests/Shared.Specs/TestingTimer.cs
```diff
@@ -6,31 +6,54 @@
...
 {
     /// <summary>
-    /// Implementation of <see cref="ITimer"/> for testing purposes only.
+    /// Implementation of <see cref="IClock"/> for testing purposes only.
     /// </summary>
-    internal class TestingTimer : ITimer
+    internal class TestClock : IClock
     {
-        private TaskCompletionSource<bool> Signal { get; } = new TaskCompletionSource<bool>();
+        private DateTime now = DateTime.Now;
```
lg2de: The `now` is quite confusing to me. The time will be changed with each invocation of `Delay`.
You should add some comments to explain the complex behavior.

Tests/Shared.Specs/TestingTimer.cs
```diff
@@ -6,31 +6,54 @@
...
 
-        Task ITimer.DelayAsync(TimeSpan delay, CancellationToken cancellationToken)
+        private TaskCompletionSource<bool> delayTask { get; } = new TaskCompletionSource<bool>();
+
+        Task IClock.DelayAsync(TimeSpan delay, CancellationToken cancellationToken)
+        {
+            return delayTask.Task;
+        }
+
+        bool IClock.Wait(Task task, TimeSpan timeout)
```
lg2de: Waiting for a task does not update `now`.
This can cause unexpected behavior.

FluentAssertions.sln
dennisdoomen: Otherwise [Subject Identification](https://fluentassertions.com/documentation/#subject-identification) won't work.

Src/FluentAssertions/Common/Clock.cs
dennisdoomen: It was already `internal`. I'm not sure the API is stable enough to be exposed as an official extension point. People can always provide their own implementation. What do you think @jnyrup ?

Src/FluentAssertions/Common/IClock.cs
dennisdoomen: I named it like that because it does more than just producing timers now.

FluentAssertions.sln
lg2de: I think, instead the project configuration should be changed to not optimize.
Otherwise it is confusing that there is a Release configuration which is never used.

Src/FluentAssertions/Common/Clock.cs
lg2de: If you think about the stability of the API, the interfaces should be internal as well.

Src/FluentAssertions/Common/IClock.cs
lg2de: All is related to **timing**, this is why I recommend IAssertion**Timing**.

FluentAssertions.sln
dennisdoomen: We could do that, but now with this PR.

Src/FluentAssertions/Common/Clock.cs
dennisdoomen: Agreed, but they are part of the constructor dependencies of classes that are already public. In short, we can't.

Tests/Shared.Specs/TestingTimer.cs
dennisdoomen: I added some comments.

Src/FluentAssertions/Common/IClock.cs
dennisdoomen: I kept `IClock`

Tests/Shared.Specs/FormatterSpecs.cs
```diff
@@ -858,6 +858,7 @@ public void When_no_formatter_scanning_is_configured_it_should_use_the_default_f
             // Arrange
             //-----------------------------------------------------------------------------------------------------------
             Services.ResetToDefaults();
+            Configuration.Current.ValueFormatterDetectionMode = ValueFormatterDetectionMode.Disabled;
```
jnyrup: ❓ Is this change relation to timing?
If yes, how?

Tests/Shared.Specs/FakeClock.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// </remarks>
+    internal class FakeClock : IClock
+    {
+        private DateTime fakeNow = DateTime.Now;
+
+        private TaskCompletionSource<bool> delayTask { get; } = new TaskCompletionSource<bool>();
+
+        Task IClock.DelayAsync(TimeSpan delay, CancellationToken cancellationToken)
+        {
+            fakeNow = fakeNow.Add(delay);
```
jnyrup: 🙃 `DateTime` has nice (in my opinion) overloaded operators such as `+=`, `-` instead of `Add` and `Subtract`.

Tests/Shared.Specs/FakeClock.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// Implementation of <see cref="IClock"/> for testing purposes only.
+    /// </summary>
+    /// <remarks>
+    /// It allows you to control the "current" time.
+    /// </remarks>
+    internal class FakeClock : IClock
+    {
+        private DateTime fakeNow = DateTime.Now;
+
+        private TaskCompletionSource<bool> delayTask { get; } = new TaskCompletionSource<bool>();
```
jnyrup: ❌ lower case property. Should this be a readonly private field instead?

Src/FluentAssertions/Specialized/FunctionAssertions.cs
```diff
@@ -11,8 +12,11 @@ namespace FluentAssertions.Specialized
     [DebuggerNonUserCode]
     public class FunctionAssertions<T> : DelegateAssertions<Func<T>>
     {
-        public FunctionAssertions(Func<T> subject, IExtractExceptions extractor) : base(subject, extractor)
+        private readonly IClock clock;
+
+        public FunctionAssertions(Func<T> subject, IExtractExceptions extractor, IClock clock) : base(subject, extractor, clock)
```
jnyrup: 🔧 This and other classes changes a public constructor to take an additional mandatory `IClock`.
We could keep a constructor without an `IClock` and default to `Clock`

```c#
public FunctionAssertions(Func<T> subject, IExtractExceptions extractor)
    : this(subject, extractor, new Clock())
{}
```

Src/FluentAssertions/Common/IClock.cs
```diff
@@ -7,8 +7,14 @@ namespace FluentAssertions.Common
     /// <summary>
     /// Represents an abstract timer that is used to make some of this library's timing dependent functionality better testable.
     /// </summary>
-    public interface ITimer
+    public interface IClock
     {
+        /// <summary>
+        /// Will block the current thread until a time delay has passed..
```
jnyrup: 🙃 double dots `..`.

Tests/Shared.Specs/FakeClock.cs
```diff
@@ -0,0 +1,63 @@
...
+{
+    /// <summary>
+    /// Implementation of <see cref="IClock"/> for testing purposes only.
+    /// </summary>
+    /// <remarks>
+    /// It allows you to control the "current" time.
+    /// </remarks>
+    internal class FakeClock : IClock
+    {
+        private DateTime fakeNow = DateTime.Now;
```
jnyrup: 🔧 How about working entirely with `TimeSpan`, starting with `TimeSpan.Zero`?
The difference between two `TimeSpan` is also a `TimeSpan` as required for `ITimer.Elapsed`.

Tests/Shared.Specs/FormatterSpecs.cs
dennisdoomen: Moved it into a separate commit.

Src/FluentAssertions/Specialized/FunctionAssertions.cs
dennisdoomen: Good catch and great idea.

Src/FluentAssertions/Specialized/DelegateAssertions.cs
```diff
@@ -15,13 +15,19 @@ namespace FluentAssertions.Specialized
     public abstract class DelegateAssertions<TDelegate> : ReferenceTypeAssertions<Delegate, DelegateAssertions<TDelegate>> where TDelegate : Delegate
     {
         private readonly IExtractExceptions extractor;
+        private readonly IClock clock;
 
         protected DelegateAssertions(TDelegate @delegate, IExtractExceptions extractor) : base(@delegate)
         {
```
jnyrup: If this constructor is used, `clock` will be `null` in the constructed instance.
Can we call ``: this(...)` as in the other classes?

Src/FluentAssertions/Specialized/ActionAssertions.cs
```diff
@@ -9,7 +10,12 @@ namespace FluentAssertions.Specialized
     [DebuggerNonUserCode]
     public class ActionAssertions : DelegateAssertions<Action>
     {
-        public ActionAssertions(Action subject, IExtractExceptions extractor) : base(subject, extractor)
+        public ActionAssertions(Action subject, IExtractExceptions extractor) : this(subject, extractor, new Clock())
+        {
+            Subject = subject;
```
jnyrup: `Subject` is set in the other constructor?

Tests/Shared.Specs/ExceptionAssertionSpecs.cs
lg2de: You resolved the "conversation" without conversation.
:(
It seems you have no interest in my thoughts.

dennisdoomen: Sorry. I must have missed that one.

dennisdoomen: Wait, what part did I forget? That file is not using the `Stopwatch` anymore.

lg2de: The confusion my be related to the fact that I just selected one line for an comment which is intended to be more an overall comment:
You have introduced here the "new fake" for timing. I expect that specific time should NOT BE used anymore.
My understanding is that the fake completely remove any times from the test code.
Maybe, internally the specified time values will be not be used anymore for waiting or similar. But this is only understood by reading the implementation code of FakeClock.
Your attempt to "simulate time" by "incrementing now" looks interesting, but reading here "waittime" I expect the test code will "sleep" for the waittime. If it does not really wait "you" may be confused.
Hope this clarifies my thoughts...

dennisdoomen: Yes, but I'm going to postpone that for a next PR. For now, my goal is to fix the stability issue with that specific set of tests that is now causing all PRs to be red.

dennisdoomen:
> But my two last comments still holds :)

Which ones? I applied all your suggestions.

Src/FluentAssertions/Specialized/ActionAssertions.cs
jnyrup: @dennisdoomen this one

Src/FluentAssertions/Specialized/DelegateAssertions.cs
jnyrup: @dennisdoomen and this one

Src/FluentAssertions/Specialized/ActionAssertions.cs
dennisdoomen: Apparently I forgot to push my changes....

Src/FluentAssertions/Specialized/DelegateAssertions.cs
```diff
@@ -15,13 +15,19 @@ namespace FluentAssertions.Specialized
     public abstract class DelegateAssertions<TDelegate> : ReferenceTypeAssertions<Delegate, DelegateAssertions<TDelegate>> where TDelegate : Delegate
     {
         private readonly IExtractExceptions extractor;
+        private readonly IClock clock;
 
-        protected DelegateAssertions(TDelegate @delegate, IExtractExceptions extractor) : base(@delegate)
+        protected DelegateAssertions(TDelegate @delegate, IExtractExceptions extractor) : this(@delegate, extractor, new Clock())
         {
             this.extractor = extractor;
```
jnyrup: @dennisdoomen Move these two assignments to the other constructor

PR #1084, opened 2019-06-16
Fix redirects

jnyrup:
The fixes in #1083 did not seem to fix the backward compatibility problem as https://fluentassertions.com/documentation is still a dead link.

Apparently Github Pages does [not support ´.htaccess´](https://help.github.com/en/articles/redirects-on-github-pages).

Instead we can use `jekyll-redirect-from`. 
I removed the redirect from `extensibility.html` as it caused an infinite redirect loop.


jnyrup:
Build fail is due to time sensitive test

PR #1087, opened 2019-06-24
Fix MemberPath so that it can exclude overriden members

### LINKED ISSUE #1077: Excluding not functioning on overriden linked property
### Description

When trying to apply an exclusion to a property that links to a different property, Fluent Assertions ignores the exclusion on said property, saying the test failed when 

Test method:

```csharp
[TestMethod]
public async Task Then()
{
	var obj1 = new Candidate{PersonalInfo = new CandidatePersonalInfo{Firstname = "John", Lastname = "Johnson", Initials = "JJ"}};
	var obj2 = new Candidate{PersonalInfo = new CandidatePersonalInfo{Firstname = "John", Lastname = "Jameson", Initials = "JJ"}};

	obj1.Should().BeEquivalentTo(obj2, opt => opt.Excluding(o => o.PersonalInfo.Fullname).Excluding(o => o.DocumentName).Excluding(o => o.PersonalInfo.Lastname));
}
```

Relevant objects:

```csharp
public class TestCandidate : TestBase
{
	public override string DocumentName => PersonalInfo.Fullname;
	public CandidatePersonalInfo PersonalInfo;
}

public abstract class TestBase
{
	public abstract string DocumentName { get; }
}

public class CandidatePersonalInfo
{
	public CandidatePersonalInfo()
	{
		Location = new AddressInfo();
	}

	public string Firstname { get; set; }

	public string Middlename { get; set; }

	public string Lastname { get; set; }
			
	public string Initials { get; set; }

	public Gender Gender { get; set; }

	public DateTimeOffset? DateOfBirth { get; set; }

	public string Fullname => ($"{Firstname} {Middlename}".Trim() + $" {Lastname}").Trim();

}
```

### Expected behavior:

The property `DocumentName` should be excluded from the assertion

### Actual behavior:

The test case fails because the properties `DocumentName` are not equal

```
Then Failed: Expected member DocumentName to be 
Expected member DocumentName to be 
"John Jameson", but 
"John Johnson" differs near "ohn" (index 6).
Expected member Info.CreatedAt to be <2019-06-07 13:31:57.077648 +2h>, but found <2019-06-07 13:31:57.0770104 +2h>.

With configuration:
- Compare enums by value
- Exclude member root.PersonalInfo.Fullname
- Exclude member root.DocumentName
- Exclude member root.PersonalInfo.Lastname

   at FluentAssertions.Execution.LateBoundTestFramework.Throw(String message) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\LateBoundTestFramework.cs:line 16
   at FluentAssertions.Execution.CollectingAssertionStrategy.ThrowIfAny(IDictionary`2 context) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\CollectingAssertionStrategy.cs:line 49
   at FluentAssertions.Equivalency.EquivalencyValidator.AssertEquality(EquivalencyValidationContext context) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Equivalency\EquivalencyValidator.cs:line 45
   at RecruitNow.Cockpit.IntegrationTests.Candidates.GivenCandidatesController.WhenManagingCandidates.Then() in C:\Projects\RecruitNow.Cockpit 2.0\RecruitNow.Cockpit.IntegrationTests\Candidates\GivenCandidatesController\WhenManagingCandidates.cs:line 149


```

### Versions

5.6.0
.NET core 2.0

### Additional Information

Evangelink:
Fix #1077 

I think that the method `IsSameOrInherits` should be updated to use `IsSubclassOf` instead of `AssignableFrom` to avoid some potential FPs (assignable checks also for cast not only direct inheritance).

### DIFF
--- a/Tests/Shared.Specs/BasicEquivalencySpecs.cs
+++ b/Tests/Shared.Specs/BasicEquivalencySpecs.cs
@@ -1918,6 +1918,18 @@ namespace FluentAssertions.Specs
             act.Should().Throw<XunitException>().Which.Message.Should().Contain("Field1").And.Contain("Property1");
         }
 
+        [Fact]
+        public void When_excluding_virtual_or_abstract_property_exclusion_works_properly()
+        {
+            var obj1 = new Derived { DerivedProperty1 = "Something", DerivedProperty2 = "A" };
+            var obj2 = new Derived { DerivedProperty1 = "Something", DerivedProperty2 = "B" };
+
+            obj1.Should().BeEquivalentTo(obj2, opt => opt
+                .Excluding(o => o.AbstractProperty)
+                .Excluding(o => o.VirtualProperty)
+                .Excluding(o => o.DerivedProperty2));
+        }
+
         #endregion
 
         #region Matching Rules
@@ -4253,6 +4265,24 @@ namespace FluentAssertions.Specs
         {
             throw new InvalidCastException();
         }
+    }
+
+    public abstract class Base
+    {
+        public abstract string AbstractProperty { get; }
+
+        public virtual string VirtualProperty => "Foo";
+    }
+
+    public class Derived : Base
+    {
+        public string DerivedProperty1 { get; set; }
+
+        public string DerivedProperty2 { get; set; }
+
+        public override string AbstractProperty => $"{DerivedProperty1} {DerivedProperty2}";
+
+        public override string VirtualProperty => "Bar";
     }
 
     #endregion
--- a/Tests/Shared.Specs/BasicEquivalencySpecs.cs
+++ b/Tests/Shared.Specs/BasicEquivalencySpecs.cs
@@ -4272,6 +4272,8 @@ namespace FluentAssertions.Specs
         public abstract string AbstractProperty { get; }
 
         public virtual string VirtualProperty => "Foo";
+
+        public virtual string NonExcludedBaseProperty => "Foo";
     }
 
     public class Derived : Base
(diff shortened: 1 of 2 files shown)


Tests/Shared.Specs/BasicEquivalencySpecs.cs
```diff
@@ -1918,6 +1918,18 @@ public void
             act.Should().Throw<XunitException>().Which.Message.Should().Contain("Field1").And.Contain("Property1");
         }
 
+        [Fact]
+        public void When_excluding_virtual_or_abstract_property_exclusion_works_properly()
```
dennisdoomen: Don't you also need an example that uses that `VirtualProperty`?

Evangelink:
@dennisdoomen What do you think about creating a new PR to change `IsAssignableFrom` to `IsSubclassOf`?

dennisdoomen:
> What do you think about creating a new PR to change `IsAssignableFrom` to `IsSubclassOf`?

I don't quite get why. `IsSameOrInherits` implies that it'll return `true` if the type is the same or inherits from it and supports interfaces as well.

Evangelink:
Current implementation is probably good enough but `IsAssignableFrom` also handles cast. So if you have type A with property Foo and type B with the same property name and B has the implicit cast operator this would return true.

Tests/Shared.Specs/BasicEquivalencySpecs.cs
Evangelink: Done

dennisdoomen: I meant another unit test. I mean, you exclude `VirtualProperty` and `DerivedProperty`, but your _actual_ and _expectation_ don't seem to set those properties.

PR #1089, opened 2019-06-25
[RFC] Localize texts (GH-316)

### LINKED ISSUE #316: Localize texts to other languages
I'm thinking about fork your repo and change texts from Assertions FailWith that today are hardcoded to resources files.
My idea about that is allow community translate these messages to any language without need to change the code. What do you think?

The steps will be:
- Fork repository
- Get all hardcoded text and put into en-us.resources file
- Create an static variable that will hold the language (default will be en-us)

So, for example, if I translate it to Brazilian Portuguese, users will only need to do something like:

``` csharp
[TestInitialize]
public void TestInitialize()
{
     FluentAssertions.Language = "pt-br";
}
```

So, when some assert fail. Instead of show:
**_Expected value to be {0} because {1} but found {2}_**

I'll get from resources, and text will be (in pt-br):
**_Esperado o valor {0} porque {1}, porém foi encontrado {2}_**

Use resources file is what I have in mind, any other ideas will be appreciated!

Evangelink:
Fix #316 

## Current implementation

There is still quite a lot of work to do but I wanted to gather a first round of feedback on my implementation.

I have tried to stick to the following patterns:

- Key ends with `Format` if it contains any format item

- Key contains X, Y, Z, W for the location of the various format items in English. This helps with making sure we pass the good number of parameters.

## Suggestions

We have a lot of redundancies between messages for the last part `, but found XXX.` and so I was thinking about extracting this into a different message. As this text is a sub part of the sentence I would expect this to work with all languages but I don't have enough knowledge in translation/globalization to be 100% sure.

## Tests

Note that I still haven't started to write any test. Regarding those, I see multiple options (not all mutually exclusive):

- Running all the tests asserting some message in all the available languages.

- Adding a test that checks that all messages are not empty for all available languages.

- [advanced] We could check that we have the same number of format items for all translations of a message.

- Any other idea?


dennisdoomen:
That's an impressive amount of work @Evangelink. 

Some questions/concerns/remarks:
* How are we going to ensure that a new resource is properly translated into all supported languages?
* Can we do something smart with naming conventions to tie a resource to a particular set of assertions?
* I wonder whether we really should apply DRY on the resources? The fact that some phrase is reused right now may not always be what we want.

Evangelink:
> How are we going to ensure that a new resource is properly translated into all supported languages?

I was thinking about something like having a collection containing the list of supported languages and then updating tests so that they are automatically run once for each language. If a new resource is then added, during the review you'll have to ensure that a test has been added.

> Can we do something smart with naming conventions to tie a resource to a particular set of assertions?

What about either having one resx per set of assertions (Collections...) or adding a prefix to the keys? 

> I wonder whether we really should apply DRY on the resources? The fact that some phrase is reused right now may not always be what we want.

For this, I tend to think that we should go for the easiest solution with what we currently know which is we factorize what we can. If something comes up which requires us to change our approach it would be the good time to wonder how to tackle it.
But then if someone has a good experience with localization and wants to share some feedback/thoughts, I am all ears.

dennisdoomen:
> I was thinking about something like having a collection containing the list of supported languages and then updating tests so that they are automatically run once for each language.

I would write a test that verifies that all known resource keys exist in all supported languages. 

> What about either having one resx per set of assertions (Collections...) or adding a prefix to the keys?

Either would work for me. 

> For this, I tend to think that we should go for the easiest solution with what we currently know which is we factorize what we can. If something comes up which requires us to change our approach it would be the good time to wonder how to tackle it.

I'm fine with that approach.

Evangelink:
> Can we do something smart with naming conventions to tie a resource to a particular set of assertions?

Just wondering which scope you want to give to "set of assertions"?

- Use folder as base for the set? So we have `Collections`, `Equivalency`, `Xml` and so on.

- Disregard folders and logically group (i.e. `Collection` but includes `CollectionMemberInfo.cs` texts and so on).

- Other

dennisdoomen:
> Just wondering which scope you want to give to "set of assertions"?

Logical groups is best I think

jnyrup:
Are translations supposed to live inside the Fluent Assertions project or in extension packages?
What happens if a translation is missing?

I would dislike if the translations lived inside FA and we had to e.g. delay a release because of a missing translation of some language I know nothing about.

Evangelink:
@jnyrup I have moved the resources in a different project named `FluentAssertions.Localization`. I will make some more tests to ensure nothing changed but AFAIK when a translation is not found, there is a fallback to the neutral language.

Evangelink:
Guys, some of the failing tests are failing because we used to concatenate the value which is now passed as argument to the format string (and so is wrapped into `\"\"`).

Just wondering whether you have some preference between:
1/ Updating the tests to expect the new format. Depending on how you handle changes in the messages, this could mean a breaking change

2/ Using `string.Format` for arguments that should not be wrapped.

Evangelink:
@dennisdoomen, @jnyrup, could you give a first round of review to my changes? Specifically, could you check that:
- the pattern I have used is ok for you?
- I didn't miss the extraction of some string/resources

I will start working on new tests related to translations but @jnyrup last comment raised an interesting point. If I start adding tests to ensure that a key is translated into all available languages that means that the CI will ALWAYS fail until all translations are provided which doesn't look like a good option.

dennisdoomen:
> Guys, some of the failing tests are failing because we used to concatenate the value which is now passed as argument to the format string (and so is wrapped into `\"\"`).

Do you have an example of those options?

FluentAssertions.sln
```diff
@@ -1,4 +1,5 @@
-﻿Microsoft Visual Studio Solution File, Format Version 12.00
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
Evangelink: Automatic change from VS

Src/FluentAssertions.Localization/FluentAssertions.Localization.csproj
```diff
@@ -0,0 +1,39 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!-- One of the netstandard should be first so that the generated *.Designer.cs are generated with the right code -->
+    <TargetFrameworks>netstandard1.3;net45;net47;netstandard1.6;netstandard2.0;netcoreapp2.0</TargetFrameworks>
```
Evangelink: Not sure whether we actually need all target frameworks.

Src/FluentAssertions.Localization/Resources.Designer.cs
```diff
@@ -0,0 +1,5783 @@
...
+            }
+            set {
+                resourceCulture = value;
+            }
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to Be strict about the order of items in byte arrays.
+        /// </summary>
+        public static string Array_BeStrictAboutItemsOrderInByteArrays {
```
Evangelink: We could either have resources as `public` (current option) or as `internal` and add a `InternalVisibleToAttribute`, any preference?

Src/FluentAssertions/AlreadyFormattedString.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace FluentAssertions
+{
+    internal class AlreadyFormattedString
```
Evangelink: This class is my workaround for all parameters that you used to concatenate with strings and that I now pass as argument to the string format. Because of the `StringFormatter` which wraps the string inside double quote, we were having failing tests. By introducing this new class and the associated formatter I have managed to reproduce the previous state.

Src/FluentAssertions/Formatting/StringValueFormatter.cs
```diff
@@ -23,7 +23,9 @@ public string Format(object value, FormattingContext context, FormatChild format
             string prefix = context.UseLineBreaks ? Environment.NewLine : "";
             string escapedString = value.ToString();
 
-            return prefix + "\"" + escapedString + "\"";
+            return escapedString == Environment.NewLine
```
Evangelink: Instead of wrapping all new line inside a `AlreadyFormattedString` I have decided to update the string formatter to avoid wrapping them (this makes more sense IMO).

Evangelink:
> 
> 
> > Guys, some of the failing tests are failing because we used to concatenate the value which is now passed as argument to the format string (and so is wrapped into `\"\"`).
> 
> Do you have an example of those options?

Instead of the 2 options I was giving, I decided to create a class called `AlreadyFormattedString` which prevents the `StringValueFormatter` to wrap the string parameter inside double quotes. This wrapper class is used in all places where you used to concatenate the strings.

Evangelink:
It took me quite some time but TBH I am more and more wondering if we should proceed further on this PR. 

I personally prefer exceptions in English even if that's not my native language but I decided to go further with the implementation as I know some people like to have localized development experience and because that's actually something done for .net framework. But I have just seen that Microsoft is currently wondering whether they will localize exceptions for .Net Core or not. Besides, I think that the localization might slow down / make harder for you the development of this great tool.

The best thing would be to have some stats about how many percent of FA users really care about localized messages.

One last thing is that localization might be useful for teams where FA is used also for functional tests who are reviewed by non-dev people.

dennisdoomen:
I have the exact same concerns. Honestly, I never saw the value of this, but felt like "hey, if the community asks for this".

Evangelink:
Alright! That was a good training for me :D And if we/you need to ever do it you have some basis to start from.

I'll let you decide whether you want to keep the PR and/or ticket open to gather feedback.

jnyrup:
I'll close this PR for now as we all seem to agree on the potential problems of it.
I'm glad to hear you take this as training.

dennisdoomen:
Yeah, I agree. But @Evangelink definitely deserves kudos for trying.

Evangelink:
Thank you guys!

PR #1090, opened 2019-06-25
Updating AssertionScope.current from ThreadStatic to AsyncLocal to su…

conklinb:
Adding support for async code used within using (new AssertionScope())

This will have no impact on current functionality - however, this should prevent unintended consequences when running the code on multiple threads

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Linq;
 using FluentAssertions.Common;
+using Microsoft.VisualStudio.Threading;
 
 namespace FluentAssertions.Execution
 {
@@ -21,8 +22,8 @@ namespace FluentAssertions.Execution
         private Func<string> reason;
         private bool useLineBreaks;
 
-        [ThreadStatic]
-        private static AssertionScope current;
+        
+        private static AsyncLocal<AssertionScope> current = new AsyncLocal<AssertionScope>();
 
         private AssertionScope parent;
         private Func<string> expectation;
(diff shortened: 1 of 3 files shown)


Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Linq;
 using FluentAssertions.Common;
+using Microsoft.VisualStudio.Threading;
```
dennisdoomen: Hmm, this might be the wrong namespace since it doesn't compile.

jnyrup:
[`AsyncLocal`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1) isn't available until net46?

conklinb:
> [`AsyncLocal`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1) isn't available until net46?

Exactly, didn't even notice the 4.5 req. Let me fix that

jnyrup:
Is it possible to write a test that exposes the case, where `AsyncLocal` works and `[ThreadStatic]` doesn't?

conklinb:
> Is it possible to write a test that exposes the case, where `AsyncLocal` works and `[ThreadStatic]` doesn't?

Yes and no, what we see is: 

using(new AssertionScope()){
 //bunch of asserts in here
}

+ Running multiple tests in parallel

OCCASIONALLY you will see a case where something uses reaches out for current when Current is in the context of one of the other tests. 

This is, best I can tell, because many classes like to use AssertionScope.Current. 

I'm looking into locking methods now since CallContext doesnt work in .Net Core, and AsyncLocal doesnt work pre 4.6.

jnyrup:
It's fine by me to have something target framework dependent, if that makes it easier.
```c#
#if NET45
// Use CallContext
#else
// Use AsyncLocal
#endif
```

conklinb:
I am going to close this - and reopen.

PR #1091, opened 2019-06-26
Feature/assertionscope

conklinb:
Adds safety to tests using AssertionScope while running many tests in parallel

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -18,7 +18,9 @@ namespace FluentAssertions.Execution
     /// This class is supposed to have a very short life time and is not safe to be used in assertion that cross thread-boundaries such as when
     /// using <c>async</c> or <c>await</c>.
     /// </remarks>
+#if NET45
     [Serializable]
+#endif
     public class AssertionScope : IAssertionScope
     {
         #region Private Definitions
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -1,6 +1,13 @@
-﻿using System;
+using System;
 using System.Linq;
+#if NOTNET45
+using Microsoft.VisualStudio.Threading;
+#else
+using System.Runtime.Remoting.Messaging;
+#endif
 using FluentAssertions.Common;
+
+
 
 namespace FluentAssertions.Execution
 {
@@ -21,15 +28,15 @@ namespace FluentAssertions.Execution
         private Func<string> reason;
         private bool useLineBreaks;
 
-        [ThreadStatic]
-        private static AssertionScope current;
-
+#if NOTNET45
+        private static AsyncLocal<AssertionScope> current = new AsyncLocal<AssertionScope>();
+#endif
         private AssertionScope parent;
         private Func<string> expectation;
         private string fallbackIdentifier = "object";
         private bool? succeeded;
 
-        #endregion
+#endregion
 
         private AssertionScope(IAssertionStrategy _assertionStrategy)
         {
@@ -44,9 +51,15 @@ namespace FluentAssertions.Execution
         public AssertionScope()
             : this(new CollectingAssertionStrategy())
         {
-            parent = current;
-            current = this;
 
+#if NOTNET45
+            parent = current.Value;
+            current.Value = this;
+#else
+            
+            parent = (AssertionScope) CallContext.LogicalGetData("this");
+            CallContext.LogicalSetData("this", this);
+#endif
             if (parent != null)
             {
                 contextData.Add(parent.contextData);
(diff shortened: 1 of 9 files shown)


conklinb:
Looks like the tests need updated to support threading. On it

Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -60,6 +63,9 @@
     <PackageReference Include="System.ValueTuple" Version="4.4.0" />
     <Reference Include="System.Configuration" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' != 'net45'">
```
jnyrup: Can we use `AsyncLocal` from `System.Threading` instead of introducing a dependency on `Microsoft.VisualStudio.Threading`?

Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -1,7 +1,14 @@
-﻿using System;
+using System;
 using System.Linq;
+#if NOTNET45
```
jnyrup: No need to define a new constant, just use `!NET45`.

Src/FluentAssertions/Execution/DefaultAssertionStrategy.cs
```diff
@@ -1,43 +1,62 @@
...
 
 using FluentAssertions.Common;
 
 namespace FluentAssertions.Execution
 {
-    internal class DefaultAssertionStrategy : IAssertionStrategy
+#if NET45
+    [Serializable]
+#endif
+    internal class CollectingAssertionStrategy : IAssertionStrategy
```
jnyrup: `DefaultAssertionStrategy` is overwritten with the code of `CollectingAssertionStrategy`

Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -69,15 +85,33 @@ public AssertionScope(string context)
...
+        /// <summary>
+        /// Gets the current thread-specific assertion scope.
+        /// </summary>
+        public static AssertionScope Current => LogicalCurrent(); 
 
+        public static AssertionScope LogicalCurrent()
+        {
+            if (CallContext.LogicalGetData("this") == null)
+            {
+                CallContext.LogicalSetData("this", new AssertionScope(new DefaultAssertionStrategy()));
```
jnyrup: ❓ Currently when `current` is `null` a new `AssertionScope` is constructed and returned, but not written to `current`.
Why the difference here?

Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -44,9 +54,15 @@ private AssertionScope(IAssertionStrategy _assertionStrategy)
         public AssertionScope()
             : this(new CollectingAssertionStrategy())
         {
-            parent = current;
-            current = this;
 
+#if NOTNET45
```
jnyrup: How about if we kept the use of `CallContext` and `AsyncLocal` a single place.
E.g.
```c#
private static AssertionScope CurrentImpl // I haven't thought about a good name
{
    get
    {
#if NET45
        return (AssertionScope) CallContext.LogicalGetData("this");
#else
        return current.Value;
#endif
    }
    set
    {
#if NET45
        CallContext.LogicalSetData("this", value);
#else
        current.Value = value;
#endif
    }
}
```

Src/FluentAssertions/FluentAssertions.csproj
conklinb: Fixed

Src/FluentAssertions/Execution/AssertionScope.cs
conklinb: Fixed

Src/FluentAssertions/Execution/AssertionScope.cs
conklinb: Fix

Src/FluentAssertions/Execution/AssertionScope.cs
conklinb: done

conklinb:
Sorry about the one file per commit "commits" right now. Work VPN doesn't want to let me log in via bash or gh desktop

Tests/NetStandard13.Specs/NetStandard13.Specs.csproj
```diff
@@ -11,6 +11,7 @@
     <PackageReference Include="Chill" Version="4.0.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.0.1" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="System.Threading" Version="4.3.0" />
```
jnyrup: Is this necessary?
According to the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1?view=netstandard-1.3) `AsyncLocal` is available in `netstandard13`

conklinb: No, fixed.

Src/FluentAssertions/Execution/DefaultAssertionStrategy.cs
conklinb: Fixed

conklinb:
All comments addressed

jnyrup:
The revised implementation looks great to me 👍.

Now to verification.
I observe that if I remove the `[ThreadStatic]` from the current implementation the tests fail, whereas this PR succeeds.
So this PR seems at _least_ as good as master.

I've tried to come up with a test to verify that the changes are also an _improvement_ over master.
The test below fails in master, but succeeds with this PR.
If this test like this gets added (with better names than `First` and `Second`) I'm confident with these changes.


```c#
[Fact]
public async Task When_using_AssertionScope_across_thread_boundaries_it_should_work()
{
    using (var semaphore = new SemaphoreSlim(0, 1))
    {
        await Task.WhenAll(First(semaphore), Second(semaphore));
    }
}

private static async Task First(SemaphoreSlim semaphore)
{
    await Task.Yield();
    var scope = new AssertionScope();
    await semaphore.WaitAsync();
    scope.Should().BeSameAs(AssertionScope.Current);
}

private static async Task Second(SemaphoreSlim semaphore)
{
    await Task.Yield();
    var scope = new AssertionScope();
    semaphore.Release();
    scope.Should().BeSameAs(AssertionScope.Current);
}
```

conklinb:
Added Tests

conklinb:
Is there anything else you need in order to get this merged?

PR #1095, opened 2019-07-10
Unify API methods to check for attribute (GH-731)

### LINKED ISSUE #731: Consistency in Attribute assertions
If #727 gets merged, we should also revise other calls to
`GetCustomAttributes(bool inherit)` and
`GetCustomAttributes(Type attributeType, bool inherit)`

E.g `TypeSelector,ThatAreDecoratedWith` currently calls `GetCustomAttributes(typeof(TAttribute), true)` which would be inconsistent with the difference between `BeDecoratedWith` and `BeDecoratedWithOrInherit`.

Evangelink:
Fix #731 

I have marked the old methods as obsolete but depending on your preferences, I could get rid of them (breaking changes).

I decided to name the methods `XXXOrInherit` not `XXXOrInherits` as recommended to follow the convention used by already implemented methods.

Some of the tests are failing because of my changes on `PropertyInfoSelector.cs` but I can't figure out what's wrong with my change. I'd be grateful for some fresh eyes because I am likely missing something obvious.

### DIFF
--- a/Src/FluentAssertions/Types/PropertyInfoSelector.cs
+++ b/Src/FluentAssertions/Types/PropertyInfoSelector.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using FluentAssertions.Common;
 
 namespace FluentAssertions.Types
 {
@@ -55,7 +56,7 @@ namespace FluentAssertions.Types
         public PropertyInfoSelector ThatAreDecoratedWith<TAttribute>()
             where TAttribute : Attribute
         {
-            selectedProperties = selectedProperties.Where(property => CustomAttributeExtensions.GetCustomAttributes(property, false).OfType<TAttribute>().Any());
+            selectedProperties = selectedProperties.Where(property => property.IsDecoratedWith<TAttribute>());
             return this;
         }
 
@@ -65,7 +66,7 @@ namespace FluentAssertions.Types
         public PropertyInfoSelector ThatAreDecoratedWithOrInherit<TAttribute>()
             where TAttribute : Attribute
         {
-            selectedProperties = selectedProperties.Where(property => CustomAttributeExtensions.GetCustomAttributes(property, true).OfType<TAttribute>().Any());
+            selectedProperties = selectedProperties.Where(property => property.IsDecoratedWithOrInherit<TAttribute>());
             return this;
         }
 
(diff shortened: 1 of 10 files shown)


dennisdoomen:
Hmm, the code you touched is supposed to be used internally and should not have been `public` in the first place.

Evangelink:
@dennisdoomen Which strategy do you want to go for? Breaking change, in which case we can change the accessibility, current option (with `Obsolete`) or shall I remove the `Obsolete`?

Src/FluentAssertions/Types/PropertyInfoSelector.cs
```diff
@@ -65,7 +66,18 @@ public PropertyInfoSelector ThatAreDecoratedWith<TAttribute>()
         public PropertyInfoSelector ThatAreDecoratedWithOrInherit<TAttribute>()
             where TAttribute : Attribute
         {
-            selectedProperties = selectedProperties.Where(property => CustomAttributeExtensions.GetCustomAttributes(property, true).OfType<TAttribute>().Any());
+            // TODO: There seems to be a bug with `GetCustomAttributes` when used as extension method
```
Evangelink: It seems that there is a bug with `GetCustomAttributes`, I'll file an issue on Microsoft side.

jnyrup:
@Evangelink If I remember correct `GetCustomAttributes` behaves differently for net45. 
I'm on mobile for the next two weeks, so can't look it up at the moment, but I think I also commented it in some other FA issue/PR.

dennisdoomen:
I think we can make all of these `internal`. They were never supposed to be used outside the assembly.

Evangelink:
@jnyrup it seems that the bug only happens when using the method as extension method so I did a change in the code and left a message to explain why,

@dennisdoomen Just to be sure, are you referring to `HasAttribute` and the new methods added? All the methods of the `TypeExtensions` class or something else?

dennisdoomen:
All the methods of the `TypeExtensions` class. In fact, anything under `Common` is for internal use.

Evangelink:
Do you want me to do it in this PR or to create a new one?

dennisdoomen:
> Do you want me to do it in this PR or to create a new one?

Whatever you like ;-)

Evangelink:
@dennisdoomen Are you okay with `InternalsVisibleTo`? Because a lot of the extension methods are used in the unit tests

Evangelink:
ping @dennisdoomen

jnyrup:
I like the PR as it is right now.
We are not currently consistent in when types in `FluentAssertions.Common` should be `internal`
E.g. `CSharpAccessModifier` is inside `FluentAssertions.Common`, but `public` and used for `HaveAccessModifier`.

Evangelink:
I'll try to push a rebase before the week-end. Thank you for the review :)

Evangelink:
@dennisdoomen @jnyrup PR ready to be merged!

PR #1100, opened 2019-07-26
Argument exceptions

jnyrup:
This PR adds null guards on predicates, to avoid throwing `NullReferenceException` if one somehow manages to pass in null values to assertions.

This is a part of #1039


dennisdoomen:
Gee, where do you find the time.... ;-)

Src/FluentAssertions/Common/Guard.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+
+namespace FluentAssertions.Common
+{
+    internal static class Guard
+    {
+        public static void ThrowIfArgumentIsNull<T>(T obj, string paramName)
+            where T : class
```
dennisdoomen: Can't you do something smart with an `Expression<T>` so you can get the value _and_ name from the same parameter?

jnyrup: It seems doable by using an `Expression<Func<T>>`

```c#
public static void ThrowIfArgumentIsNull<T>(Expression<Func<T>> expression)
    where T : class
{
    Func<T> func = expression.Compile();
    T obj = func();

    if (obj is null)
    {
        var body = (MemberExpression)expression.Body;
        var variableName = body.Member.Name;
        throw new ArgumentNullException(variableName);
    }
}
```

```c#
string a = null;
ThrowIfArgumentIsNull(() => a);
```

```
Value cannot be null.
Parameter name: a
```

I see it would be feasible to avoid the duplication, but using `Expression`s feel smelly to me.

Ultimately what we want is [`[CallerArgumentExpression]`](https://github.com/dotnet/csharplang/issues/287)

PR #1104, opened 2019-07-27
Fix [Not]BeAssignableTo failure messages

### LINKED ISSUE #1103: Confusing failure message in Should().BeAssignableTo(...)
### Description

When `Should().BeAssignableTo(...)` fails, it leaves a confusing message.

```csharp
((object) null).Should().BeAssignableTo(typeof(string));
```

### Expected behavior:

`Expected an object assignable to System.String, but found <null> instead.`

### Actual behavior:

`Expected type not to be System.String, but found <null>.`

### Versions

5.7.0

.NET 4.6

### Additional Information

image

jnyrup:
This fixes #1103

### DIFF
--- a/Tests/Shared.Specs/ObjectAssertionSpecs.cs
+++ b/Tests/Shared.Specs/ObjectAssertionSpecs.cs
@@ -662,6 +662,22 @@ namespace FluentAssertions.Specs
         }
 
         [Fact]
+        public void When_a_null_instance_is_asserted_to_be_assignable_it_should_fail_with_a_descriptive_message()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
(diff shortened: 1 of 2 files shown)


Tests/Shared.Specs/ObjectAssertionSpecs.cs
```diff
@@ -661,6 +661,22 @@ public void When_an_implemented_open_generic_interface_type_instance_it_should_s
             someObject.Should().BeAssignableTo(typeof(System.Collections.Generic.IList<>));
         }
 
+        [Fact]
+        public void When_a_null_type_instance_it_should_fail_with_a_descriptive_message()
```
dennisdoomen: Bit of a weird name

PR #1106, opened 2019-07-27
cleanup

jnyrup:
* auto-properties
* whitespacing
* marking fields as `readonly`
* using builtin types (`String` vs `string`)
* avoid casting return type, e.g. `T[]` to `IEnumerable<T>`
* remove unused variables
* Use compound assignment


jnyrup:
A failing timing test `When_the_execution_time_of_an_action_is_close_to_a_limit_it_should_not_throw` :(

Src/FluentAssertions/Execution/MessageBuilder.cs
```diff
@@ -48,23 +48,25 @@ private string SubstituteIdentifier(string message, string identifier, string fa
...
             message = Regex.Replace(message, pattern, match =>
             {
-                string defaultIdentifier = match.Groups["default"].Value;
-                string result = "object";
-
+                const string result = " ";
                 if (!identifier.IsNullOrEmpty())
                 {
-                    result = identifier;
+                    return result + identifier;
```
dennisdoomen: You know i hate those intermediate `return`s, don't you?

jnyrup: I was not aware of that.
I usually code in a "return early, return often" style and a prefer Guard Clauses.

My intention here was to avoid the initialization of `defaultIdentifier` if the first branch was taken.

With a single `return` it could either be the code below, which I dislike because of additional indentation, or extracting the else part into a separate method.
```c#
if (!identifier.IsNullOrEmpty())
{
    result = identifier;
}
else
{
    string defaultIdentifier = match.Groups["default"].Value;

    if (!defaultIdentifier.IsNullOrEmpty())
	{
		result = defaultIdentifier;
	}
	else if (!fallbackIdentifier.IsNullOrEmpty())
	{
		result = fallbackIdentifier;
	}
}
```

dennisdoomen: To be fair, I don't like that either. So it's a bit of a taste thing. I've been bitten a few times by long methods that suddenly return somewhere half-way. Of course, the problem wasn't really the `return` statement but in the length of that method.

PR #1109, opened 2019-07-28
Styling

jnyrup:
The vast majority is changing implicit types to explicit, but I've put the changes into several commits, to make reviewing easier.

bea6b63 renames a parameter in a public `AssertionScope` ctor, but that ctor was made public after the latests release, so the change does not break the public API.

### DIFF
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -343,7 +343,7 @@ namespace FluentAssertions.Collections
             // from O(n^2) to O(n). For bigger tables it is necessary in order to achieve acceptable
             // execution times.
             Func<EquivalencyAssertionOptions<TExpectation>, EquivalencyAssertionOptions<TExpectation>> forceStringOrderingConfig =
-                x => config(x).WithStrictOrderingFor(s => s.SelectedMemberPath == "");
+                x => config(x).WithStrictOrderingFor(s => string.IsNullOrEmpty(s.SelectedMemberPath));
 
             BeEquivalentTo(repeatedExpectation, forceStringOrderingConfig, because, becauseArgs);
         }
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -45,7 +45,7 @@ namespace FluentAssertions.Collections
 
             Func<T, TKey> compiledPredicate = predicate.Compile();
 
-            var values = Subject
+            T[] values = Subject
                 .Where(e => compiledPredicate(e) is null)
                 .ToArray();
 
--- a/Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
@@ -777,7 +777,7 @@ namespace FluentAssertions.Collections
         /// Returns an enumerable consisting of all items in the first collection also appearing in the second.
         /// </summary>
         /// <remarks>Enumerable.Intersect is not suitable because it drops any repeated elements.</remarks>
-        private IEnumerable<TValue> RepetitionPreservingIntersect(
+        private static IEnumerable<TValue> RepetitionPreservingIntersect(
             IEnumerable<TValue> first, IEnumerable<TValue> second)
         {
             var secondSet = new HashSet<TValue>(second);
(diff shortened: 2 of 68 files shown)


lg2de:
Why do you prefer explicit types? There maybe parts getting more readable, but I think most times "var" is better for overall readability. 

Why do you disable warnings without explanation (documentation/comment)? 

Why you do not like "this."? I think it helps to distinguish between local variables and members.

jnyrup:
I tried to add [Microsoft.CodeAnalysis](https://www.nuget.org/packages/Microsoft.CodeAnalysis) and [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers) to see what they would point out.
I would like to add analyzers that catch styling "typos" as I hope it would make it easier to contribute if an analyzer warn/error about it instead of human reviewers.

> Why do you prefer explicit types?

That's used style for this the project.
When I joined I was also skeptic about the explicit types, but I have to admit that it makes reviewing easier for me.

> Why do you disable warnings without explanation (documentation/comment)?

Good point, I'll fix that.

> Why you do not like "this."?

I think `this` should be used to qualify members when they are hidden by other variables.

dennisdoomen:
In short, we follow the https://csharpcodingguidelines.com/

Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
```diff
@@ -355,7 +355,7 @@ public GenericDictionaryAssertions(IDictionary<TKey, TValue> dictionary) : base(
                         additionalKeys);
             }
 
-            foreach (var key in expected.Keys)
+            foreach (TKey key in expected.Keys)
```
lg2de: var and TKey are generic.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -43,7 +43,7 @@ public AndConstraint<GenericCollectionAssertions<T>> NotContainNulls<TKey>(Expre
 
             Func<T, TKey> compiledPredicate = predicate.Compile();
 
-            var values = Subject
+            T[] values = Subject
```
lg2de: According AV1520 the type is clear enough.

Src/FluentAssertions/Common/ExpressionExtensions.cs
```diff
@@ -46,7 +46,7 @@ internal static class ExpressionExtensions
                 throw new ArgumentNullException(nameof(expression), "Expected a property expression, but found <null>.");
             }
 
-            var memberInfo = AttemptToGetMemberInfoFromCastExpression(expression) ??
+            MemberInfo memberInfo = AttemptToGetMemberInfoFromCastExpression(expression) ??
```
lg2de: According AV1520 `RepositoryFactory.Get()` can use var. I think, here the type is clear, too.

Src/FluentAssertions/Common/NoSynchronizationContextScope.cs
```diff
@@ -7,7 +7,7 @@ internal static class NoSynchronizationContextScope
     {
         public static DisposingAction Enter()
         {
-            var context = SynchronizationContext.Current;
+            SynchronizationContext context = SynchronizationContext.Current;
```
lg2de: According AV1520 the type is clear.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
dennisdoomen: Knowing that it's an array is valuable info.

Src/FluentAssertions/Common/ExpressionExtensions.cs
dennisdoomen: Debatable. I'd prefer an explicit type here.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
lg2de: The line ends with `ToArray`. Isn't clear enough?

Src/FluentAssertions/Common/ExpressionExtensions.cs
lg2de: Can you give a reason? (I would like to understand. )
"MemberInfo" is part of the method name.
What is the difference to the repository sample?

dennisdoomen: IMO, a variable is similar to a relationship/association in UML. It has a type and a role that type plays in the relationship. When there's no need to differentiate or when the type is obvious, `var` is fine. Otherwise, I prefer both.

PR #1112, opened 2019-08-09
Add Match method to Numeric and NullableNumeric assertions

### LINKED ISSUE #372: Support Match() for non-reference types?
I have an existing library I'm updating to use Fluent Assertions instead of the built-in xunit `Assert` class.

I have an xunit assert that looks like this, where the `UserId` property is of type `int?` which needs migrating:

``` csharp
Assert.True(project.Item.UserId == null || project.Item.UserId > 0);
```

There doesn't appear to be a way to do an equivalent with Fluent Assertions. I found the `Match<T>()` method, but this only works for reference types.

I'm currently doing this as a workaround:

``` csharp
if (project.Item.UserId.HasValue)
{
    project.Item.UserId.Should().BeGreaterThan(0);
}
```

but it would be much nicer to do something this:

``` csharp
project.Item.UserId.Should().Match((p) => p == null || p > 0);
```

Liklainy:
Fix #372 

- Added `Match` for `NumericAssertions` and `NullableNumericAssertions`
- Added simple tests


dennisdoomen:
You may want to update the docs in the same PR.

Src/FluentAssertions/Numeric/NumericAssertions.cs
```diff
@@ -384,5 +385,35 @@ public AndConstraint<NumericAssertions<T>> NotBeOfType(Type unexpectedType, stri
...
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because"/>.
+        /// </param>
+        public AndConstraint<NumericAssertions<T>> Match(Expression<Func<T, bool>> predicate,
+            string because = "",
+            params object[] becauseArgs)
+        {
+            if (predicate == null)
+            {
+                throw new ArgumentNullException(nameof(predicate), "Cannot match an value against a <null> predicate.");
```
jnyrup: *a* value

Src/FluentAssertions/Numeric/NullableNumericAssertions.cs
```diff
@@ -81,5 +83,35 @@ public AndConstraint<NullableNumericAssertions<T>> BeNull(string because = "", p
...
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because"/>.
+        /// </param>
+        public AndConstraint<NullableNumericAssertions<T>> Match(Expression<Func<T?, bool>> predicate,
+            string because = "",
+            params object[] becauseArgs)
+        {
+            if (predicate == null)
+            {
+                throw new ArgumentNullException(nameof(predicate), "Cannot match an value against a <null> predicate.");
```
jnyrup: *a* value

docs/_pages/nullabletypes.md
```diff
@@ -13,12 +13,15 @@ sidebar:
...
 theInt.Should().HaveValue();
 theInt.Should().NotBeNull();
+theInt.Should().Match(x => x.HasValue && x > 0);
 
 DateTime? theDate = null;
 theDate.Should().NotHaveValue();
 theDate.Should().BeNull();
-```
\ No newline at end of file
+theDate.Should().Match(x => !x.HasValue);
```
jnyrup: Remove the example for `theDate`

Tests/Shared.Specs/NullableNumericAssertionSpecs.cs
```diff
@@ -1243,6 +1243,65 @@ public void When_asserting_not_approximately_and_nullable_decimal_is_approximati
...
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            int? nullableInteger = 1;
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => nullableInteger.Should().Match(o => !o.HasValue, "it is not initialized yet");
```
jnyrup: Please include `becauseArgs` to verify they are properly passed along to `BecauseOf`.
E.g.
```c#
Match(o => !o.HasValue, "because we want to test the failure {0}", "message");
...
WithMessage("Expected value to match Not(o.HasValue) because we want to test the failure message.")
```

docs/_pages/nullabletypes.md
```diff
@@ -13,12 +13,15 @@ sidebar:
 short? theShort = null;
 theShort.Should().NotHaveValue();
 theShort.Should().BeNull();
+theShort.Should().Match(x => !x.HasValue);
```
jnyrup: I would prefer some examples that better reflects the use cases that `Match` is intended to cover, i.e. assertions that can't be expressed with existing ones.
E.g. `Match(x => !x.HasValue)` could be written as `NotHaveValue()`.

How about `Match(x => !x.HaveValue || x > 0)`?

Src/FluentAssertions/Numeric/NullableNumericAssertions.cs
```diff
@@ -81,5 +83,35 @@ public AndConstraint<NullableNumericAssertions<T>> BeNull(string because = "", p
...
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because"/>.
+        /// </param>
+        public AndConstraint<NullableNumericAssertions<T>> Match(Expression<Func<T?, bool>> predicate,
+            string because = "",
+            params object[] becauseArgs)
+        {
+            if (predicate == null)
```
jnyrup: 🙃 If you're branching out from master, I've added a `Guard.ThrowIfArgumentIsNull` helper.

PR #1117, opened 2019-08-18
Null checks on delegates

jnyrup:
When asserting on delegates we do not check if the delegate is non-null before invoking it.

`FailIfSubjectIsAsyncVoid` is changed so it does not run for `AsyncFunctionAssertions` via a new `CanHandleAsync` property.
The reason it worked before, was that `ExecuteInDefaultSynchronizationContext` wrapped an `async Task` method in a `Task` method.


Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -57,10 +59,15 @@ protected override void InvokeSubject()
             params object[] becauseArgs)
             where TException : Exception
         {
-            Exception exception = await InvokeSubjectWithInterceptionAsync();
-
             Type expectedType = typeof(TException);
 
+            Execute.Assertion
+                .ForCondition(!ReferenceEquals(Subject, null))
```
dennisdoomen: Isn't there a reverse version of `subject is null`?

jnyrup: The only exact opposite of `p is null` is `!(p is null)`.
The closest syntax without a negation operator is `p is object`.
The difference between the two is that:
* `is null` *only* works on reference types as value types cannot be `null`.
* `is object` also works on value types, as every types inherits `object`, but is always `true` and gives a CS0183 warning.

In C# 8 with recursive patterns, you could even write
`p is {} x` as a shorthand for `p is MyClass x`.

But I would be happy to replace as many `ReferenceEquals(p, null)` with `p is object`/`p is null` in the code base.

PR #1122, opened 2019-08-26
Replace Cake with Nuke

matkoch:


dennisdoomen:
The build failed.

matkoch:
https://twitter.com/matkoch87/status/1166026259039522818

dennisdoomen:
Beware, we customized the Cake bootstrapper script to first download .NET Core 3.0.100-preview7. This is needed for the solution to compile

robertcoltheart:
What's the reasoning for switching to Nuke? Just curious.

dennisdoomen:
* Downloading .NET Core is built-in
* Can run targets without its dependencies or skip individual dependencies
* Nice look-and-feel in the output
* Quickly jump to a target using familiar techniques (F12)
* You are forced to think about dependencies a target has instead of just defining a list of targets and hope they are executed in order
* Targets can define 
		○ required parameters
		○ dynamic conditions
* Tools are added as real Nuget dependnecies (with explicit versions)
* You can hide targets that are only used internally
* Parameters can provide a help text
* Can run multiple independent targets
* Can request required parameters at the command-line
* Every built-in command logs automatically 
* Command-line help lists all (listed) targets as well as all parameters
* Very nice method to download nuget packages and use its tools
* A lot of built-in methods follow the same name as the corresponding .NET Framework methods.
* Target dependencies are type-safe instead of magic strings
* Can use breakpoints to debug individual targets with the Rider/Resharper plugins
* Support through a Slack channel
* Plans for parallel builds

Build/_build.csproj
```diff
@@ -0,0 +1,40 @@
...
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+    <RootNamespace></RootNamespace>
+    <IsPackable>False</IsPackable>
+    <NoWarn>CS0649;CS0169</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageDownload Include="GitVersion.CommandLine.DotNetCore" Version="[4.0.1-beta1-53]" />
```
jnyrup: Why depending on a beta?
It seems `GitVersion.CommandLine.DotNetCore` comes with batteries included.
Can we remove `Lib/`?

matkoch: You want it updated?

jnyrup:
I wanted to try this out on my linux machine at home before commenting, but here goes.
This is awesome!

Comparing the artifacts of 
https://ci.appveyor.com/project/dennisdoomen/fluentassertions/builds/26952990/artifacts
image


https://ci.appveyor.com/project/dennisdoomen/fluentassertions/builds/26999396/artifacts
image

I would *love* if it only updated the assembly version numbers when running the `Pack` target, so the `AssemblyInfo.cs` isn't changed when testing locally.

Build/_build.csproj
jnyrup: I would rather rely on a "stable" release if it works.
Mostly I just wondered why `beta1-53`, it seemed so arbitrary to me, as if there was something particular about exactly that version.

matkoch:
@jnyrup 	there's no clean way of updating the AssemblyInfo only for the `Pack` target. I wonder why you even have the `AssemblyInfo.cs`, afaik, all the version numbers can be just passed from command-line.

Build/_build.csproj
matkoch: It just was the latest available at that time I wrote the template.

jnyrup:
@matkoch I quite sure it's just legacy.
I would be happy to nuke `AssemblyInfo.cs` if we can either setup the csproj to fetch the numbers automatically or inject them via cli parameters.
Do you have any ideas/recommendations?

robertcoltheart:
Pretty sure Nuke can handle passing in versions using the `DotNetBuild` syntax. All other values in AssemblyInfo.cs can be specified in your csproj.

jnyrup:
Just to be clear, as it was getting late for me yesterday when I looked at this yesterday.

If we can get rid of `AssemblyInfo.cs` by:
* moving attributes to csproj, and
* let e.g. `GitVersion` inject the AssemblyVersion/AssemblyFileVersion/AssemblyInformationalVersion numbers from cli

then I'm good with this PR.

dennisdoomen:
@matkoch any chance you want to wrap up with the remarks from @jnyrup ?

matkoch:
Updated, please take a look. AssemblyInfo.cs seemed outdated.

jnyrup:
The differences between the assembly from this PR now limits to whether using normalized numbers.
I'm not very familiar with the implications (if any) of using `major.minor.0.0` over `major.minor.pacth.0` as the AssemblyVersion while also using strong naming.
To put it another way, whether we should:
* keep the current versioning scheme (safe-than-sorry), or
* if the normalized numbers are non-breaking and have benefits.

Old assembly:
AssemblyVersion: 5.9.1.0
AssemblyFileVersion: 5.9.1.0

New assembly:
AssemblyVersion: 5.9.0.0
AssemblyFileVersion: 5.9.1

https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/versioning#assembly-version
> ✔️ CONSIDER only including a major version in the AssemblyVersion.
e.g. Library 1.0 and Library 1.0.1 both have an AssemblyVersion of 1.0.0.0, while Library 2.0 has AssemblyVersion of 2.0.0.0. 
When the assembly version changes less often, it reduces binding redirects.

https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming
> ✔️ CONSIDER incrementing the assembly version on only major version changes to help users reduce binding redirects, and how often they're updated.
Read more about versioning and the assembly version.

https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/versioning#assembly-file-version
> ✔️ DO use the format Major.Minor.Build.Revision for file version.
While the file version is never used by .NET, Windows expects the file version to be in the Major.Minor.Build.Revision format. 
A warning is raised if the version doesn't follow this format.

dennisdoomen:
As long as the full assembly name remains `FluentAssertions, Version=5.9.1.0, Culture=neutral`, I'm fine with either.

matkoch:
Please just tell me which format you want. I have my own reasoning behind this.

dennisdoomen:
> I have my own reasoning behind this.

Care to explain? Maybe we're missing something.

matkoch:
@dennisdoomen Not really. Don't remember the post. And I guess for assemblies participating in testing/building it doesn't even matter.

dennisdoomen:
Let's stick the full version as provided by GitVersoin

Build/Build.cs
```diff
@@ -0,0 +1,140 @@
...
+[UnsetVisualStudioEnvironmentVariables]
+[DotNetVerbosityMapping]
+class Build : NukeBuild
+{
+    /// Support plugins are available for:
+    ///   - JetBrains ReSharper        https://nuke.build/resharper
+    ///   - JetBrains Rider            https://nuke.build/rider
+    ///   - Microsoft VisualStudio     https://nuke.build/visualstudio
+    ///   - Microsoft VSCode           https://nuke.build/vscode
+    public static int Main() => Execute<Build>(
```
jnyrup: Does not work locally for me, as it compiles in debug mode, but tries to pack in release mode.
@dennisdoomen you had some opinions about default invocation of `build.ps1`?

dennisdoomen: Everything should be compiled in _release_ mode, except for the unit tests. We need the full symbols to be able to get proper identifier names.

jnyrup:
Thanks a lot @matkoch!
I have had "replace Cake" on my todo for quite some time now.

matkoch:
Hoping you're happy with the result :>

PR #1131, opened 2019-09-09
Update NotThrowAfter tests to use FakeClock (GH-1116) 

### LINKED ISSUE #1116: NotThrowAfter tests does not use FakeClock
### Description

Two tests of `NotThrowAfter` and `NotThrowAfterAsync`, respectively, does not utilize the new `FakeClock` introduced in #1076 and are sensitive to timing.

* `When_no_exception_should_be_thrown_for_async_func_executed_with_wait_after_wait_time_but_it_was_it_should_throw`
* `When_no_exception_should_be_thrown_for_async_func_executed_with_wait_after_wait_time_and_none_was_it_should_not_throw`
* `When_no_exception_should_be_thrown_for_async_func_after_wait_time_but_it_was_it_should_throw`
* `When_no_exception_should_be_thrown_for_async_func_after_wait_time_and_none_was_it_should_not_throw`

Let's fix 'em!

Evangelink:
Fix #1116

### DIFF
--- a/Tests/Shared.Specs/AsyncFunctionExceptionAssertionSpecs.cs
+++ b/Tests/Shared.Specs/AsyncFunctionExceptionAssertionSpecs.cs
@@ -1042,7 +1042,7 @@ namespace FluentAssertions.Specs
             //-----------------------------------------------------------------------------------------------------------
             // Assert
             //-----------------------------------------------------------------------------------------------------------
-            action.Should(clock).Throw<XunitException>()
+            action.Should().Throw<XunitException>()
                 .WithMessage("Did not expect any exceptions after 2s because we passed valid arguments*");
         }
 
@@ -1076,7 +1076,7 @@ namespace FluentAssertions.Specs
             //-----------------------------------------------------------------------------------------------------------
             // Assert
             //-----------------------------------------------------------------------------------------------------------
-            act.Should(clock).NotThrow();
+            act.Should().NotThrow();
         }
 
         #endregion
@@ -1210,7 +1210,7 @@ namespace FluentAssertions.Specs
             //-----------------------------------------------------------------------------------------------------------
             // Assert
             //-----------------------------------------------------------------------------------------------------------
-            action.Should(clock).Throw<XunitException>()
+            action.Should().Throw<XunitException>()
                 .WithMessage("Did not expect any exceptions after 2s because we passed valid arguments*");
         }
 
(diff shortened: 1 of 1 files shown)


Tests/Shared.Specs/AsyncFunctionExceptionAssertionSpecs.cs
```diff
@@ -1177,13 +1183,17 @@ public void When_no_exception_should_be_thrown_for_async_func_after_wait_time_bu
...
             // Arrange
             //-----------------------------------------------------------------------------------------------------------
-            var watch = Stopwatch.StartNew();
             var waitTime = 2.Seconds();
             var pollInterval = 10.Milliseconds();
 
+            var clock = new FakeClock();
+            var timer = clock.StartTimer();
+            clock.Delay(waitTime);
+            clock.CompletesBeforeTimeout();
```
Evangelink: This test seems to be the only one requiring this call (note that the RunsIntoTimeout method) also works here.

Tests/Shared.Specs/AsyncFunctionExceptionAssertionSpecs.cs
```diff
@@ -1033,13 +1036,13 @@ public void
...
             //-----------------------------------------------------------------------------------------------------------
-            Action action = () => throwLongerThanWaitTime.Should()
+            Action action = () => throwLongerThanWaitTime.Should(clock)
                 .NotThrowAfter(waitTime, pollInterval, "we passed valid arguments");
 
             //-----------------------------------------------------------------------------------------------------------
             // Assert
             //-----------------------------------------------------------------------------------------------------------
-            action.Should().Throw<XunitException>()
+            action.Should(clock).Throw<XunitException>()
```
jnyrup: 🤔 `Should().Throw()` and `Should().NotThrow()` should not need an `ITimer` as they aren't timing related assertions?

Tests/Shared.Specs/AsyncFunctionExceptionAssertionSpecs.cs
jnyrup: Whether you use `CompletesBeforeTimeout` or `RunsIntoTimeout` seems to only matter for sync assertions, e.g. `CompleteWithin()`.
See `When_task_completes_fast_it_should_succeed`.

I frankly don't known if that is intended.

Tests/Shared.Specs/AsyncFunctionExceptionAssertionSpecs.cs
Evangelink: Indeed!

PR #1132, opened 2019-09-09
[WIP] Add support for IImmutableDictionary and IReadOnlyDictionary

### LINKED ISSUE #568: Support for IReadOnlyDictionary, IReadOnlyCollection

Evangelink:
Fix #568 and overrides #963 

I still need to add some tests for `IImmutableDictionary` and `IReadOnlyDictionary` but there are multiple ways I could achieve it and I am not sure which one is best:

1/ Add only one test for each type (total new tests = 2).

2/ Add one test per assertion method for each type (total new tests = 2 * number of assertions).

3/ Duplicate the test class for the other 2 interfaces.

4/ (Maybe more complex) Create some helper method that takes a `Dictionary`, an act `Action` and an assert `Action` and runs them onto the `Dictionary` as itself, as `IDictionary`, as `IReadOnlyDictionary` and as `IImmutableDictionary`.

Any preference?
@dennisdoomen, @jnyrup

### DIFF
--- a/Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
@@ -15,9 +15,9 @@ namespace FluentAssertions.Collections
     /// </summary>
     [DebuggerNonUserCode]
     public class GenericDictionaryAssertions<TKey, TValue> :
-        ReferenceTypeAssertions<IDictionary<TKey, TValue>, GenericDictionaryAssertions<TKey, TValue>>
+        ReferenceTypeAssertions<IEnumerable<KeyValuePair<TKey, TValue>>, GenericDictionaryAssertions<TKey, TValue>>
     {
-        public GenericDictionaryAssertions(IDictionary<TKey, TValue> dictionary) : base(dictionary)
+        public GenericDictionaryAssertions(IEnumerable<KeyValuePair<TKey, TValue>> keyValuePairs) : base(keyValuePairs)
         {
         }
 
@@ -44,7 +44,7 @@ namespace FluentAssertions.Collections
                     .FailWith("Expected {0} item(s){reason}, but found {1}.", expected, Subject);
             }
 
-            int actualCount = Subject.Count;
+            int actualCount = CountSubjectItems();
 
             Execute.Assertion
                 .ForCondition(actualCount == expected)
@@ -74,7 +74,7 @@ namespace FluentAssertions.Collections
                     .FailWith("Expected {context:dictionary} to not have {0} item(s){reason}, but found <null>.", unexpected);
             }
 
-            int actualCount = Subject.Count;
+            int actualCount = CountSubjectItems();
 
             Execute.Assertion
                 .ForCondition(actualCount != unexpected)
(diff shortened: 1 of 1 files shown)


jnyrup:
Is the intention of this PR to cause less breaking changes than #963?

This PR remaps `IEnumerable<KeyValuePair<TKey, TValue>>` 
from `GenericCollectionAssertions<KeyValuePair<Tkey, TValue>>`
to `GenericDictionaryAssertions<TKey, TValue>`.

So this e.g. breaks
```c#
MyEnumerable.Should().HaveElementAt(0, myItem);
```

dennisdoomen:
> I was also thinking that another way to handle this was to keep this implementation and to provide all assertions from `GenericCollectionAssertions` to the `GenericDictionaryAssertions`.

That could work indeed and prevent breaking changes.

jnyrup:
I think I quite like this approach as it:
* Adds support for `IReadOnlyDictionary`
* Keeps support for `IDictionary`
* With `GenericDictionaryAssertions` inheriting from `GenericCollectionAssertions` has minimal impact for `IEnumerable<KeyValuePair>`

If we further redefine `GenericCollectionAssertions` as
```c#
public class GenericCollectionAssertions<T> :
    GenericCollectionAssertions<T, GenericCollectionAssertions<T>>

public class GenericCollectionAssertions<T, TAssertions> :
    SelfReferencingCollectionAssertions<T, TAssertions>
    where TAssetions : GenericCollectionAssertions<T, TAssertions>
```

`GenericDictionaryAssertions` will be able to inherit a lot of assertions while still returning `AndConstraint<GenericDictionaryAssertions>`. (only 15 tests fail, and that because of slightly different failure messages)

Evangelink:
ALright! I'll work on making this change happen then!

jnyrup:
@Evangelink Hold your horses. 
This PR is competing with #963 and I would like to avoid you or @krajek working needlessly on something that will be abandoned. That would be a waste of your valuable contributions.

jnyrup:
@Evangelink I've chatted with the others and please continue this approach. 
Regarding tests, I'm very satisfied if we cover each line in `CountSubjectItems` etc.
No need to duplicate each and every test.

jnyrup:
Fiddled a bit around with this PR in https://github.com/jnyrup/fluentassertions/commit/6773886bdf8247955af39eb08fba7b0aefb9bedb
* Let `GenericDictionaryAssertions` derive from `GenericCollectionAssertions` to inherit all counting related assertions for free.
* Abstract `IEnumerable<T>` into `TCollection : IEnumerable<KeyValuePair<K, V>>` such that assertions on an `IDictionary<K, V>` still returns an `IDictionary<K, V` instead of `IEnumerable<KeyValuePair<K, V>>`.

dennisdoomen:
> Fiddled a bit around with this PR in [jnyrup@6773886](https://github.com/jnyrup/fluentassertions/commit/6773886bdf8247955af39eb08fba7b0aefb9bedb)

What are the pros and cons compared to the other two approaches?

jnyrup:
It's a continuation of this PR, so the base idea of extending on `IEnumerable<KeyValuePair<K, V>>` instead of `IDictionary<K, V>` is still the same, but taken one step further by abstracting `IEnumerable<KeyValuePair<K, V>>` in to `TCollection : IEnumerable<KeyValuePair<K, V>>`.
This way the return type of `dictionary.Should().Subject` will still be `IDictionary<K, V>` instead of "only" `IEnumerable<KeyValuePair<K, V>>`.

Pros:
* `[Not]HaveCount`, `[Not]BeEmpty` are inherited from `GenericCollectionAssertions`.
* Retains the specific type of dictionary instead of just `IEnumerable<KeyValuePair<K, V>>.
* Let `[Not]Equal` take an `TCollection`, such that it supports both `IDictionary` and `IReadOnlyDictionary`.

Cons:
* ?

PR #1134, opened 2019-09-09
Add StringAssertions::NotBeEquivalentTo

### LINKED ISSUE #997: Missing StringAssertions::NotBeEquivalentTo
Is there any reason why `StringAssertions::NotBeEquivalentTo` does not exist? It would seem an useful addition to the current library and should be easy to implement (it could simply be the negation of BeEquivalentTo).

Or is it simply a case of "nobody has implemented it yet"? I wouldn't mind providing a PR for the functionality in this case.

danielmpetrov:
Closes #997


Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -100,6 +100,38 @@ public AndConstraint<StringAssertions> BeOneOf(IEnumerable<string> validValues,
...
+            using (var scope = new AssertionScope())
+            {
+                Subject.Should().BeEquivalentTo(unexpected, because, becauseArgs);
+                notEquivalent = scope.Discard().Any();
+            }
+
+            Execute.Assertion
+                .ForCondition(notEquivalent)
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected {context:string} not to be equivalent to {0}, but they are.", unexpected);
```
jnyrup: Use `{0}{reason}, but they are` to include the `because` and `becauseArgs` in the final failure message.

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -2043,6 +2043,138 @@ public void When_the_actual_string_equivalent_to_the_expected_but_with_trailing_
...
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action action = () => actual.Should().NotBeEquivalentTo(unexpected);
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            action.Should().Throw<XunitException>()
```
jnyrup: At least one test that call `Throw<XunitException>()` needs to verify that `because` and `becauseArgs` are included in the failure message.

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -2043,6 +2043,138 @@ public void When_the_actual_string_equivalent_to_the_expected_but_with_trailing_
...
+#endif
+        }
+
+        [Fact]
+        public void When_strings_differ_other_than_by_case_it_should_not_throw()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => "ADC".Should().NotBeEquivalentTo("abc", "we will test {0} + {1}", 1, 2);
```
jnyrup: When calling `NotThrow` you can omit `because` and `becauseArgs`.
(I've read that you love to delete code)

danielmpetrov: Fair point. Since I copied all the StringAssertions::BeEquivalentTo test and inverted them, things got left over. 👍

Src/FluentAssertions/Primitives/StringAssertions.cs
danielmpetrov: Thanks for letting me know, still learning the source. 👍

danielmpetrov:
> Thanks for taking up this task!
> I've only got a few things before everything looks good to me.
> If any of my comments are unclear, don't hesitate asking.

Thanks for the review.

Since it's getting late now, I'll resolve these tomorrow and let you know.

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -2043,6 +2043,138 @@ public void When_the_actual_string_equivalent_to_the_expected_but_with_trailing_
...
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_non_null_string_is_expected_to_be_equivalent_to_null_it_should_not_throw()
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => "ABCDEF".Should().NotBeEquivalentTo(null);
```
dennisdoomen: Is this a valid case? Why would you want to compare to `null`?

danielmpetrov: In practice you're right. I don't see why anyone would want to use this method instead of `NotBeNull()` in this scenario, but the API allows it, so might as well test for it.

If this case is not valid, then the one in the opposite direction is not valid too. https://github.com/fluentassertions/fluentassertions/blob/6dd60e7eba9d17c296e6901a9731d9fdd427bbce/Tests/Shared.Specs/StringAssertionSpecs.cs#L1954

jnyrup: I would like to keep this, at least to test that no `NullReferenceException` is thrown.

PR #1138, opened 2019-09-15
Add StringCollectionAssertions::ContainMatch

### LINKED ISSUE #482: Check if a string collection contains a wildcard match
I would like to be able to check if a string in a collection matches a wildcard pattern:
`(new string[] {"build succeded", "test failed"}).Should().HaveMatch("* failed")` or similar

danielmpetrov:
Closes #482

It's an old one, but better late than never, _right_?


Src/FluentAssertions/Collections/StringCollectionAssertions.cs
```diff
@@ -190,5 +191,50 @@ public AndConstraint<StringCollectionAssertions> Contain(IEnumerable<string> exp
...
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<StringCollectionAssertions> ContainMatch(string wildcardPattern, string because = "",
+            params object[] becauseArgs)
+        {
+            if (Subject is null)
```
dennisdoomen: You can use the `ForCondition` instead of the `if`. See https://github.com/fluentassertions/fluentassertions/blob/af2830f66184d6e53be78b3a0e858ac3536904ee/Src/FluentAssertions/Equivalency/GenericDictionaryEquivalencyStep.cs#L188

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
```diff
@@ -190,5 +191,50 @@ public AndConstraint<StringCollectionAssertions> Contain(IEnumerable<string> exp
...
+        /// The wildcard pattern with which the subject is matched, where * and ? have special meanings.
+        /// </param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<StringCollectionAssertions> ContainMatch(string wildcardPattern, string because = "",
```
jnyrup: ❓ Should we instead of returning an `AndConstraint` return an `AndWhichConstraint`, such that we include the matched `item`?
Downside: If there are multiple matches, we would only return the first match.

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
danielmpetrov: I looked into the example you provided, but I'm not sure I'm understanding. We want the method to return early, I still need the `if` clause.

jnyrup: You should be able to write something like this
```c#
Execute.Assertion
    .ForCondition(Subject is object)
    .FailWith("Expected {context:collection} to contain a match of {0}{reason}, but found <null>.", wildcardPattern);
```

If the condition in `ForCondition` fails, then `FailWith` throws a test framework exception (for the FA tests an `XunitException`)

danielmpetrov: @jnyrup That I understood, but if there is no early return there, the method execution continues, and the tests fail.

jnyrup: What test fails? 
I just tried to:
* Check out this branch
* Replace the `if` with inner `return` to the code above
* All tests succeeded.

danielmpetrov: I'm sorry, can you post the full snippet. The only way I can think of is saving the `isMatch` to a method scoped variable and adding another `ForCondition` in the final `Assertion`.

jnyrup: You can replace
```c#
if (Subject is null)
{
	Execute.Assertion
		.FailWith("Expected {context:collection} to contain a match of {0}{reason}, but found <null>.", wildcardPattern);

	return new AndConstraint<StringCollectionAssertions>(this);
}
```
with
```c#
Execute.Assertion
    .ForCondition(Subject is object)
    .FailWith("Expected {context:collection} to contain a match of {0}{reason}, but found <null>.", wildcardPattern);
```

The part below using `isMatch` looks great 👍

danielmpetrov: Oh, I saw my mistake. I had put `.ForCondition(Subject is null)` because that was the `if` condition. Got a bit mislead there, `ForCondition` takes the opposite...

Thanks for the help.

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
```diff
@@ -190,5 +191,50 @@ public AndConstraint<StringCollectionAssertions> Contain(IEnumerable<string> exp
...
+            if (Subject is null)
+            {
+                Execute.Assertion
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith("Expected {context:collection} to contain a match of {0}{reason}, but found <null>.", wildcardPattern);
+
+                return new AndConstraint<StringCollectionAssertions>(this);
+            }
+
+            using (var scope = new AssertionScope())
```
jnyrup: I just realized that this could be written as:
```c#
public AndConstraint<StringCollectionAssertions> ContainMatch(string wildcardPattern, string because = "",
    params object[] becauseArgs)
{
    Execute.Assertion
        .ForCondition(Subject is object)
        .FailWith("Expected {context:collection} to contain a match of {0}{reason}, but found <null>.", wildcardPattern)
        .Then
        .ForCondition(ContainsMatch(wildcardPattern))
        .FailWith("Expected {context:collection} {0} to contain a match of {1}{reason}.", Subject, wildcardPattern);

    return new AndConstraint<StringCollectionAssertions>(this);
}

private bool ContainsMatch(string wildcardPattern)
{
    using (var scope = new AssertionScope())
    {
        foreach (string item in Subject)
        {
            item.Should().Match(wildcardPattern);
            bool isMatch = !scope.Discard().Any();
            if (isMatch)
            {
                return true;
            }
        }
    }

    return false;
}
```

That avoids the extra `return` statement in the bottom, which is effectively dead-code as it will never be hit.

Tests/Shared.Specs/GenericCollectionAssertionOfStringSpecs.cs
```diff
@@ -2223,5 +2223,128 @@ public void When_using_StringCollectionAssertions_the_AndConstraint_should_have_
...
+        {
+            //-----------------------------------------------------------------------------------------------------------
+            // Arrange
+            //-----------------------------------------------------------------------------------------------------------
+            IEnumerable<string> collection = null;
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action action = () => collection.Should().ContainMatch("* failed");
```
jnyrup: Please add `because` and `becauseArgs` to this test as well, as it hits another `FailWith` in the assertion implementation.

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
danielmpetrov: That's great, I was thinking of something similar with extracting the private method, but your version is more succinct.

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
danielmpetrov: That's a good point. 

Another possible alternative is to return all matches.

The downside to that is that the whole string enumerable will have to be iterated, instead of returning at the first match.

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
jnyrup: I just updated the example above to use `Any` instead in the `ContainsMatch` helping method.

danielmpetrov: Even better.

If this gets any shorter, I'd have to rename the file to `StringCollectionAssertions.py` 😄

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
jnyrup: Aha! 
`SelfReferencingCollectionAssertions.Contain(Expression<Func<T, bool>> predicate` returns an `AndWhichConstraint`.
https://github.com/fluentassertions/fluentassertions/blob/dbc0d5f6e301f05b2b096e2bb29414c8cf66c546/Src/FluentAssertions/Collections/SelfReferencingCollectionAssertions.cs#L461-L466

If invoking `.Which` on the returned `AndWhichConstraint` it *then* checks if there was more than one result.

danielmpetrov: Should I provide an example in the docs about `.Which` if this is the accepted API?

danielmpetrov:
Any additional concerns or comments in order to wrap this up? @jnyrup @dennisdoomen

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
jnyrup: I'm sorry, I forgot to include the complete desired signature.
My thought was that the return type should be `AndWhichConstraint<StringCollectionAssertions, string>`. 
From what I could see, that's whats the most places in FA uses.

danielmpetrov: So return first match only?

jnyrup: Yes.
That seems most consistent with the API as that's what all other assertions use, at least as far as I could see.
Any input @dennisdoomen?

So 
```c#
return new AndWhichConstraint<StringCollectionAssertions, string>(this, AllThatMatch(wildcardPattern));
```

danielmpetrov: If that is the case, do you prefer implementation A or B:

A)
```csharp
string match;
using (var scope = new AssertionScope())
{
    match = Subject?.FirstOrDefault(item =>
    {
        item.Should().Match(wildcardPattern);
        return !scope.Discard().Any();
    });
}

Execute.Assertion
    .ForCondition(Subject is object)
    .FailWith("Expected {context:collection} to contain a match of {0}{reason}, but found <null>.", wildcardPattern)
    .Then
    .ForCondition(match is object)
    .FailWith("Expected {context:collection} {0} to contain a match of {1}{reason}.", Subject, wildcardPattern);

return new AndWhichConstraint<StringCollectionAssertions, string>(this, match);
```

B)
```csharp
public AndWhichConstraint<StringCollectionAssertions, string> ContainMatch(string wildcardPattern, string because = "",
    params object[] becauseArgs)
{
    Execute.Assertion
        .ForCondition(Subject is object)
        .FailWith("Expected {context:collection} to contain a match of {0}{reason}, but found <null>.", wildcardPattern)
        .Then
        .ForCondition(ContainsMatch(wildcardPattern))
        .FailWith("Expected {context:collection} {0} to contain a match of {1}{reason}.", Subject, wildcardPattern);

    return new AndWhichConstraint<StringCollectionAssertions, string>(this, FirstMatch(wildcardPattern));
}

private bool ContainsMatch(string wildcardPattern)
{
    using (var scope = new AssertionScope())
    {
        return Subject.Any(item =>
        {
            item.Should().Match(wildcardPattern);
            return !scope.Discard().Any();
        });
    }
}

private string FirstMatch(string wildcardPattern)
{
    return Subject.FirstOrDefault(item =>
    {
        using (var scope = new AssertionScope())
        {
            item.Should().Match(wildcardPattern);
            return !scope.Discard().Any();
        }
    });
}
```

danielmpetrov: @jnyrup Oh, I just realized if the `AndWhichConstraint` is passed an enumerable with more than 1 element, when accessing `Which` FA throws an exception stating `More than one object found. FluentAssertions cannot determine which object is meant.` Is this the desired behavior you're after?

jnyrup: Yes.
Again sorry for the confusion, I somehow thought you had seen that.

danielmpetrov: Haha, no worries, it's part of learning the framework.

Since we were talking about 'first match only' I had assumed first match, discard other matches. Did not consider throwing an exception, but it seems like this is the default behavior for `.Which`

dennisdoomen:
> Any additional concerns or comments in order to wrap this up?

Not from me. @jnyrup has been thorough enough 😀

jnyrup:
Thanks @danielmpetrov for taking the time to implement this and the additional changes 👍

PR #1140, opened 2019-09-18
Continuation of PR #989

### LINKED ISSUE #296: Collection Assertion NotBeInAscendingOrder with lambda not available but documentation says it is

danielmpetrov:
This PR is an attempt to close out #989.

Changelog since the original PR:
- Update docs
- Add more tests based on https://github.com/fluentassertions/fluentassertions/pull/989#discussion_r250120652 by @jnyrup
- The behavior implemented in regards to empty/single element collections is as suggested by @jnyrup and agreed on by @dennisdoomen (see below)

> I came to think of two inherent properties of empty/single element collections.
> They are:
> 
> - always ordered, and
> - both in ascending and descending order at the same time.

- Organized the tests for _Be In Ascending/Descending Order_ in `GenericCollectionAssertionsSpecs.cs` into sub-regions for better navigation, since it was hard to find my way around. I kept the original tests, and added on a few more as stated earlier, but please check them that they make sense. 👍 

Closes #296 and Resolves #989


Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -936,9 +936,26 @@ public AndConstraint<TAssertions> BeInDescendingOrder(IComparer<object> comparer
...
+        /// using their <see cref="IComparable.CompareTo(object)" /> implementation.
+        /// </summary>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> NotBeInAscendingOrder(string because = "", params object[] becauseArgs)
```
dennisdoomen: I think we should extend the docs for all the new methods to include the "rules" for empty and single-element collections in the `<remarks>`. 

E.g., today, a colleague asked me what happens if you call `collection.Should().OnlyContain` on an empty collection. I had to look at the implementation to find the answer.

docs/_pages/collections.md
```diff
@@ -80,8 +80,10 @@ collection.Should().NotIntersectWith(anotherCollection);
 
 collection.Should().BeInAscendingOrder();
 collection.Should().BeInDescendingOrder();
-collection.Should().NotBeAscendingInOrder();
-collection.Should().NotBeDescendingInOrder();
+// collection.Should().NotBeAscendingInOrder(); - deprecated in 5.x, to be removed in future version
```
dennisdoomen: Let's just remove these.

Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -936,9 +948,32 @@ public AndConstraint<TAssertions> BeInDescendingOrder(IComparer<object> comparer
         /// <param name="becauseArgs">
         /// Zero or more objects to format using the placeholders in <see cref="because" />.
         /// </param>
+        /// <remarks>
+        /// Empty and single element collections are considered to be ordered both in ascending and descending order at the same time.
+        /// </remarks>
+        [Obsolete("Use NotBeInAscendingOrder instead")]
```
jnyrup: For other methods marked with `Obsolete`, we also added 
```
[EditorBrowsable(EditorBrowsableState.Never)]
```
This will hide the method when using IntelliSense to browse available methods.
Could I ask you add those as well?

jnyrup:
Thanks @david-a-jetter and @danielmpetrov for contributing this improvement

PR #1141, opened 2019-09-19
Add Should().BeEnumerable() assertions

aalmada:
This PR adds a `BeEnumerable()` assertion to validate the implementation of collections and operations on collections.

A collection to be enumerable by a `foreach` loop does not have to implement any interface. It just needs to have a `GetEnumerator()` method that returns a type that has a property `Current` and a method `MoveNext()`that returns `bool`. 

Here's the [minimal implementation of an enumerable](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAgrgOwD4AEBMAjALABQZuAzAAQHUCiYADjAJ72bpiITAA2cADwAVAHxkA3mWozaNRi3aduCXgJGjqAcTgwOXHjAD2CABQBKagF5NMABYBLKAG4y02VWrDqAYXQIEOEwYa00AEzgAMwh0PhhXclJZOWpgIyM+agBZIwA3OAA5OAAPGAtQ6mi+KDgEgF83UgIaSXcZAnxqNuopJOTZSJM4CABjO1NciARqSb50OGoHTGpMOAB3BmY2fRU1ISWYUQtzbuSASFxCAE4JiDm4cwTkhtI6oA===):

```csharp
public class EmptyEnumerable<T>
{
    public EmptyEnumerable<T> GetEnumerator() => this;

    public T Current => default;

    public bool MoveNext() => false;
}
```

Here's an [implementation of an enumerable that returns values from 0 to `count`](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAgrgOwD4AEBMAjALABQZuAzAAQJwQAmA9pgDYCe1sGAxjNQFl2AJQiYA5nDIBvMtXm16zNpwCWmfjyZYYAbjkKD8qoJFjJACnWbtGgJRHqs0gtfUYAC1VQAdFp3UALzU/hr6LgoAvo6OJnSMLBzUAKKY6GCIEDBMCNQA4nAwqemZ2QgWdkEAfNSYcADuKWkZCFk5FqEwduGusTTc6HxNJa1ljs5uCvHKSdYhtnqOrnM86Ah0YUuGEZMmxS1t5SsLDjtuE5Nunt5+C0HzOj2XrqvrcBr3ALSET5PRZ70AQo4kpEmoPgBhNYbfiBGqvGG/NxbYw0YBMJisQRMABucAAcnAAB4wCrVagAagpCPe/AAPA9Nmd/v8KPgaDJYvh8NRxijqAAzHL0HgeCw4iC5CWsdBwajqWoNUyiCRwCyEAAMdlOzwAkLhCABOcUQGVwbqOFlAA):

```csharp
public readonly struct MyRange
{
    readonly int count;
    
    public MyRange(int count)
    {
        this.count = count;
    }
    
    public readonly Enumerator GetEnumerator() => new Enumerator(count);
    
    public struct Enumerator
    {
        readonly int count;
        int current;
        
        public Enumerator(int count)
        {
            this.count = count;
            current = -1;
        }
        
        public readonly int Current => current;
        
        public bool MoveNext() => ++current < count;
    }
}
```

_NOTE: The advantage here is performance. The enumerator is a value type so the method calls are not virtual. The use of interfaces would box the enumerator and turn method calls into virtual. The enumerator is not disposable, so the `foreach` does not generate a `try`/`finally` clause, making it inlinable. The example also uses C# 8 'struct read-only members' feature to avoid defensive copies._

FluentAssertions implements collection assertions in `NonGenericCollectionAssertions` and `GenericCollectionAssertions<>` which can only handle collections that implement the interfaces `IEnumerable` or `IEnumerable<T> `. 

These also only test one form of enumerating the collection. This is fine when assuming that the collection is well implemented, but not when testing its implementation. For example, a collection that implements `IReadOnlyList<T>` can be enumerated using the indexer, using `IEnumerable<T>.GetEnumerator()`, using `IEnumerable.GetEnumerator()` and using a public `GetEnumerator()` that is not an override of any of these interfaces. There's no guarantee that they all are correctly implemented. The `Count` property can also return the wrong value.

Here's an example of [a collection with multiple possible enumerations](https://sharplab.io/#v2:EYLgtghgzgLgpgJwD4AEBMBGAsAKBQBgAIUMA6AYQHsAbauAYxgEtKA7KAblwOIwBYuObgGZCCOBAAmbagE9CsBAFdGhALKyAShFYBzOIRCEAkpomSA8qzkAZJrAA8TVjAB8uAN6FvhXD+KiGtp6cAAUzjCE9JRKLgCUfj4eif7eVLGRALxRMS6C/gC+Kb44/iii4lIy8hGE6S6EXvowHIRFpT7F5YS1MAAW9gDatc6ScAAeALrFyR3eAJDNxf6zqWs9AGaE4axj44QOhERISD27E4Su2fUwCXPrawAqfQiUAO7G5+MWSjAWG8F9ABRcb0OAAB2YbFCcUEy3WKAA7Gc9nD7g8UHxCM9Xh8vj8/gCdMDQRCoawYYRMq5CP1cYRWHA3iZ8b9/oC4CCwZCWBTYfDvO1CsUuhVzNVCEDWEowIgIDBKAhCABxOAwKUyuUKhCU6kMpmS6WyhDyxWhG78+6VaTWeTGDXG00IJwuGkkYQutykVXqo1as1xKk0xnMgAi9nBlCgEGAdAd/p1FvyPmtEvtfpN2t4wm9avjmYDQf1YYjUZjcYzTvNuVuye8ooUMGUqnzTpmArE4ttZ0i0QyddStXoSgQ4jyHY73Vb2p2vZrdweqwePn69lIfYa2Q3LQ7/mHo7gm8IAFoMAPhejOpfvN1U93auQR2OsjT98/z1eMaJgJQaOpKAAbnAAByEwwLqNIANSQW+h6RIc24DkKn4+OghDhlAkbRrGnKVlmRjppqBbOhE7j3Euax3nIPY5P2HZDk+cEfvW14BOhpbYRWRFVkO84dhRy6rlA641lStHjqxe6MUep7MW0E6sd0D7SS+UQqXJlDAAAVgwkSEY62oUCpRawRJDyTt+v7UP+QGgeM4GBnq0GmfB4k7qxFnEFiZhQGqEFqQeMlngpX5eexmFRmEgYeMhgq4O03BoKInhdGgaAlCsHYbIqEj0H0oQARASqFdQSgGM4xbqFoxJhBg+BxAu6zzCQACcBUQKVcCWheBRAA):

```csharp
public readonly struct MyRange : IReadOnlyList<int>
{    
    public MyRange(int count)
    {
        Count = count;
    }
    
    public readonly int Count { get; }
    
    public int this[int index]
    {
    	get
        {
            if (index < 0 || index >= Count)
                ThrowIndexOutOfRangeException();

            return index;

            void ThrowIndexOutOfRangeException() => throw new IndexOutOfRangeException();
        }
    }
    
    public readonly Enumerator GetEnumerator() => new Enumerator(Count);
    readonly IEnumerator<int> IEnumerable<int>.GetEnumerator() => new DisposableEnumerator(Count);
    readonly IEnumerator IEnumerable.GetEnumerator() => new DisposableEnumerator(Count);
    
    public struct Enumerator
    {
        readonly int count;
        int current;
        
        public Enumerator(int count)
        {
            this.count = count;
            current = -1;
        }
        
        public readonly int Current => current;
        
        public bool MoveNext() => ++current < count;
    }
    
    class DisposableEnumerator : IEnumerator<int>
    {
        readonly int count;
        int current;
        
        public DisposableEnumerator(int count)
        {
            this.count = count;
            current = -1;
        }
        
        public int Current => current;
        object IEnumerator.Current => current;
        
        public bool MoveNext() => ++current < count;
        
        public void Reset() => current = -1;
        
        public void Dispose() {}
    }
}
```

_NOTE: The indexer uses a local function so that the accessor does not throw an exception, making it inlinable. The local function does not add dependencies to the example... ;)_

It has two enumerators to take advantage of the performance features described above but also be casted to an interface and take advantage of extension methods for collections (like LINQ). It can also be enumerated using the indexer.

New interfaces can add even more possible enumerations. Here's an interface that restricts the enumerator to a value type and adds a `GetEnumerator()` that explicitly returns the enumerator type. This allows the use of an enumerable interface without boxing the enumerator:

```csharp
public interface IValueEnumerable<T, TEnumerator>
    : IEnumerable<T>
    where TEnumerator : struct, IEnumerator<T>
{
    new TEnumerator GetEnumerator();
}
```

Here's a possible [implementation](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAgrgOwD4AEBMAjALABQuADAAS6EB0AwgPYA2LcAxjAJZOZQBuMpRqEALENLCAzFW6Z4CAGYQOcKgEkAahBbo4AUUzowiCMHYAeACpIq1oybMwmCAHxkAkCFHSbH0k8AdwALRHUHY1MECBcEKh9YDC47DUdo2Nd/MgBvL0w4IPt051cqAHE4GBKYuIAKAEpJAF8yGSoEOAgAEz4WAE8qJPQuKgBZfoAlCEwAc3UyKiWEzR09QyizCzhLeRg7Cem5uDoAEW4oAAcmKHN2Gsz3XOWqL06evsG9qg4mLBhJC9FstcLJDjN5nVvr9/g1gUs8qQXi8YCELnQYQoqABeH5/BSA5atJHIpbwmiyd69TADKgPOIVKr01yNHFuKgFIrMhB1TEwJrkqmfKjnK43O4bJy1MraXT6B7bXYKOyi663bbctx0SrVTbSnkNNkcwoii5qiXc3n4/mEpZCmmDNJ6x5KmDs2h+PZanWWw3Y9mc01i9X3Z31PkCkl2rrU2lOqWPXzaplhll+gMm1XijWpnkR22vKMUoYwZIwOm58mI0nRj4OuRYvkFl7Q9AIToE8lAosvUEVhP1aHWuE95bVmsotFQDHWnF4/7NmscNsd8u4gC0hEXS2JE67IMpMeF3wYK7gWP9PzPndHZNvxeATFY4yYADc4AA5OAADxgrMvADUAHLu257lpY843i8u7LOSfbDKMWYhpKGQMj48aoVkXpVvutaxl8jbWtuDbliBq7EbhxZIRauZQoRsKUeOE5LKi6J8nOTaUS8ZFgXOm7ETBpKUX2J7Xmu7I8VBNZMMAABWnDlhhpQIIwYlGpJAKUcJsiPs+Yxvp+P5/umVBARpVAQZx97aTQYhUJMcBQFU/4SWpG5blp959rgdlIXArI5IJVDEsSwj4LIuRwfg+CFi8THIkorhdBwIR1K+EDxOl6wNsaRTgscdSEBQDQjhOni0AAnGlcpwJG0FkM0QA):

```csharp
public readonly struct MyRange 
    : IValueEnumerable<int, MyRange.DisposableEnumerator>
{    
    readonly int count;
    
    public MyRange(int count)
    {
        this.count = count;
    }
        
    public readonly Enumerator GetEnumerator() => new Enumerator(count);
    readonly DisposableEnumerator IValueEnumerable<int, DisposableEnumerator>.GetEnumerator() => new DisposableEnumerator(count);
    readonly IEnumerator<int> IEnumerable<int>.GetEnumerator() => new DisposableEnumerator(count);
    readonly IEnumerator IEnumerable.GetEnumerator() => new DisposableEnumerator(count);
    
    public struct Enumerator
    {
        readonly int count;
        int current;
        
        public Enumerator(int count)
        {
            this.count = count;
            current = -1;
        }
        
        public readonly int Current => current;
        
        public bool MoveNext() => ++current < count;
    }
    
    public struct DisposableEnumerator : IEnumerator<int>
    {
        readonly int count;
        int current;
        
        public DisposableEnumerator(int count)
        {
            this.count = count;
            current = -1;
        }
        
        public int Current => current;
        object IEnumerator.Current => current;
        
        public bool MoveNext() => ++current < count;
        
        public void Reset() => current = -1;
        
        public void Dispose() {}
    }
}
```

This new assertion should be able to handle all these possible cases...


aalmada:
Current implementation can assert if a `object` is enumerable and if, when enumerated directly by a `foreach` loop, returns the expected sequence of items. It includes unit tests for all these cases.

It's still missing validation of enumeration using other possible enumerations not used directly by `foreach`.

jnyrup:
The overload
```c#
BeEnumerable(IEnumerable expected, Func<object, object, bool> equalityComparison, ...)
```
doesn't read very fluently to me.
Probably because it does two thing at a time:
* Verifies that the `Subject` is enumerable, and
* compares it with the provided `expected`.

Would the following be possible?
A single overload, that verified enumerability and on success returned a regular `IEnumerable` for further assertions? 
```c#
public AndConstraintWhich<ObjectAssertions, IEnumerable> BeEnumerable(string because = "", params object[] becauseArgs)
```

Such that one could write:
```c#
object someObject = new RangeEnumerable(5);

someObject.Should().BeEnumerable()
    .Which.Should().Equal(new[] { 0, 1, 2, 3, 4 }, (a, e) => int.Equals(a, e));
```

aalmada:
Hi @jnyrup. Thanks for the feedback!

I've already been wondering how I could support continuation as an `IEnumerable`. I didn't know about `AndWhichConstraint<,>`. I already [made the change](https://github.com/aalmada/fluentassertions/blob/e2a015147761aeb10129e35f50a6104ba0a1f025/Src/FluentAssertions/Primitives/ObjectAssertions.cs#L267) but it's missing the unit tests.

If I understand correctly your code example for `Which`, it will use [`CollectionAssertions.Equal(IEnumerable)`](https://github.com/fluentassertions/fluentassertions/blob/a139144b47ae3b6a9341f1ca856108e0887b7f9c/Src/FluentAssertions/Collections/CollectionAssertions.cs#L219). This ends up converting the original enumerable into a [`List<TActual>`](https://github.com/fluentassertions/fluentassertions/blob/a139144b47ae3b6a9341f1ca856108e0887b7f9c/Src/FluentAssertions/Common/EnumerableExtensions.cs#L17), its content compared using [`IndexOfFirstDifferenceWith()`](https://github.com/fluentassertions/fluentassertions/blob/a139144b47ae3b6a9341f1ca856108e0887b7f9c/Src/FluentAssertions/Common/EnumerableExtensions.cs#L39), which only uses `IEnumerable<>.GetEnumerator()`to get its content. I want to assert all possible forms of enumeration.

I see two possible solutions:
1. Develop another assertion but I don't know what name I should give to it.
2. Change `CollectionAssertions.Equal(IEnumerable)` so that it tests all possible enumerations, making it perform slower. I would also need to add a `ObjectAssertion.Equal(object, IEnumerable)` to support collections that don't implement any know enumerable interface.

Src/FluentAssertions/Primitives/ObjectAssertions.cs
```diff
@@ -251,6 +254,139 @@ public AndConstraint<ObjectAssertions> NotBe(object unexpected, string because =
...
+                }
+
+                AssertIsEnumerable(Subject, because, becauseArgs);
+
+                var subjectType = Subject.GetType();
+
+                // assert equality using the public GetEnumerator()
+                AssertEquality(Subject, subjectType, expected, equalityComparison, because, becauseArgs);
+
+                // assert equality using the GetEnumerator() of implemented enumerable interfaces
```
dennisdoomen: I wasn't expecting this method to enumerate through all supported interfaces, but I do realize the confusion when an object has multiple ways of enumerating. Not sure yet how to deal with this. cc @jnyrup

Src/FluentAssertions/Primitives/ObjectAssertions.cs
```diff
@@ -251,6 +254,139 @@ public AndConstraint<ObjectAssertions> NotBe(object unexpected, string because =
...
+                foreach (var @interface in subjectType.GetInterfaces())
+                {
+                    if (@interface.IsEnumerable())
+                        AssertEquality(Subject, @interface.GetType(), expected, equalityComparison, because, becauseArgs);
```
dennisdoomen: Please use explicit curly braces.

Src/FluentAssertions/Primitives/ObjectAssertions.cs
```diff
@@ -251,6 +254,139 @@ public AndConstraint<ObjectAssertions> NotBe(object unexpected, string because =
...
+               }
+            }
+
+            return new AndConstraint<ObjectAssertions>(this);
+        }
+
+        internal static void AssertIsEnumerable(object subject, string because = "", params object[] becauseArgs)
```
dennisdoomen: Why are these methods `internal`?

Src/FluentAssertions/Primitives/ObjectAssertions.cs
```diff
@@ -251,6 +254,139 @@ public AndConstraint<ObjectAssertions> NotBe(object unexpected, string because =
...
+        /// Zero or more values to use for filling in any <see cref="string.Format(string,object[])" /> compatible placeholders.
+        /// </param>
+        public AndWhichConstraint<ObjectAssertions, IEnumerable> BeEnumerable(string because = "", params object[] becauseArgs)
+        {
+            AssertIsEnumerable(Subject, because, becauseArgs);
+
+            return new AndWhichConstraint<ObjectAssertions, IEnumerable>(this, new ObjectEnumerable(Subject));
+        }
+
+        public AndConstraint<ObjectAssertions> BeEnumerable(IEnumerable expected, Func<object, object, bool> equalityComparison, string because = "", params object[] becauseArgs)
```
dennisdoomen: Same remark as @jnyrup made: drop the `expected` parameter and return an `AndWhichConstraint` instead

Src/FluentAssertions/TypeExtensions.cs
```diff
@@ -68,5 +68,71 @@ public static PropertyInfoSelector Properties(this TypeSelector typeSelector)
         {
             return new PropertyInfoSelector(typeSelector.ToList());
         }
+
+        public static bool IsEnumerable(this Type type)
```
dennisdoomen: Keep all these methods `internal` please.

jnyrup:
I don't think we should try to verify all ways you can iterate an enumerable.
For `IDictionary` we could either use `ContainsKey+Indexer` or `TryGetValue` to safely access a value, but when asserting that
```c#
dict.Should().ContainKey(key)
    .WhichValue.Should().Be(42);
```
it should be of no concern to neither the user nor us how we access the value.

Of course to reliably use a custom enumerable like these, I would test the individual ways explicit.

I see it as tests of two different concerns:
* Testing that the data structure. 
   * It works consistently no matter how it is enumerated.
* Testing that functions *using* the data structure works. 
  * For `MyRange<int> Foo()` testing that `Foo()` returns an instance containing `1, 2, 3`.


I hope this makes any sense, otherwise don't hesitate asking for clarification.

dennisdoomen:
Maybe we can talk the following approach:
* If an object implements `IEnumerable`, use its `GetEnumerator` method
* If it doesn't, see if it has a direct `GetEnumerator` method, and use that one
* Otherwise...fail

aalmada:
Let me give you an example from _corefx_. The collection `System.Collections.Immutable.ImmutableArray<>` [returns two types of enumerators](https://github.com/aalmada/corefx/blob/master/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray_1.Enumerator.cs), just like some of the examples I give in the description. One is a value-type and the other is a reference-type. [Notice that its test calls the two GetEnumerator()](https://github.com/aalmada/corefx/blob/master/src/System.Collections.Immutable/tests/ImmutableArrayTest.cs#L653) and is a lot less readable than a simple `source.Should().BeEnumerable(expected)`.

I implemented dozens of enumerators for [my own project](https://github.com/NetFabric/NetFabric.Hyperlinq). I don't want to copy/paste unit tests and end up having to test them too.

I understand that most people don't need this assertion but, for the one that do, it saves a lot of time.

Src/FluentAssertions/Common/EnumerableExtensions.cs
```diff
@@ -30,18 +31,77 @@ public static IList<T> ConvertOrCastToList<T>(this IEnumerable source)
...
+
+            Type enumeratorType = getEnumerator.ReturnType;
+            PropertyInfo current = enumeratorType.GetPublicOrExplicitProperty("Current");
+            MethodInfo moveNext = enumeratorType.GetPublicOrExplicitParameterlessMethod("MoveNext");
+            MethodInfo dispose = enumeratorType.GetPublicOrExplicitParameterlessMethod("Dispose");
+
+#if !NETSTANDARD2_1
+            // 'Current' may return by-ref but reflection only supports its invocation on netstandard 2.1
+            if (current.PropertyType.IsByRef)
+                return -1; // what should we do here?????
```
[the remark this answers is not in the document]
aalmada: It is possible to implement an enumerator that returns a reference to the item. [I do that in my project](https://github.com/NetFabric/NetFabric.Hyperlinq/blob/master/NetFabric.Hyperlinq/Filtering/Where/Where.Array.cs#L62). The issue is that, I'm using reflection for this assertion and reflection only started supporting invocation of reference returning methods with _netstandard 2.1_. The only option is to throw an exception saying that the feature is not supported (that's what reflection does anyway if this condition is removed). What's the protocol in FluentAssertions for this?

Src/FluentAssertions/Common/EnumerableExtensions.cs
```diff
@@ -30,18 +31,77 @@ public static IList<T> ConvertOrCastToList<T>(this IEnumerable source)
...
+                    {
+                        bool isFirstCompleted = !(bool)moveNext.Invoke(firstEnumerator, new object[0]);
+                        bool isSecondCompleted = !secondEnumerator.MoveNext();
+
+                        if (isFirstCompleted && isSecondCompleted)
+                        {
+                            return -1;
+                        }
+
+                        if (isFirstCompleted ^ isSecondCompleted)
```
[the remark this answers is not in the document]
aalmada: To be honest, I copied it from already existing code in this same file. I was also surprised when I first saw it. ;)

Src/FluentAssertions/Common/EnumerableExtensions.cs
```diff
@@ -30,18 +31,77 @@ public static IList<T> ConvertOrCastToList<T>(this IEnumerable source)
...
+        }
+
+        /// <summary>
+        /// Searches for the first different element in two sequences using specified <paramref name="equalityComparison" />
+        /// </summary>
+        /// <param name="first">The first sequence to compare.</param>
+        /// <param name="second">The second sequence to compare.</param>
+        /// <param name="equalityComparison">Method that is used to compare 2 elements with the same index.</param>
+        /// <returns>Index at which two sequences have elements that are not equal, or -1 if enumerables are equal</returns>
+        public static int IndexOfFirstDifferenceWith(this IEnumerable first, IEnumerable second, Func<object, object, bool> equalityComparison)
```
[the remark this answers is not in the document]
aalmada: I changed strategy while developing this PR but I was afraid I would end up in a dead end, so I left these to clean up a bit later.

Src/FluentAssertions/Execution/GivenSelectorExtensions.cs
```diff
@@ -60,6 +62,15 @@ public static ContinuationOfGiven<ICollection<T>> AssertCollectionHasEnoughItems
                 .AssertCollectionHasNotTooManyItems(length);
         }
 
+        public static ContinuationOfGiven<IEnumerable> AssertForEachEnumerablesHaveSameItems(
+            this GivenSelector<object> givenSelector, IEnumerable expected, Func<object, IEnumerable, int> findIndex)
+            => givenSelector
```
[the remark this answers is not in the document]
aalmada: I can do that but I think that's highly arguable. Fluent syntax and expression-bodied format are the perfect match. Why add a `return` and a pair of brackets?

Src/FluentAssertions/Primitives/ObjectAssertions.cs
aalmada: I never use them on single lines. I was being careful to always add them in this PR but I missed this one. Doh!... ;)

Src/FluentAssertions/Primitives/ObjectAssertions.cs
aalmada: These were to be used on other assertions but I changed strategy. I'll remove it.

Src/FluentAssertions/TypeExtensions.cs
aalmada: All the other methods in this class are `public`...

Src/FluentAssertions/Execution/GivenSelectorExtensions.cs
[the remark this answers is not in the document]
dennisdoomen: It's our coding convention. Fluent Assertions follows https://csharpcodingguidelines.com/, in this case https://csharpcodingguidelines.com//layout-guidelines/#av2410

Src/FluentAssertions/Common/EnumerableExtensions.cs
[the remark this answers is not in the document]
dennisdoomen: Throwing an exception is fine. Looking at your example tells me you're doing something pretty edgy already.

Src/FluentAssertions/Common/EnumerableExtensions.cs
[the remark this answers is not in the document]
dennisdoomen: Then let's rewrite it to avoid more confusion.

dennisdoomen:
I agree with @jnyrup. There are two concerns. One, that an object is enumerable. And two, that multiple ways of enumerating all return the same result. Trying to combine the two behaviors in a single method is going to be too much magic for most devs. 

The second can also be solved using:

```csharp
var array = source.ToImmutableArray();
((IEnumerable<int>)array).Should().BeEquivalentTo(array, o => o.WithStrictOrdering);
```

Src/FluentAssertions/TypeExtensions.cs
dennisdoomen: Yeah, there's some confusion _and_ technical debt. The `TypeExtensions` in the root of the project is part of the public API of the type assertions. There's another one under `Common`, which is supposed to be used within the framework and _should_ have been ìnternal`.

aalmada:
My initial plan was to only have an assertion that would validate all possible enumerations. That's what I need for my project. While doing it, making public the enumerable check, was very simple. The second overload was born and all my troubles started... :(

dennisdoomen:
But why did you close it? You could still implement what was suggested in 
https://github.com/fluentassertions/fluentassertions/pull/1141#issuecomment-533907426

PR #1143, opened 2019-09-21
Handle comparing enumerables with null values

### LINKED ISSUE #1139: Collections containing nulls do not compare equivalent
# Description

Collections containing nulls do not always appear to compare equivalent.

```csharp
{
    var actual = new int?[] { null, 1 };
    actual.Should().BeEquivalentTo(new int?[] { 1, null }); // passes
    actual.Should().BeEquivalentTo(new int?[] { null, 1 }); // passes
}
{
    var actual = new int?[] { 1, null };
    actual.Should().BeEquivalentTo(new int?[] { 1, null }); // passes
    actual.Should().BeEquivalentTo(new int?[] { null, 1 }); // passes
}
{
    var actual = new object[] { null, 1 };
    actual.Should().BeEquivalentTo(new object[] { 1, null }); // passes
    actual.Should().BeEquivalentTo(new object[] { null, 1 }); // passes
}
{
    var actual = new object[] { 1, null };
    actual.Should().BeEquivalentTo(new object[] { 1, null }); // passes
    actual.Should().BeEquivalentTo(new object[] { null, 1 }); // FAILS
}
```

### Expected behavior:

I expect all 8 of the very similar assertions to succeed. Should I?

### Actual behavior:

8th assertion fails with: `Expected item[0] to be <null>, but found 1.`

### Versions

* Which version of Fluent Assertions are you using? - *5.9.0*
* Which .NET runtime and version are you targeting? E.g. .NET framework 4.6.1 or .NET Core 2.0. - *.Net Framework 4.7.2*

jnyrup:
Fixes #1139


Tests/Shared.Specs/BasicEquivalencySpecs.cs
```diff
@@ -3929,6 +3929,46 @@ public void When_asserting_instances_of_Object_are_equivalent_it_should_fail()
...
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Act
+            //-----------------------------------------------------------------------------------------------------------
+            Action act = () => actual.Should().BeEquivalentTo(expected);
+
+            //-----------------------------------------------------------------------------------------------------------
+            // Assert
+            //-----------------------------------------------------------------------------------------------------------
+            act.Should().Throw<XunitException>();
```
dennisdoomen: Don't we want to verify at least some part of the exception message?

PR #1145, opened 2019-09-22
Add more specific variants of contain (continued)

### LINKED ISSUE #818: Add version of StringAssertions.Contain that allows specifying the number of times a substring should occur.

danielmpetrov:
Hi,

I am attempting to finish off #962, since @mkolumb seems to be MIA.

There was quite the conversation around this change, and the agreed non-breaking API was described here - https://github.com/fluentassertions/fluentassertions/pull/962#issuecomment-451376681

Now, I picked off where the last PR was at, and changed the API to be backwards compatible as described by @jnyrup, but I need some help on what would be the best way to get good failure messages for predefined delegates.

Closes #818 and Resolves #962


Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -660,7 +660,144 @@ public AndConstraint<StringAssertions> ContainEquivalentOf(string expected, stri
...
+
+            Execute.Assertion
+                .ForCondition(Contains(Subject, expected, StringComparison.Ordinal))
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected {context:string} {0} to contain {1}{reason}, but not found.", Subject, expected)
+                .Then
+                .ForCondition(predicate(actual))
+                .BecauseOf(because, becauseArgs)
+                .FailWith(
+                    $"Expected {{context:string}} {{0}} to contain {{1}} {nameof(predicate)} {{reason}}, but found {(actual == 1 ? "1 time" : $"{actual} times")}.",
```
danielmpetrov: Instead of `nameof(predicate)` the message needs to be, for example, `at most 2 times`

dennisdoomen:
Looks promising, although I don't yet see what you intend to do with `StringContainmentConstraints`

danielmpetrov:
@dennisdoomen I intend to remove `StringContainmentConstraints.cs` and `StringWithOccurenceConstraint.cs` once the PR is ready ready to merge. Those two classes are left over from the old breaking state of this change.

Now the problem I'm having is how to get good failure messages as in the original PR.

Any ideas welcome.

danielmpetrov:
Using reflection we can read whether the passed delegate method was one of the predefined by FA options and display appropriate message.

If the user passes a lambda instead, a more generic message will be shown, such as `'expected the number of times {expected} is contained within {actual} to match custom predicate'`

Sounds good?

danielmpetrov:
@dennisdoomen @jnyrup What do you think about this implementation?

dennisdoomen:
It most definitely looks promising.

dennisdoomen:
Is there anything left before we can do a proper review?

danielmpetrov:
@dennisdoomen As far as `StringAssertions` is concerned, there's nothing left to do.

From what I remember, last I was working on this I was concerned about the possibility of making this functionality reusable as suggested by 
@jnyrup https://github.com/fluentassertions/fluentassertions/pull/962#issuecomment-450485140

I believe this should be discussed before this PR is merged.

dennisdoomen:
The `TimesConstraints` looks pretty reusable to me. It's not tied to any specific assertion method other than using hard-coded strings in its `FailWith` method. But this is something we might be able to refactor later on. Just make sure it's `internal`.

danielmpetrov:
@dennisdoomen Fair enough.

One more question that I have is in which namespace should `TimesConstraint`, `AtLeast`, `AtMost`, `MoreThan`, `LessThan`, and `Exactly` reside? Currently they're in `FluentAssertions.Primitives`, but if they'll be reused for collections as well, should they be placed in the root `FluentAssertions` namespace?

Other than that, you guys are free to review this PR. 👍

dennisdoomen:
> One more question that I have is in which namespace should `TimesConstraint`, `AtLeast`, `AtMost`, `MoreThan`, `LessThan`, and `Exactly` reside

I'll take a look

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -660,7 +693,40 @@ public AndConstraint<StringAssertions> ContainEquivalentOf(string expected, stri
             Execute.Assertion
                 .ForCondition(Contains(Subject, expected, StringComparison.CurrentCultureIgnoreCase))
                 .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context:string} to contain equivalent of {0}{reason} but found {1}.", expected, Subject);
+                .FailWith("Expected {context:string} {0} to contain equivalent of {1}{reason}, but not found.", Subject, expected);
```
dennisdoomen: That's a weird failure message. Sloppy paste?

Src/FluentAssertions/Primitives/TimesConstraint.cs
```diff
@@ -0,0 +1,169 @@
+﻿using System;
+using FluentAssertions.Common;
+using FluentAssertions.Execution;
+
+namespace FluentAssertions.Primitives
+{
+    public abstract class TimesConstraint
```
dennisdoomen: I think we should keep the type `internal` for now. 

Also, since it's not a _verb_, I think the word _times_ is a bit weird. Maybe we should name this `OccurrenceConstraint`.

This class needs some documentation love.

Src/FluentAssertions/Primitives/TimesConstraint.cs
```diff
@@ -0,0 +1,169 @@
+﻿using System;
+using FluentAssertions.Common;
+using FluentAssertions.Execution;
+
+namespace FluentAssertions.Primitives
```
dennisdoomen: I agree that this must be part of the `FluentAssertions` namespace and thus needs to be in the root of the project folder.

Src/FluentAssertions/Primitives/TimesConstraint.cs
```diff
@@ -0,0 +1,169 @@
...
+{
+    public abstract class TimesConstraint
+    {
+        private int? actualCount;
+
+        protected readonly int expectedCount;
+
+        public TimesConstraint(int expectedCount)
```
dennisdoomen: Can be `protected`.

Src/FluentAssertions/Primitives/TimesConstraint.cs
```diff
@@ -0,0 +1,169 @@
...
+
+        protected abstract bool IsMatch { get; }
+
+        private string Expected { get; set; }
+
+        private string Subject { get; set; }
+
+        private StringComparison StringComparison { get; set; }
+
+        private static string Times(int count) => count == 1 ? "1 time" : $"{count} times";
```
dennisdoomen: Please move this one below the caller. We like Clean Code 😅

Src/FluentAssertions/Primitives/TimesConstraint.cs
```diff
@@ -0,0 +1,169 @@
...
+
+        internal void AssertContain(string subject, string expected, string because, params object[] becauseArgs)
```
dennisdoomen: Since this class knows nothing about the algorithm that you use (_contains_), I think just naming it `Assert` would be enough. This also opens up the path for reuse (even though we would have to solve the hard-coded string in `FailWith` )

Src/FluentAssertions/Primitives/TimesConstraint.cs
```diff
@@ -0,0 +1,169 @@
...
+            Execute.Assertion
+                .ForCondition(IsMatch)
+                .BecauseOf(because, becauseArgs)
+                .FailWith(
+                    $"Expected {{context:string}} {{0}} to contain equivalent of {{1}} {Mode} {Times(expectedCount)}{{reason}}, but found {Times(ActualCount)}.",
+                    Subject, expected);
+        }
+    }
+
+    internal sealed class AtLeastTimesConstraint : TimesConstraint
```
dennisdoomen: I think this class can be nested in the `AtLeast` class and moved into a separate file. Same for the other combinations.

docs/_pages/strings.md
```diff
@@ -35,12 +35,50 @@ theString.Should().BeOneOf(
 );
 
 theString.Should().Contain("is a");
+theString.Should().Contain("is a", Exactly.Once());
```
dennisdoomen: I think having a single example per type of constraint will be enough.

Src/FluentAssertions/Primitives/TimesConstraint.cs
danielmpetrov: 🤔 I agree on the naming, but the type cannot be marked `internal` since it's part of the public API https://github.com/fluentassertions/fluentassertions/pull/1145/files#diff-1ceb6d097a48e71dff2ed5ad377a1009R658

Src/FluentAssertions/Primitives/StringAssertions.cs
danielmpetrov: This is something left over from the original PR. Looking at it, I don't see why the message was changed at all in the first place.

danielmpetrov: Ok, so the messages for `Contain` and `ContainEquivalentOf` are not consistent, so the original author probably decided to make them the same, but ended up with a weird mixture of both.

The choice here is between the following 2 options
1. "Expected string to contain "XYZ" because that is required but found "ABCDEF"
2. "Expected string "ABCDEF" to contain "XYZ" because that is required."

In my opinion the second option sounds more clear and natural, so I'd go with that personally.

Src/FluentAssertions/Primitives/TimesConstraint.cs
danielmpetrov: I see where you're going, but the current implementation is specifically targeting strings. What shall I do with the other `internal` method `AssertContainEquivalentOf`? That one wouldn't make sense for something like an `int[]`. This is where the re-usability of the current implementation is lacking in my opinion. Maybe `OccuranceConstraint` should be made generic? I'm not sure the best way to go about this.

Src/FluentAssertions/Primitives/StringAssertions.cs
dennisdoomen: I agree

Src/FluentAssertions/Primitives/TimesConstraint.cs
dennisdoomen: Oh, you're right. In that case I think we should introduce an interface that only contains that single Assert method. The rest can be hidden from the public contract.

Src/FluentAssertions/Primitives/TimesConstraint.cs
dennisdoomen: True. So maybe it's fine if we keep it all locked down as much as possible until we find a proper we to reuse it.

danielmpetrov: @dennisdoomen I've dumbed down the `OccurrenceConstraint` type, so that it only knows about the expected count, actual count, and the relationship between them, in the form of an `Assert` method. This way the responsibility of counting and producing an assertion message is given back to the FA assertion method, and `OccurrenceConstraint` can safely be reused in the public API.

Can you have a look and tell me what you think? 👍

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -1638,640 +1680,1747 @@ public void When_the_actual_string_equivalent_to_the_expected_but_with_trailing_
...
+            Action act = () => actual.Should().Contain(expectedSubstring, AtMost.Once());
+
+            // Assert
+            act.Should().Throw<ArgumentException>()
+                .WithMessage("Cannot assert string containment against an empty string.*")
+                .And.ParamName.Should().Be("expected");
+        }
+
+        [Fact]
+        public void When_string_containment_at_most_once_is_asserted_and_actual_value_is_null_then_it_should_throw()
```
jnyrup: function name says `should_throw` implementation has `NotThrow()`

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -636,6 +636,46 @@ public AndConstraint<StringAssertions> Contain(string expected, string because =
...
+                throw new ArgumentException("Cannot assert string containment against an empty string.", nameof(expected));
+            }
+
+            int actual = Subject.CountSubstring(expected, StringComparison.Ordinal);
+
+            Execute.Assertion
+                .ForCondition(occurrenceConstraint.Assert(actual))
+                .BecauseOf(because, becauseArgs)
+                .FailWith(
+                    $"Expected {{context:string}} {{0}} to contain {{1}} {occurrenceConstraint.Mode} {occurrenceConstraint.ExpectedCount.Times()}{{reason}}, but found {actual.Times()}.",
```
jnyrup: ❓ Why the eager formatting as opposed to providing all arguments to `FailWith` for lazy formatting?

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -660,7 +700,47 @@ public AndConstraint<StringAssertions> ContainEquivalentOf(string expected, stri
...
                 .ForCondition(Contains(Subject, expected, StringComparison.CurrentCultureIgnoreCase))
                 .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context:string} to contain equivalent of {0}{reason} but found {1}.", expected, Subject);
+                .FailWith("Expected {context:string} {0} to contain equivalent of {1}{reason}.", Subject, expected);
+
+            return new AndConstraint<StringAssertions>(this);
+        }
+
+        /// <summary>
+        /// Asserts that a string contains another (fragment of a) string a set amount of times.
```
jnyrup: Include the part from the existing `ContainEquivalent` about what constitutes an equivalent (casing).

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -1638,640 +1680,1747 @@ public void When_the_actual_string_equivalent_to_the_expected_but_with_trailing_
...
+
+            // Act
+            Action act = () => actual.Should().Contain(expectedSubstring, LessThan.Twice());
+
+            // Assert
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_containment_less_than_once_is_asserted_against_null_it_should_throw_earlier()
```
jnyrup: The two tests of a null and empty `expected` are independent of the constraint, so there is no need to duplicate them for each of the constraints.
The same goes for testing the `because` and `becauseArgs`.

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -1638,640 +1680,1747 @@ public void When_the_actual_string_equivalent_to_the_expected_but_with_trailing_
...
+        #endregion
+
+        #region Contain Equivalent Of
+
+        #region Default
+
+        [InlineData("aa", "A")]
+        [InlineData("aCCa", "acca")]
+        [Theory]
+        public void Should_pass_when_contans_equivalent_of(string actual, string equivalentSubstring)
```
jnyrup: contans -> contains

danielmpetrov:
> I found it somewhat difficult to review the tests and their seems to have been quite some rearrangement of the tests (unless it's just git diff confusing me).

It seems like the tests were rearranged in #962, which is the basis of this PR. That's something that confused me as well honestly.

jnyrup:
@danielmpetrov Let me know if I should do the tedious work of untangling the re-arrangement of the tests.
I'll might have time this Sunday.

Src/FluentAssertions/Primitives/StringAssertions.cs
danielmpetrov: Just tried this, but `FailWith` puts quotation marks around the strings this way, so the message looks improper.

e.g.

> Expected actual "ABCDEBCDF" to contain "BCD" "more than" "2 times" because that is required, but found "2 times".

Not sure if I'm missing something.

danielmpetrov:
> @danielmpetrov Let me know if I should do the tedious work of untangling the re-arrangement of the tests.
> I'll might have time this Sunday.

I'll give it a try 😅

danielmpetrov:
@jnyrup This is a git diff issue...

I've reverted the test file to match the one on `master` and then copy pasted all the new tests only nested under the `contain` and `contain equivalent of` regions of the test file, but the diff is still messed up.

danielmpetrov:
@jnyrup Ok, good news. I cleaned up some duplicate tests, and the diff appears properly now. I guess the duplicated code snippets messed git up.

Src/FluentAssertions/Primitives/StringAssertions.cs
jnyrup: Of course, makes sense.
No need to investigate further.

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -660,7 +700,48 @@ public AndConstraint<StringAssertions> ContainEquivalentOf(string expected, stri
...
+        public AndConstraint<StringAssertions> ContainEquivalentOf(string expected, OccurrenceConstraint occurrenceConstraint, string because = "", params object[] becauseArgs)
+        {
+            Guard.ThrowIfArgumentIsNull(expected, nameof(expected), "Cannot assert string containment against <null>.");
+
+            if (expected.Length == 0)
+            {
+                throw new ArgumentException("Cannot assert string containment against an empty string.", nameof(expected));
+            }
+
+            int actual = Subject.CountSubstring(expected, StringComparison.CurrentCultureIgnoreCase);
```
jnyrup: #1184 is switching working on switching from `StringComparison.CurrentCultureIgnoreCase` to `StringComparison.OrdinalIgnoreCase` in the other string equivalency assertions.
So it should (now) be `StringComparison.OrdinalIgnoreCase` here as well.

Tests/Shared.Specs/StringAssertionSpecs.cs
```diff
@@ -1432,12 +1432,354 @@ public void When_string_containment_is_asserted_and_actual_value_is_null_then_it
...
+            // Arrange
+            string actual = "ABCDEF";
+            string expectedSubstring = "XYS";
+
+            // Act
+            Action act = () => actual.Should().Contain(expectedSubstring, Exactly.Once(), "that is {0}", "required");
+
+            // Assert
+            act.Should().Throw<XunitException>()
+                .WithMessage("Expected * \"ABCDEF\" to contain \"XYS\" exactly 1 time because that is required, but found 0 times.");
```
jnyrup: Something sounds odd to me about the part `but found 0 times.`
Brainstorming some alternatives:
*  `but found it 0 times.`
*  `but found 0 occurences.`

dennisdoomen: I would stick to _times_. If you would like to use the word _occurrence_, the entire message should change.

Src/FluentAssertions/Primitives/TimesConstraint.cs
dennisdoomen: Looks good enough for me.

Src/FluentAssertions/AtLeast.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        public static OccurrenceConstraint Once() => new AtLeastTimesConstraint(1);
+
+        public static OccurrenceConstraint Twice() => new AtLeastTimesConstraint(2);
+
+        public static OccurrenceConstraint Thrice() => new AtLeastTimesConstraint(3);
+
+        public static OccurrenceConstraint Times(int expected) => new AtLeastTimesConstraint(expected);
+
+        internal sealed class AtLeastTimesConstraint : OccurrenceConstraint
```
dennisdoomen: You could make these `private`.

Tests/Shared.Specs/StringAssertionSpecs.cs
jnyrup: @dennisdoomen So you would change the current wording to "but found it 0 times"?

dennisdoomen: Yep

danielmpetrov:
It's your favorite failing test - `When_the_execution_time_of_an_action_is_close_to_a_limit_it_should_not_throw` 🎉 

Please tell me if anything else needs work on this PR @jnyrup @dennisdoomen

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -660,7 +700,48 @@ public AndConstraint<StringAssertions> ContainEquivalentOf(string expected, stri
             Execute.Assertion
                 .ForCondition(Contains(Subject, expected, StringComparison.CurrentCultureIgnoreCase))
                 .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context:string} to contain equivalent of {0}{reason} but found {1}.", expected, Subject);
+                .FailWith("Expected {context:string} {0} to contain the equivalent of {1}{reason}.", Subject, expected);
```
jnyrup: For consistency the message of `ContainEquivalentOf` should match `CollectionAssertions.ContainEquivalentOf`.
But I would leave that change out of this PR :).

dennisdoomen: @jnyrup  If there's nothing else, can you please approve the PR?

jnyrup: I just re-read my words above and see that they might read opposite of my intentions.
I was meant to be:

Drop the changes from this PR as that makes it inconsistent with `CollectionAssertions.ContainEquivalentOf`. 
The pedantic me also think it should be *an equivalent* and not *the equivalent*, as there are several different equivalents.

danielmpetrov: @jnyrup Wait, I'm confused. This change is so that consistency _**is**_ introduced with `CollectionAssertions.ContainEquivalentOf`.

https://github.com/fluentassertions/fluentassertions/blob/c477f8bcdb6613e47792d7729bfa253785d9384d/Src/FluentAssertions/Collections/CollectionAssertions.cs#L636

The only difference is the extra word `the`, which can be removed, or alternatively replaced with `an`.

dennisdoomen: Yeah, now I'm confused as well.

jnyrup: Now I see the other changes :)
I only saw the addition of `the`, which seems to match `StringWildcardMatchingValidator.cs:62`.

Issue #172, opened 2014-10-21
Should().Be() / NotBe() prefers IComparable<T>.Compare over object.Equals


ulrichb:
Given the following class with _value semantics_.

``` C#
public class Ranked<T> : IComparable<Ranked<T>>
{
    public Ranked (int ranking, T name) { Ranking = ranking; Item = name; }

    public int Ranking { get; private set; }
    public T Item { get; private set; }

    public override bool Equals (object obj)
    {
        if (ReferenceEquals (null, obj))
            return false;
        if (obj.GetType() != this.GetType())
            return false;
        var other = (Ranked<T>) obj;
        return Ranking == other.Ranking && Equals (Item, other.Item);
    }

    public override int GetHashCode () { /* ... */ }

    public int CompareTo (Ranked<T> other)
    {
        if (other == null) return 1;

        return this.Ranking.CompareTo (other.Ranking);
    }
}
```

At the moment (v3.2.1) the assertion `new Ranked<string>(1, "Some item").Should().Be (new Ranked<string>(1, "Other item"))` passes whereas casting the first part to object, `((object)new Ranked<string>(1, "Some item")).Should().Be (new Ranked<string>(1, "Other item"))`, fails. The first assertion calls `CompareTo`, which just compares the `Ranking` value and the second one calls `Equals`.

The reason for the different behavior of the two assertions is the `IComparable<>` overload of `Should()`. The idea of this overload is to be able to offer fluent methods for `IComparable<>` values, like `BeInRange()`. Unfortunately also `Be()` and `NotBe()` are "overridden" to check for `CompareTo() == 0` resp. `CompareTo() != 0`.

In my opinion this is not a good idea, because `Be()` / `NotBe()` are the _standard methods_ in FluentAssertions to check for _equality / inequality_. And sort ordering is not the same as equality. The example above shows a legitimate case, where just a subset of the properties is part of the ordering.

Thinking of another example of a `Person`-_entity_ where equality is defined by ID-equality (e.g. a `SocialSecurityNumber` property) and the ordering should be applied for `LastName / GivenName` shows that the set of properties used for equality / ordering don't even have to intersect.

We should think about a (breaking) change of the behavior of `ComparableTypeAssertions<T>.Be()`. So that `Should().Be()` always checks for equality and introduce a new `Should().HaveSameOrderingPositionAs()` (probably there's better name), which checks for `CompareTo() == 0`.

[...]

PR #1152, opened 2019-09-26
Add assertions for Equality when type implements IComparable #172

rubenrorije:
During creating this pull request I stumbled upon the following points of attention.

1. In the original code `Be` checks for reference equality, but `NotBe` does not. I changed the `NotBe` to also check reference equality.
2. It would be useful to differ the error messages that are shown to the user. Ideally there would need to be three different error messages: for reference equality, for 'normal' equality, and one for comparison. I am not sure how to change the code in such a way that different error messages are returned for the different cases. I can also imagine that a discussion about the correct wording might be necessary first.
For example (NotBe):
    - Did not expect subject to point to the same reference ...
    - Did not expect subject to return 0 for comparison ...
    - Did not expect subject to be equal to ...
3. I added tests for reference equality, which seemed to be missing.
4. I added tests for the new behaviour, but am not sure whether it conforms to the standards and especially whether the naming is that intuitive, I could use an exta pair of eyes.

Part of #172


Src/FluentAssertions/Numeric/ComparableTypeAssertions.cs
```diff
@@ -35,7 +36,7 @@ public ComparableTypeAssertions(IComparable<T> value) : base(value)
         public AndConstraint<ComparableTypeAssertions<T>> Be(T expected, string because = "", params object[] becauseArgs)
         {
             Execute.Assertion
-                .ForCondition(ReferenceEquals(Subject, expected) || (Subject.CompareTo(expected) == Equal))
+                .ForCondition(ReferenceEquals(Subject, expected) || ((Subject.CompareTo(expected) == Equal) && Subject.IsSameOrEqualTo(expected)))
```
dennisdoomen: I think we should _only_ use `ISameOrEqual` since that will ensure `Be` behaves like other methods

Src/FluentAssertions/Numeric/ComparableTypeAssertions.cs
```diff
@@ -125,7 +127,7 @@ public AndConstraint<ComparableTypeAssertions<T>> Be(T expected, string because
         public AndConstraint<ComparableTypeAssertions<T>> NotBe(T unexpected, string because = "", params object[] becauseArgs)
         {
             Execute.Assertion
-                .ForCondition(Subject.CompareTo(unexpected) != Equal)
+                .ForCondition(!ReferenceEquals(Subject, unexpected) && (Subject.CompareTo(unexpected) != Equal || !Subject.IsSameOrEqualTo(unexpected)))
```
dennisdoomen: Same here. Just use `!IsSameOrEqualTo`

Src/FluentAssertions/Numeric/ComparableTypeAssertions.cs
rubenrorije: Do I understand correctly that the easiest solution then would be to just remove these methods `Be` and `NotBe` altogether from `ComparableTypeAssertions`? It would fall back on the ObjectAssertions. This would have the downside, though, of not checking the implementation of `CompareTo` anymore.

dennisdoomen: Yep. That's why we [suggested](https://github.com/fluentassertions/fluentassertions/issues/172#issuecomment-59910838) adding another method for that.

rubenrorije: From the thumbs up on my suggestion I (wrongly) assumed that was the way to go. Maybe I jumped the gun on this.

dennisdoomen: You're almost there, aren't you?

rubenrorije: Now I am confused. My understanding is that you only want Equality and no Reference Equality and Comparison checks?
What I implemented is checking all three, to make sure that people who are depending on the behaviour that the comparison is checked is not implicitly removed in a new version. Most of the implementation is checking these three cases.

dennisdoomen: Reference quality is usually checked with `BeSameAs`.

rubenrorije:
Is this what you had in mind?

dennisdoomen:
Any change you'll continue with this?

rubenrorije:
Removed most code I added, because of the changes requested. Now the code only checks equality and does not check comparison. Which meant a lot of the tests where no longer needed. Now the new implementation is identical to the one in ObjectAssertions.

dennisdoomen:
Your PR now also includes all the other's commits that were added recently. You either need to re-base or create a new branch and cherry-pick/squash your own changes.

rubenrorije:
Started over in a new pull request to tidy up commits

PR #1155, opened 2019-10-01
Chain assertions

jnyrup:
We have great support for chaining assertions and by using `WithExpectation` we can ensure that failure messages from assertions with a shared prefix are in sync.
E.g. instead of writing
```c#
Execute.Assertion
    .ForCondition(condition1)
    .BeCauseOf(because, becauseArgs)
    .FailWith("Expected collection to be something, but found <null>.");

Execute.Assertion
    .ForCondition(condition2)
    .BeCauseOf(because, becauseArgs)
    .FailWith("Expected collection to be something, but it was empty.");
```
we can
```c#
Execute.Assertion
    .BeCauseOf(because, becauseArgs)
    .WithExpectation("Expected collection to be something, ")
    .ForCondition(condition1)
    .Then
    .ForCondition(condition2)
    .FailWith("but it was empty.")
    .Then
    .ClearExpectation();
```

Also this PR remove unnecessary extra calls to `BecauseOf` for chained assertions.

### DIFF
--- a/Src/FluentAssertions/Primitives/DateTimeAssertions.cs
+++ b/Src/FluentAssertions/Primitives/DateTimeAssertions.cs
@@ -379,13 +379,12 @@ namespace FluentAssertions.Primitives
         public AndConstraint<DateTimeAssertions> HaveYear(int expected, string because = "", params object[] becauseArgs)
         {
             Execute.Assertion
+                .BecauseOf(because, becauseArgs)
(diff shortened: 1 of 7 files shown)


dennisdoomen:
I didn't quite get what this PR does. Maybe make sure your commits explain the _why_ and _what_ and not the _what_.

jnyrup:
`When_the_execution_time_of_an_action_is_close_to_a_limit_it_should_not_throw ` that's failing :(

Issue #1085, opened 2019-06-16
Abstract time away from ExecutionTime


jnyrup:
#1076 abstracted time (`Wait`, `Delay` and more) to make testing `DelegationAssertions` reliable.

We should do the same for `ExecutionTime`.

jnyrup:
I'm giving this a try here https://github.com/jnyrup/fluentassertions/tree/ExecutionTime
I could use an extra pair of eyes as it's not working yet.

Evangelink:
@jnyrup I suspect that the problem is coming from the [line 45](https://github.com/jnyrup/fluentassertions/commit/796982b1f1e7dc01491ce72ed7798096128e718d#diff-b92251b31975b253d34fed16bed41833R45) when executing `!execution.Wait(rate);`.

If you try to debug the test described [here](https://github.com/jnyrup/fluentassertions/commit/796982b1f1e7dc01491ce72ed7798096128e718d#diff-dd0f96bbe45e01b1feafc897a60a299dR437), you will end up with the following workflow:
1. Set result of the `delayTask` to `true`
2. Fake wait for 100ms (time given in the `BeGreaterThan`)
3. Return `true`, result that was set in #1
4. `isRunning` is now `false` so we stop the loop
5. We execute the assertion with the currently elapsed time
6. We fake sleep for 200ms
7. Test fails because 100ms not strictly greater than 100ms

Serg046:
Clock doesn't have `Wait(task, timeout)` anymore so we cannot set sleeping time automatically as we do it with Delay/DelayAsync

jnyrup:
@Serg046 If cracking the case involves re-introducing members on Clock that's fine.
The whole `IClock/Clock` mechanism was introduced solely to be able to abstract time for our unit tests.

jnyrup:
`When_the_execution_time_of_a_member_is_close_to_a_limit_it_should_not_throw` failed in https://ci.appveyor.com/project/dennisdoomen/fluentassertions/builds/40080954.

```cs
Action act = () => subject.ExecutionTimeOf(s => s.Sleep(210)).Should().BeCloseTo(200.Milliseconds(),
    150.Milliseconds());
```

I read from #1620 that you decided to keep using a real `StopWatch` for all remaining tests - I assume that's to keep tests close to the actual executed paths?

It would be convenient to get rid of all remaining calls to `Thread.Sleep`, but it would either:
* Leave `StopwatchTimer` untested, or 
[...]

PR #1158, opened 2019-10-07
Support assertions for methods returning ValueTask and ValueTask<T>

danielcweber:
Explicitly take netcoreapp2.1 and netstandard2.1 into account in FluentAssertions.csproj.


Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net45;net47;netstandard1.3;netstandard1.6;netstandard2.0;netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks>net45;net47;netstandard1.3;netstandard1.6;netstandard2.0;netstandard2.1;netcoreapp2.0;netcoreapp2.1</TargetFrameworks>
```
dennisdoomen: If we target just `.NET Standard 2.1`, isn't that enough?

jnyrup: According to [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1?view=netcore-1.0) it's available from netcore1.0 and netstandard2.1

dennisdoomen: Which means we don't need to explicitly change the targets.

jnyrup: As dotnet core 3.0 is the first platform to implement netstandard2.1 and unit test projects must target a platform, i.e. they cannot target netstandard, I agree it should not be necessary to change the targets.

jnyrup:
If we start to support `ValueTask<T>`, we should also add support for the non-generic `ValueTask`.
And probably add support for all overloads that currently takes `Task`/`Task<T>`.

We should also make sure that we follow the [guidelines](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1?view=netcore-3.0#remarks) about usage of `ValueTask`
>The following operations should never be performed on a ValueTask<TResult> instance:
> * Awaiting the instance multiple times.
> * Calling AsTask multiple times.
> * Using .Result or .GetAwaiter().GetResult() when the operation hasn't yet completed, or using them multiple times.
> * Using more than one of these techniques to consume the instance.

Src/FluentAssertions/FluentAssertions.csproj
danielcweber: There's a dicrepancy about the availability of ValueTask and ValueTask<T> in .netstandard and .netcore. They are not equally available (or unavailable). Thus the two new targets.

danielcweber:
Non-generic ValueTask is indeed supported in my PR. The "Awaiting" assertions don't change the return types though, and call AsTask just once, so it's safe according to the guidelines you quoted.

Src/FluentAssertions/FluentAssertions.csproj
jnyrup: Ahh, I just assumed that `ValueTask` `ValueTask<T>` was introduced at the same time, but [`ValueTask`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask?view=netcore-2.1) is not available until netcore2.1

What's the difference in availability between netstandard2.1 and netcore2.1?

danielcweber: As for .netstandard, both are just available from 2.1. As for core, ValueTask<T> has always been around (according to docs) while ValueTask is around since .net core 2.1.

However, it is always possible to reference the discrete library for all unsupported platforms. You'd still have to import it based on target framework, but you could get rid of the #if directives in the code. I didn't do that because taking additional dependencies as part of a pull request is almost always and for every maintainer a no-go.

dennisdoomen: So what if just stick to .NET Standard 2.1?

dennisdoomen:
So what's the status of this PR?

danielcweber:
>So what's the status of this PR?
Sorry, did I miss something you want me to change? Do you consider this change useful?

dennisdoomen:
> So what if just add .NET Standard 2.1, and skip .NET Core 2.0? It's a new feature, so I'm fine if it only works for that platform.

I think we should do this one. 

@jnyrup are your concerns addressed in a satisfactory way?

Src/FluentAssertions/AssertionExtensions.cs
```diff
@@ -67,6 +67,30 @@ public static Func<Task> Awaiting<T>(this T subject, Func<T, Task> action)
...
+        }
+#endif
+
+#if NETCOREAPP2_0 || NETCOREAPP2_1 || NETSTANDARD2_1
+        /// <summary>
+        /// Invokes the specified action on a subject so that you can chain it
+        /// with any of the assertions from <see cref="AsyncFunctionAssertions"/>
+        /// </summary>
+        [Pure]
+        public static Func<Task<TResult>> Awaiting<T, TResult>(this T subject, Func<T, ValueTask<TResult>> action)
```
jnyrup: Overload is never referenced?

dennisdoomen:
> With the proposed changes in `TargetFrameworks`
> NetCore30.Specs changes from `netcoreapp2.1` to `netcoreapp2.1` `TargetFramework`, but already tested via `NetCore20.Specs` leaving `netstandard2.1` untested.
> 
> Would any be left out, if we only add `netcoreapp2.1`? (Xamarin, UWP, ???)

Not sure if I understand you correctly, but if we switch the NetCore20.csproj to NetCore21.csproj, we should be fine right. We don't have to have to directly test .NET Standard 2.1. That's a compiler concern./

jnyrup:
If we change NetCore20.Specs to NetCore21.specs, then nothing tests the `netcoreapp20` target.

dennisdoomen:
And since that one is [no longer supported](https://github.com/eVisionSoftware/PermitVision/pull/10924) by Microsoft, I think that is fine.

danielcweber:
Pushed 3 more commits to address the required changes. Not sure I understood your intentions correctly though....

danielcweber:
Apparently Nuke doesn't like it...

dennisdoomen:
> Apparently Nuke doesn't like it...

It's not Nuke, it's XUnit

danielcweber:
It's not a failed test though....is it ?

dennisdoomen:
Try to run the `build.ps1` locally.

danielcweber:
>It's not Nuke, it's XUnit

It was due to the build script. Fixed by https://github.com/fluentassertions/fluentassertions/pull/1158/commits/9182d1997016e88a6015823d25c38ca3a0b8ea3e.

Tests/Shared.Specs/AsyncFunctionExceptionAssertionSpecs.cs
```diff
@@ -735,6 +989,29 @@ public void When_async_method_of_T_succeeds_and_expected_not_to_throw_particular
             action.Should().NotThrow();
         }
 
+#if NETCOREAPP2_1 || NETSTANDARD2_1
```
jnyrup: `NETSTANDARD2_1` -> `NETCOREAPP3_0` four times in this document

dennisdoomen:
> We're still supporting net45, even though it's been [EOL](https://devblogs.microsoft.com/dotnet/support-ending-for-the-net-framework-4-4-5-and-4-5-1/) for almost four years.

We'll drop the explicit .NET Core 2.0 target, but we're still supporting it through our .NET Standard 2.0 target. I might be wrong, but I didn't see anything _specifically_ using .NET Core 2.0 stuff.

jnyrup:
@dennisdoomen the following does not work with `netstandard2_0`, but with `netcoreapp2_0`:

* `EventAssertions`
* `EventHandlerFactory`
* `EventMonitor`
* `EventRecorder`
* `IMonitor`
* `Monitor<T>(this T eventSource, Func<DateTime> utcNow = null)`

dennisdoomen:
You've tried? If so, that's annoying. Maybe we should then just target both. We'll clean-up the targets with 6.0 anyway/

jnyrup:
I just took master and changed the target framework from `netcoreapp2_0` to `netcoreapp2_1`. 
Then `NetCore20.Specs` won't compile as `Monitor(this T eventSource, ...)` is excluded from the `NETSTANDARD2_0` target.
That is ultimately excluded because `Netstandard2.0` does not include `DynamicMethod` or `ILGenerator`. See #859

danielcweber:
Got a little bit lost on direction here....if you need any additional work in this PR, just @ me. Otherwise I think you know better which targets to support and not support in the future...

dennisdoomen:
@jnyrup I propose we keep the changes in the PR and merge it. We'll clean up the targets in 6.0 anyway.

jnyrup:
@danielcweber Sorry for the confusion about the targets.
Could you re-add the `netcoreapp2.0` target, such the changes effectively will be addition of `netstandard2.1` and `netcoreapp2.1`?

Just to be sure, let's have both:
* `NetCore20.Specs`
* `NetCore21.Specs`

Tests/Shared.Specs/AsyncFunctionExceptionAssertionSpecs.cs
```diff
@@ -735,6 +989,29 @@ public void When_async_method_of_T_succeeds_and_expected_not_to_throw_particular
             action.Should().NotThrow();
         }
 
+#if NETCOREAPP2_0 || NETCOREAPP2_1 || NETCOREAPP3_0 || NETSTANDARD2_1
```
jnyrup: Is the ` || NETSTANDARD2_1` for automatically including potentially future test projects?

Tests/Shared.Specs/EventAssertionSpecs.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if !NETCOREAPP1_1 && !NETSTANDARD1_3 && !NETSTANDARD1_6 && !NETSTANDARD2_0
+﻿#if !NETCOREAPP1_1 &&  !NETCOREAPP2_0 && !NETSTANDARD1_3 && !NETSTANDARD1_6 && !NETSTANDARD2_0 && !NETSTANDARD2_1
```
jnyrup: Shouldn't `NETCOREAPP2_0` still be included?

Tests/Shared.Specs/AsyncFunctionExceptionAssertionSpecs.cs
danielcweber: Shouldn't be there. Fixed.

Tests/Shared.Specs/EventAssertionSpecs.cs
danielcweber: For some reason, it would complain this morning about "Monitor", now it doesn't. Fixed.

jnyrup:
Just so I don't forget about it:
We should also the following docs with the added target frameworks:
* `docs/index.html:29+53`
* `docs/_config.yml:22`
* `docs/_pages/about.md:42+43+46+92`
* `Src/FluentAssertions/FluentAssertions.csproj:16`
*

danielcweber:
There was some odd "this thing took 300ms too long" test fail. Test went well now. But there's maybe a flaky test around.

jnyrup:
@danielcweber It's the test that keeps on giving... #1085
When the docs are updated, I'll be happy to punch the merge button.

dennisdoomen:
@danielcweber if you update the docs as @jnyrup request, we should be ready to merge it.

danielcweber:
^

jnyrup:
@danielcweber Could you update the index.html line 53 as well?

danielcweber:
Must have slipped through.

jnyrup:
@danielcweber Thanks for contributing this!
... and for hanging in there while we made up our minds.

Issue #1154, opened 2019-09-29
TypeAssertions.HaveAccessModifier and TypeAssertions.NotHaveAccessModifier return AndConstraint<Type> instead of AndConstraint<TypeAssertions>


adrianiftode:
### Description

These two methods are the only ones with this return and I believe it should be AndConstraint<TypeAssertions>

MethodBaseAssertions returns AndConstraint<TAssertions>

### Versions

* Which version of Fluent Assertions are you using? 5.9.0

PR #1159, opened 2019-10-08
Fix the return type of TypeAssertions.HaveAccessModifier and NotHaveAccessModifier

adrianiftode:
#1154

_The changes don't include a unit test, because I didn't see a common practice in the code base to test that the AndConstraint is closed with the correct type. The only test I've seen that checks for this particular situation is `When_using_StringCollectionAssertions_the_AndConstraint_should_have_the_correct_type`. I could add something similar._

There are no references in docs to these two assertions


dennisdoomen:
I wondering whether we should treat this as a binary compatibility issue.

adrianiftode:
It will be an issue, because the return type is changed and it will throw a `MethodNotFoundException`. The clients will need to recompile.

So for let's say if the current usage is something like:

`type.Should().HaveAccessModifier(CSharpAccessModifier.Internal).And.Should().BeAbstract();`

After this change the required step would be:

`type.Should().HaveAccessModifier(CSharpAccessModifier.Internal).And.BeAbstract();`

If the `AndConstraint` is not used, then the clients don't have to change the source code, just to recompile.

dennisdoomen:
Yeah, so technically this would be a breaking change. But then again, how likely is this going to be a realistic problem. @jnyrup what do you think?

dennisdoomen:
Can we continue with this one now that we're accepting potentially breaking changes again?

jnyrup:
Let's continue with this!

@adrianiftode  Note that since #1217 any API change must be changed in the relevant [`approved.txt` files](https://github.com/fluentassertions/fluentassertions/tree/develop/Tests/Approval.Tests/ApprovedApi/FluentAssertions).
When you run the tests locally, `received.txt` files are generated in `Tests/Approval.Tests/ApprovedApi/FluentAssertions` with the API changes.
Copy these into `approved.txt` files to make the approval tests pass.

adrianiftode:
@jnyrup I've run the tests as instructed.

When I first time run the tests my editor added an extra line at the end of the files and the tests failed. I wonder if these lines could be relaxed/ignored while asserting it.

jnyrup:
Sorry, the flow has been improved since I suggested to "copy over the changes".
It should not be necessary to involve an IDE.
In the root folder of the project we now have `AcceptApiChanges.{sh,ps1}` to do the changes.
Would that avoid the problems you had?

adrianiftode:
I just tried it and looks fine.

jnyrup:
While I remember it...
Could I ask you to add a line to `docs/_pages/releases.md` under "Breaking Changes" with the changes done?

jnyrup:
@adrianiftode Thanks again for reporting *and* contributing a fix for this issue.

adrianiftode:
It's been a pleasure, thanks for guidance

vin., 21 feb. 2020, 16:57 Jonas Nyrup <notifications@github.com> a scris:

PR #1164, opened 2019-10-13
Replace the package downloads with normal package references.

dennisdoomen:


Build/_build.csproj
```diff
@@ -10,12 +10,12 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageDownload Include="GitVersion.CommandLine.DotNetCore" Version="[5.0.1]" />
-    <PackageDownload Include="NSpec" Version="[1.0.13]" />
-    <PackageDownload Include="NSpec" Version="[2.0.1]" />
-    <PackageDownload Include="NSpec" Version="[3.1.0]" />
-    <PackageDownload Include="nunit.runners" Version="[2.7.1]" />
-    <PackageDownload Include="xunit.runner.console" Version="[2.4.1]" />
+    <PackageReference Include="GitVersion.CommandLine.DotNetCore" Version="[5.0.1]" />
```
lg2de: Wondering whether "[" and "]" here are really intended or just left over.
I could not found documentation for these characters for `PackageReference`.
Further, your change is inconsistent on this. Is this really your intention?

eNeRGy164: [Package versioning - Version ranges and wildcards](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges-and-wildcards)


Notation | Applied rule | Description
-- | -- | --
[1.0] | x == 1.0 | Exact version match

dennisdoomen:
I'll close it. It doesn't fix the original problem anyway.

matkoch:
What is the original problem?

PR #1169, opened 2019-10-25
Fix comparison of empty multi-dimensional arrays

### LINKED ISSUE #1167: Equivalency of multidimensional arrays with zero size width
Following code throws an exception (`System.IndexOutOfRangeException : Index was outside the bounds of the array.`):

```csharp
[Fact]
public void Test()
{
    Array a = new long[1,0] { { } };
    Array b = new long[1,0] { { } };

    a
        .Should().BeEquivalentTo(b);
}
```

I'd expect `a` to be equivalent to `b`.

I'm using version 5.9.0 of FluentAssertions.

jnyrup:
When comparing multi-dimensional `Array`s using `BeEquivalentTo`
we weren't handling the case of empty `Array`s, which threw an
`IndexOutOfRangeException` when it tried to get an element.
This fixes #1167


jnyrup:
The failing test is `When_the_execution_time_of_an_async_action_is_less_than_a_limit_it_should_not_throw`...

dennisdoomen:
> The failing test is `When_the_execution_time_of_an_async_action_is_less_than_a_limit_it_should_not_throw`...

Sigh. I thought we fixed this by now.

jnyrup:
`ExecutionTime()` #1085 is the missing timing assertion to be fixed.

PR #1170, opened 2019-10-25
Detect self-closing xml tags with attributes

### LINKED ISSUE #1166: XDocument BeEquivalentTo assertion returns incorrect path to node
### Description

XDocument BeEquivalentTo assertion returns incorrect path to node

```csharp
// Arrange
XDocument expected = XDocument.Parse("<?xml version=\"1.0\" encoding=\"utf-8\"?><root><a><aa><aaa attr=\"x\"/></aa></a><b><bb id=\"\"/></b></root>");
XDocument actual = XDocument.Parse("<?xml version=\"1.0\" encoding=\"utf-8\"?><root><a><aa><aaa attr=\"x\"/></aa></a><b><bb id=\"value\"/></b></root>");

// Assert
```

### Expected behavior:

Assertion should fail with the following message:
>Expected attribute "id" at "/root/b/bb" to have value "", but found "value".

### Actual behavior:

Assertion fails with the following message:
>Expected attribute "id" at "/root/a/b/bb" to have value "", but found "value".

The 'a' node is not in the correct path to the difference between the two xmls.

### Versions

* Version of Fluent Assertions: 5.9.0
* .NET framework 4.7

### Additional Information

If the attr attributes are removed, that is, the xml is modified to the following format:
```csharp
XDocument expected = XDocument.Parse("<?xml version=\"1.0\" encoding=\"utf-8\"?><root><a><aa><aaa/></aa></a><b><bb id=\"\"/></b></root>");
XDocument actual = XDocument.Parse("<?xml version=\"1.0\" encoding=\"utf-8\"?><root><a><aa><aaa/></aa></a><b><bb id=\"value\"/></b></root>");
```
then the test behaves as expected, the message is: 
>Expected attribute "id" at "/root/b/bb" to have value "", but found "value".

jnyrup:
When traversing the structure of an `XDocument` and into
the `Attributes` of an xml tag, we need to move out again
using `MoveToElement()` to properly detect the xml tag,
and not the attributes, as self-closing.
This fixes #1166


Tests/Shared.Specs/XDocumentAssertionSpecs.cs
```diff
@@ -659,6 +659,21 @@ public void When_asserting_an_xml_document_is_equivalent_to_different_xml_docume
...
         }
 
+        [Fact]
+        public void when_an_xml_element_has_attributes_and_is_self_closing_it_should_succeed()
+        {
+            // Arrange
+            XDocument actual = XDocument.Parse("<xml><a attr=\"x\"/><b id=\"foo\"/></xml>");
+            XDocument expected = XDocument.Parse("<xml><a attr=\"x\"/><b id=\"bar\"/></xml>");
+
+            // Act
```
dennisdoomen: Maybe we should start using this "light" AAA format everywhere

jnyrup: Pure muscle memory, cause the the light AAA format is what I use at work.
I'll push an update to this PR to use the other format for consistency.

But I'll be happy to create a another PR converting to light AAA syntax. I don't recall having seen this heavier syntax anywhere else.

dennisdoomen: I introduced the heavy format myself years ago, but I don't see a lot of value in it anymore.

danielmpetrov: +1 on the 'light' AAA syntax. I find the current format over the top, and many times the comment to code ratio is very high, making it extra hard to navigate around the tests.

Issue #1149, opened 2019-09-24
.NET Native causing NullReferenceException in CallerIdentity class


borrrden:
### Description

When using certain functions in UWP tests, null reference exceptions occur *only when compiling with .NET Native* (which is a requirement for the Windows Store). 

### Complete minimal example reproducing the issue

Complete means the code snippet can be copied into a unit test method in a fresh C# project and run.
Minimal means it is stripped from code not related to reproducing the issue.

E.g.

```csharp
var arr1 = new[] {1, 2, 3, 4, 5};
var arr2 = new[] {1, 2, 3, 4, 5, 6};
arr1.Should().NotBeEquivalentTo(arr2);
```

### Expected behavior:

The assertion should pass

### Actual behavior:

Message: Test method ... threw exception: 
System.NullReferenceException: Object reference not set to an instance of an object.

Test Name:	TestReKey
Test FullName:	Test.DatabaseEncryptionTest.TestReKey
Test Source:	C:\Development\couchbase-lite-net-ee\couchbase-lite-net\src\Couchbase.Lite.Tests.Shared\DatabaseEncryptionTest.cs : line 169
Test Outcome:	Failed
Test Duration:	0:00:00.2356786

Result StackTrace:	
```
at FluentAssertions.CallerIdentifier.DetermineCallerIdentity()
   at FluentAssertions.Execution.AssertionScope.<>c__DisplayClass30_0.<FailWith>b__0()
   at FluentAssertions.Execution.AssertionScope.FailWith(Func<System.String> failReasonFunc)
   at FluentAssertions.Execution.AssertionScope.FailWith(Func<FluentAssertions.Execution.FailReason> failReasonFunc)
   at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args)
   at Continuation FluentAssertions.Equivalency.EnumerableEquivalencyValidatorExtensions.AssertCollectionHasEnoughItems[T](IAssertionScope, ICollection`1, ICollection`1) + 0x100
   at Continuation FluentAssertions.Equivalency.EnumerableEquivalencyValidator.AssertCollectionsHaveSameCount[T](ICollection`1, ICollection`1) + 0xe8
   at Void FluentAssertions.Equivalency.EnumerableEquivalencyValidator.Execute[T](Object[], T[]) + 0x90
[...]

PR #1171, opened 2019-10-25
Catch exceptions from DetermineCallerIdentity

jnyrup:
Apparently when compiling an UWP project with .NET Native
`StackFrame.GetMethod` can return `null`. As determination
of caller identity is just a nice-to-have and not a need-to-have
I think we should wrap it in a `try/catch` instead of throwing
exceptions.
Related to #1149

### DIFF
--- a/Src/FluentAssertions/CallerIdentifier.cs
+++ b/Src/FluentAssertions/CallerIdentifier.cs
@@ -21,17 +21,29 @@ namespace FluentAssertions
         {
             string caller = null;
 
-            StackTrace stack = new StackTrace(true);
-
-            foreach (StackFrame frame in stack.GetFrames())
+            try
             {
-                logger(frame.ToString());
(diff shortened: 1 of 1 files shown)


jnyrup:
I thought of one thing...
If it throws an exception, but `caller` is non-null, should we return that or just null?

dennisdoomen:
> If it throws an exception, but `caller` is non-null, should we return that or just null?

When would this happen?

jnyrup:
I can't give a realistic example, but if `caller` is set to something during one iteration, and then an exception is thrown during a subsequent iteration.

dennisdoomen:
Can we merge this one?

jnyrup:
Yes, I'll merge it.
I just re-read the code and hadn't seen the `break` inside the loop.
I thought `caller` could be assigned multiple times since it had `?? caller`.

PR #1174, opened 2019-10-28
Fix self-closing XML element is not equivalent to an element without content

lg2de:
While testing equivalence of XML documents using the XDocument/XElement extension of FA, I found that "empty elements" are identified as different to "self-closing elements", but e.g. `<child></child>` should be equivalent to `<child />`, e.g. according to https://www.w3schools.com/xml/xml_elements.asp

My proposed solution does not look great. But I did not have any other idea to "move forward" and "rewind" the `XmlReader`.

What do you think?


Src/FluentAssertions/Xml/XmlReaderValidator.cs
```diff
@@ -103,8 +104,41 @@ private ValidationResult Validate()
                     return validationResult;
                 }
 
-                subjectReader.Read();
-                otherReader.Read();
+                var readOnSubjectReader = true;
```
dennisdoomen: I don't really get how this works together with the `switch` statement. Shouldn't this be part of a `case`?

Src/FluentAssertions/Xml/XmlReaderValidator.cs
```diff
@@ -103,8 +104,41 @@ private ValidationResult Validate()
...
+                var readOnOtherReader = true;
+                if (subjectReader.IsEmptyElement && !otherReader.IsEmptyElement && !otherReader.HasValue)
```
dennisdoomen: You could make this code easier to read if we wrap the `XmlReader` in another class that exposes specific methods and properties instead of directly interacting with the `XmlReader`

Src/FluentAssertions/Xml/XmlReaderValidator.cs
```diff
@@ -103,8 +104,41 @@ private ValidationResult Validate()
...
+                if (otherReader.IsEmptyElement && !subjectReader.IsEmptyElement && !subjectReader.HasValue)
+                {
+                    // advance subject reader to EndElement to sync empty element with self-closing element
+                    subjectReader.Read();
+                    if (subjectReader.NodeType != XmlNodeType.EndElement || subjectReader.LocalName != otherReader.LocalName)
+                    {
+                        // advancing failed
+                        // skip reading subject reader below to "simulate" rewind reader
+                        readOnSubjectReader = false;
+                    }
```
dennisdoomen: With my previous remark, this entire block can be encapsulated.

jnyrup:
Thanks for taking a stab at this.
With all this conditional logic, the number of tests doesn't seem covering.
E.g. I was able to comment out both `!otherReader.IsEmptyElement` and `&& !otherReader.HasValue` and all tests still passed.

Src/FluentAssertions/Xml/XmlReaderValidator.cs
lg2de: You're right, this checks should be done only while `XmlNodeType.Element` is current.

Src/FluentAssertions/Xml/XmlReaderValidator.cs
lg2de: I understand what you mean. But, I think it is not really possible.
Both readers must get synchronized. (Advancing "subject reader" is only done in specific state of "other reader" and vice versa.

lg2de:
> Thanks for taking a stab at this.
> With all this conditional logic, the number of tests doesn't seem covering.
> E.g. I was able to comment out both `!otherReader.IsEmptyElement` and `&& !otherReader.HasValue` and all tests still passed.

I think, the conditions are required (logically).
I'll try to extend tests.

Src/FluentAssertions/Xml/XmlReaderValidator.cs
dennisdoomen: I get that, but wrapping the reader you can make that same code much more intention-revealing.

lg2de: Please have a look on the new version.

Src/FluentAssertions/Xml/XmlReaderValidator.cs
```diff
@@ -79,11 +77,23 @@ private ValidationResult Validate()
 
                         locationStack.Push(subjectReader.LocalName);
                         validationResult = ValidateAttributes();
+
                         if (subjectReader.IsEmptyElement)
                         {
+                            // TODO Why? Please help to explain
```
lg2de: I'm not sure why this is done.
Can you help please to give a good comment?

dennisdoomen: Is it used to detect cyclic references or something?

jnyrup: Thought: could it be a leftover after I added `reader.MoveToElement();` to `GetAttributes`in #1170?

lg2de: Once again I tried to understand the block.
I think, I understood.
Please check my comment whether you agree.

jnyrup:
I tried to do some manual mutation testing to get a better gut feeling about the logic.
I wanted to verify that no condition can be omitted or negated.
All cases should have test catching the mutation.

The 6 cases marked with "no test" are the ones that didn't fail.
* 3 of those failed as expected when negating the condition instead of omitting, so I'm not concerned about those.
* It seems odd to me that the remaining 3 aren't caught by any tests.

By adding a test with both `subject` and `expected` being `<parent><child></child></parent>` we can kill two mutants. (Unless I messed something up during my testing)

`this.reader.NodeType != XmlNodeType.EndElement`

* omitting: no test
* mutate to `==`: `When_asserting_a_selfclosing_xml_element_is_equivalent_to_a_different_empty_xml_element_it_should_succeed`

`this.reader.LocalName != currentName`

* omitting:
* mutate to `==`: `When_asserting_a_selfclosing_xml_element_is_equivalent_to_a_different_empty_xml_element_it_should_succeed`

`||` in `MoveToEndElement`
* mutate to `&&`: no test
* mutate to `^`: `When_asserting_an_xml_node_is_equivalent_to_different_xml_node_which_contains_extra_elements_it_should_fail_with_descriptive_message`

`if (subjectReader.IsEmptyElement && !otherReader.IsEmptyElement)`

* mutate to `otherReader.IsEmptyElement`: `When_asserting_a_selfclosing_xml_element_is_equivalent_to_a_different_empty_xml_element_it_should_succeed`
* omitting `!otherReader.IsEmptyElement`: no test
* mutate to `!subjectReader.IsEmptyElement`: `When_asserting_a_selfclosing_xml_element_is_equivalent_to_a_different_empty_xml_element_it_should_succeed`
* omitting `subjectReader.IsEmptyElement`: subject: `"<parent><child></child></parent>"`, expected: `"<parent><child></child></parent>"`
* remove entire condition: `When_asserting_a_selfclosing_xml_element_is_equivalent_to_a_different_empty_xml_element_it_should_succeed`

`else if (otherReader.IsEmptyElement && !subjectReader.IsEmptyElement)`

* mutate to `subjectReader.IsEmptyElement`: `When_asserting_an_empty_xml_element_is_equivalent_to_a_different_selfclosing_xml_element_it_should_succeed`
* omitting `!subjectReader.IsEmptyElement`: no test
* mutate to `!otherReader.IsEmptyElement`: `When_asserting_an_empty_xml_element_is_equivalent_to_a_different_selfclosing_xml_element_it_should_succeed`
* omitting `otherReader.IsEmptyElement`: subject: `"<parent><child></child></parent>"`, expected: `"<parent><child></child></parent>"`
* remove entire condition: `When_asserting_an_empty_xml_element_is_equivalent_to_a_different_selfclosing_xml_element_it_should_succeed`


`skipOnce = true;` in `MoveToEndElement()`

* remove: `When_asserting_an_xml_node_is_equivalent_to_different_xml_node_which_contains_extra_elements_it_should_fail_with_descriptive_message`

`reader.MoveToContent();` in `MoveToEndElement()`

* remove: no test

`skipOnce = false;` in `Read()`

* remove: no test

`return true;` in `Read()`

* remove: `When_asserting_an_xml_node_is_equivalent_to_different_xml_node_which_contains_extra_elements_it_should_fail_with_descriptive_message`

dennisdoomen:
@jnyrup you don't have Netflix over there? 😂

lg2de:
@jnyrup, I added two tests and removed a statement which is not really required.
So, I reduced the number of "surviving mutants".

There are left-overs due to `skipOnce`.
I have no idea so far to generate a user-level test (public API) to validate.
I tried to create tests directly on `XmlReaderWrapper`, but failed because internal classes cannot be tested currently in FA. :(

So, I kindly ask to take the current version. The wrapper may (!!!) not need the code when used only in `XmlReaderValidator` (in the current form). But it is required logically and will be required when it is reused in other parts of FA.

PR #1177, opened 2019-11-06
Changed Be/NotBe on IComparable to use Object.Equals and added BeRankedEquallyTo

rubenrorije:
A new pull request for #172 in which all commits have been cleaned up. I struggled with getting Git to behave the way I wanted, but I think this should be it. 

Just to reiterate, this might be a breaking change when released since it does not check CompareTo == 0 anymore. This might be something people depend upon.

### DIFF
--- a/Src/FluentAssertions/Numeric/ComparableTypeAssertions.cs
+++ b/Src/FluentAssertions/Numeric/ComparableTypeAssertions.cs
@@ -19,12 +19,11 @@ namespace FluentAssertions.Numeric
         {
         }
 
+
         /// <summary>
-        /// Asserts that the subject is considered equal to another object according to the implementation of <see cref="IComparable{T}"/>.
+        /// Asserts that an object equals another object using its <see cref="object.Equals(object)" /> implementation.
         /// </summary>
-        /// <param name="expected">
-        /// The object to pass to the subject's <see cref="IComparable{T}.CompareTo"/> method.
-        /// </param>
+        /// <param name="expected">The expected value</param>
         /// <param name="because">
         /// A formatted phrase as is supported by <see cref="string.Format(string,object[])"/> explaining why the assertion
         /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
@@ -35,7 +34,7 @@ namespace FluentAssertions.Numeric
         public AndConstraint<ComparableTypeAssertions<T>> Be(T expected, string because = "", params object[] becauseArgs)
         {
             Execute.Assertion
-                .ForCondition(ReferenceEquals(Subject, expected) || (Subject.CompareTo(expected) == Equal))
+                .ForCondition(Subject.IsSameOrEqualTo(expected))
                 .BecauseOf(because, becauseArgs)
                 .FailWith("Expected {context:object} to be equal to {0}{reason}, but found {1}.", expected, Subject);
 
@@ -110,22 +109,20 @@ namespace FluentAssertions.Numeric
         }
 
         /// <summary>
-        /// Asserts that the subject is not equal to another object according to its implementation of <see cref="IComparable{T}"/>.
+        /// Asserts that an object does not equal another object using its <see cref="object.Equals(object)" /> method.
         /// </summary>
-        /// <param name="unexpected">
-        /// The object to pass to the subject's <see cref="IComparable{T}.CompareTo"/> method.
-        /// </param>
+        /// <param name="unexpected">The unexpected value</param>
         /// <param name="because">
         /// A formatted phrase as is supported by <see cref="string.Format(string,object[])"/> explaining why the assertion
         /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
         /// </param>
         /// <param name="becauseArgs">
-        /// Zero or more objects to format using the placeholders in <see cref="because"/>.
+        /// Zero or more values to use for filling in any <see cref="string.Format(string,object[])" /> compatible placeholders.
         /// </param>
         public AndConstraint<ComparableTypeAssertions<T>> NotBe(T unexpected, string because = "", params object[] becauseArgs)
         {
             Execute.Assertion
-                .ForCondition(Subject.CompareTo(unexpected) != Equal)
+                .ForCondition(!Subject.IsSameOrEqualTo(unexpected))
                 .BecauseOf(because, becauseArgs)
                 .FailWith("Did not expect {context:object} to be equal to {0}{reason}.", unexpected);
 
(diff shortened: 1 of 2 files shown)


jnyrup:
I agree on the change and think the potential runtime breakage sounds reasonable.

From [`IComparer<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.icomparable-1?view=netframework-4.8)
> Defines a generalized comparison method that a value type or class implements to create a type-specific comparison method for ordering or sorting its instances.

Can we offer an alternative or guidance for users who still want to test that two items are equal using their `CompareTo` method?
Ugly workarounds would be
```c#
subject.Should().BeLessOrEqualTo(expected)
    .And.BeGreaterOrEqualTo(expected);

subject.Should().BeInRange(expected, expected);
```

Explicit test but inferior failure message.
```c#
subject.CompareTo(expected).Should().Be(0);
```

rubenrorije:
How about naming it `subject.Should().BeRankedEqualTo(expected)`?
About guidance, when such a method exists an explanation and reference can be included in the comments of `Be` and `NotBe`.

Tests/Shared.Specs/ComparableSpecs.cs
```diff
@@ -538,6 +571,43 @@ public override string ToString()
         }
     }
 
+    public class SameInstanceIsNotEqualClass
```
dennisdoomen: I would keep these helper classes right after the last test method that uses it. We did this differently in the past, but I think this aligns closer to the "reads like a book" from Clean Code.

dennisdoomen:
> How about naming it `subject.Should().BeRankedEqualTo(expected)`?

Love that one. Probably should be `BeRankedAsEqualTo`

dennisdoomen:
@rubenrorije are you planning to add that `BeRankedAsEqualTo`?

rubenrorije:
Sorry, made the same git mistakes again. I am obviously still a noob on the git part. I will try to fix this.

rubenrorije:
Could you give me some pointers to resolve the git issues here? I assume that the rebase I did resulted in the commits 9050e37 and c477f8b to be included. But I am a bit lost how to resolve this.

jnyrup:
@rubenrorije It looks like you pulled `master` after rebasing on it.
When rebasing, you'll need to push force.
`git push --force`/`git push -f` (There's also a `force-with-lease` option that does some additional checking)

In git commits are identified by a hash value computed on the difference between the commit and its parent commit.

Before rebasing both your local branch and the remote branch on github looks like:
```
master --> ... --> HEAD
 |
 + --> PR1177
```

After rebasing your local branch the picture is now
local:
```
master --> ... --> HEAD
                    |
                    + --> PR1177' // note the '
```

remote:
```
master --> ... --> HEAD
 |
 + --> PR1177
```

As `PR1177` and `PR1177'` has different parent commits locally and remote, git sees them as two different commits.
This step can be confusing as git will now report that there are remote changes, which your local branch does not have. 
It's technically true, due to different hash values, but we *know* it's because of our rebase.

So if you pull master instead of force pushing the local picture becomes:
```
master --> ... --> HEAD
                    |
                    + --> PR1177 --> PR1177'
```

If you rebase again on `master`, you can remove the unwanted commits from the history of this branch.
I usually do this via terminal
```
git rebase -i master
```
The terminal opens some editor (for me it's Vim)
I then write replace `pick` with `drop` for the commits I want to remove.
Save the file and exit.
The rewritten history can be seen with
`git log master...`
Then `git push -f` to push the local branch to the remote repository.

rubenrorije:
Thanks for the extensive explanation. That made things much clearer. I was able to fix the issues now. I did not want to screw everything up again by rebasing again. When the pull request is not accepted directly it must be done again anyway.

jnyrup:
I haven't fallen in love with `BeRankedAsEqualTo` (yet), so I would like to brainstorm a bit on the naming before settling.

* `BeComparableTo`
* `CompareEqualTo`
* `RankEqualTo`
* `BeEqualTo`

`BeEqualTo` is not used anywhere else and it's in line with naming of e.g. `BeGreaterThan`, but I'm not sure whether it's confusingly close to `Be`

rubenrorije:
I agree discussing this a bit more might be useful. I do not have a strong preference. 

> `BeEqualTo` is not used anywhere else and it's in line with naming of e.g. `BeGreaterThan`, but I'm not sure whether it's confusingly close to `Be`

If I would see the options `Be` and `BeEqualTo` after typing `Should()` I would (wrongfully) assume from the method names that `Be` checks reference equality and `BeEqualTo` equality. When also considering the similarly named `BeEquivalentTo`, it might not be that intuitive which one does what.

Slightly related and maybe against the whole idea of the project, but I think it would be nice to have one (additional?) function to check all comparisons:
`subject.Should().BeComparable(instanceLess, instanceEqual, instanceGreater, because)` that combines all checks.  I assume this is done now by multiple test methods and/or multiple assertions within a test method.

dennisdoomen:
I could live with `BeComparableTo`, but not with the others.

dennisdoomen:
What about `BeRankedEquallyTo`?

jnyrup:
I'm fine with `BeRankedEquallyTo`.

rubenrorije:
I do not have any preference, so I renamed the methods.

Src/FluentAssertions/Numeric/ComparableTypeAssertions.cs
```diff
@@ -122,12 +171,12 @@ public AndConstraint<ComparableTypeAssertions<T>> Be(T expected, string because
         /// <param name="becauseArgs">
         /// Zero or more objects to format using the placeholders in <see cref="because"/>.
         /// </param>
-        public AndConstraint<ComparableTypeAssertions<T>> NotBe(T unexpected, string because = "", params object[] becauseArgs)
+        public AndConstraint<ComparableTypeAssertions<T>> NotBeRankedEquallyTo(T expected, string because = "", params object[] becauseArgs)
```
jnyrup: `expected` --> `unexpected`

dennisdoomen:
What's stopping us from merging this one?

Src/FluentAssertions/Numeric/ComparableTypeAssertions.cs
```diff
@@ -110,7 +111,55 @@ public AndConstraint<ComparableTypeAssertions<T>> Be(T expected, string because
...
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more values to use for filling in any <see cref="string.Format(string,object[])" /> compatible placeholders.
+        /// </param>
+        public AndConstraint<ComparableTypeAssertions<T>> BeRankedEquallyTo(T expected, string because = "", params object[] becauseArgs)
+        {
+            Execute.Assertion
+                .ForCondition(Subject.CompareTo(expected) == Equal)
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected {context:object} {0} to be be ranked as equal to {1}{reason}.", Subject, expected);
```
jnyrup: `be be`

dennisdoomen:
Now the unit test is failing on the changed message.

PR #1178, opened 2019-11-06
Add Have/NotHaveSameCount for Dictionaries

### LINKED ISSUE #1176: GenericDictionaryAssertions doesn't have many of the GenericCollectionAssertions methods
### Description

I was confused when working with a dictionary that `HaveSameCount` didn't seem to be resolving properly.  I would expect that `GenericDictionaryAssertions` would derive from `GenericCollectionAssertions` to provide the same methods.

It looks like most of these methods are just explicitly implemented on DictionaryAssertions, would it be reasonable to add a few more, or is there a reason that these methods aren't included?

veleek:
Adding support for `HaveSameCount` and `NotHaveSameCount` for dictionaries.  This was almost directly cribbed from the existing `CollectionAssertions` (including test cases).

Resolves #1176 

## IMPORTANT 

* [X] If the contribution affects the documentation, please include your changes to [**documentation.md**](https://github.com/fluentassertions/fluentassertions/blob/master/docs/_pages/documentation.md) in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com).

### DIFF
--- a/Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
```diff
@@ -242,6 +243,87 @@ public GenericDictionaryAssertions(IDictionary<TKey, TValue> dictionary) : base(
...
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<GenericDictionaryAssertions<TKey, TValue>> HaveSameCount(IEnumerable otherCollection, string because = "",
+            params object[] becauseArgs)
+        {
+            Guard.ThrowIfArgumentIsNull(otherCollection, nameof(otherCollection), "Cannot compare dictionary count against a <null> collection.");
+
+            if (ReferenceEquals(Subject, null))
```
dennisdoomen: Considering using `is null`.

Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
```diff
@@ -242,6 +243,87 @@ public GenericDictionaryAssertions(IDictionary<TKey, TValue> dictionary) : base(
...
+        /// </summary>
+        /// <param name="otherCollection">The other collection with the same expected number of elements</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndConstraint<GenericDictionaryAssertions<TKey, TValue>> HaveSameCount(IEnumerable otherCollection, string because = "",
```
dennisdoomen: Shouldn't this method take a dictionary?

jnyrup: That would prevent you from writing the following test

```c#
// Arrange
var subject = new List<int> { 1, 2, 3};

// Act
Dictionary<int, int> actual = ConvertToDictionary(subject);

// Assert
actual.Should().HaveSameCount(subject);
```

Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
jnyrup: The reason `is null` isn't used in `CollectionAssertions.HaveSameCount` is because `TSubject` is not constrained to reference types and `is null` can only be used on reference types.

Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
veleek: Agreed, this is the main scenario for me, converting to/from dictionaries and then validating the results.

veleek:
If there are no other concerns, what's the process for getting this merged?

Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
dennisdoomen: SonarQube keeps telling me (in another project) to compare to `default(T)` ;-)

jnyrup:
@veleek Could I ask you to change the two `ReferenceEquals(Subject, null)` to `Subject is null`. 
That will align with the rest of the project where using (generically constrained) reference types?
Then I'll be happy to push to the merge button 👍

jnyrup:
@veleek Thanks for the contribution!

dennisdoomen:
> Thanks for the contribution!

Indeed. Another nice little gem.

PR #1180, opened 2019-11-07
Add new namespace assertions to TypeSelectorAssertions

### LINKED ISSUE #1175: Add new namespace assertions to TypeSelectorAssertions
### Description

Currently, TypeSelectorAssertions only contains methods for asserting that types are (not) decorated with specific attributes, e.g., `BeDecoratedWith<>`. However, TypeSelector contains more selectors that would also be useful as assertions.

One such group of selectors filter on the namespace of types and contain the methods `ThatAreInNamespace`, `ThatAreNotInNamespace`, `ThatAreUnderNamespace`, and `ThatAreNotUnderNamespace`. Corresponding assertions could be named `BeInNamespace`, `NotBeInNamespace`, `BeUnderNamespace`, and `NotBeUnderNamespace`.

### Use case

A concrete use case I have that prompted this suggestion is that the JsonSubTypes library, which is used to deserialize JSON into one of a number of subclasses based on a discriminator value, [only supports deserializing into subclasses in the same namespace as the base class](https://github.com/manuc66/JsonSubTypes#deserializeobject-with-custom-type-property-name). Therefore, I would like to be able to write a test that asserts that all subclasses of a specific class are in the same namespace as it, à la:

``` cs
AllTypes.From(baseClassAssembly).ThatDeriveFrom<BaseClass>().Should().BeInNamespace(baseClassNamespace);
```

### Draft implementation as custom extension

To support the use case above, I have implemented the `BeInNamespace` assertion in our project as the following custom extension:

``` cs
public static class TypeSelectorAssertionsExtensions
{
    public static AndConstraint<TypeSelectorAssertions> BeInNamespace(this TypeSelectorAssertions typeSelectorAssertions, string @namespace, string because = "", params object[] becauseArgs)
    {
        Execute.Assertion
               .ForCondition(@namespace != null)
               .FailWith("Types cannot have a null namespace")
               .Then
               .Given(() => typeSelectorAssertions.Subject)
               .ForCondition(types => types.All(type => type.Namespace == @namespace))
               .FailWith("Expected types to be in namespace {0}{reason}, but found {1}.",
                   _ => @namespace, types => types.Where(type => type.Namespace != @namespace));

        return new AndConstraint<TypeSelectorAssertions>(typeSelectorAssertions);
    }
}
```

kimsey0:
## IMPORTANT 

Fixes #1175


kimsey0:
@jnyrup, @dennisdoomen: Does this follow the right pattern? Then I'll add the rest of the methods.

jnyrup:
Looks good to me 👍

Tests/Shared.Specs/TypeAssertionSpecs.cs
```diff
@@ -1445,6 +1446,50 @@ public void When_injecting_a_null_predicate_into_TypeSelector_NotBeDecoratedWith
...
+            var types = new TypeSelector(new[]
+            {
+                typeof(ClassInDummyNamespace),
+                typeof(ClassNotInDummyNamespace),
+                typeof(OtherClassNotInDummyNamespace)
+            });
+
+            // Act
+            Action act = () =>
+                types.Should().BeInNamespace(nameof(DummyNamespace), "because reasons");
```
jnyrup: One minor thing, please also include a single `becauseArgs`.
We usually use something alike
`"because we want to test the error {0}", "message"`

kimsey0: Will do.

kimsey0:
I have added the rest of the assertions together with tests for the oddities discussed in #1175. I'm still missing some documentation.

Tests/Shared.Specs/TypeAssertionSpecs.cs
```diff
@@ -1445,6 +1448,451 @@ public void When_injecting_a_null_predicate_into_TypeSelector_NotBeDecoratedWith
...
+            act.Should().Throw<XunitException>()
+                .WithMessage("Expected all types to be in namespace <null>" +
+                             " because we want to test the error message," +
+                             " but the following types are in a different namespace:*" +
+                             "*ClassInDummyNamespace*" +
+                             "*ClassNotInDummyNamespace*" +
+                             "*OtherClassNotInDummyNamespace*");
+        }
+
+        [Fact]
```
jnyrup: Please put the `BeInNamespace` and the `NotBeInNamespace` tests in separate `#regions`s

kimsey0: Done!

Src/FluentAssertions/Types/TypeSelectorAssertions.cs
```diff
@@ -292,6 +292,131 @@ public AndConstraint<TypeSelectorAssertions> NotBeDecoratedWithOrInherit<TAttrib
...
+        public AndConstraint<TypeSelectorAssertions> NotBeInNamespace(string @namespace, string because = "", params object[] becauseArgs)
+        {
+            Type[] typesInNamespace = Subject
+                .Where(t => t.Namespace == @namespace)
+                .ToArray();
+
+            Execute.Assertion
+                .ForCondition(!typesInNamespace.Any())
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected all types to not be in namespace {0}{reason}," +
```
dennisdoomen: _Expected none of the types to be in namespace_

jnyrup:
@dennisdoomen I don't think it's ready for review yet?

Src/FluentAssertions/Types/TypeSelectorAssertions.cs
kimsey0: One could consider changing the wording in TypeAssertionSpecs as well, for example at https://github.com/fluentassertions/fluentassertions/blob/767a86bbf74ba5894102e24b63dbc004fc0d308a/Tests/Shared.Specs/TypeAssertionSpecs.cs#L1254

kimsey0:
@jnyrup, @dennisdoomen: I've adjusted this for the null and prefix namespace handling changes. How does it look?

kimsey0:
(Documentation is still missing. I'll get to that.)

kimsey0:
The _Type, Method, and Property assertions_ documentation section doesn't document all type assertions, but I've added a call to `BeInNamespace`, just to make readers aware of the namespace assertions.

Tests/Shared.Specs/TypeAssertionSpecs.cs
```diff
@@ -1445,6 +1448,470 @@ public void When_injecting_a_null_predicate_into_TypeSelector_NotBeDecoratedWith
...
+            var types = new TypeSelector(new[]
+            {
+                typeof(ClassInDummyNamespace),
+                typeof(ClassNotInDummyNamespace),
+                typeof(OtherClassNotInDummyNamespace)
+            });
+
+            // Act
+            Action act = () =>
+                types.Should().BeInNamespace(nameof(DummyNamespace),
```
jnyrup: `BeInNamespace` -> `NotBeInNamespace`?

dennisdoomen: @kimsey0 would be nice to wrap this up

dennisdoomen: @kimsey0 ?

dennisdoomen:
@jnyrup What should we do with this PR? Finish it ourselves? Looks like @kimsey0 has abandoned it.

jnyrup:
@dennisdoomen I'll wrap it up.

PR #1181, opened 2019-11-07
Add index to XPath information of XDocument equivalence failure message

lg2de:
I think, the XDocument equivalence failure message is not helpful when comparing documents with repeating element names, like HTML documents. By just providing the simple XPath it may be very difficult to identify the element with difference:
`/html/body/table/tr/td`

This PR is a proposal to improve the failure message by writing index number (according XPath syntax) when a specific element was found at least second time on same level:
`/html/body/table[2]/tr[17]/td[5]`

Additionally
* I recommend a new setting for `editorconfig` according to the current code style
* I changed a test name according to the test content

### DIFF
--- a/Src/FluentAssertions/Xml/XmlReaderValidator.cs
+++ b/Src/FluentAssertions/Xml/XmlReaderValidator.cs
@@ -13,32 +13,8 @@ namespace FluentAssertions.Xml
         private readonly XmlReaderWrapper subjectReader;
         private readonly XmlReaderWrapper otherReader;
 
-        private string GetCurrentLocation()
-        {
-            var keyBuilder = new StringBuilder();
-            var resultBuilder = new StringBuilder();
-            foreach (var location in locationStack.Reverse())
-            {
-                keyBuilder.Append($"/{location}");
-                resultBuilder.Append($"/{location}");
-                var locationCount = locationCounts[keyBuilder.ToString()];
-                if (locationCount > 1)
-                {
-                    resultBuilder.Append($"[{locationCount}]");
-                }
-            }
-
-            string result = resultBuilder.ToString();
-            if (string.IsNullOrEmpty(result))
-            {
-                return "/";
-            }
-
-            return result;
-        }
-
         private readonly Stack<string> locationStack = new Stack<string>();
-        private readonly Dictionary<string, int> locationCounts = new Dictionary<string, int>();
+        private readonly Dictionary<string, int> countPerElementPath = new Dictionary<string, int>();
 
         public XmlReaderValidator(XmlReader subjectReader, XmlReader otherReader, string because, object[] reasonArgs)
         {
@@ -46,19 +22,6 @@ namespace FluentAssertions.Xml
 
             this.subjectReader = new XmlReaderWrapper(subjectReader);
             this.otherReader = new XmlReaderWrapper(otherReader);
-        }
-
-        private class ValidationResult
-        {
-            public ValidationResult(string formatString, params object[] formatParams)
-            {
-                FormatString = formatString;
-                FormatParams = formatParams;
-            }
-
-            public string FormatString { get; }
-
-            public object[] FormatParams { get; }
         }
 
         public void Validate(bool expectedEquivalence)
@@ -83,7 +46,7 @@ namespace FluentAssertions.Xml
                 if (subjectReader.NodeType != otherReader.NodeType)
                 {
                     return new ValidationResult("Expected node of type {0} at {1}{reason}, but found {2}.",
-                        otherReader.NodeType, GetCurrentLocation(), subjectReader.NodeType);
+                        otherReader.NodeType, GetCurrentXPath(), subjectReader.NodeType);
                 }
 
                 ValidationResult validationResult = null;
@@ -104,9 +67,7 @@ namespace FluentAssertions.Xml
                         locationStack.Push(subjectReader.LocalName);
 
                         // building key according XPath to count repeating element names
-                        var locationKey = "/" + string.Join("/", locationStack.Reverse());
-                        locationCounts.TryGetValue(locationKey, out var locationCount);
-                        locationCounts[locationKey] = ++locationCount;
+                        UpdateElementPathCount();
 
                         validationResult = ValidateAttributes();
 
@@ -138,7 +99,7 @@ namespace FluentAssertions.Xml
                         validationResult = ValidateText();
                         break;
                     default:
-                        throw new NotSupportedException($"{subjectReader.NodeType} found at {GetCurrentLocation()} is not supported for equivalency comparison.");
+                        throw new NotSupportedException($"{subjectReader.NodeType} found at {GetCurrentXPath()} is not supported for equivalency comparison.");
                 }
 
                 if (validationResult != null)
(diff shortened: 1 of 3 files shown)


Src/FluentAssertions/Xml/XmlReaderValidator.cs
```diff
@@ -77,7 +101,15 @@ private ValidationResult Validate()
                             return validationResult;
                         }
 
+                        // starting new element, add local name to location stack
+                        // to build XPath info
                         locationStack.Push(subjectReader.LocalName);
+
+                        // building key according XPath to count repeating element names
+                        var locationKey = "/" + string.Join("/", locationStack.Reverse());
```
dennisdoomen: I don't get this. Aren't you building the entire path expression? How can that be used as a key?

Src/FluentAssertions/Xml/XmlReaderValidator.cs
```diff
@@ -12,9 +13,32 @@ internal class XmlReaderValidator
         private readonly XmlReader subjectReader;
         private readonly XmlReader otherReader;
 
-        private string GetCurrentLocation() => "/" + string.Join("/", locationStack.Reverse());
+        private string GetCurrentLocation()
```
dennisdoomen: I recommend renaming it to `GetCurrentXPath`. We also prefer to order methods by invocation so you can read from top to bottom. In this case, it means moving this method under `ValidateText`.

Src/FluentAssertions/Xml/XmlReaderValidator.cs
```diff
@@ -12,9 +13,32 @@ internal class XmlReaderValidator
...
+                resultBuilder.Append($"/{location}");
+                var locationCount = locationCounts[keyBuilder.ToString()];
+                if (locationCount > 1)
+                {
+                    resultBuilder.Append($"[{locationCount}]");
+                }
+            }
+
+            string result = resultBuilder.ToString();
+            if (string.IsNullOrEmpty(result))
```
dennisdoomen: This can never be `null`, right?

Src/FluentAssertions/Xml/XmlReaderValidator.cs
```diff
@@ -12,9 +13,32 @@ internal class XmlReaderValidator
...
+            {
+                return "/";
+            }
+
+            return result;
+        }
 
         private readonly Stack<string> locationStack = new Stack<string>();
+        private readonly Dictionary<string, int> locationCounts = new Dictionary<string, int>();
```
dennisdoomen: Isn't this the count per element? The current name confused me a bit.

Src/FluentAssertions/Xml/XmlReaderValidator.cs
jnyrup: Given the xml `<a><b /><b /></a>`
* `<a>`: 
  * push `a` on the stack
  * update the dictionary with `/a` = 1
* `<b />`
  * push `b` on the stack
  * update the dictionary with `/a/b` = 1 <-- first time we see `<b>` on the `/a` xpath
  * pop `b` from the stack
* `<b />`
  * push `b` on the stack
  * update the dictionary with `/a/b` = 2 <-- second time we see `<b>` on the `/a` xpath
  * pop `b` from the stack
* `</a>`
  * pop `a` from the stack

Src/FluentAssertions/Xml/XmlReaderValidator.cs
```diff
@@ -77,7 +101,15 @@ private ValidationResult Validate()
...
                         }
 
+                        // starting new element, add local name to location stack
+                        // to build XPath info
                         locationStack.Push(subjectReader.LocalName);
+
+                        // building key according XPath to count repeating element names
+                        var locationKey = "/" + string.Join("/", locationStack.Reverse());
+                        locationCounts.TryGetValue(locationKey, out var locationCount);
+                        locationCounts[locationKey] = ++locationCount;
```
jnyrup: 🔧How about extracting these lines into a method with some appropriate method name?

Src/FluentAssertions/Xml/XmlReaderValidator.cs
jnyrup: Right, it cannot be `null`, at most empty. 

Could also be written as
```c#
if (resultBuilder.Length == 0)
{
    return "/";
}

string result = resultBuilder.ToString();
return result;
```

Src/FluentAssertions/Xml/XmlReaderValidator.cs
lg2de: I will do so as soon you have merged #1174.

dennisdoomen: Done

lg2de:
Do you think that this change of error message is a breaking change? 
If users asserting XPath using `WithMessage` existing tests may fail.

jnyrup:
While it technically can be a breaking change, as in tests failing after upgrading the package, I don't recall we've considered improvements/changes to failure messages as breaking changes.

Tests/Shared.Specs/XDocumentAssertionSpecs.cs
```diff
@@ -490,8 +490,23 @@ public void when_an_xml_element_has_attributes_and_is_self_closing_it_should_suc
...
+        [Fact]
+        public void When_asserting_equivalence_of_document_with_repeating_element_names_but_differs_it_should_fail_with_index_xpath_to_difference()
+        {
+            // Arrange
+            XDocument actual = XDocument.Parse(
+                "<xml><xml2 /><xml2 /><xml2><a x=\"y\"/><b><sub /></b><a x=\"y\"/></xml2></xml>");
+            XDocument expected = XDocument.Parse(
+                "<xml><xml2 /><xml2 /><xml2><a x=\"y\"/><b><sub /></b><a x=\"z\"/></xml2></xml>");
+
+            // Arrange
```
jnyrup: `// Arrange` -> `// Act`

Src/FluentAssertions/Xml/XmlReaderValidator.cs
lg2de: Mmh, this is the only usage, but it reduces the LOC in the `switch`.
Maybe, more statements should be moved to method (in another PR)!?

Tests/Shared.Specs/XDocumentAssertionSpecs.cs
lg2de: Nice catch!
Copied from existing test introduced by fa46521 from ... ;)

dennisdoomen:
> While it technically can be a breaking change, as in tests failing after upgrading the package, I don't recall we've considered improvements/changes to failure messages as breaking changes.

Why would a test fail? Because somebody observe FA's failure messages in an exception assertion?

lg2de:
> > While it technically can be a breaking change, as in tests failing after upgrading the package, I don't recall we've considered improvements/changes to failure messages as breaking changes.
> 
> Why would a test fail? Because somebody observe FA's failure messages in an exception assertion?

While creating an example I found that my question is not relevant.
Only when asserting FA failure message the change can cause failing tests. But this is not a normal use case.

jnyrup:
I came to think of another way to combine the `Stack` and `Dictionary` into a single graph that keeps track of the xpath indices.

https://github.com/jnyrup/fluentassertions/commit/711a7f992a2cf0fb464d7180a76e36c1bab0afdf

Some very primitive benchmarks:
* `N` is the depth of the xml tree.
* xml tags are named `a0...aN`


# 10 children per node.

## This PR
```
|                     Method | N |         Mean |      Error |     StdDev | Ratio |    Gen 0 | Gen 1 | Gen 2 |  Allocated |
|--------------------------- |-- |-------------:|-----------:|-----------:|------:|---------:|------:|------:|-----------:|
| BeEquivalentTo             | 1 |     5.234 us |  0.1092 us |  0.1214 us |  1.00 |   1.3351 |     - |     - |    4.11 KB |
|                            |   |              |            |            |       |          |       |       |            |
| BeEquivalentTo             | 2 |    25.672 us |  0.2558 us |  0.2393 us |  1.00 |   4.6082 |     - |     - |   14.17 KB |
|                            |   |              |            |            |       |          |       |       |            |
| BeEquivalentTo             | 5 | 3,828.342 us | 30.7972 us | 24.0445 us |  1.00 | 648.4375 |     - |     - | 1995.18 KB |
```

## My branch
```
|                     Method | N |         Mean |      Error |     StdDev | Ratio |    Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------- |-- |-------------:|-----------:|-----------:|------:|---------:|------:|------:|----------:|
| BeEquivalentTo             | 1 |     2.987 us |  0.0256 us |  0.0227 us |  1.00 |   1.0529 |     - |     - |   3.24 KB |
|                            |   |              |            |            |       |          |       |       |           |
| BeEquivalentTo             | 2 |    12.788 us |  0.2393 us |  0.2122 us |  1.00 |   2.9602 |     - |     - |   9.14 KB |
|                            |   |              |            |            |       |          |       |       |           |
| BeEquivalentTo             | 5 | 1,558.388 us | 16.2188 us | 13.5434 us |  1.00 | 289.0625 |     - |     - | 890.32 KB |
```


# 4 children per node.

## This PR
```
|                     Method |  N |            Mean |         Error |        StdDev | Ratio |       Gen 0 | Gen 1 | Gen 2 |     Allocated |
|--------------------------- |--- |----------------:|--------------:|--------------:|------:|------------:|------:|------:|--------------:|
| BeEquivalentTo             |  3 |        71.90 us |      1.375 us |      1.351 us |  1.00 |     11.8408 |     - |     - |      36.59 KB |
|                            |    |                 |               |               |       |             |       |       |               |
| BeEquivalentTo             |  5 |     1,346.27 us |     36.026 us |     30.084 us |  1.00 |    224.6094 |     - |     - |     694.78 KB |
|                            |    |                 |               |               |       |             |       |       |               |
| BeEquivalentTo             | 10 | 1,871,360.79 us | 27,158.727 us | 24,075.504 us |  1.00 | 328000.0000 |     - |     - | 1009116.03 KB |
```

## My branch
```
|                     Method |  N |          Mean |         Error |       StdDev |        Median | Ratio |       Gen 0 | Gen 1 | Gen 2 |    Allocated |
|--------------------------- |--- |--------------:|--------------:|-------------:|--------------:|------:|------------:|------:|------:|-------------:|
| BeEquivalentTo             |  3 |      34.01 us |      0.678 us |     1.170 us |      33.47 us |  1.00 |      7.0190 |     - |     - |     21.63 KB |
|                            |    |               |               |              |               |       |             |       |       |              |
| BeEquivalentTo             |  5 |     534.87 us |      6.195 us |     5.492 us |     535.04 us |  1.00 |    101.5625 |     - |     - |    312.93 KB |
|                            |    |               |               |              |               |       |             |       |       |              |
| BeEquivalentTo             | 10 | 592,578.85 us | 10,082.794 us | 8,419.590 us | 594,102.50 us |  1.00 | 103000.0000 |     - |     - | 317702.62 KB |
```

dennisdoomen:
@jnyrup do you want to do anything with those performance results? Or can this be merged?

lg2de:
> @jnyrup do you want to do anything with those performance results? Or can this be merged?

Good question.

So far, I was unsure what to do.
I did not have time to analyzed in detail was @jnyrup has changed.
If performance is better, then I'm fine.
Should I merge it into "my" PR?

jnyrup:
@lg2de You can merge the commit into this PR.
Its commit message (hopefully) describes the idea behind the visit count graph.
In the end it does the same as your algorithm, but more lazily to avoid stack iterations, string allocations and dictionary lookups.

> ### Use a counting graph to keep track of xpath index
> 
> Currently we use a stack to keep track of the location and a dictionary to keep track of the xpath index.
> That has multiple downsides:
> 
> * We need to keep two separate data structure in sync.
> * We're a lot of string allocations on each iterations when computing path as key for the dictionary.
> * When creating the failure message we're doing lookups in two data structures
> 
> Instead we can combine their purposes in a single graph where each node has a `Name` and a `Count`, where the latter is updated each time we visit a node with the same path.
> 
> On each xml traversal we now only need to find the matching sibling or insert a new child.
> 
> Creating the failure message is simply the travel from the current node to the root node in reverse order.

lg2de:
@jnyrup why do you prefer
```
while (current.parent is object)
```
over
```
while (current.parent != null)
```
?

dennisdoomen:
The second one can trigger an `!= operator`. The first one is a new C# construct that makes it immediately clear what you're expecting (providing you know the construct ;-))

jnyrup:
https://intellitect.com/check-for-null-not-null/ has a nice write-up of different ways to compare against `null`.

By using `is object`, `is null` or `is {}` I don't have to care about whether a potentially overloaded `==` or `!=` operator has proper null checking.

Here's an example of the differences for value and reference types:
(The `{}` is a )
https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LpGYYBGA9uQDboBqAhlQK4CmAKgJ4AOLcAFAEsAdgBd0AgJToAvAD5x4gM7pypAFYsAxiIDchYvqJp0FanUatOPGINHipchQEJp6IUypU9SYukMkTShoGZnZuFlRbMUkZeQElfABfb19/f2NTGgAlFgAzFgAnFiFNMJ5+XBjHeIFlVQ1tFIMfYgyg9Bz8opKylhtKhzj0FzcPLzSWozJ2zsLi0qsIvgHYhVqkpqJ/AHpt9AApJkUxEXJ0FgAPFgBbLioBXI50ehMmAHM3p4FbqhvikXoIgE5CE6QwUDgADZAmZyDwCoDyAVpNJls8ADQ4EyDdAiAAWBXIAHc3CwSQA5MazbqlACiF1KXCBIL4Ek2AQh0MyKnhiIKLjR9ExuFIOKcfBeKOx3kSQA===

lg2de:
Mmh, understood. It feels overengineered to me.

I've applied jnyrup improvement (leaving the mentioned statement) and did some cleanup.
Would you like to merge now? Or do you still missing something?

jnyrup:
Is that 
> leaving the mentioned statement

referring to this?
```
// starting new element, add local name to location stack
// to build XPath info
```

If so, everything looks good to me 👍

lg2de:
No, sorry. 
I was referring to the null check. 
;)

jnyrup:
@dennisdoomen whenever you're ready ;)

lg2de:
> This class needs a bit of love. I'll take a look after merging this one.

What do you mean?

I just want to note, that I think about to provide new implementation for XDocument comparison.
Recently I worked often with XDocument/XElement and FluentAssertions. I found the output is still sometimes not helpful. I would like to have results like using `BeEquivalentTo` on complex classes. Hope I'll find time for this soon...

PR #1193, opened 2019-11-26
Fix unexpected IsDotNet behavior for non-dotnet namespaces

FLeXyo:
The check in `IsDotNet` only checks if the namespace starts with "System" which can cause non-dotnet namespaces like `SystemSomething.RestApi` to return true unexpectedly.

I struggled with this one at work today for a few hours as our product name (and therefor our namespaces) are `SystemProductName.RestApi`. I noticed it because the variable names were missing from our failing asserts even when building in debug mode.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/CallerIdentifier.cs
+++ b/Src/FluentAssertions/CallerIdentifier.cs
@@ -66,8 +66,11 @@ namespace FluentAssertions
 
         private static bool IsDotNet(StackFrame frame)
         {
-            return frame.GetMethod().DeclaringType.Namespace
-                ?.StartsWith("system", StringComparison.InvariantCultureIgnoreCase) == true;
+            var frameNamespace = frame.GetMethod().DeclaringType.Namespace;
+            var comparisonType = StringComparison.InvariantCultureIgnoreCase;
+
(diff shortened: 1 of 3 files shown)


jnyrup:
Thanks for pointing this out *and* submitting a PR.
Would you be able to create a unit test to verify the bug and the fix?

Side note: Less than three weeks ago the same flawed logic was discovered for [`ThatAreUnderNamespace`](https://github.com/fluentassertions/fluentassertions/issues/1175#issuecomment-551554516)

FLeXyo:
@jnyrup sure I'll give it a try. Any suggestions on where to write the unit tests?

jnyrup:
We don't seem to have tests targeting `CallerIdentifier` directly, so feel free to create a new test class named `CallerIdentifierSpecs.cs`.

FLeXyo:
Hopefully this does it, but I have my doubts :) I'm quite new to unit testing and this library, and testing the thing that you use to test certainly makes it a bit more confusing 🥴

No idea why the appveyor build fails as it works fine for me locally with `build.ps1`.

jnyrup:
@FLeXyo Thanks for fixing this!
The failing tests are a recurring issue #1085

PR #1201, opened 2019-12-08
Support `SatisfyRespectively` for string collections

### LINKED ISSUE #1188: StringCollectionAssertions missing SatisfyRespectively GenericCollectionAssertions
### Description

GenericCollectionAssertions has SatisfyRespectively, which is not present in the more specialized StringCollectionAssertions - I would expect more specialized Assertions classes to add, but not remove functionality.

Seems like it could've happened elsewhere, but I don't have the time to track down any other instances.

```csharp
new object[] { }.Should().SatisfyRespectively(x => { }); //Should() is GenericCollectionAssertions
new string[] { }.Should().SatisfyRespectively(...) // does not exist, Should() is StringCollectionAssertions
```

### Versions

5.9.0

.NET Core 2.2

jnyrup:
By moving `SatisfyRespectively` from `GenericCollectionAssertions` to `SelfReferencingCollectionAssertions`  we enable string collections to re-use the same assertions.

To the best of my knowledge, this should not be a breaking change.

* Moving members to parent class is not a breaking change

* The return type has changed from `AndConstraint<GenericCollectionAssertions<T>>` to `AndConstraint<TAssertions>`, but for `GenericCollectionAssertions<T>` `TAssertions` *is* exactly `GenericCollectionAssertions<T>`.

This fixes #1188


dennisdoomen:
> * The return type has changed from `AndConstraint<GenericCollectionAssertions<T>>` to `AndConstraint<TAssertions>`, but for `GenericCollectionAssertions<T>` `TAssertions` _is_ exactly `GenericCollectionAssertions<T>`.

Isn't changing the return type always a breaking change?

jnyrup:
Hmm... You're probably right.
I'll change it to have be in both `GenericCollectionAssertions<T>` and `StringCollectionAssertions` for now.
Can consolidate it in 6.0

jnyrup:
Pulling `SatisfyRespectively` up into `SelfReferencingCollectionAssertions` is not a breaking change.
Asked on SO and got an answer from Eric Lippert himself.
https://stackoverflow.com/questions/59255869/is-pulling-up-a-method-into-a-generic-base-class-a-breaking-change#comment104785181_59257698

jnyrup:
The return type of `GenericCollectionAssertions<T>.SatisfyRespectively` does not change as:
* `SelfReferencingCollectionAssertions<T, TAssertions>.SatisfyRespectively` returns `AndConstraint<TAssertions>`
* `GenericCollectionAssertions<T>` derives from `SelfReferencingCollectionAssertions<T, GenericCollectionAssertions<T>>`
* So `GenericCollectionAssertions<T>.SatisfyRespectively` will still return `AndConstraint<GenericCollectionAssertions<T>>`

I asked the question on SO and got an answer from Eric Lippert himself.
https://stackoverflow.com/questions/59255869/is-pulling-up-a-method-into-a-generic-base-class-a-breaking-change/59257698#59257698

PR #1202, opened 2019-12-08
Be more strict when trying to convert types in IsSameOrEqualTo

### LINKED ISSUE #1010: BeEquivalentTo on double collections
```c#
new List<double> { 0.02 }.Should().BeEquivalentTo(0   );
new List<double> { 0    }.Should().BeEquivalentTo(0.02);
new List<double> { 0.02 }.Should().BeEquivalentTo(0.00);
new List<double> { 0.00 }.Should().BeEquivalentTo(0.02);
```
The first line passes, all the others fail.

I can see an argument for ".BeEquivalentTo()" isn't, in general, meaningful on collections of doubles/floats, in the same way that ".Be()" or ".BeEqualTo()" isn't.
But it should at least be self-consistent, shouldn't it?

### Versions
FA v5.6.0
.Net Fmwk 4.6.1
C# 7.3

### LINKED ISSUE #1183: Should.be() fails when comparing object and true
# Description
If I have an `object` and I use `Should().Be()` to compare it to a boolean, it will always pass, if passed in `true`, but fail correctly when passed in `false`.

See the example below:

```csharp
            // For unrelated reasons I have an object as a number
            object myNumber = 10;
            // This should not work, they're not identical, but it passes
            myNumber.Should().Be(true);
            // The opposite arrangement fails properly
            ((object)true).Should().Be(myNumber);
            // Comparing the number to false, fails properly as well
            myNumber.Should().Be(false);
```

### Versions

FluentAssertions 5.9.0 targeting .NET Core 3.0

jnyrup:
When comparing whether two objects are equal, a conversion should be precision preserving.

Two examples of comparisons that currently unexpectedly passes:
* `10` is not equal to `true` even though `10` is convertible to `true` as `true` is not convertible to `10`.
* `0.02` is not equal to `0`, as that is due to truncation when converting the `double` into an `int`.

This fixes #1010 
This fixes #1183

### DIFF
--- a/Src/FluentAssertions/Common/ObjectExtensions.cs
+++ b/Src/FluentAssertions/Common/ObjectExtensions.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Globalization;
 
 namespace FluentAssertions.Common
 {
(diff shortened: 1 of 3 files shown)


jnyrup:
I came to think that this *might* break (undocumented?) runtime behavior for some.

With the current code, the second test of `When_comparing_a_custom_convertible_to_a_basic_type_it_should_fail` unexpectedly passes, but the first fails as expected. 
This is due to `CustomConvertible` being convertible into `int`.
With the changes in this PR `int` would also have to be convertible into `CustomConvertible`.

dennisdoomen:
> `When_comparing_a_custom_convertible_to_a_basic_type_it_should_fail`

Can't find that one.

eNeRGy164:
> > `When_comparing_a_custom_convertible_to_a_basic_type_it_should_fail`
> 
> Can't find that one.

https://github.com/jnyrup/fluentassertions/commit/e0631d7f9152e58f8f32d529b2d00d29c7fa00b8#diff-9b95f746d5165c406e037ab57c731715R54

dennisdoomen:
Duh. Wasn't looking in the PR itself.

dennisdoomen:
> With the changes in this PR `int` would also have to be convertible into `CustomConvertible`.

Maybe we should be much more restrictive in what we consider equal. In other words, list the types for which we apply the rules.

jnyrup:
I've updated the PR to use a white-list of numeric types, for which we will try conversion between.

Issue #1207, opened 2019-12-21
Support usage in Jupyter notebooks


ralfw:
### Description

When using FluentAssertions in a C# Jupyter notebook and the assertion fails, the exception is not pertaining to the actual mismatch, but regarding missing infrastructure.

### Complete minimal example reproducing the issue

<img width="1070" alt="fluentassertions failure in jupyter notebook" src="https://user-images.githubusercontent.com/132662/71309277-64564300-2406-11ea-99f6-9dec82a22f48.png">

### Expected behavior:

The expected exception for the above example is:

```
Expected value to be 99, but found 1.
   at FluentAssertions.Execution.LateBoundTestFramework.Throw(String message)
   at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message)
   at FluentAssertions.Execution.DefaultAssertionStrategy.HandleFailure(String message)
   ...
```

### Actual behavior:

The actual exception is:

```
Unhandled Exception
Could not load file or assembly 'System.Configuration.ConfigurationManager, Version=4.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. Reference assemblies should not be loaded for execution.  They can only be loaded in the Reflection-only loader context. (0x80131058)
   at FluentAssertions.Common.AppSettingsConfigurationStore.GetSetting(String name)
   at FluentAssertions.Common.Configuration.get_ValueFormatterAssembly() in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Common\Configuration.cs:line 88
   at FluentAssertions.Common.Configuration.DetermineFormatterDetectionMode() in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Common\Configuration.cs:line 55
   at FluentAssertions.Common.Configuration.get_ValueFormatterDetectionMode() in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Common\Configuration.cs:line 40
   at FluentAssertions.Formatting.AttributeBasedFormatter.get_IsScanningEnabled() in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Formatting\AttributeBasedFormatter.cs:line 33
   at FluentAssertions.Formatting.AttributeBasedFormatter.CanHandle(Object value) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Formatting\AttributeBasedFormatter.cs:line 28
[...]

Issue #1151, opened 2019-09-25
Configuration system throws `PlatformNotSupportedException` on Xamarin Android.


borrrden:
### Description

Configuration system throws `PlatformNotSupportedException` on Xamarin Android.

### Complete minimal example reproducing the issue

```csharp
3.Should().BeGreaterThan(4); // Probably any failing test?
```

### Expected behavior:

The normal assertion failure process

### Actual behavior:

The following Exception is thrown:

```
System.PlatformNotSupportedException : Operation is not supported on this platform.
 		  at System.Configuration.ConfigurationManager.PrepareConfigSystem () [0x0001a] in <4cd8a957e7664f4a9e27ce43dd1afeb7>:0 
 		  at System.Configuration.ConfigurationManager.GetSection (System.String sectionName) [0x0000a] in <4cd8a957e7664f4a9e27ce43dd1afeb7>:0 
 		  at System.Configuration.ConfigurationManager.get_AppSettings () [0x00000] in <4cd8a957e7664f4a9e27ce43dd1afeb7>:0 
		  at FluentAssertions.Common.AppSettingsConfigurationStore.GetSetting (System.String name) [0x00000] in <d52b982f9eb24747b8fd904dff0631fe>:0 
 		  at FluentAssertions.Common.Configuration.get_ValueFormatterAssembly () [0x00008] in <d52b982f9eb24747b8fd904dff0631fe>:0 
 		  at FluentAssertions.Common.Configuration.DetermineFormatterDetectionMode () [0x00000] in <d52b982f9eb24747b8fd904dff0631fe>:0 
 		  at FluentAssertions.Common.Configuration.get_ValueFormatterDetectionMode () [0x0001e] in <d52b982f9eb24747b8fd904dff0631fe>:0 
 		  at FluentAssertions.Formatting.AttributeBasedFormatter.get_IsScanningEnabled () [0x00005] in <d52b982f9eb24747b8fd904dff0631fe>:0 
 		  at FluentAssertions.Formatting.AttributeBasedFormatter.CanHandle (System.Object value) [0x00000] in <d52b982f9eb24747b8fd904dff0631fe>:0 
 		  at FluentAssertions.Formatting.Formatter+<>c__DisplayClass7_0.<Format>b__0 (FluentAssertions.Formatting.IValueFormatter f) [0x00000] in <d52b982f9eb24747b8fd904dff0631fe>:0 
[...]

PR #1210, opened 2019-12-29
Intercept Reflection exceptions from `AppSettingsConfigurationStore`

jnyrup:
Some platforms throw reflection exceptions when trying to use `ConfigurationManager`.
This does not fix the underlying problem of not being unable to use `ConfigurationManager`, but returns `null` instead of an exception.

Relates to #1207 and #1151

### DIFF
--- a/Src/FluentAssertions/Common/ConfigurationStoreExceptionInterceptor.cs
+++ b/Src/FluentAssertions/Common/ConfigurationStoreExceptionInterceptor.cs
@@ -1,0 +1,36 @@
+﻿#if !NETSTANDARD1_3 && !NETSTANDARD1_6
+
+namespace FluentAssertions.Common
+{
+    internal class ConfigurationStoreExceptionInterceptor : IConfigurationStore
+    {
+        private bool underlyingStoreUnavailable;
+
+        private readonly IConfigurationStore configurationStore;
+
+        public ConfigurationStoreExceptionInterceptor(IConfigurationStore configurationStore)
+        {
+            this.configurationStore = configurationStore;
+        }
+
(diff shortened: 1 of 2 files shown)


Src/FluentAssertions/Common/ConfigurationStoreExceptionInterceptor.cs
```diff
@@ -0,0 +1,36 @@
+﻿#if !NETSTANDARD1_3 && !NETSTANDARD1_6
+
+namespace FluentAssertions.Common
+{
+    internal class ConfigurationStoreExceptionInterceptor : IConfigurationStore
+    {
+        private bool throwsException;
```
dennisdoomen: I would name this `storeThrowsException` or `underlyingStoreUnavailble`. It's about the underlying store, not the current instance.

SteveBush:
Current fix returns null for GetSettings but NullConfigurationStore returns an empty string.  Seems like a more consistent fix would be to emulate NullConfigurarion store behavior.

jnyrup:
@SteveBush I see the inconsistency.
The code here was intentional, as `AppSettingsConfigurationStore` can return `null`.
Are having problems with the merged fix?

SteveBush:
Sorry for the delayed response.  I was traveling over the holidays and could only look at code diffs on GitHub.

I have verified the changes to intercept the reflection exceptions fix FluentAssertions for my Android builds.

I also reviewed to the FluentAssertions code more closely and I understand callers to AppSettingsConfigurationStore can handle a null response.

Thanks for making the change and have a happy new year.

Steve

PR #1220, opened 2020-01-05
Make more assertions chainable

jnyrup:
Change return type from `void` to `AndConstraint` to make the assertions chainable.
I deliberately didn't add tests for `ExecutionTimeAssertions` as time hasn't been abstracted away from it yet.

### DIFF
--- a/Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
@@ -512,10 +512,10 @@ namespace FluentAssertions.Collections
         /// <param name="becauseArgs">
         /// Zero or more objects to format using the placeholders in <see cref="because" />.
         /// </param>
-        public void BeEquivalentTo<TExpectation>(TExpectation expectation,
+        public AndConstraint<GenericDictionaryAssertions<TKey, TValue>> BeEquivalentTo<TExpectation>(TExpectation expectation,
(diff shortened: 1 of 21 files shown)


Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
```diff
@@ -512,10 +512,10 @@ public GenericDictionaryAssertions(IDictionary<TKey, TValue> dictionary) : base(
         /// <param name="becauseArgs">
         /// Zero or more objects to format using the placeholders in <see cref="because" />.
         /// </param>
-        public void BeEquivalentTo<TExpectation>(TExpectation expectation,
+        public AndConstraint<GenericDictionaryAssertions<TKey, TValue>> BeEquivalentTo<TExpectation>(TExpectation expectation,
```
dennisdoomen: Maybe return `AndWhichConstraint`?

jnyrup: What would the `Which` part be?

dennisdoomen: On 2nd thought, `And` would be enough. This PR merely triggered some thoughts on #1126.

jnyrup:
I'd like to avoid fiddling with AndWhich/That/Whose in this PR.

dennisdoomen:
Fair enough

PR #1221, opened 2020-01-06
Remove obsoleted members

jnyrup:
Remove members marked as `Obsolete` for removal in 6.0.


jnyrup:
Good idea, I'll do that.
Would you prefer that I split out these changes into separate PRs, such that each line in the release notes refers to different PRs?

dennisdoomen:
> Would you prefer that I split out these changes into separate PRs, such that each line in the release notes refers to different PRs?

Meh. Just let them refer to the same PR.

jnyrup:
I've updated the release notes, let me know if you have any comments.

PR #1222, opened 2020-01-08
Add script for easier approval of new API

krajek:


AcceptApiChanges.ps1
```diff
@@ -0,0 +1,17 @@
+## This script is related to approval tests that protect developers from unintentional changes to the public API
+## If your change does change the API on purpose and you double-checked correctness of the changes you can use this script to change the "approved" state of the API
+## Make sure that you run the approval tests before running this script, because the tests generate *.received.txt files.
+
+
+## Go to the directory where "approved" API is saved as .txt files
+Set-Location -Path .\Tests\Approval.Tests\ApprovedApi\FluentAssertions
```
dennisdoomen: Which means you can only run it once, right? I would make it re-runable.

krajek: Thanks, fixed.

Issue #1232, opened 2020-01-12
Dictionary assertions do not use the equality comparer of the dictionary involved


rubenrorije:
I noticed in the source code that GenericDictionaryAssertions uses `intersect` to compare the keys from the dictionary and the unexpected keys, which does not use the equality comparer of the dictionary.  The dual `ContainKeys` implementation as well as the other methods all use the `ContainsKey` on the dictionary which is correct.

To illustrate the issue I added two unit tests and changed the implementation in the pull request.

I also updated the xml documentation of all `(Not)ContainKey(s)` to reflect the code. The documentation was saying that the keys were compared using object.Equals which is not the case in the code.

dennisdoomen:
I suspect that this not only affects the dictionary assertions, but also `BeEquivalentTo`

jnyrup:
Fixed in #1233

PR #1233, opened 2020-01-12
Dictionary assertions do not always use the dictionary's comparer

rubenrorije:
See #1232.

I noticed in the source code that `GenericDictionaryAssertions` uses intersect to compare the keys from the dictionary and the unexpected keys, which does not use the equality comparer of the dictionary. The dual ContainKeys implementation as well as the other methods all use the ContainsKey on the dictionary which is correct.

To illustrate the issue I added two unit tests and changed the implementation in the pull request.

I also updated the xml documentation of all `(Not)ContainKey(s)` to reflect the code. The documentation was saying that the keys were compared using object.Equals which is not the case in the code.


jnyrup:
Seems to be a general problem for this class.
Anything that operates on a key without using the dictionary comparer.
`Except`, `Intersect`, `IsSameOrEqualTo`

jnyrup:
Looks good! 
Could I ask you to update the documentation for the other methods that still mention Keys being compared using `object.Equals`?

Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
```diff
@@ -732,7 +732,7 @@ public GenericDictionaryAssertions(IDictionary<TKey, TValue> dictionary) : base(
                     .FailWith("Expected {context:dictionary} to contain keys {0}{reason}, but found {1}.", unexpected, Subject);
             }
 
-            IEnumerable<TKey> foundKeys = unexpectedKeys.Intersect(Subject.Keys);
+            IEnumerable<TKey> foundKeys = unexpected.Where(key => Subject.ContainsKey(key));
```
dennisdoomen: Does this "bug" also affect the other methods in this class?

dennisdoomen:
@rubenrorije are you still around?

rubenrorije:
Sorry, I had other priorities so did not have time/mental bandwith to look at this earlier. I updated the xml documentation to reflect the code better. 

As far as I can see the only other method that does not use the keycomparer of the dictionary is the  `Equal` method. I did not change the implementation because I am not sure what the intended behaviour should be in some situations. 
When you for instance want to check a case insensitive dictionary but as `expected` give a case sensitive dictionary, what should be the expected behaviour?

dennisdoomen:
One more thing. Could you also add a row to the https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/releases.md#600?

PR #1234, opened 2020-01-12
Fix error description containing "}}}}"

### LINKED ISSUE #1223: String.Should().Be() output displays incorrect actual value
### Description

The output for failed string.Should().Be() shows the incorrect actual value when the value contains JSON strings ending with }}}}.

```csharp
string expect = "{\"a\":{\"b\":1}}";
string actual = "{\"a\":{\"b\":1}}}}";
actual.Should().Be(expect);
```

### Expected behavior:

```
FluentAssertions.Execution.AssertionFailedException: Expected actual to be
"{"a":{"b":1}}" with a length of 13, but
"{"a":{"b":1}}}}" has a length of 15, differs near "}}" (index 13).
```

### Actual behavior:

```
FluentAssertions.Execution.AssertionFailedException: Expected actual to be
"{"a":{"b":1}}" with a length of 13, but
"{"a":{"b":1}}" has a length of 15, differs near "}}" (index 13).
```

### Versions
Fluent Assertions 5.10.0 Targetting .Net Core 2.0/2.1/2.2/3.0/3.1
Fluent Assertions 5.10.0 Targetting .Net Framework 4.7.2 (not tried others)

krajek:
Closes #1223

### DIFF
--- a/Src/FluentAssertions/Execution/MessageBuilder.cs
+++ b/Src/FluentAssertions/Execution/MessageBuilder.cs
@@ -91,10 +91,6 @@ namespace FluentAssertions.Execution
             string[] values = failureArgs.Select(a => Formatter.ToString(a, useLineBreaks)).ToArray();
             string formattedMessage = string.Format(failureMessage, values);
 
-            var replacedCurlyBraces =
-                formattedMessage
-                    .Replace("{{{{", "{{")
-                    .Replace("}}}}", "}}");
             return formattedMessage;
         }
 
--- a/Src/FluentAssertions/Execution/MessageBuilder.cs
+++ b/Src/FluentAssertions/Execution/MessageBuilder.cs
@@ -91,7 +91,11 @@ namespace FluentAssertions.Execution
             string[] values = failureArgs.Select(a => Formatter.ToString(a, useLineBreaks)).ToArray();
             string formattedMessage = string.Format(failureMessage, values);
 
-            return formattedMessage.Replace("{{{{", "{{").Replace("}}}}", "}}");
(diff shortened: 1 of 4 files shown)


krajek:
Seems like for some reason there was a piece of code that just messed up `{{{{` and `}}}}`. The code goes back to 902c87b8. 
Removing the code fixed the problem.

Src/FluentAssertions/Execution/MessageBuilder.cs
```diff
@@ -91,7 +91,7 @@ private string FormatArgumentPlaceholders(string failureMessage, object[] failur
             string[] values = failureArgs.Select(a => Formatter.ToString(a, useLineBreaks)).ToArray();
             string formattedMessage = string.Format(failureMessage, values);
 
-            return formattedMessage.Replace("{{{{", "{{").Replace("}}}}", "}}");
+            return formattedMessage;
```
dennisdoomen: You could directly return the result of the `Format` call

krajek: If not too much of a problem I would rather leave it like it is. I like such an approach for
a) easier to debug, there is a place to take a look at the result before returning it
b) intention revealing name attached to the value

dennisdoomen: Sure It was just a suggestion. By the way, I don't know about Visual Studio, but at least in Jetbrains Rider, you don't need to do this anymore. The debugger will show you all parts of an expression.

dennisdoomen:
@jnyrup just a gentle reminder.

PR #1244, opened 2020-02-06
Added NotContainMatch method for string collections and unit tests to cover it.

tomaustin700:
Added NotContainMatch for string collections which does the opposite of ContainMatch. I recently had a need for NotContainMatch so decided to add it.


tomaustin700:
Hey @dennisdoomen, I'm targeting the `master` branch as the documentation [here](https://github.com/fluentassertions/fluentassertions/blob/develop/CONTRIBUTING.md) says to however I've noticed most other PR's are going into the `develop` branch. Should this be targeting `develop`? Thanks.

dennisdoomen:
Yeah, you're right. Please target `develop`.

PR #1246, opened 2020-02-06
Added NotContainMatch method for string collections and unit tests to cover it

tomaustin700:
Added NotContainMatch for string collections which does the opposite of ContainMatch. I recently had a need for NotContainMatch so decided to add it.

### DIFF
--- a/FluentAssertions.sln
+++ b/FluentAssertions.sln
@@ -48,12 +48,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NetCore30.Specs", "Tests\Ne
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Approval.Tests", "Tests\Approval.Tests\Approval.Tests.csproj", "{F5115158-A038-4D14-A04E-46E7863E40B9}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NetCore21.Specs", "Tests\NetCore21.Specs\NetCore21.Specs.csproj", "{0CF100F2-CE71-41F4-9523-62AD78BD6961}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NetCore21.Specs", "Tests\NetCore21.Specs\NetCore21.Specs.csproj", "{0CF100F2-CE71-41F4-9523-62AD78BD6961}"
 EndProject
 Global
 	GlobalSection(SharedMSBuildProjectFiles) = preSolution
 		Tests\Shared.Specs\Shared.Specs.projitems*{0c23c12c-899f-4ad1-b737-b967a9910c08}*SharedItemsImports = 13
-		Tests\Shared.Specs\Shared.Specs.projitems*{442c9b50-6f91-4cba-8d4e-b470e5cc3d5f}*SharedItemsImports = 4
 	EndGlobalSection
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
--- a/FluentAssertions.sln
+++ b/FluentAssertions.sln
@@ -18,7 +18,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Specs", "Specs", "{963262D0
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "Shared.Specs", "Tests\Shared.Specs\Shared.Specs.shproj", "{0C23C12C-899F-4AD1-B737-B967A9910C08}"
 EndProject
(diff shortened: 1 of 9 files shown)


FluentAssertions.sln
```diff
@@ -48,12 +48,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NetCore30.Specs", "Tests\Ne
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Approval.Tests", "Tests\Approval.Tests\Approval.Tests.csproj", "{F5115158-A038-4D14-A04E-46E7863E40B9}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NetCore21.Specs", "Tests\NetCore21.Specs\NetCore21.Specs.csproj", "{0CF100F2-CE71-41F4-9523-62AD78BD6961}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NetCore21.Specs", "Tests\NetCore21.Specs\NetCore21.Specs.csproj", "{0CF100F2-CE71-41F4-9523-62AD78BD6961}"
```
dennisdoomen: What happened here?

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
```diff
@@ -242,5 +242,58 @@ private IEnumerable<string> AllThatMatch(string wildcardPattern)
...
+        public AndWhichConstraint<StringCollectionAssertions, string> NotContainMatch(string wildcardPattern, string because = "",
+            params object[] becauseArgs)
+        {
+            Execute.Assertion
+                .BecauseOf(because, becauseArgs)
+                .ForCondition(Subject is object)
+                .FailWith("Expected {context:collection} to not contain a match of {0}{reason}, but found <null>.", wildcardPattern)
+                .Then
+                .ForCondition(NotContainsMatch(wildcardPattern))
+                .FailWith("Expected {context:collection} {0} to not contain a match of {1}{reason}.", Subject, wildcardPattern);
```
dennisdoomen: I would probably rephrase this as `Did not expect`

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
```diff
@@ -242,5 +242,58 @@ private IEnumerable<string> AllThatMatch(string wildcardPattern)
...
+            {
+                return Subject.All(item =>
+                {
+                    item.Should().NotMatch(wildcardPattern);
+                    return !scope.Discard().Any();
+                });
+            }
+        }
+
+        private IEnumerable<string> AllThatDontMatch(string wildcardPattern)
```
jnyrup: Avoid contractions.
`AllThatDontMatch` -> `AllThatDoNotMatch`

FluentAssertions.sln
tomaustin700: Not sure what happened, should now be back to normal.

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
tomaustin700: Changed to "Did not expect"

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
tomaustin700: Changed to `AllThatDoNotMatch`

tomaustin700:
@dennisdoomen @jnyrup Thanks for the review, all issues should hopefully now be resolved.

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
```diff
@@ -242,5 +242,58 @@ private IEnumerable<string> AllThatMatch(string wildcardPattern)
...
+        /// The wildcard pattern with which the subject is matched, where * and ? have special meanings.
+        /// </param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public AndWhichConstraint<StringCollectionAssertions, string> NotContainMatch(string wildcardPattern, string because = "",
```
jnyrup: Wondering whether a negative assertion should return an `AndWhichConstraint`...
I have an odd feeling about asserting that a collection does not contains something and then continue asserting on what *didn't* match.
Also because `.Which` will only(?) make sense if the collection has *exactly* one non-matching item, which can be returned from `.Which`.
Maybe I'm just struggling finding a good example.

dennisdoomen: No, I think you are right. Let's just use an `AndConstraint`

tomaustin700: Just pushed this change :)

Issue #876, opened 2018-07-19
TryConversionStep executing after Using statements


mwasson74:
### Description

It appears that the `TryConversionStep` is being executed after the `Using` statements.

### Complete minimal example reproducing the issue

```csharp
[TestMethod]
public void CompareStringToDate()
{
    //Arrange
    var classA = new ClassA { ThisIsMyDateTime = DateTime.Now };
    var classB = new ClassB { ThisIsMyDateTime = classA.ThisIsMyDateTime.ToString() };

    //Asserts
    classB.Should().BeEquivalentTo(classA,
        options => options.ExcludingMissingMembers()
            .WithAutoConversion()
            .Using<DateTime>(ctx => ctx.Subject.Should().BeCloseTo(ctx.Expectation, 1000))
            .WhenTypeIs<DateTime>());
}

private class ClassA
{
    public DateTime ThisIsMyDateTime { get; set; }
}

private class ClassB
{
    public string ThisIsMyDateTime { get; set; }
}
```

### Expected behavior:

The conversion from string to DateTime happens before the `Using` statements.

### Actual behavior:

Unit test fails with the following message:

> Expected member ThisIsMyDateTime from subject to be a System.DateTime, but found a System.String.

### Versions

* Which version of Fluent Assertions are you using?

> 5.4.1

* Which .NET runtime and version are you targeting? 

> .NET framework 4.5.2

### Additional Information
It looks like the `TryConversionStep` is placed at the end of the steps collection.

[https://github.com/fluentassertions/fluentassertions/blob/41c5e9b8a13d53794a8efe47e903c66f0c13f53e/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs#L129](url)

dennisdoomen:
I'm in doubt what would be the expected behavior. If I would provide some custom assertion rule, would one expect that to override any conversion rules. What do you think @jnyrup >

dennisdoomen:
@jnyrup ?

jnyrup:
I wouldn't expect custom assertion rules to override existing conversion rules.
I would expect that after failing all existing conversion rules, the code would try my custom rule.

For this particular example, here's how I imagine the code should behave:
1. If both `subject` and `expectation` are `DateTime`s, use the `Using`
2. As `WithAutoConversion` is enabled, try to convert `subject` into the type of `expectation`
2.1. If both `convertedSubject` and `expectation` are `DateTime`s, use the `Using`

[...]

PR #1257, opened 2020-02-19
Using/When in BeEquivalentTo will respect the conversion rules. 

dennisdoomen:
This implements the functionality as suggested by @jnyrup in #876. 

* If both subject and expectation are `DateTime`s, use the `Using`
* If `WithAutoConversion` is enabled, try to convert subject into the type of expectation
* If, after conversion, both the converted subject and the expectation are `DateTime`s, use the `Using`

Also fixes the API approval script.

### DIFF
--- a/Src/FluentAssertions/Equivalency/CollectionMemberAssertionOptionsDecorator.cs
+++ b/Src/FluentAssertions/Equivalency/CollectionMemberAssertionOptionsDecorator.cs
@@ -39,9 +39,9 @@ namespace FluentAssertions.Equivalency
 
         public ConversionSelector ConversionSelector => inner.ConversionSelector;
 
-        public IEnumerable<IEquivalencyStep> GetUserEquivalencySteps(ConversionSelector conversionSelector)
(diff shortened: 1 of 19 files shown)


Src/FluentAssertions/Equivalency/CollectionMemberAssertionOptionsDecorator.cs
```diff
@@ -39,9 +39,9 @@ public OrderingRuleCollection OrderingRules
 
         public ConversionSelector ConversionSelector => inner.ConversionSelector;
 
-        public IEnumerable<IEquivalencyStep> GetUserEquivalencySteps(ConversionSelector conversionSelector)
+        public IEnumerable<IEquivalencyStep> UserEquivalencySteps
```
jnyrup: Why converting this into a property?

dennisdoomen: It used to be a property, but then became a method because of the introduction of the `ConversionSelector`. This is no longer needed, so I switched it back to be a property.

PR #1258, opened 2020-02-22
[WIP] Reduce usage of IsSameOrEqualTo

jnyrup:
One of the original benefits of `IsSameOrEqualTo` seems to have been that it could handle `null` values but it also handles comparing boxed numerics.
In many cases the extra functionality of `IsSameOrEqualTo` is purely overhead,

The better alternatives than using `IsSameOrEqualTo`:
* Potentially null `object`s can use [`Equals(a, b)`](https://referencesource.microsoft.com/#mscorlib/system/object.cs,68).
* Generic objects can use [`EqualityComparer<T>.Default.Equals(a, b)`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.equalitycomparer-1.default?view=netframework-4.8).
* `Type` can be compared using `==`

For primitives `a.Equals(b)` and `a.CompareTo(b) == b` (for `TimeSpan`) are replaced with infix operators such as `==` and `!=`.

`(Not)Be` for `XDocument` is fixed to handle comparing two `null`s as expected.

### DIFF
--- a/Tests/Shared.Specs/Xml/XDocumentAssertionSpecs.cs
+++ b/Tests/Shared.Specs/Xml/XDocumentAssertionSpecs.cs
@@ -57,6 +57,21 @@ namespace FluentAssertions.Specs
         }
 
         [Fact]
+        public void When_a_a_null_xml_document_is_equal_to_a_null_xml_document_it_should_succeed()
+        {
+            // Arrange
+            XDocument document = null;
+            XDocument otherXDocument = null;
+
+            // Act
+            Action act = () =>
+                document.Should().Be(otherXDocument);
+
+            // Assert
+            act.Should().NotThrow<XunitException>();
+        }
+
+        [Fact]
         public void When_asserting_a_xml_document_is_equal_to_a_different_xml_document_it_should_fail_with_descriptive_message()
         {
             // Arrange
@@ -105,7 +120,7 @@ namespace FluentAssertions.Specs
         }
 
         [Fact]
-        public void When_asserting_a_null_xml_document_is_not_equal_to_some_xml_document_it_should_fail()
+        public void When_asserting_a_null_xml_document_is_not_equal_to_some_xml_document_it_should_succeed()
         {
             // Arrange
             XDocument document = null;
@@ -116,8 +131,23 @@ namespace FluentAssertions.Specs
                 document.Should().NotBe(someXDocument);
 
             // Assert
+            act.Should().NotThrow<XunitException>();
+        }
+
+        [Fact]
+        public void When_asserting_a_null_xml_document_is_not_equal_to_a_null_xml_document_it_should_fail()
+        {
+            // Arrange
+            XDocument document = null;
+            XDocument someXDocument = null;
+
+            // Act
+            Action act = () =>
+                document.Should().NotBe(someXDocument);
+
+            // Assert
             act.Should().Throw<XunitException>()
-                .WithMessage("*not expect XML document to be*, but found <null>*");
+                .WithMessage("Did not expect XML document to be <null>.");
         }
 
         [Fact]
--- a/Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
@@ -434,7 +434,7 @@ namespace FluentAssertions.Collections
             foreach (var key in expected.Keys)
             {
                 Execute.Assertion
-                    .ForCondition(Subject[key].IsSameOrEqualTo(expected[key]))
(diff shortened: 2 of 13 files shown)


Tests/Shared.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -56,6 +56,21 @@ public void When_asserting_a_null_xml_document_is_equal_to_another_xml_document_
             act.Should().Throw<XunitException>();
         }
 
+        [Fact]
+        public void When_a_a_null_xml_document_is_equal_to_another_null_xml_document_it_should_fail()
```
lg2de: `When_asserting_a_null_xml_document_is_equal_to_another_a_null_xml_document_it_should_succeed`

or your implemented assertion is wrong.

jnyrup: Thanks

dennisdoomen:
Isn't this change going to be a source of confusion? I.o.w. which method do you need in what circumstances? And what about the boxed numeric comparisons? I remember we had to jump through lots of hoops to make it work as expected. I'm not entirely sure this is properly covered by all the tests.

jnyrup:
We're not home safe for for boxed numerics or non-generic enumerables in `develop`.
Here are the 9 combinations, the resolved overload and the failure message if failing.
The results are equal for both `develop` and this PR.

```c#
[Fact]
public void generic_subject_generic_expect()
{
	new int[] { 1 }.Should().Equal(new double[] { 1d });
	CollectionAssertions.Equal(IEnumerable)
}

[Fact]
public void generic_subject_object_expect()
{
	new int[] { 1 }.Should().Equal(new object[] { 1d });
	CollectionAssertions.Equal(params object[])
}

[Fact]
public void generic_subject_ienumerable_expect()
{
	new int[] { 1 }.Should().Equal((IEnumerable)new object[] { 1d });
	CollectionAssertions.Equal(IEnumerable)
}

[Fact]
public void object_subject_generic_expect()
{
	new object[] { 1 }.Should().Equal(new double[] { 1d });
	SelfReferencingCollectionAssertions.Equal(params T[])
	// Expected collection to be equal to {{1.0}}, but {1} differs at index 0.
}

[Fact]
public void object_subject_object_expect()
{
	new object[] { 1 }.Should().Equal(new object[] { 1d });
	SelfReferencingCollectionAssertions.Equal(params T[])
	// Expected collection to be equal to {1.0}, but {1} differs at index 0.
}

[Fact]
public void object_subject_ienumerable_expect()
{
	new object[] { 1 }.Should().Equal((IEnumerable)new object[] { 1d });
	SelfReferencingCollectionAssertions.Equal(params T[])
	// Expected collection to be equal to {{1.0}}, but {1} differs at index 0.
}

[Fact]
public void ienumerable_subject_generic_expect()
{
	((IEnumerable)new object[] { 1 }).Should().Equal(new double[] { 1d });
	CollectionAssertions.Equal(IEnumerable)
}

[Fact]
public void ienumerable_subject_object_expect()
{
	((IEnumerable)new object[] { 1 }).Should().Equal(new object[] { 1d });
	CollectionAssertions.Equal(params object[])
}

[Fact]
public void ienumerable_subject_ienumerable_expect()
{
	((IEnumerable)new object[] { 1 }).Should().Equal((IEnumerable)new object[] { 1d });
	CollectionAssertions.Equal(IEnumerable)
}
```

The notorious `Equal(params T[] elements)` (again) seems to be the offender as it makes it impossible to distinguish a collection of `T`s from a single `T` containing a collection of `T`s. 

I'd really like to remove all `params object[]` and `params T[]` assertions.

dennisdoomen:
What makes you think you can compare a double and an integer like that?

jnyrup:
* `1d == 1` as `1` is promoted to a `double` and `double` overrides `==(double, double)`
* `1d.Equals(1)` `1` is again promoted to `double` and resolves to `double.Equals(double)`
* `1.Equals(1d)` is false as `int.Equals()` does not have an overload for `double`.
* `((object)1d).Equals(1)` is false, as it resolves to `object.Equals(object)`
* `(object)1d != (object)1` and the reverse are `false` as it does reference equality.

Comparing enumerables of different types can use:
* `CollectionAssertions.Equal(IEnumerable)`
* `SelfReferencingCollectionAssertions.Equal<TExpected>(IEnumerable<TExpected>, Func<T, TExpected, bool>)`


I'll think some more about this topic...

jnyrup:
Closing this PR as it turned into trying to solve multiple things at once.
Will open some smaller and more focused PR's instead.

PR #1266, opened 2020-03-07
Constrain generic type on NumericAssertions

jnyrup:
Further constraining of `T` to `IComparable<T>` gets us:
* No more runtime type check in constructor
* Avoid (un)boxing of primitives
* Type preserving `Subject`, i.e. `T`/`T?` instead of `IComparable<T>`

### DIFF
--- a/Src/FluentAssertions/Numeric/NumericAssertions.cs
+++ b/Src/FluentAssertions/Numeric/NumericAssertions.cs
@@ -10,25 +10,25 @@ using FluentAssertions.Execution;
 namespace FluentAssertions.Numeric
 {
     /// <summary>
-    /// Contains a number of methods to assert that an <see cref="IComparable"/> is in the expected state.
+    /// Contains a number of methods to assert that an <see cref="IComparable{T}"/> is in the expected state.
     /// </summary>
     [DebuggerNonUserCode]
     public class NumericAssertions<T>
-        where T : struct
+        where T : struct, IComparable<T>
     {
-        public NumericAssertions(object value)
+        public NumericAssertions(T value)
+            : this((T?)value)
         {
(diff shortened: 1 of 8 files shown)


Src/FluentAssertions/Numeric/NumericAssertions.cs
```diff
@@ -14,21 +14,21 @@ namespace FluentAssertions.Numeric
...
-                Subject = value as IComparable;
-                if (Subject is null)
-                {
-                    throw new InvalidOperationException("This class only supports types implementing IComparable.");
-                }
-            }
         }
 
-        public IComparable Subject { get; }
+        private protected NumericAssertions(T? value)
```
lg2de: Why `private protected`?
Do you want forbid extensions using this class?

jnyrup: I don't think I'm forbidding extensions on this class.
Did you see the other constructors?
There are two constructor available, which I think should be covering
```c#
public NumericAssertions(T value) { }
public NullableNumericAssertions(T? value) { }
```

jnyrup:
Glad to hear there was a reason back then.
The somewhat odd construction have stricken me every time I've seen it over the past years.

Tried to dig in history to see if I could find a cause:

1eb8f4dfec1f848148bc05e73183e4979ef43888 
* Removed `T : struct, Comparable` constraints (I'm noticing the non-generic `IComparable`)
* Changed constructor from `T?` to `T`
* Added runtime `IComparable` check
* Changed `Subject` from `T?` to `IComparable`

bddb0247760eec7440116a08bf79169529121e50
* Added `T : struct`
* Changed constructor from `T` to `object`

But I didn't find any compilation issues...

PR #1267, opened 2020-03-08
add extension to assert TaskCompletionSource

lg2de:
This is a new extension to create assertions on `TaskCompletionSource<T>` (TCS).

This PR creates a breaking change, because for instances of TCS the assertions of the currently catching `ObjectAssertions` extension will not be available anymore.
But FA is going to version 6.0, so it should be acceptable.

Further, I tried to analyze the assertions which may get lost. Here is *my* answer to the question "Do we need such assertion?":
* `ObjectAssertions`
  * `Be`/`NotBe`: No, why should we check equivalency?
  * `BeEquivalentTo`/`NotBeEquivalentTo`: No, why should we check equivalency?
* `ReferenceTypeAssertions`
  * `BeNull`/`NotBeNull`: **Unsure**, I do not see any relevance.
  * `BeSameAs`/`NotBeSameAs`: **Unsure**, could be relevant if TCS is part of an API.
  * `BeOfType`/`NotBeOfType`: No, I do not see any relevance.
  * `BeAssignableTo`/`NotBeAssignableTo`: No, I do not see any relevance.
  * `Match`: No, I do not see any relevance.

Please help to answer these questions and check whether you agree.

While creating unit tests using the clock API I was a bit confused whether the methods `CompletesBeforeTimeout` and `RunsIntoTimeout` influences the behavior of `DelayAsync`. I found, it **did not**. So I introduced new methods and changed existing usages. Hope you agree.
If not, we should think, whether `DelayAsync` or even more the `IClock` API can be adopted to work similar to `FakeClock.Wait`. Maybe `DelayAsync` should be replaced by a `WaitAsync` returning `boolean` too.


Tests/Shared.Specs/Specialized/TaskCompletionSourceAssertionSpecs.cs
```diff
@@ -0,0 +1,117 @@
+﻿using System;
+using System.Threading.Tasks;
+using FluentAssertions.Extensions;
+using Xunit;
+using Xunit.Sdk;
+
+namespace FluentAssertions.Specs
+{
+    public class TaskCompletionSourceAssertionSpecs
```
dennisdoomen: Please use `Arrange`, `Act` and `Assert` comments

Src/FluentAssertions/Specialized/TaskCompletionSourceAssertions.cs
```diff
@@ -0,0 +1,105 @@
...
+            {
+                Task completedTask;
+                using (NoSynchronizationContextScope.Enter())
+                {
+                    completedTask = await Task.WhenAny(
+                        this.subject.Task,
+                        Clock.DelayAsync(timeSpan, timeoutCancellationTokenSource.Token));
+                }
+
+                if (completedTask == this.subject.Task)
```
dennisdoomen: We prefer not to use `this`, unless it is needed.

Src/FluentAssertions/Specialized/TaskCompletionSourceAssertions.cs
```diff
@@ -0,0 +1,105 @@
...
+            string failMessage1,
+            string failMessage2,
+            object[] failArgs)
+        {
+            Execute.Assertion
+                .ForCondition(subject is object)
+                .BecauseOf(because, becauseArgs)
+                .FailWith(failMessage1, failArgs);
+
+            using (var timeoutCancellationTokenSource = new CancellationTokenSource())
```
dennisdoomen: Maybe we can bump the version of C# to 8 so we don't need those blocks anymore

Tests/Shared.Specs/Specialized/TaskCompletionSourceAssertionSpecs.cs
lg2de: Do you really need these comments?
I added blank lines for separation.
Is that not enough?

dennisdoomen: It's more a convention in our code base.

Src/FluentAssertions/Specialized/TaskCompletionSourceAssertions.cs
lg2de: Should I really change the project file this way?

lg2de:
> new `specialized.md` 

In this file all assertions from `Specialized` namespace should be in, right?

Src/FluentAssertions/Specialized/TaskCompletionSourceAssertions.cs
dennisdoomen: It's up to you. It was just a suggestion.

dennisdoomen:
> In this file all assertions from `Specialized` namespace should be in, right?

Well, if you add yours, we'll take it from there.

Src/FluentAssertions/AssertionOptions.cs
```diff
@@ -18,6 +16,7 @@ public static class AssertionOptions
         static AssertionOptions()
         {
             EquivalencySteps = new EquivalencyStepCollection();
+            TaskTimeout = 30.Seconds();
```
jnyrup: `30` seems like a rather arbitrary choice to me?

If I saw something like this snippet, my immediate thought is "when do I know that something does *not* complete?" and the halting problem.
```c#
something.Should().CompleteAsync()
```

I would need to consult implementation or documentation to know that there is a default timeout.
I prefer this being explicit in the code.

Just my thoughts.

dennisdoomen: But if we include the timeout in the failure message?

lg2de: It can also be mentioned in the documentation.

jnyrup: My idea was that one does not have to read a failure message or documentation to see what some default timeout is.
I think we already did a design mistake in `BeCloseTo(DateTime, int precision = 20)` by specifying a default precision. 
Was constitutes as two dates being close is context specific.

```c#
/// <summary>
/// Defined by company spec §1.2.3
/// </summary>
private static readonly TimeSpan DefaultTimeout = TimeSpan.FromSeconds(30);

MyTestMethod()
{
   subject.Should().CompleteWithinAsync(DefaultTimeout);
}
```

lg2de: I understand your thoghts, @jnyrup . I do not want to start a long discussion here.

You and @dennisdoomen, you are leading together. Please decide whether such default values are accepted or not.
And, **please**, document your design rules to prevent discussion and implementation effort for /dev/null.

For me, I do not see a problem with default behaviors. E.g. `BeEquivalentTo` has a lot of default behaviors too.

dennisdoomen: @jnyrup what do you propose instead? Just wait indefinitely?

lg2de: Waiting indefinitely is useless. In this case you can just use `await`.
This is why I introduced this extension: I want to prevent waiting indefinitely...

jnyrup: @dennisdoomen Removing the `CompleteAsync()` and only have the `CompleteWithinAsync(TimeSpan)`. 

Just as `GenericAsyncFunctionAssertions<TResult>` and `NonGenericAsyncFunctionAssertions` only have `CompleteWithin(TimeSpan)` and `CompleteWithinAsync(TimeSpan)` assertions.

@lg2de I'm very sorry that your efforts once again ends up in these exhausting discussions.
I'll start see if we can making some design guidelines. 
That said, I can't make any promises on:
* when they will be ready
* that they be fully covering
* that the *current* codebase lives up to those guidelines.

dennisdoomen: So the general design decision is "no default timeouts". I can live with that.

lg2de: Ok, I'll remove.
Anyway I do not really understand what is the difference between
* "default behavior A": BeEquivalentTo checks all public properties
* "default behavior B": CompleteAsync uses timeout of 30 seconds

lg2de:
Anything still missing here?

Tests/Shared.Specs/Specialized/TaskCompletionSourceAssertionSpecs.cs
```diff
@@ -0,0 +1,76 @@
...
```
jnyrup: Could you add a test for the `subject is object` check?

Tests/Shared.Specs/Specialized/TaskCompletionSourceAssertionSpecs.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        [Fact]
+        public void When_TCS_completes_in_time_and_result_is_expected_it_should_succeed()
+        {
+            // Arrange
+            var subject = new TaskCompletionSource<int>();
+            var timer = new FakeClock();
+
+            // Act
+            Func<Task> action = async() => (await subject.Should(timer).CompleteWithinAsync(1.Seconds())).Which.Should().Be(42);
```
jnyrup: 🙃 missing space between `async` and `()`

Tests/Shared.Specs/Specialized/TaskCompletionSourceAssertionSpecs.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        [Fact]
+        public void When_TCS_completes_in_time_and_result_is_not_expected_it_should_fail()
+        {
+            // Arrange
+            var subject = new TaskCompletionSource<int>();
+            var timer = new FakeClock();
+
+            // Act
+            Func<Task> action = async() => (await subject.Should(timer).CompleteWithinAsync(1.Seconds())).Which.Should().Be(42);
```
jnyrup: 🙃 missing space between `async` and `()`

docs/_pages/releases.md
```diff
@@ -15,6 +15,7 @@ sidebar:
 * Added `collectionOfStrings.Should().NotContainMatch()` to assert that the collection does not contain a string that matches a wildcard pattern 
 * The `Using`/`When` option on `BeEquivalentTo` will now use the conversion rules when trying to match the predicate.
 * Added `NotBeWritable` to `PropertyInfoSelectorAssertions` to be able to assert that properties are not writable.
+* Added extension to assert `TaskCompletionSource<T>`
```
jnyrup: Add link to this PR.
I have a PR coming up to fix it for the other bullet points.

Src/FluentAssertions/Specialized/TaskCompletionSourceAssertions.cs
```diff
@@ -0,0 +1,82 @@
...
+        public Task<AndWhichConstraint<TaskCompletionSourceAssertions<T>, T>> CompleteWithinAsync(
+            TimeSpan timeSpan, string because = "", params object[] becauseArgs)
+        {
+            return AssertAsync(timeSpan, because, becauseArgs,
+                "Expected {context} to complete within {0}{reason}, but found <null>.",
+                "Expected {context:task} to complete within {0}{reason}.",
+                new object[] { timeSpan });
+        }
+
+        private async Task<AndWhichConstraint<TaskCompletionSourceAssertions<T>, T>> AssertAsync(
```
jnyrup: You *could* inline this method again to avoid parameter names such as `failMessage1` and `failMessage2`. 
Not crucial since it's not part of the public API.

Tests/Shared.Specs/Specialized/TaskCompletionSourceAssertionSpecs.cs
lg2de: Hope, the new test is what you mean.

Src/FluentAssertions/Specialized/TaskCompletionSourceAssertions.cs
lg2de: I think, this is a good entry for future extensions.

Tests/Shared.Specs/Specialized/TaskCompletionSourceAssertionSpecs.cs
jnyrup: It was 👍

FluentAssertions.sln
```diff
@@ -18,7 +18,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Specs", "Specs", "{963262D0
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "Shared.Specs", "Tests\Shared.Specs\Shared.Specs.shproj", "{0C23C12C-899F-4AD1-B737-B967A9910C08}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AssemblyA", "Tests\AssemblyA\AssemblyA.csproj", "{7144BD9D-2A5F-45B6-AC5B-E35578D03350}"
```
lg2de: Please note. these changes were done automatically by Visual Studio.

jnyrup: I've noticed them as well locally.

dennisdoomen:
Another gem by @lg2de 👌

jnyrup:
@lg2de Forgot to complement you for the the thorough analysis of the consequences of introducing the new `Should(this TaskCompletetionSource<T>)` overload.
That makes reviewing much easier.

PR #1268, opened 2020-03-09
Switch target from .NET Core 3.0 to 3.1 because the first is no longer supported

lg2de:
The current HEAD is building package using .NET Core (SDK) 3.1, but targeting 3.0.
3.0 is already [out of support](https://dotnet.microsoft.com/platform/support/policy/dotnet-core).
Recently you have removed "old" versions like .NET Core 2.0. Why not targeting 3.1?

* switch target for package from 3.0 to 3.1
* switch target for tests from 3.0 to 3.1
* updated documentation


jnyrup:
Ah... I thought you meant *only* upgrading `_build.csproj`.

What's the benefit in retargeting the library 3.0 to 3.1?

The current 6.0 targets were chosen from ideas about:
* Reducing the number of targets
* Better alignment of available features across targets
* Fewer preprocessor directives to ease developing of FA
* Fewer workarounds to support old targets (net45, netstandard1.3, netstandard1.6).

One of the reasons we chose .net core 2.1 over .net core 2.0 is the inclusion of
* [`ValueTask`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask?view=netcore-2.1)
* [`Span<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.span-1?view=netcore-2.1)
* [`ReadOnlySpan<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.readonlyspan-1?view=netcore-2.1)

dennisdoomen:
> What's the benefit in retargeting the library 3.0 to 3.1?

The only reason could be that .NET Core 3.0 is already out of support and superseded by 3.1 (which is LTS). I think it's safe to assume that by the time we release 6.0, everybody who was on 3.0 has already upgraded to 3.1. We did this at several places and this was a painless upgrade.

jnyrup:
I fully agree that *users* of Fluent Assertions *should* upgrade their applications from 3.0 to 3.1.
What I don't see is why we as a *library* should pursue being on the strictly newest.

I'm not strictly against it, I just don't see any benefit it - but I will be happy to learn new perspectives :)

lg2de:
A user targeting 3.0 can use FA anyway while netstandard2.X or netcoreapp2.X is part of FA.

jnyrup:
@dennisdoomen 's suspicion was that one needs to use a netcoreapp target to use `System.Configuration.ConfigurationManager`, but we didn't find any cut-in-stone proofs as I recall.

dennisdoomen:
Yeah, I agree. Unless there's a technical reason to upgrade, let's stick to 3.0 for now.

lg2de:
FA 6.0 will be the first release with official support of .netcore 3.X.
Why releasing this for a target already out of support?
I do not see any reason to run tests on 3.0 while superseded by 3.1.

lg2de:
May be we have to go another step back.
While REMOVING old targets you added .NET Core 3.0.
Why not adding netstandard2.1 as target with netcoreapp3.1 in the tests?

jnyrup:
@lg2de I'm not sure I understand your question.
If you're asking why we didn't only add netstandard2.1 to the targetframeworks and added a netcoreapp3.1 test project, my current answer is:


> @dennisdoomen 's suspicion was that one needs to use a netcoreapp target to use `System.Configuration.ConfigurationManager`, but we didn't find any cut-in-stone proofs as I recall.

dennisdoomen:
@lg2de I value your passion for this topic, but we're not going to change the target frameworks again unless there's a technical reason for that. We spend quite some time on finding a good balance, but we didn't do a particular good job in capturing the rationale somewhere central. Most of that happened in private discussions between me and @jnyrup. We just agreed to move those kinds of discussion to the [#develop](https://fluentassertions.slack.com/archives/C01055CMSD9) channel in the future. 

So please revert the change to only the build project if you still want this PR to be merged.

lg2de:
I think, changing the build project you can do together with other changes.

PR #1272, opened 2020-03-19
Fix XPath index calculation for XML comparison

lg2de:
Last year I provided rework of the assertion failure message when comparing XML by introducing XPath index.
Now I found that the current implementation can generate invalid XPath. The index could should apply to the parent.

## Input
Compare
```xml
<root>
  <xml1 />
  <xml1><xml2 /><xml2 a=\"x\" /></xml1>
  <xml1>
    <xml2 />
    <xml2 a=\"x\" />
  </xml1>
</root>
```

with
```xml
<root>
  <xml1 />
  <xml1><xml2 /><xml2 a=\"x\" /></xml1>
  <xml1>
    <xml2 />
    <xml2 a=\"y\" /> <!-- note the different attribute -->
  </xml1>
</root>
```

## Expected behavior
Expected attribute "a" at "/root/xml1[3]/xml2[**2**]" to have value "y", but found "x".

## Actual behavior
Expected attribute "a" at "/root/xml1[3]/xml2[**4**]" to have value "y", but found "x".

### DIFF
--- a/Src/FluentAssertions/Xml/Equivalency/Node.cs
+++ b/Src/FluentAssertions/Xml/Equivalency/Node.cs
@@ -54,14 +54,19 @@ namespace FluentAssertions.Xml.Equivalency
 
         public Node Parent { get; }
 
-        public Node Push(string name)
+        public Node Push(string localName)
         {
-            Node node = children.Find(e => e.name == name)
-                        ?? AddChildNode(name);
+            Node node = children.Find(e => e.name == localName)
+                        ?? AddChildNode(localName);
 
             node.count++;
 
             return node;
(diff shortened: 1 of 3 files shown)


lg2de:
Maybe release notes are missing. 
Depending on the schedule for 6.0 this fix could be targeted to 5.x?

dennisdoomen:
> Depending on the schedule for 6.0 this fix could be targeted to 5.x?

But then you have to target `master` and rebase.

dennisdoomen:
We can close this one, right?

lg2de:
replaced by #1273

Issue #902, opened 2018-08-24
Add support for Span and ReadOnlySpan


sergey-rybalkin:
### Description

Add support for [Span<T>](https://docs.microsoft.com/en-us/dotnet/api/system.span-1?view=netcore-2.1) and [ReadOnlySpan<T>](https://docs.microsoft.com/en-us/dotnet/api/system.readonlyspan-1?view=netcore-2.1). Though they do not directly implement `IEnumerable<T>` interface it would be nice to be able to assert using `NotContain` etc.

### Complete minimal example reproducing the issue

```csharp
// Arrange
ReadOnlySpan<string> span = new ReadOnlySpan<string>(new[] {"a", "b", "c"});

// Act
span = span.Slice(1);

// Assert
span.Should().NotContain("a");
```

### Expected behavior:

Test should pass.

### Actual behavior:

Compilation error saying extension method Should does not exist for ReadOnlySpan<T>.

### Versions

* Which version of Fluent Assertions are you using? 5.4.1
* Which .NET runtime and version are you targeting? .NET Core 2.1.

jnyrup:
Thanks for opening this issue.
I'm not sure this is natively possible in the way Fluent Assertions works.

`Should()` works as a series of extension on types, e.g.

https://github.com/fluentassertions/fluentassertions/blob/646dd3ba45146c537b0d56e242d180936e8e22d0/Src/FluentAssertions/AssertionExtensions.cs#L513-L516

which saves the string as a member in a new `StringAssertions`

https://github.com/fluentassertions/fluentassertions/blob/646dd3ba45146c537b0d56e242d180936e8e22d0/Src/FluentAssertions/Primitives/StringAssertions.cs#L21-L24

`Span<T>` has a restriction that it cannot be used as fields/properties in classes.

On the other hand, if the `Should()` extension converts the `Span<T>` into an `IEnumerable<T>` we should be able to support it.
One minor thing is, that this would either require `netcoreapp2.1` as a new target framework or add `System.Memory` as a dependency.

dennisdoomen:
.NET Core 2.1 is already released.

jnyrup:
Yeah `Span<T>` is *also* natively available in .Net Core 2.1. 
.Net Core 2.1 does not implement .Net Standard 2.1, but 2.0. 
.Net Core 3.0 is the first to implement .Net Standard 2.1.

dennisdoomen:
[...]

PR #1276, opened 2020-03-21
[WIP] create assertions for e.g. ReadOnlySpan

lg2de:
I found the issue #902 .
Here is a very simple prove of concept.
Do you think this could be implemented this way?

BTW: The changes on sln file was done automatically by Visual Studio, may be related to recent version 16.5.0.

### DIFF
--- a/FluentAssertions.sln
+++ b/FluentAssertions.sln
@@ -18,7 +18,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Specs", "Specs", "{963262D0
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "Shared.Specs", "Tests\Shared.Specs\Shared.Specs.shproj", "{0C23C12C-899F-4AD1-B737-B967A9910C08}"
 EndProject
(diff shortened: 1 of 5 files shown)


jnyrup:
Calling `ToArray` is also my best guess as the `Span` anyways would need to be converted into something that can be stored in `Subject`.

Question: 
* If we convert the `Span<T>` to an `IEnumerable<T>`, should we just return an `SelfReferencingCollectionAssertions` instead?

Note: netcore2.1 also supports these types out of the box.

Up for discussion:
* Should `ReadOnlySpan<char>` use `StringAssertions`?
* Should we add a dependency on `System.Memory` for net47 and netstandard2.0 to support these types across all targets?

dennisdoomen:
I don't think we should not be reusing the `SelfReferencingCollection` base-class here. Looking at the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.readonlyspan-1?view=netcore-3.1), it seems Microsoft already created a lot of methods to support this model. Although I'm in doubt if the performance impact is worth it.

jnyrup:
@dennisdoomen was the double negation in your first sentence intended?

dennisdoomen:
@lg2de any intentions to complete this PR?

lg2de:
I do not have (anymore) a real-world use case to complete this.
So, I'll close the PR for now.

PR #1278, opened 2020-03-21
Remove overloads of `[Not]BeCloseTo` that has a default precision

jnyrup:
Whether two time instances are "close" is context dependent and therefore Fluent Assertions should not try to guess a precision.


dennisdoomen:
Good idea!

jnyrup:
Just a heads up, I initially forgot to update the docs, but fixed that before merging.

PR #1280, opened 2020-03-22
Reduce usage of IsSameOrEqualTo

jnyrup:
`IsSameOrEqualTo` is necessary when comparing boxed numerics, e.g. that `1` is equal to `1L`.
`Type` and `XDocument` are not numerics, so this check is pure overhead and can be replaced with either `==` or `Equals`.

`KeyValuePair.Equals` uses `ValueType.Equals(object)` and causes boxing.
This is avoided by converting the `KeyValuePair` into a 2-tuple and compare it against the default 2-tuple.

This is split out from #1258

### DIFF
--- a/Tests/Shared.Specs/Xml/XDocumentAssertionSpecs.cs
+++ b/Tests/Shared.Specs/Xml/XDocumentAssertionSpecs.cs
@@ -57,6 +57,21 @@ namespace FluentAssertions.Specs
         }
 
         [Fact]
+        public void When_a_null_xml_document_is_equal_to_a_null_xml_document_it_should_succeed()
+        {
+            // Arrange
+            XDocument document = null;
+            XDocument otherXDocument = null;
+
+            // Act
+            Action act = () =>
+                document.Should().Be(otherXDocument);
+
+            // Assert
+            act.Should().NotThrow<XunitException>();
+        }
+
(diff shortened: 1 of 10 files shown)


Tests/Shared.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -56,6 +56,21 @@ public void When_asserting_a_null_xml_document_is_equal_to_another_xml_document_
...
+        [Fact]
+        public void When_a_null_xml_document_is_equal_to_a_null_xml_document_it_should_succeed()
+        {
+            // Arrange
+            XDocument document = null;
+            XDocument otherXDocument = null;
+
+            // Act
+            Action act = () =>
+                document.Should().Be(otherXDocument);
```
dennisdoomen: Every time I see this I wonder whether this should succeed or throw an `ArgumentNullException`.

jnyrup: That would be inconsistent with `==` and `Equals(object, object)`. 
Equality in C# is indeed hard.
As often Eric Lippert has blogged about it. https://ericlippert.com/2014/01/15/inconsistent-equality/

PR #1184, opened 2019-11-08
Use ordinal comparison in StringAssertions instead of culture-sensitive comparison

kimsey0:
## IMPORTANT 

* [ ] The code complies with the [Coding Guidelines for C#](https://www.csharpcodingguidelines.com/).
* [ ] The changes are covered by a new or existing set of unit tests which follow the Arrange-Act-Assert syntax such as is used [in this example](https://github.com/fluentassertions/fluentassertions/blob/daaf35b9b59b622c96d0c034e8972a020b2bee55/Tests/FluentAssertions.Shared.Specs/BasicEquivalencySpecs.cs#L33).
* [ ] If the contribution affects the documentation, please include your changes to [**documentation.md**](https://github.com/fluentassertions/fluentassertions/blob/master/docs/_pages/documentation.md) in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com).

Fixes #1096

PR #1283, opened 2020-03-29
Use Ordinal[IgnoreCase] string comparison

### LINKED ISSUE #1096: String .BeEquivalentTo() is case-sensitive on linux
It seems that `BeEquivalentTo()` doesn't ignore case when comparing strings on **Linux** environment.

```csharp
"SomeString".Should().BeEquivalentTo("somestring");
```

Expected: verification passes.

It works as expected on **Windows**, but throws `but "SomeString" differs near "Som" (index 0)` exception on **Linux**.

On the same environment string assertion from NUnit framework works as expected.
```csharp
StringAssert.AreEqualIgnoringCase("SomeString", "somestring");
```

Info:
- FluentAssertions v5.7.0
- .netcoreapp 2.1

jnyrup:
Unit tests should behave the same no matter the culture of the machine being run on.

* [Microsoft.CodeAnalysis.FxCopAnalyzers](https://github.com/dotnet/roslyn-analyzers#microsoftcodeanalysisfxcopanalyzers)
  * Helps find usages of methods where a `StringComparison` is not specified.
* [Microsoft.CodeAnalysis.BannedApiAnalyzers](https://github.com/dotnet/roslyn-analyzers#microsoftcodeanalysisbannedapianalyzers)
  * Used in conjuction with `BannedSymbols.txt` to prohibit usage of culture aware `StringComparison` and `CultureInfo`.

As netcoreapp2.1/netcoreapp3.0/netstandard2.1 has more string overloads than net47/netstandard2.0 I added some dummy overloads in `SystemExtensions` to satisfy the analyzers.

One large difference in behavior is that `TryConversionStep` now uses `CultureInfo.Invariant` instead of `CultureInfo.CurrentCulture`.

This fixes #1096
This closes #1184 (Thanks @kimsey0 for pushing this change forward)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -121,3 +121,16 @@ csharp_space_between_method_declaration_name_and_open_parenthesis = false
 csharp_space_between_method_declaration_parameter_list_parentheses = false
 csharp_space_between_parentheses = false
 csharp_space_between_square_brackets = false
+
+# FxCop Analyzers
+dotnet_diagnostic.CA1030.severity = none
+dotnet_diagnostic.CA1034.severity = none
+dotnet_diagnostic.CA1062.severity = none
+dotnet_diagnostic.CA1304.severity = error
+dotnet_diagnostic.CA1305.severity = none
+dotnet_diagnostic.CA1307.severity = error
+dotnet_diagnostic.CA1308.severity = error
+dotnet_diagnostic.CA1309.severity = error
+
+# Banned API Analyzers
+dotnet_diagnostic.RS0030.severity = error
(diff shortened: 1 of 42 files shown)


jnyrup:
> So using ordinal case-insensitive for string comparisons should be another design guideline. Maybe we should add them to the PR template?

Would fine be. 
Nevertheless one should get a compile time error now if trying due to 
* [CA1304](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1304?view=vs-2019): Specify CultureInfo
* [CA1307](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1307?view=vs-2019): Specify StringComparison
* [CA1308](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1308?view=vs-2019): Normalize strings to uppercase
* [CA1309](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1309?view=vs-2019): Use ordinal StringComparison

PR #1284, opened 2020-03-29
Allows using an IEqualityComparer<T> with BeEquivalentTo

### LINKED ISSUE #1236: Allow using an IEqualityComparer in BeEquivalentTo
> Actually, regarding my suggestion, it would also work if you could supply an EqualityComparer for EquivalencyAssertionOptions, or is it already possible? I'm just saying because EqualityComparer is the "standard" .NET way of doing it, and often there is already a suitable equality comparer available that could be reused as is. At least in our case, I like writing EqualityComparers because they nicely encapsulate the equivalency. I don't like repeating equivalency rules in-line, because usually the same rules are needed in many places. I guess you could encapsulate them as IEquivalencySteps, but why do that if you already have EqualityComparer?

For example

`orderDto.Should().BeEquivalentTo(order, options => options
    .Using<DateTimeEqualityComparer>());`
or

`orderDto.Should().BeEquivalentTo(order, options => options
    .Using<DateTime>(new DateTimeEqualityComparer()));`

dennisdoomen:
Adds overloads to the `Using` method of `BeEquivalentTo` options parameter to allows using an implementation of `IEqualityComparer<T>` as an alternative to `IEquivalencyStep`.

Resolves #1236

### DIFF
--- a/Src/FluentAssertions/Equivalency/EqualityComparerEquivalencyStep.cs
+++ b/Src/FluentAssertions/Equivalency/EqualityComparerEquivalencyStep.cs
@@ -1,0 +1,40 @@
+using System;
+using System.Collections.Generic;
+using FluentAssertions.Execution;
+
+namespace FluentAssertions.Equivalency
+{
+    public class EqualityComparerEquivalencyStep<T> : IEquivalencyStep
+    {
+        private readonly IEqualityComparer<T> comparer;
+
+        public EqualityComparerEquivalencyStep(IEqualityComparer<T> comparer)
+        {
+            this.comparer = comparer ?? throw new ArgumentNullException(nameof(comparer));
+        }
+
+        public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAssertionOptions config)
+        {
+            return config.GetExpectationType(context) == typeof(T);
+        }
+
+        public bool Handle(IEquivalencyValidationContext context, IEquivalencyValidator parent, IEquivalencyAssertionOptions config)
+        {
+            Execute.Assertion
+                .BecauseOf(context.Because, context.BecauseArgs)
+                .ForCondition(context.Subject is T)
+                .FailWith("Expected {context:object} to be of type {0}{because}, but found {1}", typeof(T), context.Subject)
+                .Then
+                .ForCondition(comparer.Equals((T)context.Subject, (T)context.Expectation))
+                .FailWith("Expected {context:object} to be equal to {1} according to {0}{because}, but {2} was not.",
+                    comparer.ToString(), context.Expectation, context.Subject);
+
+            return true;
+        }
+
+        public override string ToString()
+        {
+            return $"Use {comparer} for objects of type {typeof(T)}";
+        }
+    }
+}
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
@@ -423,6 +423,27 @@ namespace FluentAssertions.Equivalency
         }
 
         /// <summary>
+        /// Ensures the equivalency comparison will create and use an instance of <typeparamref name="TEqualityComparer"/>
+        /// that implements <see cref="IEqualityComparer{T}"/>, any time
+        /// when a property is of type <typeparamref name="T"/>.
+        /// </summary>
+        public TSelf Using<T, TEqualityComparer>() where TEqualityComparer : IEqualityComparer<T>, new()
+        {
+            return Using(new TEqualityComparer());
+        }
+
+        /// <summary>
+        /// Ensures the equivalency comparison will use the specified implementation of <see cref="IEqualityComparer{T}"/>
+        /// any time when a property is of type <typeparamref name="T"/>.
+        /// </summary>
+        public TSelf Using<T>(IEqualityComparer<T> comparer)
+        {
+            userEquivalencySteps.Insert(0, new EqualityComparerEquivalencyStep<T>(comparer));
+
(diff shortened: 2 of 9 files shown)


Src/FluentAssertions/Equivalency/EqualityComparerEquivalencyStep.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace FluentAssertions.Equivalency
+{
+    public class EqualityComparerEquivalencyStep<T> : IEquivalencyStep
+    {
+        private readonly IEqualityComparer<T> comparer;
+
+        public EqualityComparerEquivalencyStep(IEqualityComparer<T> comparer)
+        {
+            this.comparer = comparer;
```
jnyrup:
```suggestion
            this.comparer = comparer ?? throw new ArgumentNullException(nameof(comparer));
```

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
```diff
@@ -422,6 +422,28 @@ public TSelf Using(IEquivalencyStep equivalencyStep)
...
+        /// any time when a property is of type <typeparamref name="T"/>.
+        /// </summary>
+        public TSelf Using<T>(IEqualityComparer<T> comparer)
+        {
+            userEquivalencySteps.Insert(0, new EqualityComparerEquivalencyStep<T>(comparer));
+
+            return (TSelf)this;
+        }
+
+
```
jnyrup: nit: double empty newline

Src/FluentAssertions/Equivalency/EqualityComparerEquivalencyStep.cs
```diff
@@ -0,0 +1,35 @@
...
+        public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAssertionOptions config)
+        {
+            return context.RuntimeType == typeof(T);
+        }
+
+        public bool Handle(IEquivalencyValidationContext context, IEquivalencyValidator parent, IEquivalencyAssertionOptions config)
+        {
+            Execute.Assertion
+                .BecauseOf(context.Because, context.BecauseArgs)
+                .ForCondition(comparer.Equals((T)context.Subject, (T)context.Expectation))
```
jnyrup: ❓ Does the `CanHandle` check ensures that both `Subject` and `Expectation` can be casted to `T`?

Src/FluentAssertions/Equivalency/EqualityComparerEquivalencyStep.cs
```diff
@@ -0,0 +1,35 @@
...
+                .FailWith("Expected {context:object} to be equal according to {0}{because}, but it was not.", comparer.ToString());
```
jnyrup: Does the failure message contain both the subject and expectation?

docs/_pages/releases.md
```diff
@@ -16,6 +16,7 @@ sidebar:
 * The `Using`/`When` option on `BeEquivalentTo` will now use the conversion rules when trying to match the predicate - [#1257](https://github.com/fluentassertions/fluentassertions/pull/1257).
 * Added `NotBeWritable` to `PropertyInfoSelectorAssertions` to be able to assert that properties are not writable - [#1269](https://github.com/fluentassertions/fluentassertions/pull/1269).
 * Added extension to assert `TaskCompletionSource<T>` - [#1267](https://github.com/fluentassertions/fluentassertions/pull/1267).
+* Added the ability to pass an `IEqualityComparer<T>` through `BeEquivalentTo(x => x.Using<MyComparer>())` - [1284](https://github.com/fluentassertions/fluentassertions/pull/1284)
```
```suggestion
* Added the ability to pass an `IEqualityComparer<T>` through `BeEquivalentTo(x => x.Using<MyComparer>())` - [#1284](https://github.com/fluentassertions/fluentassertions/pull/1284).
```

Src/FluentAssertions/Equivalency/EqualityComparerEquivalencyStep.cs
dennisdoomen: Now it does ;-)

Src/FluentAssertions/Equivalency/EqualityComparerEquivalencyStep.cs
dennisdoomen: Now it does.

Src/FluentAssertions/Equivalency/EqualityComparerEquivalencyStep.cs
```diff
@@ -0,0 +1,40 @@
...
+        private readonly IEqualityComparer<T> comparer;
+
+        public EqualityComparerEquivalencyStep(IEqualityComparer<T> comparer)
+        {
+            this.comparer = comparer ?? throw new ArgumentNullException(nameof(comparer));
+        }
+
+        public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAssertionOptions config)
+        {
+            return context.RuntimeType == typeof(T);
```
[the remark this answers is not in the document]
dennisdoomen: I'm not sure. That one will also unwrap nullable types.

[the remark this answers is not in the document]
dennisdoomen: But the consistency aspect is important enough to change it.

jnyrup:
It would maybe be nice with a test hitting the `UseRuntimeTypes`.

dennisdoomen:
> It would maybe be nice with a test hitting the `UseRuntimeTypes` case.

Because you insist ;-)

jnyrup:
Woohoo!

PR #1289, opened 2020-03-29
Support chaining `NotThrowAfter` and `NotThrowAsync`

### LINKED ISSUE #990: Assertions for Func<Task<T>>

jnyrup:
This PR enables you to continue asserting on the result of `Func<Task<int>>` when it does not throw.

```c#
Func<Task<int>> func = () => Task.FromResult(42);

func.Should().NotThrow()
    .Which.Should().Be(42);

(await func.Should().NotThrowAsync())
    .Which.Should().Be(42);

func.Should().NotThrowAfter(10.Seconds(), 1.Seconds())
    .Which.Should().Be(42)

(await func.Should().NotThrowAfterAsync(10.Seconds(), 1.Seconds()))
    .Which.Should().Be(42);
```

The implementations and tests are mostly copied from the non-generic or sync versions.
To avoid the redefinitions of `Subject` in each class two changes in generics:
* `DelegateAssertions` and `AsyncFunctionAssertions` are now generic in `TAssertions`.
* `AsyncFunctionAssertions` is also generic in `TTask : Task` to support
  both `Task` and `Task<T>` equally.

This fixes #990

### DIFF
--- a/Src/FluentAssertions/Specialized/GenericAsyncFunctionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/GenericAsyncFunctionAssertions.cs
@@ -6,18 +6,19 @@ using FluentAssertions.Execution;
 
 namespace FluentAssertions.Specialized
 {
-    public class GenericAsyncFunctionAssertions<TResult> : AsyncFunctionAssertions
+    public class GenericAsyncFunctionAssertions<TResult> : AsyncFunctionAssertions<Task<TResult>, GenericAsyncFunctionAssertions<TResult>>
     {
-        private readonly Func<Task<TResult>> subject;
-
         public GenericAsyncFunctionAssertions(Func<Task<TResult>> subject, IExtractExceptions extractor) : this(subject, extractor, new Clock())
         {
         }
 
-        public GenericAsyncFunctionAssertions(Func<Task<TResult>> subject, IExtractExceptions extractor, IClock clock) : base(
-            subject, extractor, clock)
+        public GenericAsyncFunctionAssertions(Func<Task<TResult>> subject, IExtractExceptions extractor, IClock clock) : base(subject, extractor, clock)
         {
-            this.subject = subject;
+        }
+
+        private new TResult InvokeSubject()
+        {
+            return Subject.ExecuteInDefaultSynchronizationContext().GetAwaiter().GetResult();
         }
 
         /// <summary>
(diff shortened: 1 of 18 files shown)


Src/FluentAssertions/Specialized/GenericAsyncFunctionAssertions.cs
```diff
@@ -90,5 +90,108 @@ public GenericAsyncFunctionAssertions(Func<Task<TResult>> subject, IExtractExcep
...
+        /// Asserts that the current <see cref="Task{T}"/> does not throw any exception.
+        /// </summary>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <see cref="because" />.
+        /// </param>
+        public new async Task<AndWhichConstraint<GenericAsyncFunctionAssertions<TResult>, TResult>> NotThrowAsync(string because = "", params object[] becauseArgs)
```
dennisdoomen: I had to read the original issue to understand why all of these new methods were necessary. Would be nice to make that a bit clearer in the PR description.

Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
```diff
@@ -129,5 +132,205 @@ public void When_task_completes_slow_async_it_should_fail()
...
+            Func<Task> action = () => throwLongerThanWaitTime.Should(clock)
+                .NotThrowAfterAsync(waitTime, pollInterval, "we passed valid arguments");
+
+            // Assert
+            action.Should().Throw<XunitException>()
+                .WithMessage("Did not expect any exceptions after 2s because we passed valid arguments*");
+        }
+
+        [Fact]
+        public void When_no_exception_should_be_thrown_for_async_func_after_wait_time_and_none_was_it_should_not_throw()
```
dennisdoomen: I would skip the `async_func` part since that should be clear from the context.

Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
```diff
@@ -129,5 +132,205 @@ public void When_task_completes_slow_async_it_should_fail()
             // Assert
             action.Should().Throw<XunitException>();
         }
+        #endregion
+
+        #region NotThrowAsync
+        [Fact]
+        public void When_subject_it_null_when_expecting_to_not_throw_async_it_should_throw()
```
dennisdoomen: Typo

jnyrup: fixed

Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
jnyrup: fixed

jnyrup:
Updated PR with more widespread changes.

jnyrup:
@dennisdoomen this should be ready for (another) review now.

In continuation of #1296 I've added more generics:
* `T` to reuse `T Subject` from `ReferenceAssertions<T>`.
* `TAssertions` to improve continuation on classes that inherits `DelegateAssertions`.

PR #1295, opened 2020-03-30
Overriding DefaultValueFormatter and EnumerableValueFormatter

jnyrup:
This PR makes  `DefaultValueFormatter` and `EnumerableValueFormatter` more suitable for extending for making it easier to create custom formatters.

### DIFF
--- a/Src/FluentAssertions/Formatting/DefaultValueFormatter.cs
+++ b/Src/FluentAssertions/Formatting/DefaultValueFormatter.cs
@@ -9,12 +9,13 @@ namespace FluentAssertions.Formatting
 {
     public class DefaultValueFormatter : IValueFormatter
     {
-        #region Private Definitions
-
         private const int RootLevel = 0;
-        private const int SpacesPerIndentionLevel = 3;
 
-        #endregion
+        /// <summary>
+        /// The number of spaces to indent the members of this object by.
+        /// </summary>
+        /// <remarks>The default value is 3.</remarks>
+        protected virtual int SpacesPerIndentionLevel { get; } = 3;
 
         /// <summary>
         /// Determines whether this instance can handle the specified value.
@@ -23,7 +24,7 @@ namespace FluentAssertions.Formatting
         /// <returns>
         /// <c>true</c> if this instance can handle the specified value; otherwise, <c>false</c>.
         /// </returns>
-        public bool CanHandle(object value)
+        public virtual bool CanHandle(object value)
         {
             return true;
         }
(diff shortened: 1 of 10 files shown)


Src/FluentAssertions/Formatting/DefaultValueFormatter.cs
```diff
@@ -9,12 +9,9 @@ namespace FluentAssertions.Formatting
 {
     public class DefaultValueFormatter : IValueFormatter
     {
-        #region Private Definitions
-
         private const int RootLevel = 0;
-        private const int SpacesPerIndentionLevel = 3;
 
-        #endregion
+        protected virtual int SpacesPerIndentionLevel { get; } = 3;
```
dennisdoomen, line 14: Can we document all public and protected properties?

Src/FluentAssertions/Formatting/DefaultValueFormatter.cs
```diff
@@ -40,20 +37,24 @@ public string Format(object value, FormattingContext context, FormatChild format
...
-                return prefix + GetTypeAndPublicPropertyValues(value, context, formatChild);
+                return prefix + GetTypeAndMemberValues(value, context, formatChild);
             }
 
             return prefix + value;
         }
 
+        protected virtual IEnumerable<SelectedMemberInfo> GetMembers(Type type) => type.GetNonPrivateMembers();
+
+        protected virtual string GetTypeName(Type type) => type.FullName;
```
dennisdoomen, line 48: Why do you need this in sub-classes?

Src/FluentAssertions/Formatting/EnumerableValueFormatter.cs
```diff
@@ -7,14 +7,16 @@ namespace FluentAssertions.Formatting
 {
     public class EnumerableValueFormatter : IValueFormatter
     {
+        protected virtual int MaxItems { get; } = 32;
```
dennisdoomen, line 10: Why do you need this?

Src/FluentAssertions/Formatting/DefaultValueFormatter.cs
jnyrup: Of course, I'll add that.

Src/FluentAssertions/Formatting/DefaultValueFormatter.cs
jnyrup: At work I made a custom formatter that only returned `type.Name` as all types where from the same namespace.
It cluttered up the message as the subject was a collection of quite nested objects.

So it went from
```
A.Very.Long.Namespace.Class1
{
   Prop1 = A.Very.Long.Namespace.Class2
   {
      ...
   }
}
```
into
```
Class1
{
   Prop1 = Class2
   {
      ...
   }
}```

But I agree it might be out of scope for most people who would want to extend this class.

Maybe it just needs a better name?
`FormatType`
`FormatTypeName`

Src/FluentAssertions/Formatting/EnumerableValueFormatter.cs
jnyrup: To limit/extend the number of formatted items for a custom formatter.
Am I just repeating what the code does, instead of why it's necessary?

Src/FluentAssertions/Formatting/DefaultValueFormatter.cs
dennisdoomen: Ah, maybe `TypeDisplayName`

Src/FluentAssertions/Formatting/EnumerableValueFormatter.cs
dennisdoomen: Ah, you made the `EnumerableValueFormatter ` also extensible. I didn't get that immediatelly.

Src/FluentAssertions/Formatting/DefaultValueFormatter.cs
jnyrup: Eureka 🎉 
DisplayName, that's the right word!
Naming is hard.

jnyrup:
All your feedback should be resolved now.

PR #1298, opened 2020-04-04
Add support for IImmutableDictionary and IReadOnlyDictionary

### LINKED ISSUE #357: IReadOnlyDictionary
Hi,

We should be able to call the ContainKey() extension on a IReadOnlyDictionary:

``` C#
IReadOnlyDictionary<TKey, TValue> dictionary;
dictionary.Should().ContainKey(...)
```

### LINKED ISSUE #568: Support for IReadOnlyDictionary, IReadOnlyCollection

jnyrup:
This PR enables using `GenericDictionaryAssertions` on `IReadOnlyDictionary`.
More precisely it enables support for any 
```
TCollection : IEnumerable<KeyValuePair<TKey, TValue>>
```

### `Should` analysis:
Added two new overloads:
```c#
Should<TKey, TValue>(IEnumerable<KeyValuePair<TKey, TValue>>)
Should<TCollection, TKey, TValue>(TCollection)
    where TCollection : IEnumerable<KeyValuePair<TKey, TValue>>
```

The first one is picked by both:
* `IEnumerable<KeyValuePair<TKey, TValue>>`
* `IReadOnlyDictionary<TKey, TValue>`

The second overload is not picked up by anything and probably never will(?), as being an interface is regarded a better match than being *generic* in an interface.
[Related article](https://dotnetfalcon.com/back-to-the-basics-overload-resolution-in-c/)
Thus not sure if having the second one is better/worse that *not* having it.

As noted in previous discussions we cannot add 
```c#
Should(IReadOnlyDictionary<TKey, TValue>)
```
because that leads to ambiguity when writing
```c#
new Dictionary<int,int>().Should()
```
as it cannot tie-break picking `IReadOnlyDictionary` or `IDictionary`.

### `GenericCollectionAssertions` inheritance
`GenericDictionaryAssertions` now inherits from `GenericCollectionAssertions` where it previously inherited from `ReferenceTypeAssertions`.
This has at least two benefits:
All `Count`/`Empty` assertions now comes for free.
Their failure messages were *slightly* different, which are the changes to `GenericCollectionAssertions` and failure message assertions in existing tests.

The *only* assertions for `IEnumerable<kvp>.Should()` that *might* be different 
are `[Not]Contain(KeyValuePair<TKey, TValue>` as they are marked with `new` modifier.

### `[Not]Equal`
I've made `Equal` and `NotEqual` such that they do not require the subject and expectation to be the same `TCollection`.
So an `IDictionary` can be compared against an `IReadOnlyDictionary` and vice versa.

### What's up with all the helper methods?
I wanted to avoid littering the calls to e.g. `GetValue` in assertion bodies with type parameters.
I.e. I wanted to be able to write `GetValues(Subject)` instead of `GetValues<TCollection, TKey, TValue>(Subject)`.
This way compiled and worked.
If there is a nicer way to accomplish this, I'm all eager to know.

### Tests
I've added a set of tests to verify that all cases in all helper methods are hit.
I attached a debugger to verify that the expected cases were hit.
For a proper testing of this I would normally use a mock.

If we allowed to expose `internal` members to the test project it could also be tested with the existing `TrackingTestDictionary` class.
Any opinion/concerns?

A huge thanks to *everyone* who participated in the discussions over the years(!) on how to solve this.
@Evangelink @krajek @somewhatabstract 

This fixes #568 
This fixes #357
This closes #1132 
This closes #963

### DIFF
--- a/Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
@@ -1,402 +1,40 @@
 ﻿using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-using System.Linq.Expressions;
 using FluentAssertions.Common;
 using FluentAssertions.Equivalency;
 using FluentAssertions.Execution;
-using FluentAssertions.Primitives;
 
 namespace FluentAssertions.Collections
 {
     /// <summary>
-    /// Contains a number of methods to assert that an <see cref="IDictionary{TKey,TValue}"/> is in the expected state.
+    /// Contains a number of methods to assert that a <typeparamref name="TCollection"/> is in the expected state.
     /// </summary>
     [DebuggerNonUserCode]
-    public class GenericDictionaryAssertions<TKey, TValue> :
-        GenericDictionaryAssertions<TKey, TValue, GenericDictionaryAssertions<TKey, TValue>>
+    public class GenericDictionaryAssertions<TCollection, TKey, TValue> :
+        GenericDictionaryAssertions<TCollection, TKey, TValue, GenericDictionaryAssertions<TCollection, TKey, TValue>>
+        where TCollection : IEnumerable<KeyValuePair<TKey, TValue>>
     {
-        public GenericDictionaryAssertions(IDictionary<TKey, TValue> dictionary) : base(dictionary)
+        public GenericDictionaryAssertions(TCollection keyValuePairs)
+            : base(keyValuePairs)
         {
         }
     }
 
     /// <summary>
-    /// Contains a number of methods to assert that an <see cref="IDictionary{TKey,TValue}"/> is in the expected state.
+    /// Contains a number of methods to assert that a <typeparamref name="TCollection"/> is in the expected state.
     /// </summary>
     [DebuggerNonUserCode]
-    public class GenericDictionaryAssertions<TKey, TValue, TAssertions> :
-        ReferenceTypeAssertions<IDictionary<TKey, TValue>, TAssertions>
-        where TAssertions : GenericDictionaryAssertions<TKey, TValue, TAssertions>
+    public class GenericDictionaryAssertions<TCollection, TKey, TValue, TAssertions> :
+        GenericCollectionAssertions<TCollection, KeyValuePair<TKey, TValue>, TAssertions>
+        where TCollection : IEnumerable<KeyValuePair<TKey, TValue>>
+        where TAssertions : GenericDictionaryAssertions<TCollection, TKey, TValue, TAssertions>
     {
-        public GenericDictionaryAssertions(IDictionary<TKey, TValue> dictionary) : base(dictionary)
+        public GenericDictionaryAssertions(TCollection keyValuePairs)
+            : base(keyValuePairs)
         {
         }
-
-        #region HaveCount
-
-        /// <summary>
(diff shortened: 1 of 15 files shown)


Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
```diff
@@ -1275,5 +920,28 @@ public AndConstraint<TAssertions> NotContain(params KeyValuePair<TKey, TValue>[]
         /// Returns the type of the subject the assertion applies on.
         /// </summary>
         protected override string Identifier => "dictionary";
+
+        private static IEnumerable<TKey> GetKeys(TCollection collection) =>
```
dennisdoomen, line 924: Do you still need these if you already have the helper extension methods?

Src/FluentAssertions/Common/DictionaryHelpers.cs
```diff
@@ -0,0 +1,71 @@
...
+                IDictionary<TKey, TValue> dictionary => dictionary.Keys,
+                IReadOnlyDictionary<TKey, TValue> readOnlyDictionary => readOnlyDictionary.Keys,
+                _ => collection.Select(kvp => kvp.Key),
+            };
+        }
+
+        public static IEnumerable<TValue> GetValues<TCollection, TKey, TValue>(this TCollection collection)
+            where TCollection : IEnumerable<KeyValuePair<TKey, TValue>>
+        {
+            return collection switch
```
dennisdoomen, line 22: Although I don't like switching on types, it's a nice example of using the new C# syntax.

Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
jnyrup: Yes.
Using e.g. `Subject.GetKeys()` from the helper class directly in an assertion body gives me `CS0411`

>	The type arguments for method 'DictionaryHelpers.GetKeys<TCollection, TKey, TValue>(TCollection)' cannot be inferred from the usage. 
> Try specifying the type arguments explicitly.

The purpose of all these methods in the bottom is to explicitly pass in the type arguments to the helper class, to avoid needing to specify them in the assertion bodies.

Src/FluentAssertions/Common/DictionaryHelpers.cs
jnyrup: I'm curious why :)
Would your normally prefer regular conditionals when needing to pattern match?

```c#
if (collection is IDictionary<TKey, TValue> dictionary)
{
   return dictionary.Values;
}

if (IReadOnlyDictionary<TKey, TValue> readOnlyDictionary)
{
   return readOnlyDictionary.Values;
}

return collection.Select(kvp => kvp.Value);
```

eNeRGy164: For me, I think a possible reason could be that my eyes don't pick it up as a switch statement yet. They layout of the code is not implying it. But that might just take time.

jnyrup: @eNeRGy164 as in the GH markdown formatting not coloring the keyword switch in the switch expression?
```c#
switch (a)
{
    case I i: return i;
}
```

```c#
return a switch
{
    I i => i;
}
```

dennisdoomen: > I'm curious why :)

I prefer polymorphistic behavior ;-)

PR #1307, opened 2020-04-17
Remove unnecessary usage of GetTypeInfo

jnyrup:
Since we've moved away from netstandard 1.3 and 1.6 it's no longer
necessary to call `GetTypeInfo` as `Type` now has all the members we
need.

Also removed some usage of LINQ by using the library method `IsDefined`.

### DIFF
--- a/Src/FluentAssertions/Common/MethodInfoExtensions.cs
+++ b/Src/FluentAssertions/Common/MethodInfoExtensions.cs
@@ -24,10 +24,7 @@ namespace FluentAssertions.Common
         internal static IEnumerable<TAttribute> GetMatchingAttributes<TAttribute>(this MemberInfo memberInfo, Expression<Func<TAttribute, bool>> isMatchingAttributePredicate)
             where TAttribute : Attribute
         {
-            var customAttributes = memberInfo.GetCustomAttributes(
-                typeof(TAttribute), false)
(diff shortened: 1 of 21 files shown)


Src/FluentAssertions/Common/MethodInfoExtensions.cs
```diff
@@ -24,10 +24,7 @@ internal static bool IsAsync(this MethodInfo methodInfo)
         internal static IEnumerable<TAttribute> GetMatchingAttributes<TAttribute>(this MemberInfo memberInfo, Expression<Func<TAttribute, bool>> isMatchingAttributePredicate)
             where TAttribute : Attribute
         {
-            var customAttributes = memberInfo.GetCustomAttributes(
-                typeof(TAttribute), false)
-                .Cast<TAttribute>()
-                .ToList();
+            var customAttributes = memberInfo.GetCustomAttributes<TAttribute>(false).ToList();
```
lg2de, line 27: I would recommend to always add parameter name for boolean parameters.
For reviewer (like me just now) it is unclear what `false` does mean.

jnyrup: I agree. 
I'll create another PR with changes through out the code base.

PR #1309, opened 2020-04-18
Use `using` statements to reduce indentation

jnyrup:
This is a clear statement about using `using` statements!

Best viewed ~using IE6 in 1024x768~ ignoring whitespace.


lg2de:
> Best viewed ignoring whitespace.

I did not know that github has an option therefore.
Thanks!

PR #1310, opened 2020-04-18
NRE for GetPropertyInfo with unexpected Expression

jnyrup:
Not all `Operand`s of `UnaryExpression` are `MemberExpression`s.

I was confused with the naming `AttemptToGetMemberInfoFromCastExpression`, so had to read up on when the `Body` of `Expression<Func<T, TValue>>` would be a `MemberExpression` vs `UnaryExpression`.
https://stackoverflow.com/a/12975480/1087627
Apparently it happens when `TValue` is a value type, that's why I added `int SomeOtherProperty`.

`GetSelectedMemberInfo` wasn't used anywhere, so out it goes.

### DIFF
--- a/Tests/FluentAssertions.Specs/Events/EventAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Events/EventAssertionSpecs.cs
@@ -489,6 +489,22 @@ namespace FluentAssertions.Specs
             ((object)innerScope).Should().NotBeSameAs(outerScope);
         }
 
+        [Fact]
+        public void When_monitoring_an_object_with_invalid_property_expression_it_should_throw()
+        {
+            // Arrange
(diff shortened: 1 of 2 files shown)


Tests/FluentAssertions.Specs/Events/EventAssertionSpecs.cs
```diff
@@ -537,6 +537,22 @@ public void When_nesting_monitoring_requests_scopes_should_be_isolated()
             }
         }
 
+        [Fact]
+        public void When_monitoring_an_object_invalid_property_expression_should_throw()
```
dennisdoomen, line 541: Are you missing the _with_ here?

jnyrup: yes :)

Issue #1165, opened 2019-10-14
DateTime Should().BeLessThan(...).After(b) not check After


SvetlanaNesterova:
Hello!

It seems strange for me that DateTimeAssertions works in this way:

```c#
[Test]
public void Test_before_and_after()
{
    var a = new DateTime(2019, 01, 01, 12, 0, 0, DateTimeKind.Utc);
    var b = new DateTime(2019, 01, 01, 12, 30, 0, DateTimeKind.Utc);
    a.Should().BeLessThan(1.Hours()).Before(b);
    a.Should().BeLessThan(1.Hours()).After(b);
}
```
These methods check that difference is less than value, and it is true when difference is negative. But is not correct to say that `a` is less than an hour **after** `b` when `a - b == -30.Minutes()`
 
I think there should be assertion that difference (`TimeSpan actual`) should be not negative in methods `DateTimeRangeAssertions.Before` and `DateTimeRangeAssertions.After`.

https://github.com/fluentassertions/fluentassertions/blob/0a37562a3eebc9a3269a8f5965259cdc170306c8/Src/FluentAssertions/Primitives/DateTimeRangeAssertions.cs#L75-L80

https://github.com/fluentassertions/fluentassertions/blob/0a37562a3eebc9a3269a8f5965259cdc170306c8/Src/FluentAssertions/Primitives/DateTimeRangeAssertions.cs#L116-L121

dennisdoomen:
>  But is not correct to say that a is less than an hour after b when a - b == -30.Minutes()

I don't get it. Are you saying that the assertion API is not grammatically correct?

SvetlanaNesterova:
A = 12:00
B = 12:30
I want to know: is A less than 1 hour after B?
I think correct answer is "No", because A is before B, not after.
I expected it is equivalent to statement `A in interval (B; B + 1 hour)`. And I was surprised when I changed "Before" on "After" in my test and it didn't fail.

If you think current behaviour is correct, I will close the issue.

jonathann92:
Hey guys I'm running into this issue as well. I agree with @SvetlanaNesterova that it should fail the test in his original case. Do you know if this is being worked on?

dennisdoomen:
We kind of missed it. I've just flagged it as a bug to give it more attention.

jnyrup:
@jonathann92 To our knowledge no one is working on this issue, but we're open to contributions.

jonathann92:
[...]

PR #1313, opened 2020-04-19
date time range assertions - asserting difference is not negative

jonathann92:
## IMPORTANT 

* [x] If the contribution affects [the documentation](https://github.com/fluentassertions/fluentassertions/tree/master/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com).

Fixing issue where assertion on `DateTimeRange` will fail on `BeLessThan` or `BeWithin` when the subject and target has a negative delta.

Reported in #1165

### DIFF
--- a/Src/FluentAssertions/Primitives/DateTimeOffsetRangeAssertions.cs
+++ b/Src/FluentAssertions/Primitives/DateTimeOffsetRangeAssertions.cs
@@ -74,14 +74,21 @@ namespace FluentAssertions.Primitives
             {
                 TimeSpan actual = target - subject.Value;
 
-                if (!predicate.IsMatchedBy(actual, timeSpan))
-                {
-                    Execute.Assertion
-                        .BecauseOf(because, becauseArgs)
-                        .FailWith(
-                            "Expected {context:the date and time} to be " + predicate.DisplayText +
-                            " {1} before {2}{reason}, but {0} differs {3}.", subject, timeSpan, target, actual);
-                }
+                Execute.Assertion
+                    .ForCondition(actual >= TimeSpan.Zero)
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith(
+                        "Expected {context:the date and time} {0} to be " + predicate.DisplayText +
+                        " {1} before {2}{reason}, but it is after by {3}.",
+                        subject, timeSpan, target, actual.Duration());
+
+                Execute.Assertion
+                    .ForCondition(predicate.IsMatchedBy(actual, timeSpan))
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith(
+                        "Expected {context:the date and time} to be " + predicate.DisplayText +
+                        " {1} before {2}{reason}, but {0} differs {3}.", subject, timeSpan, target, actual);
+
             }
 
             return new AndConstraint<TAssertions>(parentAssertions);
@@ -112,15 +119,22 @@ namespace FluentAssertions.Primitives
             {
                 TimeSpan actual = subject.Value - target;
 
-                if (!predicate.IsMatchedBy(actual, timeSpan))
-                {
-                    Execute.Assertion
-                        .BecauseOf(because, becauseArgs)
-                        .FailWith(
-                            "Expected {context:the date and time} to be " + predicate.DisplayText +
-                            " {0} after {1}{reason}, but {2} differs {3}.",
-                            timeSpan, target, subject, actual);
-                }
+                Execute.Assertion
+                    .ForCondition(actual >= TimeSpan.Zero)
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith(
+                        "Expected {context:the date and time} {0} to be " + predicate.DisplayText +
+                        " {1} after {2}{reason}, but it is before by {3}.",
+                        subject, timeSpan, target, actual.Duration());
+
+                Execute.Assertion
+                    .ForCondition(predicate.IsMatchedBy(actual, timeSpan))
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith(
+                        "Expected {context:the date and time} to be " + predicate.DisplayText +
+                        " {0} after {1}{reason}, but {2} differs {3}.",
+                        timeSpan, target, subject, actual);
+                
             }
 
             return new AndConstraint<TAssertions>(parentAssertions);
--- a/Src/FluentAssertions/Primitives/DateTimeOffsetRangeAssertions.cs
+++ b/Src/FluentAssertions/Primitives/DateTimeOffsetRangeAssertions.cs
@@ -75,20 +75,12 @@ namespace FluentAssertions.Primitives
                 TimeSpan actual = target - subject.Value;
 
                 Execute.Assertion
-                    .ForCondition(actual >= TimeSpan.Zero)
-                    .BecauseOf(because, becauseArgs)
-                    .FailWith(
-                        "Expected {context:the date and time} {0} to be " + predicate.DisplayText +
-                        " {1} before {2}{reason}, but it is after by {3}.",
-                        subject, timeSpan, target, actual.Duration());
-
-                Execute.Assertion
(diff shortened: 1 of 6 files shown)


jnyrup:
Thanks for looking in this issue 👍 

I like the idea about replacing `differs` with the explicit `before`/`after`.
It's more human-friendly to read `before by 10s`, than `differs by -10s`.

Could this be simplified by moving the `actual >= TimeSpan.Zero` into the dictionary of `predicates` 
E.g.
```c#
ts1 >= TimeSpan.Zero && ts1 > ts2
```

Then if `IsMatched` fails, whether `actual` is positive or negative can control whether the failure message should be `but it is after` or `but it is before`.

Src/FluentAssertions/Primitives/DateTimeOffsetRangeAssertions.cs
```diff
@@ -112,15 +119,22 @@ public AndConstraint<TAssertions> After(DateTimeOffset target, string because =
...
-                            "Expected {context:the date and time} to be " + predicate.DisplayText +
-                            " {0} after {1}{reason}, but {2} differs {3}.",
-                            timeSpan, target, subject, actual);
-                }
+                Execute.Assertion
+                    .ForCondition(actual >= TimeSpan.Zero)
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith(
+                        "Expected {context:the date and time} {0} to be " + predicate.DisplayText +
+                        " {1} after {2}{reason}, but it is before by {3}.",
```
dennisdoomen, line 127: I think it should be something like _but it is behind by_.

Src/FluentAssertions/Primitives/DateTimeRangeAssertions.cs
```diff
@@ -75,15 +75,22 @@ public class DateTimeRangeAssertions<TAssertions>
...
-                            "Expected date and/or time {0} to be " + predicate.DisplayText +
-                            " {1} before {2}{reason}, but it differs {3}.",
-                            subject, timeSpan, target, actual);
-                }
+                Execute.Assertion
+                    .ForCondition(actual >= TimeSpan.Zero)
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith(
+                        "Expected {context:the date and time} {0} to be " + predicate.DisplayText +
+                        " {1} before {2}{reason}, but it is after by {3}.",
```
dennisdoomen, line 83: Instead of _after by_, use _ahead by_

Tests/FluentAssertions.Specs/Primitives/DateTimeAssertionSpecs.cs
```diff
@@ -1902,6 +1902,144 @@ public void When_time_is_less_than_30s_after_another_time_it_should_not_throw()
             subject.Should().BeLessThan(TimeSpan.FromSeconds(30)).After(target);
         }
 
+        [Fact]
+        public void When_asserting_subject_be_morethan_10_seconds_after_target_but_subject_is_before_target_should_throw()
```
dennisdoomen, line 1906: (Nitpicking) Missing _  between _more_ and _than_

Tests/FluentAssertions.Specs/Primitives/DateTimeAssertionSpecs.cs
```diff
@@ -1902,6 +1902,144 @@ public void When_time_is_less_than_30s_after_another_time_it_should_not_throw()
...
+        {
+            // Arrange
+            var target = 1.January(0001).At(0, 0, 30);
```
dennisdoomen, line 1909: Instead of _target_ we use the term _expectation_ everywhere.

Tests/FluentAssertions.Specs/Primitives/DateTimeAssertionSpecs.cs
jonathann92: I was debating if i should have a `be_more_than` or `bemorethan` because the method name is `BeMoreThan` but somehow I went in the middle lol. I'll put an underscore to between the words.

jonathann92:
@dennisdoomen 

> I think it should be something like _but it is behind by_.

> Instead of _after by_, use _ahead by_

My reasoning for using `before` and `after` is that its the same wording as the method name. 

Also ahead can mean "before" in some contexts other than time and this can lead to confusion. For example, there are 2 people in a line: 

- The 1st person is _before_ the 2nd person
- The 1st person is _ahead_ of the 2nd person

dennisdoomen:
> _but it is before by 2 ms_

I'm not a native speaker, but somehow it feels wrong. I'm wondering what is the _correct_ English wording. Giving you're from CA, I guess you should know better than me.

Tests/FluentAssertions.Specs/Primitives/DateTimeAssertionSpecs.cs
dennisdoomen: > Don’t refer to technical names such as the names of methods, variables and UI elements in the name of a test. In my opinion, the name of a test should explain the functional scenario, whereas the cause and effect should clearly illustrate the technicalities of that scenario. If some class or member is renamed, it shouldn’t affect the functional name of the test.

From https://www.continuousimprover.com/2015/11/12-tips-to-write-unit-tests-that-dont.html

In other words, I never use the literal name of a specific method in a test name.

jnyrup:
The no native speaker disclaimer applies to me as well.

"but it is before by 2 ms" sounds like a mix of
"but it is behind by 2 ms" and "but it is 2 ms before".

I think behind/ahead is used with time if e.g. your watch is out of sync, then 
"it is behind by 2s".

jonathann92:
Ok it makes sense when we're talking about times. I changed the messaging to ahead/behind instead of after/before.

dennisdoomen:
Are you willing to also update the [release notes](https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/releases.md#600)?

Tests/FluentAssertions.Specs/Primitives/DateTimeAssertionSpecs.cs
```diff
@@ -1902,6 +1902,144 @@ public void When_time_is_less_than_30s_after_another_time_it_should_not_throw()
...
+            var subject = 1.January(0001).At(0, 0, 15);
+
+            // Act / Assert
+            Action action = () => subject.Should().BeMoreThan(10.Seconds()).After(target);
```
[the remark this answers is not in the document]
dennisdoomen: > We do use target elsewhere in DateTimeOffsetAssertionSpecs.cs

@jnyrup legacy code ;-)

Tests/FluentAssertions.Specs/Primitives/DateTimeAssertionSpecs.cs
jonathann92: @jnyrup I can change `target` to `expectation` in the files I touched. Would you guys also like to me to change it in `DateTimeOffsetRangeAssertions.cs` and `DateTimeOffsetRangeAssertions.cs`?

jonathann92:
> Are you willing to also update the [release notes](https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/releases.md#600)?

Does this wording sound right?
```Asserting DateTime(Offset) `Before` will fail when subject is after and vice versa```

jnyrup:
@jonathann92 I would write it in past tense, since it's now fixed.
@dennisdoomen Any comments on how to formulate this in the release notes?

dennisdoomen:
> Any comments on how to formulate this in the release notes?

Taking from the original bug report:

    Ensures that date time assertions like "a is less than an hour after b" don't succeed when `a - b == -30.Minutes()`

jnyrup:
@jonathann92 again, thanks for taking the time to look this issue and contributing a fix ❤️

dennisdoomen:
Kudos for @jonathann92 indeed. With all these contributions, 6.0 is going to be awesome.

PR #1315, opened 2020-04-19
Constrain parameter of [Not]HaveFlag struct enums

jnyrup:
In C# 7.3 the enum constraint was introduced, which lets us restrict the parameter of `[Not]HaveFlag` to a struct enum.

This prevents an NRE if one passed in `null` - as it no longer compiles.

### DIFF
--- a/Src/FluentAssertions/Primitives/ObjectAssertions.cs
+++ b/Src/FluentAssertions/Primitives/ObjectAssertions.cs
@@ -208,8 +208,9 @@ namespace FluentAssertions.Primitives
         /// <param name="becauseArgs">
         /// Zero or more values to use for filling in any <see cref="string.Format(string,object[])" /> compatible placeholders.
         /// </param>
-        public AndConstraint<ObjectAssertions> HaveFlag(Enum expectedFlag, string because = "",
+        public AndConstraint<ObjectAssertions> HaveFlag<TEnum>(TEnum expectedFlag, string because = "",
             params object[] becauseArgs)
+            where TEnum : struct, Enum
         {
             Execute.Assertion
                 .BecauseOf(because, becauseArgs)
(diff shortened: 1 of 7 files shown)


dennisdoomen:
What does that do to the binary and source-code compatibility?

jnyrup:
Currently one can pass in null to `HaveFlag`, as the parameter is `Enum` is a reference type, causing an NRE.

```
using System;
public class Program {
    enum MyEnum
    {
        A
    }
    
    static bool Foo(Enum t)
        => MyEnum.A.HasFlag(t);
    
    static bool Bar<TEnum>(TEnum t)
        where TEnum : struct, Enum 
        => MyEnum.A.HasFlag(t);

    static bool Baz<TEnum>(TEnum t)
        where TEnum : Enum 
        => MyEnum.A.HasFlag(t);
   
    
    static void Main()
    {
        MyEnum nativeEnum = MyEnum.A;
        Enum classEnum = MyEnum.A;
        Enum nullEnum = null;
        
        Foo(nativeEnum);
        Foo(classEnum);
        Foo(nullEnum); // NRE
        
        Bar(nativeEnum);
        //Bar<MyEnum>(classEnum); // does not compile
        //Bar<MyEnum>(nullEnum); // does not compile

        Baz(nativeEnum);
        Baz(classEnum);
        Baz(nullEnum); // NRE
    }       
}
```

https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAgrgOwD4AEBMAjALABQuAzAAQE2EDsVA3mVW1XJumFQLICeAUS5hW7FqXZSqAQTFsAvvKrLchAGxVgAe20AbKgDFdACmHcqMAJTKpAXgB8fISIB0M1wAlohvRADmJtYA3MqqGlq6BgBCEAgAPAAq5mAOJskiljaS0mwA7gAWiHBUGRYgVLAYAMYwSFQpKjm5js4p7l4+foEhZOGaOvpUsQBeSSlpZTzWtuyFxaWNFY2zbK0C7R7eUL4BQVahzWHNdJq4ACx8EACWmCbZUhK5bBuZmBAw1wBucI12bW4ZIdng1MgQ/gDuO5gc9Glw9HoIfC9DDcqsjKZ3p8fikDujjNoTOCRHiTlICSZkbjglQAPS0qgAOQASoJ0ejYghKR9vr8SajpPTOfFXtw0sTuAc6QyACbaOBQKiYbQwKjVbRgAAO1z0cHRQriIpcYsp6AR1OlVDlCqVKrVGu1ur6ZPYo252L5koFUjdErApJBbqp/MtLLZJwUzzICiAA

Reading up on the initial implementation in #250 it seems the only/main reason this wasn't moved to a separate `EnumAssertions` class was the lack of enum constraints which we now have.

It has nice potential.
E.g.
```c#
`MyEnum.A.Should().Be(OtherEnum.A)`
`MyEnum.A.Should().Be((object)42)`
```
will simply be turned into compile time errors.

NN---:
Will it be merged? Why the PR is still draft.

PR #1318, opened 2020-04-26
Add NotContainEquivalentOf

### LINKED ISSUE #1287: Add `NotContainEquivalentOf`
### Description

Add support for `CollectionAssertions.NotContainEquivalentOf`.
Support for `CollectionAssertions.ContainEquivalentOf` was added in #950.

Would let you write

```csharp
string[] input = new[] { "foo", "FOO" };

input.Should().NotContainEquivalentOf("foo");
```

vanashimko:
Fixes #1287.

## IMPORTANT 

* [x] If the contribution affects [the documentation](https://github.com/fluentassertions/fluentassertions/tree/master/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com).

### DIFF
--- a/Src/FluentAssertions/Collections/CollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/CollectionAssertions.cs
@@ -640,6 +640,117 @@ namespace FluentAssertions.Collections
         }
 
         /// <summary>
+        /// Asserts that a collection of objects does not contain any object equivalent to another object.
+        /// </summary>
+        /// <remarks>
+        /// Objects within the collection are equivalent to the expected object when both object graphs have equally named properties with the same
+        /// value, irrespective of the type of those objects. Two properties are also equal if one type can be converted to another
+        /// and the result is equal.
+        /// Notice that actual behavior is determined by the global defaults managed by <see cref="AssertionOptions"/>.
+        /// </remarks>
+        /// <param name="because">
+        /// An optional formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the
+        /// assertion is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> NotContainEquivalentOf<TExpectation>(TExpectation unexpected, string because = "",
+            params object[] becauseArgs)
+        {
+            return NotContainEquivalentOf(unexpected, config => config, because, becauseArgs);
+        }
+
+        /// <summary>
+        /// Asserts that a collection of objects does not contain any object equivalent to another object.
+        /// </summary>
+        /// <remarks>
+        /// Objects within the collection are equivalent to the expected object when both object graphs have equally named properties with the same
+        /// value, irrespective of the type of those objects. Two properties are also equal if one type can be converted to another
+        /// and the result is equal.
+        /// Notice that actual behavior is determined by the global defaults managed by <see cref="AssertionOptions"/>.
+        /// </remarks>
+        /// <param name="config">
+        /// A reference to the <see cref="EquivalencyAssertionOptions{TSubject}"/> configuration object that can be used
+        /// to influence the way the object graphs are compared. You can also provide an alternative instance of the
+        /// <see cref="EquivalencyAssertionOptions{TSubject}"/> class. The global defaults are determined by the
+        /// <see cref="AssertionOptions"/> class.
+        /// </param>
+        /// <param name="because">
+        /// An optional formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the
+        /// assertion is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> NotContainEquivalentOf<TExpectation>(TExpectation unexpected, Func<EquivalencyAssertionOptions<TExpectation>,
+            EquivalencyAssertionOptions<TExpectation>> config, string because = "", params object[] becauseArgs)
+        {
+            Guard.ThrowIfArgumentIsNull(config, nameof(config));
+
+            Execute.Assertion
+                .BecauseOf(because, becauseArgs)
+                .ForCondition(Subject != null)
+                .FailWith("Expected {context:collection} not to contain equivalent of {0}{reason}, but collection is <null>.", unexpected);
+
+            EquivalencyAssertionOptions<TExpectation> options = config(AssertionOptions.CloneDefaults<TExpectation>());
+
+            var foundIndices = new List<int>();
+            using (var scope = new AssertionScope())
+            {
+                int index = 0;
+                foreach (object actualItem in Subject)
+                {
+                    var context = new EquivalencyValidationContext
+                    {
+                        Subject = actualItem,
+                        Expectation = unexpected,
+                        CompileTimeType = typeof(TExpectation),
+                        Because = because,
+                        BecauseArgs = becauseArgs,
+                        Tracer = options.TraceWriter,
+                    };
+
+                    var equivalencyValidator = new EquivalencyValidator(options);
+                    equivalencyValidator.AssertEquality(context);
+
+                    string[] failures = scope.Discard();
+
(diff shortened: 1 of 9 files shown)


docs/_pages/collections.md
```diff
@@ -63,6 +63,8 @@ collection.Should().NotContain(x => x > 10);
 
 object boxedValue = 2;
 collection.Should().ContainEquivalentOf(boxedValue); // Compared by object equivalence
+object unexpectedBoxedValue = 82;
+collection.Should().ContainEquivalentOf(unexpectedBoxedValue); // Compared by object equivalence
```
jnyrup, line 67:
```suggestion
collection.Should().NotContainEquivalentOf(unexpectedBoxedValue); // Compared by object equivalence
```

Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -639,6 +639,117 @@ public AndConstraint<TAssertions> BeEquivalentTo(IEnumerable expectation, string
...
+
+            if (foundIndices.Count > 0)
+            {
+                using (new AssertionScope())
+                {
+                    Execute.Assertion
+                        .BecauseOf(because, becauseArgs)
+                        .WithExpectation("Expected {context:collection} {0} not to contain equivalent of {1}{reason}, ", Subject, unexpected)
+                        .AddReportable("configuration", options.ToString());
+                    if (foundIndices.Count == 1)
```
jnyrup, line 737: nit: add an empty line after multi line statements.

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.cs
```diff
@@ -1893,6 +1893,117 @@ public void When_collection_contains_object_equivalent_of_boxed_object_it_should
...
+            // Act
+            Action act = () => collection.Should().NotContainEquivalentOf(item, config: null);
+
+            // Assert
+            act.Should().ThrowExactly<ArgumentNullException>()
+                .Which.ParamName.Should().Be("config");
+        }
+
+        [Fact]
+        public void When_asserting_empty_collection_to_not_to_contain_equivalent_it_should_succeed()
```
jnyrup, line 1958:
```suggestion
        public void When_asserting_empty_collection_to_not_contain_equivalent_it_should_succeed()
```

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.cs
```diff
@@ -1893,6 +1893,117 @@ public void When_collection_contains_object_equivalent_of_boxed_object_it_should
...
+            // Arrange
+            IEnumerable collection = new[] { 1, 2, 3 };
+            int item = 4;
+
+            // Act / Assert
+            collection.Should().NotContainEquivalentOf(item);
+        }
+
+        [Fact]
+        public void When_asserting_collection_to_not_to_contain_equivalent_it_should_respect_config()
```
jnyrup, line 1980:
```suggestion
        public void When_asserting_collection_to_not_contain_equivalent_it_should_respect_config()
```

vanashimko:
> Nice work 👍
> Could I ask you to add a line to https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/releases.md as well?

Thanks @jnyrup! Should I put it into 6.0.0 section?

jnyrup:
> Thanks @jnyrup! Should I put it into 6.0.0 section?
Yes, the `develop` branch is what will eventually become 6.0.0.

vanashimko:
@jnyrup sorry, I've accidentally pushed "trim" of release.md. Let me fix this, don't like unnecessary changes.

vanashimko:
@jnyrup thanks for the review! I've reverted trim, it should be good now.

Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -639,6 +639,118 @@ public AndConstraint<TAssertions> BeEquivalentTo(IEnumerable expectation, string
...
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> NotContainEquivalentOf<TExpectation>(TExpectation unexpected, Func<EquivalencyAssertionOptions<TExpectation>,
+            EquivalencyAssertionOptions<TExpectation>> config, string because = "", params object[] becauseArgs)
+        {
+            Guard.ThrowIfArgumentIsNull(config, nameof(config));
+
+            Execute.Assertion
+                .BecauseOf(because, becauseArgs)
+                .WithExpectation("Expected {context:collection} not to contain equivalent of {0}{reason}, ", unexpected)
```
jnyrup, line 693: When using `WithExpectation` it should be cleared afterwards using 
```c#
.FailWith(...)
.Then
.ClearExpectation();
```

Otherwise if using

```c#
using (new AssertionScope())
{
    collection.Should().NotContainEquivalentOf(another); // assume this fails
    collection.Should().HaveCount(4); // <-- this will contain the WithExpectation from NotContainEquivalentOf
}
```

vanashimko: Since I have only one `FailWith` here, I'd better insert it just there directly. Is this also applied to the second place I use `WithExpectation` under `AssertionScope`? Or under assertion scope `ClearExpectation` is not needed?

vanashimko: (I assume that under `new AssertionScope()` it is not needed, since it is a brand new scope, but asking to make sure)

Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -639,6 +639,117 @@ public AndConstraint<TAssertions> BeEquivalentTo(IEnumerable expectation, string
...
+                        foundIndices.Add(index);
+                    }
+
+                    index++;
+                }
+            }
+
+            if (foundIndices.Count > 0)
+            {
+                using (new AssertionScope())
```
jnyrup, line 730: I think this could be written as:
```c#
if (foundIndices.Count > 0)
{
    Assertion assertion = Execute.Assertion
        .WithExpectation("Expected {context:collection} {0} not to contain equivalent of {1}{reason}, ", Subject, unexpected)
        .AddReportable("configuration", options.ToString());

    if (foundIndices.Count == 1)
    {
        assertion
            .FailWith("but found one at index {0}.", foundIndices[0])
            .Then
            .ClearExpectation();
    }
...
}
```

Src/FluentAssertions/Collections/CollectionAssertions.cs
jnyrup: I haven't tried it out, but I think creating a new `AssertionScope` might not produce the intended failure message when combing multiple failures.
If my example on how to rewrite it works, I would just go with that.

Src/FluentAssertions/Collections/CollectionAssertions.cs
vanashimko: I tried similar approach earlier and it did not work unfortunately. 

Firstly I was surprised that `AddReportable` is a `void` method, however I've extracted return value from `.WithExpectation` and just called `AddReportable` on it (that's not an issue of course).

The problem I faced is that `AddReportable` does not work when called outside `new AssertionScope()`. I'm not familiar with the codebase much, but quick look brought me to conclusion that "With" stuff is only attached on `Dispose` which seems to be not called on regular non-lazy assertion scope.

Am I missing something? What do you think?

jnyrup: I haven't used `AddReportable` much, so @dennisdoomen might be better to answer this than me.

vanashimko: Ok thanks.

BTW I added a test with using `NotContainEquivalentOf`  inside `AssertionScope`: https://github.com/fluentassertions/fluentassertions/pull/1318/commits/05b5f662771cf15fd1817b26268b74e81fba76d6. I did it in a separate commit, because I don't know how this aligns with your testing approach. Looks like it works fine (if I understand "expected" correctly).

dennisdoomen: `AddReportable` was indeed only used by `EquivalencyValidator` to initialize an `AssertionScope` within which the entire equivalency engine runs.

jnyrup:
@vanashimko I saw you added a test to verify using `new AssertionScope()`.
Unless you have anything else, I'm fine merging this.

vanashimko:
@jnyrup I'm good

PR #1320, opened 2020-04-30
Add StyleCop and CSharpGuideline analyzers

jnyrup:
This PR adds analyzers to help enforcing the style guidelines up front.
This hopefully helps PR reviewers to focus on content instead of style.

The `JetBrainsAnnotations.cs` produced a gazilion style warnings, so I've replaced it with [JetBrains.Annotations](https://www.nuget.org/packages/JetBrains.Annotations) instead.

I changed the way we exclude files from preprocessor directives in the class to conditional removal in the csproj file.
This helped the analyzers with files containing preprocessor directives and regions.

I removed `EventRaisingExtensions`, `IEventRecorder` and `RecordedEvent` from the netstandard2.0 target.
They are (as I see it) only relevant for `EventAssertions` which was already excluded for netstandard2.0.

I only included the analyzers for netcoreapp3.0 to keep the compilation time somewhat down.

It is fine balance when to obey the analyzer and when to silence it.
I hope I've found a reasonable balance.


Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -118,7 +118,9 @@ internal static IEnumerable<TAttribute> GetMatchingOrInheritedAttributes<TAttrib
             return GetCustomAttributes(type, isMatchingAttributePredicate, inherit: true);
         }
 
+#pragma warning disable AV1564 // Parameter in public or internal member is of type bool or bool?
```
lg2de, line 121: In general I would recommend to always comment the REASON for disabling a rule.
Especially here I would like to know the reason

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
```diff
@@ -29,13 +29,21 @@ public abstract class SelfReferenceEquivalencyAssertionOptions<TSelf> : IEquival
...
+
+        private readonly List<Type> referenceTypes = new List<Type>();
+
+        private readonly List<Type> valueTypes = new List<Type>();
+
+        private readonly Func<Type, EqualityStrategy> getDefaultEqualityStrategy = null;
 
         [DebuggerBrowsable(DebuggerBrowsableState.Never)]
-#pragma warning disable CA1051 // TODO: fix in 6.0
+#pragma warning disable CA1051, SA1401 // TODO: fix in 6.0
```
lg2de, line 41: Would you like to add this in the issue collecting the 6.0 tasks?

Src/FluentAssertions/NumericAssertionsExtensions.cs
```diff
@@ -893,6 +895,7 @@ public static class NumericAssertionsExtensions
             params object[] becauseArgs)
         {
             Execute.Assertion
+
```
lg2de, line 898: An accidental blank line?

Src/FluentAssertions/ObjectAssertionsExtensions.cs
```diff
@@ -9,6 +8,7 @@
 using FluentAssertions.Execution;
 using FluentAssertions.Primitives;
 
+#pragma warning disable CA1062 // Validate arguments of public methods
```
lg2de, line 11: Why?

lg2de:
I really like this PR. I'm a fan of such tools helping consistent style.

But I've a few questions:
1. Did you though about start using SonarQube? It's available for free for open source projects on sonarcloud.io.
2. SonarQube comes with SonarAnalyzer which help for local analysis. For me with SonarQube/Analyzer there is no need to use StyleCop, FxCop, etc.
3. You disabled several rules. I would like to know how did you decide whether keep enabled or to disable a rule? I think at least all default rules should be enabled.

dennisdoomen:
Hmm, we used to have this in the past, but it was such an annoyance for contributors. How is this different?

lg2de:
> but it was such an annoyance for contributors

What you mean?

> How is this different?

It is difficult to give such a general answer.
I guess it has the same target: improve code quality by checking "many" rules.
For me, it is one tool for multiple checks: style, possible bugs, code smells, security checks, etc.
And you have a trend analysis.

dennisdoomen:
I think this attempt is different as in that you get immediate feedback. Previously, we used a build step that invoked StyleCop explicitly. It was also very slow.

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
jnyrup: I've added it to the preparation issue, thanks.

Src/FluentAssertions/NumericAssertionsExtensions.cs
jnyrup: fixed

Src/FluentAssertions/ObjectAssertionsExtensions.cs
jnyrup: I was unsure how to handle CA1062 warnings.
CA1062 should be at least treated as warning, as methods of public methods *should* be validated, but it needs to be fixed in separate commit/PR as it currently produces 350+ warnings.
So I've made CA1062 a suggestion instead and removed this `#pragma`

jnyrup:
> You disabled several rules. I would like to know how did you decide whether keep enabled or to disable a rule? I think at least all default rules should be enabled.

@lg2de I started out with having all rules enabled and then went over them one by one.
Some where in direct contrasts to our existing code, e.g. `SA1101: Prefix local calls with this`. Those I sat to `severity = none`.

Some rules we're rightfully triggered, but triggered so many times I think they would overshadow more import ones. E.g. warnings about missing xml-summary for generic types.
Those I sat to `severity = suggestion`.

We're probably going to tweak this more, i.e. loosen or tighten some rules, but I wanted a start where one can focus on errors an warnings and optionally look at suggestions for the files they're working in.

> @jnyrup any change you can split the work in separate commits? Maybe my [recent blog post](https://www.continuousimprover.com/2020/03/keep-source-control-history-clean.html) can help here.

@dennisdoomen here you go.

PR #1322, opened 2020-05-02
Fix all build warnings

jnyrup:
This fixes the remaining build warnings not fixed in #1320 .


Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -5,6 +5,7 @@
     <AssemblyOriginatorKeyFile>FluentAssertions.snk</AssemblyOriginatorKeyFile>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <NoWarn>1591;1573</NoWarn>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
lg2de, line 8: I would recommend to set this property only in the build.
Otherwise it is very complicated to get testing versions running while development.

jnyrup: Makes sense 👍 
Would restricting it to `Release` do it?

dennisdoomen: No, because then it will only pop-up in your PR. And that's a situation we had with StyleCop and was the main reason I moved away from it.

jnyrup: I see both points.
I suggest keeping the PR as it is for now and evaluate how it turns out.

PR #1324, opened 2020-05-08
Keep synchronization context of executing thread when asserting asynchronous operations

lg2de:
As [discussed on slack](https://fluentassertions.slack.com/archives/C039Y2H7U/p1588259915012400) `SynchronizationContext.Current` should not be changed by FluentAssertions.
In addition synchronous and asynchronous operations should not be mixed.

Please note the following articles:
* https://docs.microsoft.com/de-de/dotnet/csharp/programming-guide/concepts/async/
* https://docs.microsoft.com/de-de/dotnet/csharp/async
* https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
* https://medium.com/rubrikkgroup/understanding-async-avoiding-deadlocks-e41f8f2c6f5d


lg2de:
Hey @dennisdoomen and @jnyrup , the PR is ready for review.
The failed test is IMHO not related to my changed, but a common timing issue.

Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
```diff
@@ -23,65 +25,65 @@ public void When_getting_the_subject_it_should_remain_unchanged()
...
 
             // Assert
-            testAction.Should().Throw<XunitException>()
-                .WithMessage("*because we want to test the failure message*found <null>*");
+            await act2.Should().ThrowAsync<XunitException>();
         }
 
-        [Fact]
-        public async Task When_async_method_throws_an_empty_AggregateException_it_should_fail()
+        [UIFact]
```
dennisdoomen, line 85: So why do we still need this?

lg2de: These tests a duplicated tests. There are always two tests with the **same test content** but **different thread context**.

dennisdoomen: But why do you need this now?

lg2de: To ensure that the library is working fine with and without "valid" synchronization context.

Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
```diff
@@ -874,16 +788,16 @@ public void When_async_method_throws_the_expected_inner_exception_it_should_succ
             };
 
             // Act
-            Action action = () => task
-                .Should().Throw<AggregateException>()
-                .WithInnerException<InvalidOperationException>();
+            Func<Task> action = () => task
+                .Should().ThrowAsync<AggregateException>()
+                .WithInnerException<AggregateException, InvalidOperationException>(); // TODO How to remove duplicate generic argument?
```
jnyrup, line 793: I think that would require some sort of [partial type inference](https://github.com/dotnet/csharplang/issues/1349)

Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
```diff
@@ -612,259 +581,204 @@ public async Task When_subject_throws_expected_async_exact_exception_it_should_s
...
+                .Should().NotThrowAsync<ArgumentException>();
 
             // Assert
-            action.Should().Throw<XunitException>()
+            await action.Should().ThrowAsync<XunitException>()
                 .WithMessage("Did not expect System.ArgumentException, but found System.ArgumentException*");
         }
 
         [Fact]
-        public void When_ValueTask_async_method_of_T_throws_exception_expected_not_to_be_thrown_it_should_fail()
```
jnyrup, line 845: Not sure why this test was deleted?

Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
jnyrup: Can a test be annotated with both `[Fact]` and `[UIFact]` to avoid duplicating the content?
I have no strong opinions on which path we take, so just asking for now.

lg2de: No, at least not with the current version of StaFact:

```
1>...\AsyncFunctionExceptionAssertionSpecs.cs(86,27,86,107): error xUnit1013: Public method 'When_async_method_throws_an_empty_AggregateException_on_UI_thread_it_should_fail' on test class 'AsyncFunctionExceptionAssertionSpecs' should be marked as a Fact.
1>...\AsyncFunctionExceptionAssertionSpecs.cs(74,27,74,94): error xUnit1002: Test methods cannot have multiple Fact or Theory attributes
```

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -84,7 +47,7 @@ protected override void InvokeSubject()
                 .FailWith("Expected {context:task} to complete within {0}{reason}, but found <null>.", timeSpan);
 
             using var timeoutCancellationTokenSource = new CancellationTokenSource();
-            TTask task = Subject.ExecuteInDefaultSynchronizationContext();
+            TTask task = Subject.Invoke();
```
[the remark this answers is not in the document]
lg2de: Here I think, the text "Invoke" helps to understand that the action (behind the subject) is invoked.
I would like to keep it, or even more to use this more consequent.
What do you think?

Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
```diff
@@ -1,12 +1,14 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Threading.Tasks;
 using FluentAssertions.Extensions;
 using Xunit;
 using Xunit.Sdk;
 
-namespace FluentAssertions.Specs
+namespace FluentAssertions.Specs // TODO Move to FluentAssertions.Specs.Exceptions to follow file structure?
```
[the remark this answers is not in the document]
lg2de: Ok.
Can we keep the TODO comment?

Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
lg2de: I do not remember. :(
I've re-added and adopted to async API.

Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
lg2de: I've no idea...

Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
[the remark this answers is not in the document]
dennisdoomen: Yes, normally I use `REFACTOR` and `SMELL` for these kind of refactoring trails.

Src/FluentAssertions/Specialized/DelegateAssertionsBase.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions.Common;
+using FluentAssertions.Execution;
+using FluentAssertions.Primitives;
+
+namespace FluentAssertions.Specialized
+{
+    public abstract class DelegateAssertionsBase<TDelegate, TAssertions> : ReferenceTypeAssertions<TDelegate, DelegateAssertionsBase<TDelegate, TAssertions>>
```
dennisdoomen, line 10: Can you add some XML docs to explain the purpose of this class?

Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
lg2de: "partial type inference" seams to be a feature in (far) future. How can this help today?
I'll try again to find another solution. Do you have any suggestion?

jnyrup: My comment about partial type inference was to indicate that, until we might see that feature in the language, I can't think of a prettier way to implement that feature.
So if we can't make it better, there's no TODO and the note can be removed.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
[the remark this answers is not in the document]
jnyrup: While I don't have a strong preference in `()` vs `.Invoke()` (other than the former being shorter) I think it's easier to contribute to the library if we are consistent in our coding style.
@dennisdoomen any thoughts?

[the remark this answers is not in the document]
dennisdoomen: To be honest, it never even occurred to me to use `Invoke`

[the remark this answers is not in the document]
lg2de: @dennisdoomen , please just make a decision.
I've explained my thoughts.
If you do not like it, just tell me.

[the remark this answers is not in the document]
dennisdoomen: Let's keep it. It's not important enough to have a debate about.

Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
lg2de: removed

jnyrup:
> Nothing strikes out to me, but as mentioned in Slack discussions this is not my strongest topic.
> I assume we have the complementary async tests for the removed sync tests.
> 
> If I e.g. have existing code that for whatever reason mixes sync and async code, can this PR make it deadlock on a CI-build, where it would previously have passed?

@lg2de Do you have any comments on this case?
I'd like to avoid causing hanging tests due to bad, yet "working" user code.

lg2de:
> Do you have any comments on this case?
> I'd like to avoid causing hanging tests due to bad, yet "working" user code.

I thought, we had already discussed this.
Yes, it may happen that "wrong code" will run into deadlock.
But it is not the scope of FluentAssertion to work-around issues of user code.

dennisdoomen:
Shall we squash this PR or do you plan to clean-up the commits so we can keep some history?

lg2de:
All the recent cleanups I collected into single (force-pushed) "cleanup commit". In total we have 5 commits in this PR. I hope this is ok. It shows the development steps.

dennisdoomen:
Now I'm wondering whether you should ship this as an alpha version...

PR #1329, opened 2020-05-17
Add [Not]Be{In,Under}Namespace assertions to TypeSelectorAssertions

### LINKED ISSUE #1175: Add new namespace assertions to TypeSelectorAssertions

jnyrup:
closes #1180 
fixes #1175 

Credits to @kimsey0 for all the work.

### DIFF
--- a/Tests/FluentAssertions.Specs/Types/TypeAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Types/TypeAssertionSpecs.cs
@@ -1,6 +1,9 @@
 using System;
 using System.Linq;
 using System.Reflection;
+using DummyNamespace;
+using DummyNamespace.InnerDummyNamespace;
+using DummyNamespaceTwo;
 using FluentAssertions.Common;
 using FluentAssertions.Primitives;
 using FluentAssertions.Types;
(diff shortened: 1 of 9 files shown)


Tests/FluentAssertions.Specs/Types/TypeAssertionSpecs.cs
```diff
@@ -1540,6 +1543,390 @@ public void When_injecting_a_null_predicate_into_TypeSelector_NotBeDecoratedWith
 
         #endregion
 
+        #region BeInNamespace
+
+        [Fact]
+        public void When_asserting_a_selection_of_types_in_a_namespace_is_in_that_namespace_it_succeeds()
```
dennisdoomen, line 1549: I would rewrite those tests like:

`When_a_type_is_in_the_expected_namespace_it_should_not_throw`

Same with the others.

jnyrup: Fixed.

PR #1333, opened 2020-05-24
Improved the message when the PropertyChanged event was raised for the wrong property

### LINKED ISSUE #919: RaisePropertyChangeFor produces unhelpful failure message
### Description

RaisePropertyChangeFor produces unhelpful failure message.

```csharp
public class Bar
{
    public event PropertyChangedEventHandler PropertyChanged;

    public string Value { get; set; }

    public string OtherValue { get; set; }

    public void OnPropertyChanged([CallerMemberName] string propertyName = null)
    {
        this.PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
    }
}
```
```csharp
var bar = new Bar();
using (var monitor = bar.Monitor())
{
    bar.OnPropertyChanged(nameof(Bar.OtherValue));
    monitor.Should().RaisePropertyChangeFor(b => b.Value);
}
```

### Expected behavior:

Clear failure message including the property name of the PropertyChangedEvent that was expected, but not raised.
Maybe also a list of PropertyChangedEvents that were raised.

### Actual behavior:

Unreadable message without the property name:

Expected at least one event with arguments matching (args.PropertyName == value(FluentAssertions.Events.EventAssertions`1+<>c__DisplayClass5_0[MSTestFluentAssertions.UnitTest1+Bar]).propertyName), but found none.

### Versions

FluentAssertions 5.4.2.0
.NET Framework 4.6.1

dennisdoomen:
Fixes #919

### DIFF
--- a/Src/FluentAssertions/Events/EventAssertions.cs
+++ b/Src/FluentAssertions/Events/EventAssertions.cs
@@ -91,18 +91,31 @@ namespace FluentAssertions.Events
         public IEventRecording RaisePropertyChangeFor(Expression<Func<T, object>> propertyExpression,
             string because = "", params object[] becauseArgs)
         {
-            IEventRecording recording = monitor.GetRecordingFor(PropertyChangedEventName);
             string propertyName = propertyExpression?.GetPropertyInfo().Name;
+
+            IEventRecording recording = monitor.GetRecordingFor(PropertyChangedEventName);
 
             if (!recording.Any())
             {
                 Execute.Assertion
                     .BecauseOf(because, becauseArgs)
-                    .FailWith("Expected object {0} to raise event {1} for property {2}{reason}, but it did not.",
+                    .FailWith("Expected object {0} to raise event {1} for property {2}{reason}, but it did not raise that event at all.",
                         monitor.Subject, PropertyChangedEventName, propertyName);
             }
 
-            return recording.WithArgs<PropertyChangedEventArgs>(args => args.PropertyName == propertyName);
+            var actualPropertyNames = recording
+                .SelectMany(@event => @event.Parameters.OfType<PropertyChangedEventArgs>())
+                .Select(eventArgs => eventArgs.PropertyName)
+                .Distinct()
+                .ToArray();
+
+            Execute.Assertion
+                .ForCondition(actualPropertyNames.Contains(propertyName))
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected object {0} to raise event {1} for property {2}{reason}, but it was only raised for {3}.",
+                    monitor.Subject, PropertyChangedEventName, propertyName, actualPropertyNames);
+
+            return recording;
         }
 
         /// <summary>
--- a/Tests/FluentAssertions.Specs/Events/EventAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Events/EventAssertionSpecs.cs
@@ -491,7 +491,7 @@ namespace FluentAssertions.Specs.Events
             // Assert
             act.Should().Throw<XunitException>().WithMessage(
                 "Expected object " + Formatter.ToString(subject) +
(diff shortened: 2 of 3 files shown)


Src/FluentAssertions/Events/EventAssertions.cs
```diff
@@ -91,18 +91,31 @@ public void NotRaise(string eventName, string because = "", params object[] beca
...
-                    .FailWith("Expected object {0} to raise event {1} for property {2}{reason}, but it did not.",
+                    .FailWith("Expected object {0} to raise event {1} for property {2}{reason}, but it did not raise that event at all.",
                         monitor.Subject, PropertyChangedEventName, propertyName);
             }
 
-            return recording.WithArgs<PropertyChangedEventArgs>(args => args.PropertyName == propertyName);
+            var actualPropertyNames = recording
+                .SelectMany(@event => @event.Parameters.OfType<PropertyChangedEventArgs>())
+                .Select(@args => @args.PropertyName)
+                .Distinct()
```
jnyrup, line 109: Is there a tests that exercises the distinctiveness?

Src/FluentAssertions/Events/EventAssertions.cs
```diff
@@ -91,18 +91,31 @@ public void NotRaise(string eventName, string because = "", params object[] beca
...
                     .BecauseOf(because, becauseArgs)
-                    .FailWith("Expected object {0} to raise event {1} for property {2}{reason}, but it did not.",
+                    .FailWith("Expected object {0} to raise event {1} for property {2}{reason}, but it did not raise that event at all.",
                         monitor.Subject, PropertyChangedEventName, propertyName);
             }
 
-            return recording.WithArgs<PropertyChangedEventArgs>(args => args.PropertyName == propertyName);
+            var actualPropertyNames = recording
+                .SelectMany(@event => @event.Parameters.OfType<PropertyChangedEventArgs>())
+                .Select(@args => @args.PropertyName)
```
jnyrup, line 108:
```suggestion
                .Select(arg => arg.PropertyName)
```

Src/FluentAssertions/Events/EventAssertions.cs
```diff
@@ -91,18 +91,31 @@ public void NotRaise(string eventName, string because = "", params object[] beca
...
                     .BecauseOf(because, becauseArgs)
-                    .FailWith("Expected object {0} to raise event {1} for property {2}{reason}, but it did not.",
+                    .FailWith("Expected object {0} to raise event {1} for property {2}{reason}, but it did not raise that event at all.",
                         monitor.Subject, PropertyChangedEventName, propertyName);
             }
 
-            return recording.WithArgs<PropertyChangedEventArgs>(args => args.PropertyName == propertyName);
+            var actualPropertyNames = recording
+                .SelectMany(@event => @event.Parameters.OfType<PropertyChangedEventArgs>())
+                .Select(eventArgs => eventArgs.PropertyName)
```
jnyrup, line 108: Shouldn't it be singular?

Src/FluentAssertions/Events/EventAssertions.cs
jnyrup: I don’t think changed test exercises the call to `Distinct`. Such a test would need to raise the same event twice and check that the failure message only contains the name of the event once.

Src/FluentAssertions/Events/EventAssertions.cs
dennisdoomen: Nope. The object itself is named like that.

Src/FluentAssertions/Events/EventAssertions.cs
dennisdoomen: Adapted the test.

Src/FluentAssertions/Events/EventAssertions.cs
jnyrup: Aha, finally it makes sense to me

PR #1337, opened 2020-05-26
Allow using custom IOrderingRule

### LINKED ISSUE #1336: Allow to configure custom IOrderingRule
### Description

The [extension documentation](https://fluentassertions.com/extensibility/#building-your-own-extensions) claims that I can implement a custom `IOrderingRule` for `Should().BeEquivalentTo()`. But it's impossible to hook it into the the options. Neither for the global `AssertionOptions` nor explicitly for a particular invocation of `BeEquivalentTo()`.

```csharp
namespace XbrlPlus
{

    public class ReadOnlyListOrderingRule : IOrderingRule
    {
        public OrderStrictness Evaluate(IMemberInfo memberInfo) =>
            memberInfo.CompileTimeType.IsSameOrInherits(typeof(IReadOnlyList<>))
                ? OrderStrictness.Strict
                : OrderStrictness.Irrelevant;

        public override string ToString() => "Be strict about the order of items in read-only lists";
    }
}

AssertionOptions.AssertEquivalencyUsing(options => options.Using(new ReadOnlyListOrderingRule()));

object x = "";
x.Should().BeEquivalentTo("", options => options.Using(new ReadOnlyListOrderingRule()));
```

### Expected behavior:

Respect the custom rule.

### Actual behavior:

Compiler error.

### Versions

5.10.3
* Which .NET runtime and version are you targeting?
netstandard2.1

### Additional Information

philippdolder:
Fixes #1336 
## IMPORTANT 

* [ ] If the contribution affects [the documentation](https://github.com/fluentassertions/fluentassertions/tree/master/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com).

### DIFF
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
@@ -417,6 +417,14 @@ namespace FluentAssertions.Equivalency
         }
 
         /// <summary>
+        /// Adds an ordering rule to the ones already added by default, and which is evaluated after all existing rules.
+        /// </summary>
+        public TSelf Using(IOrderingRule orderingRule)
+        {
+            return AddOrderingRule(orderingRule);
+        }
+
+        /// <summary>
         /// Adds an equivalency step rule to the ones already added by default, and which is evaluated before previous
         /// user-registered steps
         /// </summary>
@@ -705,6 +713,12 @@ namespace FluentAssertions.Equivalency
         private TSelf AddMatchingRule(IMemberMatchingRule matchingRule)
         {
             matchingRules.Insert(0, matchingRule);
+            return (TSelf)this;
+        }
+
+        private TSelf AddOrderingRule(IOrderingRule orderingRule)
+        {
+            orderingRules.Add(orderingRule);
             return (TSelf)this;
         }
 
--- a/Tests/FluentAssertions.Specs/Equivalency/ExtensibilityRelatedEquivalencySpecs.cs
+++ b/Tests/FluentAssertions.Specs/Equivalency/ExtensibilityRelatedEquivalencySpecs.cs
@@ -154,6 +154,50 @@ namespace FluentAssertions.Specs
 
         #endregion
 
+        #region Ordering Rules
+
+        [Fact]
+        public void When_an_ordering_rule_is_added_it_should_be_evaluated_after_all_existing_rules()
+        {
+            // Arrange
+            var subject = new[] { "First", "Second" };
+            var expected = new[] { "First", "Second" };
+
+            // Act
+            Action act = () => subject.Should().BeEquivalentTo(
+                expected,
+                options => options.Using(new StrictOrderingRule()));
+
+            // Assert
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_an_ordering_rule_is_added_it_should_appear_in_the_exception_message()
+        {
+            // Arrange
+            var subject = new[] { "First", "Second" };
+            var expected = new[] { "Second", "First" };
+
+            // Act
+            Action act = () => subject.Should().BeEquivalentTo(
+                expected,
+                options => options.Using(new StrictOrderingRule()));
+
+            act.Should().Throw<XunitException>()
+                .WithMessage(string.Format("*{0}*", typeof(StrictOrderingRule).Name));
+        }
+
+        internal class StrictOrderingRule : IOrderingRule
+        {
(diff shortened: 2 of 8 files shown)


Tests/FluentAssertions.Specs/Equivalency/ExtensibilityRelatedEquivalencySpecs.cs
```diff
@@ -154,6 +154,50 @@ public SelectedMemberInfo Match(SelectedMemberInfo expectedMember, object subjec
...
+            var subject = new[] { "First", "Second" };
+            var expected = new[] { "Second", "First" };
+
+            // Act
+            Action act = () => subject.Should().BeEquivalentTo(
+                expected,
+                options => options.Using(new StrictOrderingRule()));
+
+            act.Should().Throw<XunitException>()
+                .WithMessage(string.Format("*{0}*", typeof(StrictOrderingRule).Name));
```
dennisdoomen, line 188: Why don't you use $"*{typeof(StrictOrderingRule).Name}*"?

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
```diff
@@ -416,6 +416,14 @@ public TSelf Using(IMemberMatchingRule matchingRule)
             return AddMatchingRule(matchingRule);
         }
 
+        /// <summary>
+        /// Adds an ordering rule to the ones already added by default, and which is evaluated after all existing rules.
+        /// </summary>
+        public TSelf Using(IOrderingRule orderingRule)
```
dennisdoomen, line 422: Should a custom ordering rule not override the other ordering rules?

jnyrup: That would be inconsistent with
```cs
public TSelf WithStrictOrderingFor(Expression<Func<IMemberInfo, bool>> predicate)
{
    orderingRules.Add(new PredicateBasedOrderingRule(predicate));
    return (TSelf)this;
}
```

Tests/FluentAssertions.Specs/Equivalency/ExtensibilityRelatedEquivalencySpecs.cs
philippdolder: I sticked to the way it's been done in the other tests in the same file for consistency. But I'd happily adjust it, maybe for all places in this file for consistency?
I was also thinking to use `nameof`, though I'm not sure this works with all supported frameworks.

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
philippdolder: and with a custom ordering rule we also wouldn't want to get rid of the strict byte array ordering rule.
I can also imagine having more than 1 ordering rule. e.g. for specific collection types you want to enforce strict ordering, and you could solve that with a rule per collection type

dennisdoomen: Ah yeah, you're right. If there's a rule following `ByteArrayOrderingRule` that returns `OrderStrictness.NotStrict`, it will overrule the byte array ordering rule.

Tests/FluentAssertions.Specs/Equivalency/ExtensibilityRelatedEquivalencySpecs.cs
dennisdoomen: > Maybe for all places in this file for consistency?

Only if you feel like it.

> I was also thinking to use nameof, though I'm not sure this works with all supported frameworks.

If it compiles, it most likely runs.

jnyrup: `nameof` expressions is a C# 6 language construct independent of the runtime.
Once compiled it's just a string.

jnyrup:
Please add a line to `docs/_pages/releases.md` about the addition.
I'll add it to `CONTRIBUTING.md` immediately.

Tests/FluentAssertions.Specs/Equivalency/ExtensibilityRelatedEquivalencySpecs.cs
philippdolder: done consistently in the whole file

dennisdoomen:
Thanks for this contribution @philippdolder

Issue #318, opened 2015-12-07
support for NotContainInOrder


SurajGupta:
I would be great to get an generic collection assertion `NotContainInOrder()`

dennisdoomen:
Sure. Would you be willing to send me a PR?

blairconrad:
I wasn't sure what `NotContainInOrder` meant, so I went searching the web to find out if it's a thing that I just hadn't heard of. I found

StackOverflow question [How to assert two lists are not equivalent if in different order using Fluent Assertions](http://stackoverflow.com/questions/29281224/how-to-assert-two-lists-are-not-equivalent-if-in-different-order-using-fluent-as), which used the term to mean a test that would pass if

1. two collections contained the same elements, but
2. the two collections' orders were not the same

so

|collection 1|collection 2|result|
|-------------|--------------|------|
|1, 2, 3|1, 4, 3, 2|fail|
|1, 2, 3|3, 1, 2| pass|
|1, 2, 3|1, 2, 3|fail|

Essentially, it would be the same as checking

if the lists are equivalent, without regard to order, and if the lists are NOT equivalent when considered with strict ordering.

Am I on the right track?

dennisdoomen:
SurajGupta, can you confirm?

blairconrad:
However @SurajGupta replies, I'd suggest a different name for the constraint. I'm a native speaker of English, and "not contain in order" does not mean anything to me. Every time I tried to use such a method, I'd have to refer to the documentation.

dennisdoomen:
You can't say *this collection should not contain the other collection in order*?

blairconrad:
Oh! I think you're understanding the intent better than I am. Let's augment my table from above, with my understanding.

|actual|expected|result|
|-------------|--------------|------|
|1, 2, 3|1, 4, 3, 2|fail|
|1, 2, 3|3, 1, 2| pass|
|1, 2, 3|1, 2, 3|fail|
|1, 2, 3, 4, 5|2, 3, 4| ??? |

I had thought that last entry would be a "pass" because I thought @SurajGupta wanted to reject only expected values that were exactly the same as actual, in order. But you are saying it should be "fail" because actual _contained expected's elements in order_.

If that's correct, then `NotContainInOrder` is not so bad. I still would not figure it out without docs, but it might stick in my head after I read them.

I'd consider `NotContainSequence`, as "sequence" suggest to me an ordered collection with no gaps. But I wouldn't fight very strongly for that as a new name.

[...]

PR #1339, opened 2020-05-27
Add NotContainInOrder

rynkevich:
This PR adds `NotContainInOrder` assertion as requested in #318.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Collections/CollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/CollectionAssertions.cs
@@ -978,7 +978,7 @@ namespace FluentAssertions.Collections
             }
 
             var actualItemsSkipped = 0;
-            for (int index = 0; index < unexpectedItems.Count && actualItems.Any(); index++)
+            for (int index = 0; index < unexpectedItems.Count; index++)
             {
                 object unexpectedItem = unexpectedItems[index];
 
@@ -1001,6 +1001,10 @@ namespace FluentAssertions.Collections
                     }
 
                     actualItems = actualItems.Skip(1).ToArray();
+                }
+                else
+                {
+                    return new AndConstraint<TAssertions>((TAssertions)this);
                 }
             }
 
--- a/Src/FluentAssertions/Collections/CollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/CollectionAssertions.cs
@@ -934,6 +934,80 @@ namespace FluentAssertions.Collections
         }
 
         /// <summary>
+        /// Asserts the current collection does not contain the specified elements in the exact same order, not necessarily consecutive.
+        /// </summary>
+        /// <remarks>
+        /// Elements are compared using their <see cref="object.Equals(object)" /> implementation.
+        /// </remarks>
+        /// <param name="unexpected">A <see cref="System.Array"/> with the unexpected elements.</param>
+        public AndConstraint<TAssertions> NotContainInOrder(params object[] unexpected)
+        {
+            return NotContainInOrder(unexpected, string.Empty);
+        }
+
+        /// <summary>
+        /// Asserts the current collection does not contain the specified elements in the exact same order, not necessarily consecutive.
+        /// </summary>
+        /// <remarks>
+        /// Elements are compared using their <see cref="object.Equals(object)" /> implementation.
+        /// </remarks>
+        /// <param name="unexpected">An <see cref="IEnumerable"/> with the unexpected elements.</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> NotContainInOrder(IEnumerable unexpected, string because = "",
+            params object[] becauseArgs)
+        {
+            Guard.ThrowIfArgumentIsNull(unexpected, nameof(unexpected), "Cannot verify absence of ordered containment against a <null> collection.");
+
+            if (ReferenceEquals(Subject, null))
+            {
+                return new AndConstraint<TAssertions>((TAssertions)this);
+            }
+
+            IList<object> unexpectedItems = unexpected.ConvertOrCastToList<object>();
+            IList<object> actualItems = Subject.ConvertOrCastToList<object>();
+
+            if (unexpectedItems.Count > actualItems.Count)
+            {
+                return new AndConstraint<TAssertions>((TAssertions)this);
+            }
+
+            var actualItemsSkipped = 0;
+            for (int index = 0; index < unexpectedItems.Count && actualItems.Any(); index++)
+            {
+                object unexpectedItem = unexpectedItems[index];
+
+                actualItems = actualItems.SkipWhile(actualItem =>
+                {
+                    actualItemsSkipped++;
+                    return !actualItem.IsSameOrEqualTo(unexpectedItem);
+                }).ToArray();
+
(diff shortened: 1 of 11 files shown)


Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -933,6 +933,80 @@ public AndConstraint<TAssertions> ContainInOrder(params object[] expected)
...
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> NotContainInOrder(IEnumerable unexpected, string because = "",
+            params object[] becauseArgs)
+        {
+            Guard.ThrowIfArgumentIsNull(unexpected, nameof(unexpected), "Cannot verify absence of ordered containment against a <null> collection.");
+
+            if (ReferenceEquals(Subject, null))
```
dennisdoomen, line 967: Not sure what should happen here. @jnyrup do you remember if we ever agreed on some basic behavior on this?

Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -933,6 +933,80 @@ public AndConstraint<TAssertions> ContainInOrder(params object[] expected)
...
+
+            var actualItemsSkipped = 0;
+            for (int index = 0; index < unexpectedItems.Count && actualItems.Any(); index++)
+            {
+                object unexpectedItem = unexpectedItems[index];
+
+                actualItems = actualItems.SkipWhile(actualItem =>
+                {
+                    actualItemsSkipped++;
+                    return !actualItem.IsSameOrEqualTo(unexpectedItem);
```
dennisdoomen, line 988: You're saying you're using `object/Equals` in the docs, but this method will first try to do a reference equality check, before considering the `Equals` implementation of the involved object.

Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -933,6 +933,80 @@ public AndConstraint<TAssertions> ContainInOrder(params object[] expected)
...
+
+            if (ReferenceEquals(Subject, null))
+            {
+                return new AndConstraint<TAssertions>((TAssertions)this);
+            }
+
+            IList<object> unexpectedItems = unexpected.ConvertOrCastToList<object>();
+            IList<object> actualItems = Subject.ConvertOrCastToList<object>();
+
+            if (unexpectedItems.Count > actualItems.Count)
```
dennisdoomen, line 975: What if the collection of unexpected items is smaller?

Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -933,6 +933,80 @@ public AndConstraint<TAssertions> ContainInOrder(params object[] expected)
...
+            {
+                return new AndConstraint<TAssertions>((TAssertions)this);
+            }
+
+            var actualItemsSkipped = 0;
+            for (int index = 0; index < unexpectedItems.Count && actualItems.Any(); index++)
```
dennisdoomen, line 981: I find it quite hard to read the algorithm.

jnyrup:
I'm occupied all weekend, but I'll try to have a look at this Monday or Tuesday

Src/FluentAssertions/Collections/CollectionAssertions.cs
rynkevich: Then there's a chance that actual collection contains unexpected sequence and we should check it.

Src/FluentAssertions/Collections/CollectionAssertions.cs
rynkevich: I can replace this line with `else` clause inside the loop. That's how I initially wrote it, but decided to make that a little bit more compact.

Src/FluentAssertions/Collections/CollectionAssertions.cs
rynkevich: Yes, and the same goes for `ContainInOrder`. Should I mention this behavior in docs?

P.S. I have been noticing various inaccuracies in documentation of different assertions while writing the code for the latest PRs. Probably, he documentation verification and adjustment could be planned for the future versions of FluentAssertions?

Src/FluentAssertions/Collections/CollectionAssertions.cs
dennisdoomen: OK, most other methods will throw an assertion exception when the subject is `null`. So I think we should do the same. BTW, you can replace the use of `ReferenceEquals` with `Subject is null`

Src/FluentAssertions/Collections/CollectionAssertions.cs
dennisdoomen: > Yes, and the same goes for ContainInOrder. Should I mention this behavior in docs?

No, it's totally fine. In the end, the result is the same, because if both the subject and the expectation are the _same_ object, then it doesn't matter if their `Equals` override does something non-standard.

Src/FluentAssertions/Collections/CollectionAssertions.cs
dennisdoomen: Agree with that. But maybe we should also fail the assertion if the parameter contains more items than the subject. I would assume the developer was expecting at least that many items.

rynkevich: Sorry, maybe I got your idea wrong, but isn't is just what happens here?

The parameter collection (`unexpected`) and, probably, the subject collection are enumerables. When we convert them to lists we can get their size without the need in iterating and counting elements (as they are already counted). So if the number of `unexpected` items is greater than the number of `Subject` items, we can fail the assertion.

dennisdoomen: It must not be my day. You're totally right of course....

Src/FluentAssertions/Collections/CollectionAssertions.cs
dennisdoomen: You replaced the comparison, but you're still not failing the assertion.

rynkevich: Fixed

Src/FluentAssertions/Collections/CollectionAssertions.cs
```diff
@@ -933,6 +933,84 @@ public AndConstraint<TAssertions> ContainInOrder(params object[] expected)
...
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> NotContainInOrder(IEnumerable unexpected, string because = "",
+            params object[] becauseArgs)
+        {
+            Guard.ThrowIfArgumentIsNull(unexpected, nameof(unexpected), "Cannot verify absence of ordered containment against a <null> collection.");
+
+            Execute.Assertion
```
dennisdoomen, line 967: If somebody uses an `AssertionScope`, `FailWith` will not throw immediately and return a `bool`

rynkevich: Does that mean we should also return from the method if subject is `null`?

dennisdoomen: That has nothing to do with my comment. What I meant is that you need to avoid executing the remainder of that method if the first `FailWith` returns  `false`. 

If the subject itself is `null`, it should also fail the assertion.

rynkevich: Pushed changes for this case, please check if that does what you expect it to.

dennisdoomen:
> Ensuring that the subject does not contain an unexpected [subsequence](https://en.wikipedia.org/wiki/Subsequence)

Yes, that put me off-guard a couple of times as well.

PR #1343, opened 2020-06-03
Added XSD validation for XDocument assertions.

mrcopperbeard:
## IMPORTANT 

* [X] If the contribution adds a feature or fixes a bug, please update [**the release notes**](https://github.com/fluentassertions/fluentassertions/tree/master/docs), which are published on the [website](https://fluentassertions.com/releases).
* [X] If the contribution changes the public API the changes needs to be included by running `AcceptApiChanges.ps1`/`AcceptApiChanges.sh`.
* [X] If the contribution affects [the documentation](https://github.com/fluentassertions/fluentassertions/tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).

### DIFF
--- a/Src/FluentAssertions/Xml/XDocumentAssertions.cs
+++ b/Src/FluentAssertions/Xml/XDocumentAssertions.cs
@@ -1,7 +1,9 @@
 using System;
 using System.Diagnostics;
+using System.IO;
 using System.Xml;
 using System.Xml.Linq;
+using System.Xml.Schema;
 
 using FluentAssertions.Common;
 using FluentAssertions.Execution;
@@ -112,6 +114,44 @@ namespace FluentAssertions.Xml
                 var xmlReaderValidator = new XmlReaderValidator(subjectReader, otherReader, because, becauseArgs);
                 xmlReaderValidator.Validate(shouldBeEquivalent: false);
             }
+
+            return new AndConstraint<XDocumentAssertions>(this);
+        }
+
+        /// <summary>
+        /// Asserts that document is valid by XSD schema.
+        /// </summary>
+        /// <param name="xsd">XDS schema.</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        /// <returns>Instance of <see cref="AndConstraint{T}"/>.</returns>
+        public AndConstraint<XDocumentAssertions> BeValidByXsd(string xsd, string because = "", params object[] becauseArgs)
+        {
+            var isValid = true;
+            var errorMessage = default(string);
+            var schemaSet = new XmlSchemaSet();
+
+            using var textReader = new StringReader(xsd);
+            using var xmlReader = XmlReader.Create(textReader);
+
+            schemaSet.Add(string.Empty, xmlReader);
+            schemaSet.Compile();
+
+            Subject.Validate(schemaSet, (o, e) =>
+            {
+                isValid = false;
+                errorMessage = e.Message;
+            });
+
+            Execute.Assertion
+                .BecauseOf(because, becauseArgs)
+                .ForCondition(isValid)
+                .FailWith("Expected XML document to match XSD{reason}, but failed with {0}.", errorMessage);
 
             return new AndConstraint<XDocumentAssertions>(this);
         }
--- a/Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
@@ -9,6 +9,24 @@ namespace FluentAssertions.Specs
 {
     public class XDocumentAssertionSpecs
     {
+        private const string XsdSample = @"<xsd:schema xmlns:xsd='http://www.w3.org/2001/XMLSchema' elementFormDefault='qualified'>
+                <xsd:element name='PurchaseOrder' type='PurchaseOrderType'/>
+                <xsd:complexType name='PurchaseOrderType'>
+                    <xsd:sequence>
+                        <xsd:element name='ShipTo' type='USAddress' maxOccurs='2'/>
+                        <xsd:element name='BillTo' type='USAddress'/>
+                    </xsd:sequence>
+                    <xsd:attribute name='OrderDate' type='xsd:date'/>
+                </xsd:complexType>
+                <xsd:complexType name='USAddress'>
+                    <xsd:sequence>
+                        <xsd:element name='name' type='xsd:string'/>
+                        <xsd:element name='street' type='xsd:string'/>
+                    </xsd:sequence>
+                    <xsd:attribute name='country' type='xsd:NMTOKEN' fixed='US'/>
+                </xsd:complexType>
+            </xsd:schema>";
+
         #region Be / NotBe
 
         [Fact]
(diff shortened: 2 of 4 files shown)


Src/FluentAssertions/Xml/XDocumentAssertions.cs
```diff
@@ -116,6 +118,44 @@ public AndConstraint<XDocumentAssertions> NotBeEquivalentTo(XDocument unexpected
...
+        /// <param name="xsd">XDS schema.</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        /// <returns>Instance of <see cref="AndConstraint{T}"/>.</returns>
+        public AndConstraint<XDocumentAssertions> BeValidByXsd(byte[] xsd, string because = "", params object[] becauseArgs)
```
lg2de, line 133: Why do you expect the XSD in binary format?
Just a byte array is very unspecific. It is unclear what kind of "binary represented XSD schema" is expected.
I strongly recommend to accept here `string` or maybe `XDocument` only.

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -1009,5 +1028,55 @@ public void When_asserting_a_document_has_a_null_element_it_should_fail()
...
+                        <name>Shepard</name>
+                        <street>Street</street>
+                    </ShipTo>
+                </PurchaseOrder>");
+
+            // Act
+            Action act = () => document.Should().BeValidByXsd(XsdSample);
+
+            // Assert
+            act.Should().Throw<XunitException>();
```
lg2de, line 1077: Please add `WithMessage` check to validate correctly build failure message sentence including the XSD validation error message and `because` string.

Src/FluentAssertions/Xml/XDocumentAssertions.cs
```diff
@@ -116,6 +118,44 @@ public AndConstraint<XDocumentAssertions> NotBeEquivalentTo(XDocument unexpected
...
+            Subject.Validate(schemaSet, (o, e) =>
+            {
+                isValid = false;
+                errorMessage = e.Message;
+            });
+
+            Execute.Assertion
+                .BecauseOf(because, becauseArgs)
+                .ForCondition(isValid)
+                .FailWith("Given XML is invalid by XSD-schema. {0}{reason}.", errorMessage);
```
lg2de, line 154: My recommendation:
`Expected XML document to match XSD{reason}, but failed with {0}.`

mrcopperbeard:
Thanks @lg2de for review! Fixed all conversations

lg2de:
> Fixed all conversations

But you forgot (to answer) my general question: How about `NotBeValidByXsd`?

Src/FluentAssertions/Xml/XDocumentAssertions.cs
```diff
@@ -116,6 +118,44 @@ public AndConstraint<XDocumentAssertions> NotBeEquivalentTo(XDocument unexpected
...
+            var errorMessage = default(string);
+            var schemaSet = new XmlSchemaSet();
+
+            using var textReader = new StringReader(xsd);
+            using var xmlReader = XmlReader.Create(textReader);
+
+            schemaSet.Add(string.Empty, xmlReader);
+            schemaSet.Compile();
+
+            Subject.Validate(schemaSet, (o, e) =>
```
dennisdoomen, line 145: Will this event only ever be raised once? If not, consider collecting all the error messages.

Src/FluentAssertions/Xml/XDocumentAssertions.cs
```diff
@@ -116,6 +118,44 @@ public AndConstraint<XDocumentAssertions> NotBeEquivalentTo(XDocument unexpected
...
+        /// <param name="xsd">XDS schema.</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        /// <returns>Instance of <see cref="AndConstraint{T}"/>.</returns>
+        public AndConstraint<XDocumentAssertions> BeValidByXsd(string xsd, string because = "", params object[] becauseArgs)
```
dennisdoomen, line 133: What happens if `xsd` is `null` or empty?

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -1009,5 +1027,55 @@ public void When_asserting_a_document_has_a_null_element_it_should_fail()
         }
 
         #endregion
+
+        #region xsd
+
+        [Fact]
+        public void Xsd_validation_for_valid_xml_should_work()
```
dennisdoomen, line 1034: We prefer to use the `When`... `then_it_should` convention.

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -1009,5 +1027,55 @@ public void When_asserting_a_document_has_a_null_element_it_should_fail()
...
+                        <name>Shepard</name>
+                        <street>Street</street>
+                    </ShipTo>
+                </PurchaseOrder>");
+
+            // Act
+            Action act = () => document.Should().BeValidByXsd(XsdSample, " because we want to test the failure message");
+
+            // Assert
+            act.Should().Throw<XunitException>().WithMessage("Expected XML document to match XSD because we want to test the failure message*");
```
dennisdoomen, line 1076: I would at least assert that a fragment of the expected error message from the XSD validation is there.

dennisdoomen:
> How about `NotBeValidByXsd`?

I think this is a nice to have. Why would you ever want to do this?

mrcopperbeard:
I thought this is not quite a production case, but maybe it will be useful for some pre-assertions or testing validation. I'll add that thing after my workday

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -1009,5 +1027,55 @@ public void When_asserting_a_document_has_a_null_element_it_should_fail()
...
+                        <street>Street</street>
+                    </BillTo>
+                    <ShipTo>
+                        <name>Shepard</name>
+                        <street>Street</street>
+                    </ShipTo>
+                </PurchaseOrder>");
+
+            // Act
+            Action act = () => document.Should().BeValidByXsd(XsdSample, " because we want to test the failure message");
```
lg2de, line 1073: The space character before "because" should not be required.

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
lg2de: On one hand I agree.
On the other hand I want to note, that the validation message will be language specific. So the test will depend on the locale settings of the executing host.
Maybe the test should be executed within fixed thread locale.

lg2de:
> > How about `NotBeValidByXsd`?
>
> I think this is a nice to have. Why would you ever want to do this?

Sure, it can be added easily as soon it is required.

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
mrcopperbeard: Yep, I didn't assert internal xsd error, because I have it in Russian

dennisdoomen: You can use the `[CulturedFact]` attribute to force

jnyrup:
You've marked
> If the contribution changes the public API the changes needs to be included by running AcceptApiChanges.ps1/AcceptApiChanges.sh.
but I do not the see changed api files.
Am I missing something?

lg2de:
> Am I missing something?

No, there is new method in public API.
This should cause API validation test to fail. But it succeed!

PR #1354, opened 2020-07-16
Make culture sensitive test less fragile to OS/TFM changes

jnyrup:
When performing culture sensitive string comparisons:
* .Net Framework and .Net Core on Windows uses NLS, while 
* .Net Core on Linux/MacOS and the upcoming .Net 5.0 uses ICU.
See https://docs.microsoft.com/en-us/dotnet/core/compatibility/globalization#net-50

The current test for case sensitive string comparison only worked with NLS, i.e. it is fragile if changing OS or upgrading target framework.

I searched for another pair of characters that showed same behavior as the Mandarin ones, but which works with both ICU and NLS.
So I don't have an explaining why they work, I only observe that serve my purpose and hope they continue to do so.

To verify this I've checked locally that the unit tests works on
* .Net 5.0 preview 6 on windows 10 Version 2004 (the change to ICU happened in [preview 4](https://docs.microsoft.com/en-us/dotnet/core/compatibility/globalization#version-introduced))
* Ubuntu 20.04 using WSL 2 (I only checked .net core app 3.0 tests here though)

I've named the two characters after their corresponding unicode names.

### DIFF
--- a/Tests/FluentAssertions.Specs/Primitives/StringComparisonSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Primitives/StringComparisonSpecs.cs
@@ -20,12 +20,12 @@ namespace FluentAssertions.Specs.Primitives
 #pragma warning restore CA1309
 
             // Assert
-            ordinal.Should().Be(!currentCulture, "Turkish distinguishes between a dotted and a non-dotted 'i'");
+            ordinal.Should().NotBe(currentCulture, "Turkish distinguishes between a dotted and a non-dotted 'i'");
         }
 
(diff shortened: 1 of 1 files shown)


dennisdoomen:
> * Ubuntu 20.04

You're running Ubuntu?

jnyrup:
On my personal computer Linux has been my main OS for the past ~10 years. Currently using the Ubuntu derivative Linux Mint.
On my working computer I'm using [WSL](https://docs.microsoft.com/en-us/windows/wsl/about) (which uses Ubuntu) when Git Bash comes to short.

PR #1357, opened 2020-07-17
Implemented string upper/lower case assertions

### LINKED ISSUE #1356: Feature request: lower/upper case assertions
### Description

Please introduce lowercase and uppercase assertions for strings or chars, for example:

```c#
// Arrange
var input = "A";

// Act
var actual = input.ToUpper();

// Assert
actual.Should().BeUpper();
actual.Should().NotBeLower();
```

weitzhandler:
Adds the following `Should`s to `string`, to assert its casing:

- `BeUpper`
- `NotBeUpper`
- `BeLower`
- `NotBeLower`

Fixes #1356

### DIFF
--- a/Src/FluentAssertions/Primitives/StringAssertions.cs
+++ b/Src/FluentAssertions/Primitives/StringAssertions.cs
@@ -1175,7 +1175,7 @@ namespace FluentAssertions.Primitives
         }
 
         /// <summary>
-        /// Asserts that all characters in a string are not in upper casing.
+        /// Asserts that all characters in a string are not in lower casing.
         /// </summary>
         /// <param name="because">
         /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
--- a/Src/FluentAssertions/Primitives/StringAssertions.cs
+++ b/Src/FluentAssertions/Primitives/StringAssertions.cs
@@ -1115,7 +1115,7 @@ namespace FluentAssertions.Primitives
         }
 
         /// <summary>
-        /// Asserts that a string is uppercase.
+        /// Asserts that all characters in a string are in upper casing.
         /// </summary>
         /// <param name="because">
         /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
@@ -1124,18 +1124,18 @@ namespace FluentAssertions.Primitives
         /// <param name="becauseArgs">
         /// Zero or more objects to format using the placeholders in <paramref name="because" />.
         /// </param>
-        public AndConstraint<TAssertions> BeUpper(string because = "", params object[] becauseArgs)
+        public AndConstraint<TAssertions> BeUpperCased(string because = "", params object[] becauseArgs)
         {
             Execute.Assertion
                 .ForCondition(Subject?.All(char.IsUpper) == true)
                 .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context:string} to be upper{reason}, but found {0}.", Subject);
+                .FailWith("Expected all characters in {context:string} to be upper cased{reason}, but found {0}.", Subject);
 
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
         /// <summary>
-        /// Asserts that a string is not uppercase.
+        /// Asserts that all characters in a string are not in upper casing.
         /// </summary>
         /// <param name="because">
         /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
@@ -1144,18 +1144,18 @@ namespace FluentAssertions.Primitives
         /// <param name="becauseArgs">
         /// Zero or more objects to format using the placeholders in <paramref name="because" />.
         /// </param>
-        public AndConstraint<TAssertions> NotBeUpper(string because = "", params object[] becauseArgs)
+        public AndConstraint<TAssertions> NotBeUpperCased(string because = "", params object[] becauseArgs)
         {
             Execute.Assertion
                 .ForCondition(Subject is null || Subject.Any(ch => !char.IsUpper(ch)))
                 .BecauseOf(because, becauseArgs)
-                .FailWith("Did not expect {context:string} to be upper{reason}.");
+                .FailWith("Did not expect any characters in {context:string} to be upper cased{reason}.");
 
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
         /// <summary>
-        /// Asserts that a string is lowercase.
+        /// Asserts that all characters in a string are in lower casing.
         /// </summary>
         /// <param name="because">
         /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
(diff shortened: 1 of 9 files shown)


jnyrup:
Continuing the discussion about API naming...
The BCL also has `ToTitleCase`, which _does_ use the "Case" suffix, so I'd like to keep the door open for alternative names a bit.
I think `BeUpperCase` reads closer to regular prose.
While we should get inspiration from BCL, it does not need to dictate our choices.

How does `char.IsUpper` handle culture?

I recall that `char.IsUpper` only returns true for characters that are both letters _and_ uppercase.
That clearly justifies that we should have all four combinations, as a character not being upper case does not imply that it is lower case, e.g. digits are neither lower case nor upper case.

I think the test suite would benefit from tests like these to reflect those cases.
```cs
subject = "A1";

Action act = () => subject.Should().BeUpper();

act.Should().Throw();
```

```cs
subject = "a1";

Action act = () => subject.Should().NotBeUpper();

act.Should().NotThrow();
```

and the complementary for `[Not]BeLower`.

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -1114,6 +1114,86 @@ public AndConstraint<TAssertions> BeNullOrWhiteSpace(string because = "", params
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
+        /// <summary>
+        /// Asserts that a string is uppercase.
```
dennisdoomen, line 1118: I think it would be better to phrase it like "Asserts that all characters in a string are in upper casing"

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -1114,6 +1114,86 @@ public AndConstraint<TAssertions> BeNullOrWhiteSpace(string because = "", params
...
+        /// </summary>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> BeUpper(string because = "", params object[] becauseArgs)
```
dennisdoomen, line 1127: Maybe `BeUpperCased`?

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.cs
```diff
@@ -3243,5 +3243,220 @@ public void When_a_whitespace_string_is_expected_to_not_be_null_or_whitespace_it
         }
 
         #endregion
+
+        #region (Not) Upper
+
+        [Fact]
+        public void Should_succeed_when_asserting_upper_string_to_be_upper()
```
dennisdoomen, line 3250: Our convention is `When_something_it_should_do_something`. I know we have some older tests that don't follow that convention, but we need to change that,

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -1114,6 +1114,86 @@ public AndConstraint<TAssertions> BeNullOrWhiteSpace(string because = "", params
...
+        {
+            Execute.Assertion
+                .ForCondition(Subject?.All(char.IsUpper) == true)
```
dennisdoomen, line 1130: What if the string is empty?

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -1114,6 +1114,86 @@ public AndConstraint<TAssertions> BeNullOrWhiteSpace(string because = "", params
...
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected {context:string} to be upper{reason}, but found {0}.", Subject);
```
dennisdoomen, line 1132: I would be more specific, e.g. `Expected all characters in {context:string} to be upper cased`

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.cs
weitzhandler: Do you mean `When_asserting_uppercase_string_to_be_upper_should_succeed`, in this case? Just to be clear.

Src/FluentAssertions/Primitives/StringAssertions.cs
weitzhandler: That means it's not upper case. And I've actually asked this question [here](https://github.com/fluentassertions/fluentassertions/issues/1356#issuecomment-659963900).

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.cs
dennisdoomen: Almost ;-)

`When_an_upper_case_string_is_supposed_to_be_in_upper_case_only_it_should_not_throw`

weitzhandler:
https://github.com/fluentassertions/fluentassertions/commit/cca1d01bd64121dca56783c2ff4dbabc25077d59:

- Resolved comments
- Rebased on `upstream/develop`
- Updated docs

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -1114,6 +1114,86 @@ public AndConstraint<TAssertions> BeNullOrWhiteSpace(string because = "", params
...
+            Execute.Assertion
+                .ForCondition(Subject?.All(char.IsLower) == true)
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected all characters in {context:string} to be lower cased{reason}, but found {0}.", Subject);
+
+            return new AndConstraint<TAssertions>((TAssertions)this);
+        }
+
+        /// <summary>
+        /// Asserts that all characters in a string are not in upper casing.
```
jnyrup, line 1178:
```suggestion
        /// Asserts that all characters in a string are not in lower casing.
```

docs/_pages/releases.md
```diff
@@ -34,6 +34,7 @@ sidebar:
 * Added `AllBe` to `StringCollectionAssertions` to be able to assert that all strings in collection are equal to the specified string - [#1332](https://github.com/fluentassertions/fluentassertions/pull/1332).
 * Added `ForConstraint` method to `AssertionScope` to open up `OccurenceConstraint` for usage in custom assertion extensions - [#1341](https://github.com/fluentassertions/fluentassertions/pull/1341).
 * Added `NotContainInOrder` to `CollectionAssertions` and `StringCollectionAssertions` to be able to assert that the collection does not contain the specified elements in the exact same order, not necessarily consecutive - [#1339](https://github.com/fluentassertions/fluentassertions/pull/1339).
+* Added `[Not]BeUpperCase` and `[Not]BeLowerCase` to `StringAssertions` to be able to assert that a string is in upper or lower casing or not - [#1357](https://github.com/fluentassertions/fluentassertions/pull/1357).
```
jnyrup, line 37:
```suggestion
* Added `[Not]BeUpperCased` and `[Not]BeLowerCased` to `StringAssertions` to be able to assert that a string is in upper or lower casing or not - [#1357](https://github.com/fluentassertions/fluentassertions/pull/1357).
```

docs/_pages/strings.md
```diff
@@ -18,6 +18,10 @@ theString.Should().NotBeEmpty("because the string is not empty");
 theString.Should().HaveLength(0);
 theString.Should().BeNullOrWhiteSpace(); // either null, empty or whitespace only
 theString.Should().NotBeNullOrWhiteSpace();
+theString.Should().BeUpperCase();
+theString.Should().NotBeUpperCase();
+theString.Should().BeLowerCase();
+theString.Should().NotBeLowerCase();
```
jnyrup, line 24:
```suggestion
theString.Should().BeUpperCased();
theString.Should().NotBeUpperCased();
theString.Should().BeLowerCased();
theString.Should().NotBeLowerCased();
```

Issue #1352, opened 2020-07-15
Add `async`-support for `Where()` 


voroninp:
It looks like there is no counterpart of `Where` method to check exception properties for async delegates.

Would be nice to have an option to check exception properties:

```
await foo.Awaiting(_ => _.AsyncMethodThroingException())
                .Should().ThrowAsync<Exception>().Where(ex => ...);
```

jnyrup:
As `ThrowAsync` returns a `Task<ExceptionAssertions<TException>>` you have to `await` the task before calling `Where` as `.` binds stronger than `await`.
```cs
(await foo.Awaiting(_ => _.AsyncMethodThroingException())
    .Should().ThrowAsync<Exception>())
    .Where(ex => ...);
```

We got an this extension method for `WithMessage` in #1055 to avoid the extra parentheses 
https://github.com/fluentassertions/fluentassertions/blob/c192ca014ae920c96f76a747b72bf110c31ba153/Src/FluentAssertions/AssertionExtensions.cs#L791-L799

and I see no reason why we couldn't also have similar extension methods for the remaining three methods on `ExceptionAssertions<TException>`
* `WithInnerException`
* `WithInnerExceptionExactly`
* `Where`

lg2de:
> and I see no reason why we couldn't also have similar extension methods for the remaining three methods

In #1324 I already tried to create an extension for `WithInnerException`. The result is not as good as expected because it requires additional generic parameter. Hope, somebody will find better solution for a compact fluent API.

The extension for `Where` should not cause such problem and should be added.

voroninp:
Yet there is a small syntactic issue with `WithInnerException` and `WithInnerExceptionExactly`:

They are parameterized explicitly with `TInnerExcpetion` and implicitly from the class with `TException`. Hence, extension method must specify two generic type arguments, and despite C# will inference the type of `TException` from `this` parameter, it will force the user to write the type anyway =(

```CSharp
await foo.Invoking(_ => foo.Bar()).ThrowsAsync<MyException>().WithInnerException<MyException, MyInnerException>(...);
```
Maybe it's worth changing the API here a bit, and add: `ThrowsWithInnerExcetionAsync<TException,TInnerException>()`

What is your opinion?

jnyrup:
As `ThrowsWithInnerExceptionAsync` asserts more than one thing at a time, my initial gut feeling is that it moves in a direction we've tried to avoid in Fluent Assertions.
[...]

PR #1362, opened 2020-07-23
Add async support for Where method (#1352)

voroninp:
#1352


Src/FluentAssertions/AssertionExtensions.cs
```diff
@@ -816,6 +816,31 @@ public static TTo As<TTo>(this object subject)
             return (await task).WithInnerException<TInnerException>(because, becauseArgs);
         }
 
+#pragma warning restore AV1755
```
lg2de, line 819: Why restoring warning for the documentation?
Instead I would recommend to disable this warning for the whole file.
All these extension will not follow the rule because of the fluent api.

voroninp:
@dennisdoomen Yep, but which release version should it be?

Src/FluentAssertions/AssertionExtensions.cs
voroninp: Ouch, this should not be here. Should I amend or just add new commit?

dennisdoomen: Whatever works for you. I follow [this procedure](https://www.continuousimprover.com/2020/03/keep-source-control-history-clean.html).

dennisdoomen:
> Yep, but which release version should it be?

6

PR #1371, opened 2020-08-09
Make ObjectAssertions generic in TSubject

jnyrup:
When creating a custom assertions class, deriving from `ObjectAssertions<TSubject, TAssertions>` instead of `ReferenceTypeAssertions<TSubject, TAssertions>` gives you `[Not]Be` and `[Not]BeEquivalentTo` for free.

### DIFF
--- a/Src/FluentAssertions/Primitives/ObjectAssertions.cs
+++ b/Src/FluentAssertions/Primitives/ObjectAssertions.cs
@@ -9,15 +9,27 @@ namespace FluentAssertions.Primitives
     /// <summary>
     /// Contains a number of methods to assert that an <see cref="object"/> is in the expected state.
     /// </summary>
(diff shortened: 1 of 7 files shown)


jnyrup:
@lg2de Can you think of why `When_task_is_checking_synchronization_context_on_UI_thread_it_should_succeed` would fail?

lg2de:
I'm currently away from office, so I have no chance for detailed analysis. 

I've no explanation. There is no logical reason according to these changes.
So, I fear the problem can occur any time. 

Maybe we should increase the timeouts. If it is a deadlock, the test will fail anyway. But we must ensure that the problem is not caused by accidental overload of the build machine.

jnyrup:
Ahh... It probably just needs to use a `FakeClock`.

jnyrup:
I've rerun the CI-build successfully and tracked the flaky test in #1372 1372

PR #1376, opened 2020-08-13
Internalized some of the type extensions only meant to be used by FA

dennisdoomen:
Quite some extension methods where part of a `public` class, which caused some people to use them even though they were not designed for that. I had to keep some of them public because they are covered by tests directly.

### DIFF
--- a/Src/FluentAssertions/Common/CSharpAccessModifierExtensions.cs
+++ b/Src/FluentAssertions/Common/CSharpAccessModifierExtensions.cs
@@ -3,7 +3,7 @@ using System.Reflection;
 
 namespace FluentAssertions.Common
 {
-    public static class CSharpAccessModifierExtensions
+    internal static class CSharpAccessModifierExtensions
     {
         internal static CSharpAccessModifier GetCSharpAccessModifier(this MethodBase methodBase)
         {
--- a/Src/FluentAssertions/Common/InternalObjectExtensions.cs
+++ b/Src/FluentAssertions/Common/InternalObjectExtensions.cs
@@ -3,7 +3,8 @@ using System.Globalization;
 
 namespace FluentAssertions.Common
 {
-    public static class ObjectExtensions
+    // NOTE: These extension methods are covered by tests separately and therefore need to be public.
+    public static class InternalObjectExtensions
     {
         public static bool IsSameOrEqualTo(this object actual, object expected)
         {
--- a/Tests/FluentAssertions.Specs/Common/TypeExtensions.cs
+++ b/Tests/FluentAssertions.Specs/Common/TypeExtensions.cs
@@ -1,0 +1,30 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace FluentAssertions.Specs.Common
+{
+    internal static class TypeExtensions
+    {
+        private const BindingFlags AllMembersFlag =
+            BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance;
+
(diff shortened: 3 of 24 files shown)


Src/FluentAssertions/Common/InternalObjectExtensions.cs
```diff
@@ -3,7 +3,8 @@
 
 namespace FluentAssertions.Common
 {
-    public static class ObjectExtensions
+    // NOTE: These extension methods are covered by tests separately and therefore need to be public.
```
jnyrup, line 6: We could add [`[InternalsVisibleTo]`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=netcore-3.1)?

Tests/FluentAssertions.Specs/Common/TypeExtensions.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace FluentAssertions.Specs.Common
+{
+    internal static class TypeExtensions
```
jnyrup, line 8: Are these methods duplicated in core library or only present i test project?

Src/FluentAssertions/Common/CSharpAccessModifierExtensions.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace FluentAssertions.Common
 {
-    public static class CSharpAccessModifierExtensions
+    internal static class CSharpAccessModifierExtensions
```
jnyrup, line 6: How a test like this pseudo code?
```cs
AllTypes
    .UnderNamespace("FluentAssertions.Common")
    .ThatAreClasses
    .Should().BeInternal();
```

Src/FluentAssertions/Common/InternalObjectExtensions.cs
dennisdoomen: We could, but then the Specs project must get a strong name, which requirement then bubbles to Chill, etc.

Tests/FluentAssertions.Specs/Common/TypeExtensions.cs
dennisdoomen: They are duplicated. IMO, there's no need for the specs project to reuse internal code of the core project.

Src/FluentAssertions/Common/CSharpAccessModifierExtensions.cs
dennisdoomen: I don't get your point. Those extensions where only used by internal code.

jnyrup: My idea was a test to verify that types in the `Common` namespace indeed are marked as `internal`.
But it might be superfluous given we have the API approval tests to verify API changes.

Tests/FluentAssertions.Specs/Common/TypeExtensions.cs
jnyrup: My potential concern is if the methods get out-of-sync and are no longer 1:1 duplicates.

Src/FluentAssertions/Common/InternalObjectExtensions.cs
jnyrup: 😢

Tests/FluentAssertions.Specs/Common/TypeExtensions.cs
dennisdoomen: For the purpose of the tests, they don't have to be in sync. Reusing them is IMO an example of overusing DRY.

PR #1378, opened 2020-08-13
Bump xUnit.StaFact to the latest version

dennisdoomen:
Let's see if it keeps running without problems.


dennisdoomen:
Just testing if this fixes the weird problem I reported in Slack.

PR #1381, opened 2020-09-09
Add an AssertionCount property on AssertionScope

### LINKED ISSUE #1380: How to count the number of assertions inside an AssertionScope?
I'm trying to count the number of assertions inside an `AssertionScope` but I currently can't find a solution to achieve this goal.

If I subclass `AssertionScope`, I can't override `ForCondition` since it's not virtual.

```csharp
public class CountingAssertionScope : AssertionScope
{
    public int Count { get; private set; }

    // compilation error, ForCondition is not virtual
    public override IAssertionScope ForCondition(bool condition)
    {
        Count++;
        return base.ForCondition(condition);
    }
}
```

If I try to implement `IAssertionScope` and delegate to `AssertionScope` then I can't have my own implementation set as the current scope since `AssertionScope.Current` has a private setter.

```csharp
public class CountingAssertionScope : IAssertionScope
{
    private readonly IAssertionScope _assertionScope;

    public int Count { get; private set; }

    public CountingAssertionScope() : this(new AssertionScope())
    {
    }

    public CountingAssertionScope(IAssertionScope assertionScope)
    {
        _assertionScope = assertionScope ?? throw new ArgumentNullException(nameof(assertionScope));
        // compilation error, private setter
        AssertionScope.Current = this;
    }

    public void Dispose()
    {
        _assertionScope.Dispose();
    }

    public IAssertionScope ForCondition(bool condition)
    {
        Count++;
        return _assertionScope.ForCondition(condition);
    }
    
    // [...]
}
```

Can you think of a solution to this problem?

0xced:
Fixes #1380

Marked the pull request as draft because the `When_disposing_the_assertion_scope_new_assertions_should_not_count` does not pass. Getting *Expected value to be 2, but found 3.* on the last assertion. I don't know why yet.

### DIFF
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -85,6 +85,11 @@ namespace FluentAssertions.Execution
         public string Context { get; set; }
 
         /// <summary>
+        /// Gets the number of assertions in the assertion scope.
+        /// </summary>
+        public int AssertionCount { get; private set; }
+
+        /// <summary>
         /// Gets the current thread-specific assertion scope.
         /// </summary>
         public static AssertionScope Current
@@ -178,6 +183,8 @@ namespace FluentAssertions.Execution
 
         public AssertionScope ForCondition(bool condition)
         {
+            AssertionCount++;
+
             succeeded = condition;
 
             return this;
--- a/Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.approved.txt
+++ b/Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.approved.txt
@@ -942,6 +942,7 @@ namespace FluentAssertions.Execution
         public AssertionScope() { }
         public AssertionScope(FluentAssertions.Execution.IAssertionStrategy assertionStrategy) { }
         public AssertionScope(string context) { }
+        public int AssertionCount { get; }
         public string Context { get; set; }
         public bool Succeeded { get; }
         public FluentAssertions.Execution.AssertionScope UsingLineBreaks { get; }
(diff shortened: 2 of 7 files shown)


0xced:
I figured out why the test was not passing (it was an error in the test itself). This pull request is now ready to be reviewed.

There is [one test failing](https://ci.appveyor.com/project/dennisdoomen/fluentassertions/builds/35118969#L264) on continuous integration which is totally unrelated: `When_task_is_checking_synchronization_context_on_UI_thread_it_should_succeed`.

> Expected task to complete within 1s.

Maybe you should add a little margin, complete within 1.1 seconds for a task waiting 1 second?

dennisdoomen:
> Maybe you should add a little margin, complete within 1.1 seconds for a task waiting 1 second?

Agreed. In async code, delays are never exact.

Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -84,6 +84,11 @@ public AssertionScope(string context)
         /// </summary>
         public string Context { get; set; }
 
+        /// <summary>
+        /// Gets the number of assertions in the assertion scope.
```
dennisdoomen, line 88: I think it would be better to describe it as "Gets the number of assertions that were executed since the AssertionScope was created and the last time DiscardAny was called.

Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -84,6 +84,11 @@ public AssertionScope(string context)
...
+        /// </summary>
+        public int AssertionCount { get; private set; }
```
dennisdoomen, line 90: Would also be nice to have a `FailureCount`.

dennisdoomen (approved):
Just a couple of suggestions.

Src/FluentAssertions/Execution/AssertionScope.cs
jnyrup: Good you mention it, I was thinking about how this would behave if a chain of assertions fails before executing all assertions in the chain.

dennisdoomen:
@0xced how do you want to proceed?

0xced:
The reason why I opened this pull request was because of https://github.com/ericsink/SQLitePCL.raw/pull/370. But as it stands, it seems @ericsink doesn't want to take a dependency on FluentAssertions. So I'm going to close this pull request, I have enough outstanding pull requests open across many repositories. 😉 I'll reopen if https://github.com/ericsink/SQLitePCL.raw/pull/370 is ever merged.

PR #1383, opened 2020-09-10
add ComparingByMembers(Func<Type, bool> predicate) to SelfReferenceEq…

ursenzler:
…uivalencyAssertionOptions<TSelf>

Adds `ComparingByMembers(Func<Type, bool> predicate)` to `SelfReferenceEquivalencyAssertionOptions<TSelf>` so that it is possible to define the strategy used for comparison with a callback. This simplifies scenarios when a group of types should be handled differently than FluentAssertion's standard behaviour. 

This is useful when for example structurally comparing F# record types that contain C# classes that do not implement equality.
F# records are by default treated as value types because they implement equality. As a result, the containing values are treated as value types, making structural comparison of the inner C# objects difficult (every type would have to be configured).
With the new callback, it is possible to define the behaviour for all record types at once:

```fsharp
actual.Should()
      .BeEquivalentTo(expected,
                      Func<_, _>(fun (config : EquivalencyAssertionOptions<'a>) ->
                          config.ComparingByMembers(fun t -> FSharpType.IsRecord t)),
                      "",
                      [])
```

## IMPORTANT 


Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
```diff
@@ -538,6 +546,12 @@ public TSelf ComparingByMembers<T>()
             return (TSelf)this;
         }
 
+        public TSelf ComparingByMembers(Func<Type, bool> predicate)
```
dennisdoomen, line 549: ❌ Please document this new member in the same style as the other overloads.

Tests/FluentAssertions.Specs/Equivalency/BasicEquivalencySpecs.cs
```diff
@@ -596,6 +596,56 @@ public void When_the_graph_contains_guids_it_should_properly_format_them()
             act.Should().Throw<XunitException>().WithMessage("Expected item[0].Id to be *-*, but found *-*");
         }
 
+        [Fact]
+        public void When_defining_comparing_by_members_with_callback_that_returns_false_then_the_normal_equality_strategy_process_should_be_used()
```
dennisdoomen, line 600: 🙃 I'm nitpicking here, but we prefer to keep those test names functional and never refer to the literal name of an identifier, e.g. `When_none_of_the_members_should_be_compared_by_members_it_should_succeed`

Tests/FluentAssertions.Specs/Equivalency/BasicEquivalencySpecs.cs
```diff
@@ -596,6 +596,56 @@ public void When_the_graph_contains_guids_it_should_properly_format_them()
...
+            // Act
+            Action act = () => subject.Should().BeEquivalentTo(expected, options => options
+                .ComparingByMembers(t => false));
+
+            // Assert
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_defining_comparing_by_members_with_callback_it_should_use_force_members_strategy_if_a_callback_returning_true_is_found()
```
dennisdoomen, line 624: 🤔 So is there some implicit or explcit order? Would be nice to capture that in a functional test name.

dennisdoomen (changes requested):
Would you be willing to also update the release notes and documentation as listed in the PR template? Otherwise we have to pick up the PR ourselves.

ursenzler:
I close this PR because the change how `ComparingByMember<object>` handles basic types solves the same problem.

Issue #1023, opened 2019-04-15
ComparingByMembers for open generic type


oakio:
### Description

It would be nice to have a method like `ComparingByMembers` for open generic types.

```csharp
struct Container<T>
{
    public T Data;
    public Container(T data)
    {
        Data = data;
    }
}

class Foo 
{
}

var actual = new Container<Foo>(new Foo());
var expected = ...

actual.Should().BeEquivalentTo(expected, o => o.ComparingByMembers(typeof(Container<>)))

```

PR #1389, opened 2020-09-16
Added `CompareBy{Members,Value}(Type)`

### LINKED ISSUE #1388: AssertEquivalencyUsing support for unbound generic types
### Description

I came across [this question on Stack Overflow](https://stackoverflow.com/questions/63919524/testing-optional-equivalence-with-fluentassertions/) where someone uses [Optional](https://github.com/nlkl/Optional) in combination with FA.

Optional provides a generic struct `Option<T>` that overrides `Equals`, so FA assumes value semantics.

Now if one would want to override this behavior globally, so `Option<T>` uses `ComparingByMembers` for **any** `T`, one had to add a call to `AssertionOptions.AssertEquivalencyUsing` **for each** `T`.

A (simplistic) workaround for this is an extension method like this:

```cs
public static class FluentAssertionsExtensions
{
    public static void BeEquivalentByMembers<TExpectation>(
        this ComparableTypeAssertions<TExpectation> actual,
        TExpectation expectation)
    {
        actual.BeEquivalentTo(
            expectation,
            options => options.ComparingByMembers<TExpectation>());
    }
}
```

If an override that supports a `Type` argument existed, the following would be possible:

`AssertionOptions.AssertEquivalencyUsing(options => options.ComparingByMembers(typeof(Option<>))`

jnyrup:
The new overloads allows you to use an open generic type.
If both an open and closed type are specified, the closed type takes precedence.

This fixes #1388, #1023

### DIFF
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
@@ -172,6 +172,14 @@ namespace FluentAssertions.Equivalency
             {
                 strategy = EqualityStrategy.ForceEquals;
             }
+            else if (referenceTypes.Any(type.IsAssignableToOpenGeneric))
+            {
+                strategy = EqualityStrategy.ForceMembers;
+            }
+            else if (valueTypes.Any(type.IsAssignableToOpenGeneric))
+            {
+                strategy = EqualityStrategy.ForceEquals;
+            }
             else
             {
                 if (getDefaultEqualityStrategy != null)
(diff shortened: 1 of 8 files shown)


Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
```diff
@@ -164,11 +164,13 @@ EqualityStrategy IEquivalencyAssertionOptions.GetEqualityStrategy(Type type)
         {
             EqualityStrategy strategy;
 
-            if (referenceTypes.Any(type.IsSameOrInherits))
+            if (referenceTypes.Any(type.IsSameOrInherits)
+                || (referenceTypes.Any(type.IsAssignableToOpenGeneric) && !valueTypes.Any(type.IsSameOrInherits)))
```
jnyrup, line 168: Could also be structured this way for fewer calls to `Any`

```cs
if (referenceTypes.Any(type.IsSameOrInherits))
{
    strategy = EqualityStrategy.ForceMembers
}
else if (valueTypes.Any(type.IsSameOrInherits))
{
    strategy = EqualityStrategy.ForceEquals;
}
else if (referenceTypes.Any(type.IsAssignableToOpenGeneric)) 
{
    strategy = EqualityStrategy.ForceMembers
}
else if (valueTypes.Any(type.IsAssignableToOpenGeneric)) 
{
    strategy = EqualityStrategy.ForceEquals
}
```

dennisdoomen (approved):
Nice!

Issue #1374, opened 2020-08-12
ComparingByMembers<object> lets structural equality assertions always succeed


ursenzler:
### Description

I want to be able to configure FluentAssertions' structural equality so that it uses **ComparingByMembers** for all types.
The background for this is that I have a mix of F# and C# code and all F# records implement Equals by default. In case of an object-graph with a mix starts using `Equals` when hitting the first F# record, as well for the nested C# classes.

I tried `ComparingByMembers<object>`, but that results in the tests always being succeeded. And I guess that should be fixed.
 

### Complete minimal example reproducing the issue

```
using FluentAssertions;
using Xunit;

namespace CSharp.FluentAssertionWrapper
{
    public class Facts
    {
        public struct C
        {
            public D D { get; set; }
        }
        public class D
        {
            public string S { get; set; }
            public int I { get; set; }
        }

        [Fact]
        public void Foo()
        {
            var a = new C { D = new D { S = "a", I = 17 } };
            var b = new C { D = new D { S = "a", I = 18 } };

            a.Should().BeEquivalentTo(   // this should fail, but succeeds
                b,
                config => config.ComparingByMembers<object>());
        }
    }
}
```

### Expected behavior:

All types are structurally compared correctly.

### Actual behavior:

Assertion always succeeds.

### Versions

* Which version of Fluent Assertions are you using? 5.10.3
* Which .NET runtime and version are you targeting? netcoreapp3.1

### Additional Information

None.

ursenzler:
After some debugging through FluentAssertions, I found out that the above test fails when the string values for `S` differ, but not when the integer values for `I` differ.

I think the problem is that when `ComparingByMembers<object>` is used then FluentAssertion tries to compare basic types by their properties. They have none, so they are treated es structurally equal.

Adding a `ComparingByValue<int>` does not work because either FluentAssertion says that the type is already set up to be compared by members (when put below) or because it is ignored (when put ahead).

This leads me to the conclusion that I would need something like: compare all reference types by members.
But as it looks to me, that would be very hard to be implemented in the current code base :-(

dennisdoomen:
Is there a way to detect that a type is F#-specific?

ursenzler:
Yes. From F# code you can use e.g. `FSharpType.IsRecord`

ursenzler:
Another solution for my problem would also be when I could register a callback with the following signature:

[...]

PR #1394, opened 2020-09-26
Exclude primitive types when comparing by members

jnyrup:
As primitive types do not have members, it does not make sense to try to compare them by members.
On all but one of the target frameworks two different primitives are currently incorrectly evaluated as being equivalent (#1374).
On net472 it will even hit `"The maximum recursion depth was reached"`.

This PR excludes primitive types from being compared by members.

Thought:
Should we add a guard check in `ComparingByMembers` if someone writes
`ComparingByMembers<int>()` and wonders why `int`s are still not compared by members?

For those interested in the complete set of primitive types, here are the facts from in .netcoreapp3.0 
```cs
AllTypes
  .From(typeof(int).Assembly)
  .Where(e => e.IsPrimitive)
  .Select(e => e.Name)
  .ToList();
```

```
"Boolean"
"Byte"
"Char"
"Double"
"Int16"
"Int32"
"Int64"
"IntPtr"
"SByte"
"Single"
"UInt16"
"UInt32"
"UInt64"
"UIntPtr"
```

### DIFF
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
@@ -544,6 +544,11 @@ namespace FluentAssertions.Equivalency
         {
             Guard.ThrowIfArgumentIsNull(type, nameof(type));
 
+            if (type.IsPrimitive)
+            {
+                throw new InvalidOperationException($"Cannot compare a primitive type such as {type.Name} by its members");
+            }
+
             if (valueTypes.Any(type.IsSameOrInherits))
             {
                 throw new InvalidOperationException(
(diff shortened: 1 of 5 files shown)


dennisdoomen:
> Should we add a guard check in ComparingByMembers if someone writes
ComparingByMembers<int>() and wonders why ints are still not compared by members?

Yes, I think it should throw an `InvalidOperationException`

dennisdoomen (approved):
Don't we need some update to the docs and release notes?

jnyrup:
🤦 Completely forgot about docs and release notes .

dennisdoomen:
What about https://fluentassertions.com/objectgraphs/#value-types?

jnyrup:
Added notes to the object graphs section.

PR #1401, opened 2020-10-04
Guard  [Not]Match, [Not]MatchEquivalentOf, [Not]MatchRegex and [Not]ContainMatch against null or empty patterns

### LINKED ISSUE #1243: string.NotMatch() and Match() don't fail consistently with nulls
### Description

If you attempt to do NotMatch(null) on a string it will fail when the string is not null

```csharp
// Arrange
string testString = "fdgkdfgkjdfg";

// Assert
testString.Should().NotMatch(null);
```

### Expected behavior:

Test should pass as string is not null

### Actual behavior:

Test fails

### Versions

FluentAssertions 5.10.0
net core 3.1

eNeRGy164:
For `String` assertions, `[Not]Match`, `[Not]MatchEquivalentOf`, `[Not]MatchRegex` and `[Not]ContainMatch` no longer accept a `null` or empty string as parameter.

The exception suggests to add a pattern or use one of the `[Not]BeNull`, `[Not]BeEmpty` or`[Not]Contain` methods if that behavior was intended.

Should also close #1243

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Collections/StringCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/StringCollectionAssertions.cs
@@ -327,6 +327,13 @@ namespace FluentAssertions.Collections
         public AndWhichConstraint<TAssertions, string> ContainMatch(string wildcardPattern, string because = "",
             params object[] becauseArgs)
         {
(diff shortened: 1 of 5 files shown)


jnyrup (approved):
Looks good and really nice with the guidance to users of which assertion method they should rather be using.

dennisdoomen (approved):

dennisdoomen:
> Looks good and really nice with the guidance to users of which assertion method they should rather be using.

I second that. Thanks.

Issue #1404, opened 2020-10-08
When chaining ContainsSingle, assertion failed because of FormatException inside FluentAssertion instead of actual difference


gcelet:
### Description

When chaining `ContainsSingle`, assertion failed because of `FormatException` inside FluentAssertion instead of actual difference.

### Complete minimal example reproducing the issue

```csharp
using FluentAssertions;
using Microsoft.AspNetCore.Mvc.ModelBinding;

	[Test]
	public void Failed_On_FormatException_Inside_FluentAssertions_When_Strings_Are_Different()
	{
		string propertyName = "APropertyName";
		string value = "A value";
		ModelStateDictionary modelState = new ModelStateDictionary();

		modelState.AddModelError(propertyName, $"The value '{value}' is invalid and more characters to be different than expected.");

		modelState.Should().HaveCount(1)
			.And.ContainSingle(kvp => kvp.Key == propertyName)
			.Which.Value.Errors.Should().HaveCount(1)
			.And.ContainSingle(modelError =>
				string.Equals(modelError.ErrorMessage, $"The value '{value}' is invalid."));
	}

	[Test]
	public void Succeed_When_Strings_Are_Same()
	{
		string propertyName = "APropertyName";
		string value = "A value";
		ModelStateDictionary modelState = new ModelStateDictionary();

		modelState.AddModelError(propertyName, $"The value '{value}' is invalid.");

		modelState.Should().HaveCount(1)
			.And.ContainSingle(kvp => kvp.Key == propertyName)
			.Which.Value.Errors.Should().HaveCount(1)
			.And.ContainSingle(modelError =>
				string.Equals(modelError.ErrorMessage, $"The value '{value}' is invalid."));
	}
```

### Expected behavior:

`Failed_On_FormatException_Inside_FluentAssertions_When_Strings_Are_Different` test should failed because actual string is different of expected string.

### Actual behavior:

`Failed_On_FormatException_Inside_FluentAssertions_When_Strings_Are_Different` test failed on `System.FormatException`.

```text
System.FormatException : Index (zero based) must be greater than or equal to zero and less than the size of the argument list.
   at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.Format(String format, Object[] args)
[...]

Issue #1363, opened 2020-07-25
Inherited property cannot be excluded or included when comparing interface collections using runtime types


aayjaychan:
### Description

When a derive class implements an interface through an inherited property, the property cannot be excluded by `.Excluding()` or included by `.Including()` when comparing collections of the interface using runtime types.

### Complete minimal example reproducing the issue

```csharp
public interface Interface
{
    int Value1 { get; set; }
    int Value2 { get; set; }
}

public class Base
{
    public int Value1 { get; set; }
    public int Value2 { get; set; }
}

public class Derived : Base, Interface
{
}

[Test]
public void Exclude()
{
    var actual = new Interface[]
    {
        new Derived() { Value1 = 1, Value2 = 2 }
    };
    var expected = new Interface[]
    {
        new Derived() { Value1 = 999, Value2 = 2 }
    };

    actual.Should().BeEquivalentTo(
        expected,
        options => options
            .Excluding(a => a.Value1)
            .RespectingRuntimeTypes());
}

[Test]
public void Include()
{
    var actual = new Interface[]
    {
        new Derived() { Value1 = 1, Value2 = 2 }
    };
    var expected = new Interface[]
    {
        new Derived() { Value1 = 999, Value2 = 2 }
    };

    actual.Should().BeEquivalentTo(
        expected,
        options => options
            .Including(a => a.Value2)
            .RespectingRuntimeTypes());
}
```

### Expected behavior:

Both tests pass

### Actual behavior:

`Exclude` fails with

```
  Error Message:
   Expected item[0].FromBase to be 999, but found 1.                     

With configuration:                                                      
- Use runtime types and members                                          
- Compare enums by value                                                 
- Include all non-private properties                                     
- Include all non-private fields                                         
- Exclude member root.FromBase                                           
- Match member by name (or throw)                                        
- Be strict about the order of items in byte arrays                      
- Without automatic conversion. 

  Stack Trace:                                                           
     at FluentAssertions.Execution.LateBoundTestFramework.Throw(String message)                                                                   
   at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message)                                                                      
   at FluentAssertions.Execution.CollectingAssertionStrategy.ThrowIfAny(IDictionary`2 context)                                                    
   at FluentAssertions.Equivalency.EquivalencyValidator.AssertEquality(EquivalencyValidationContext context)                                      
[...]

PR #1406, opened 2020-10-10
Issue 1404: Escape brackets for ContainSingle(...)

Serg046:
## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Collections/SelfReferencingCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/SelfReferencingCollectionAssertions.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Text.RegularExpressions;
 using FluentAssertions.Common;
 using FluentAssertions.Execution;
 
@@ -646,8 +647,9 @@ namespace FluentAssertions.Collections
         {
             Guard.ThrowIfArgumentIsNull(predicate, nameof(predicate));
 
+            var body = Regex.Replace(predicate.Body.ToString(), @"{(\d+)}", @"{{$1}}");
             string expectationPrefix =
-                string.Format("Expected {{context:collection}} to contain a single item matching {0}{{reason}}, ", predicate.Body);
+                $"Expected {{context:collection}} to contain a single item matching {body}{{reason}}, ";
 
             if (Subject is null)
             {
--- a/Src/FluentAssertions/Collections/SelfReferencingCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/SelfReferencingCollectionAssertions.cs
@@ -646,22 +646,21 @@ namespace FluentAssertions.Collections
         {
             Guard.ThrowIfArgumentIsNull(predicate, nameof(predicate));
 
-            var body = predicate.Body.ToString().EscapePlaceholders();
             string expectationPrefix =
-                $"Expected {{context:collection}} to contain a single item matching {body}{{reason}}, ";
+                "Expected {context:collection} to contain a single item matching {0}{reason}, ";
 
             if (Subject is null)
             {
                 Execute.Assertion
                     .BecauseOf(because, becauseArgs)
-                    .FailWith(expectationPrefix + "but found {0}.", Subject);
+                    .FailWith(expectationPrefix + "but found {1}.", predicate.Body, Subject);
             }
 
             ICollection<T> actualItems = Subject.ConvertOrCastToCollection();
             Execute.Assertion
                 .ForCondition(actualItems.Any())
                 .BecauseOf(because, becauseArgs)
-                .FailWith(expectationPrefix + "but the collection is empty.");
+                .FailWith(expectationPrefix + "but the collection is empty.", predicate.Body);
 
             T[] matchingElements = actualItems.Where(predicate.Compile()).ToArray();
             int count = matchingElements.Length;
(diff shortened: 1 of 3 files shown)


Serg046:
Fixes #1404

Src/FluentAssertions/Collections/SelfReferencingCollectionAssertions.cs
```diff
@@ -646,8 +647,9 @@ public AndConstraint<TAssertions> NotContain(Expression<Func<T, bool>> predicate
         {
             Guard.ThrowIfArgumentIsNull(predicate, nameof(predicate));
 
+            var body = Regex.Replace(predicate.Body.ToString(), @"{(\d+)}", @"{{$1}}");
```
jnyrup, line 650: We have a `EscapePlaceholders()` extension method on `string`, which seems applicable here to get the same effect.

Serg046: Thanks Jonas, I replaced the call. Just one a smale note - that method doesn't check the number inside the brackets so that we can affect some valid symbols (anyway it shouldn't be a problem).

jnyrup:
@Serg046 Sorry for not getting back to you lately

As described in https://github.com/fluentassertions/fluentassertions/issues/1404#issuecomment-706536639
the proper solution is to avoid the eager formatting of `predicate.Body`.

Would you be willing to update the PR with that approach?
Please let me know if my description is imprecise.

Serg046:
@jnyrup that's done.
I will also look for other places where the issue might occur including your list from the ticket. I will do it in a separate PR to keep this one fixing the original issue. But I will do it a bit later as I already started work on https://github.com/fluentassertions/fluentassertions/issues/1363. Unfortunately the issue is more than just ReflectedType vs DeclaredType and seems there is no a simple and elegant solution.

Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionsSpecs.cs
```diff
@@ -362,6 +362,34 @@ public void When_single_item_matching_a_predicate_is_found_it_should_allow_conti
             act.Should().Throw<XunitException>().WithMessage("Expected*greater*4*2*");
         }
 
+        [Fact]
+        public void When_single_item_contains_brackets_it_should_not_fail()
```
jnyrup, line 366: The method names are a bit inconsistent with the assertions.
We do want the assertions to fail, but not due to a `FormatException`.

```suggestion
        public void When_single_item_contains_brackets_it_should_format_them_properly()
```

docs/_pages/releases.md
```diff
@@ -18,6 +18,7 @@ sidebar:
 **Fixes**
 * Guard against negative precision arguments for `BeCloseTo` and `BeApproximately` - [#1386](https://github.com/fluentassertions/fluentassertions/pull/1386)
 * Guard against implicitly or explicitly trying to compare primitive types by members - [#1394](https://github.com/fluentassertions/fluentassertions/pull/1394).
+* Escaped brackets in expressions passed to ContainSingle(...) - [#1406](https://github.com/fluentassertions/fluentassertions/pull/1406).
```
jnyrup, line 21:
```suggestion
* Fixed formatting of brackets in expressions passed to ContainSingle(...) - [#1406](https://github.com/fluentassertions/fluentassertions/pull/1406). 
```

jnyrup (changes requested):
Thanks for the updates, I just have two small suggestions and then it should be ready to merge.

Serg046:
Renamed the second test (validate it too pls) and applied the notes, thank you 👍

jnyrup (approved):

Src/FluentAssertions/Collections/SelfReferencingCollectionAssertions.cs
```diff
@@ -647,34 +647,34 @@ public AndConstraint<TAssertions> NotContain(Expression<Func<T, bool>> predicate
             Guard.ThrowIfArgumentIsNull(predicate, nameof(predicate));
 
             string expectationPrefix =
-                string.Format("Expected {{context:collection}} to contain a single item matching {0}{{reason}}, ", predicate.Body);
+                "Expected {context:collection} to contain a single item matching {0}{reason}, ";
```
dennisdoomen, line 650: @jnyrup what's wrong with:

```csharp
string expectationPrefix =
                $"Expected {{context:collection}} to contain a single item matching {predicate.Body.ToString().EscapePlaceholders()}{{reason}}, ";
```

Serg046: JFYI guys, I personally support the idea not to format the string twice. So the current approach looks better for me.

dennisdoomen: I get that, but this is what that version gives me:

```Expected collection to contain a single item matching (item == "{123}"), but no such item was found.```

Comment (jnyrup) [reply Src/FluentAssertions/Collections/SelfReferencingCollectionAssertions.cs:650]: Using `EscapePlaceholders` will work in many cases, but e.g. this tests fails with `EscapePlaceholders()` but passes with the current PR changes.
```cs
// Arrange
IEnumerable<string> collection = null;

// Act
Action act = () => collection.Should().ContainSingle(item => item == "{reason}");

// Assert
act.Should().Throw<XunitException>().WithMessage("*{reason}*");
```

It's equivalent to keeping query and input separated in SQL statements.

dennisdoomen (approved):

Issue #1412, opened 2020-10-22
Improve messages for numeric assertions to show the difference of the comparison


augustoproiete:
I'd like to discuss a possible improvement to the error messages on numeric assertions (_which I'm happy to take a stab and send a PR once a design is agreed_):

Given a test similar to this:

```csharp
var expectedResult = 832168464;
var actualResult = 832168447;

actualResult.Should().Be(expectedResult);
```

The error message currently would be:

```
Expected actualResult to be 832168464, but found 832168447.
```

This is great, but given we're comparing numbers, it would be more useful if the error message included the difference, instead of the developer having to calculate it separately from the error message. e.g.

```
Expected actualResult to be 832168464, but found 832168447. Difference: -17
```

The difference when comparing numeric values is usually a good clue when trying to understand why the test failed, and displaying it as part of the error message makes the experience better.

---

Is there a desire to have something like that implemented?

dennisdoomen:
Sure, but it'll have to be done on all numeric assertions.

augustoproiete:
My initial thinking is to make `NumericAssertions<T>` and `NullableNumericAssertions<T>` classes abstract and create individual concrete classes for each numeric primitive supported today (inheriting from `NA<T>` and `NNA<T>` respectively) so that we can perform a statically typed comparison without having to compile expressions or use reflection.

- `NumericAssertions<T>`
  - `ByteAssertions : NumericAssertions<byte>`
  - `DecimalAssertions : NumericAssertions<decimal>`
  - `DoubleAssertions : NumericAssertions<double>`
  - `FloatAssertions : NumericAssertions<float>`
  - `IntAssertions : NumericAssertions<int>`
  - `LongAssertions : NumericAssertions<long>`
  - `SByteAssertions : NumericAssertions<sbyte>`
  - `ShortAssertions : NumericAssertions<short>`
  - `UIntAssertions : NumericAssertions<int>`
  - `ULongAssertions : NumericAssertions<ulong>`
  - `UShortAssertions : NumericAssertions<ushot>`

- `NullableNumericAssertions<T>`
  - `NullableByteAssertions : NullableNumericAssertions<byte>`
  - `NullableDecimalAssertions : NullableNumericAssertions<decimal>`
  - `NullableDoubleAssertions : NullableNumericAssertions<double>`
  - `NullableFloatAssertions : NullableNumericAssertions<float>`
  - `NullableIntAssertions : NullableNumericAssertions<int>`
[...]

PR #1414, opened 2020-10-26
Update Numeric Assertions messages to display the difference between Actual vs Expected

augustoproiete:
Initial spike on #1412 that includes difference between actual and expected for `NumericAssertions` & `NullableNumericAssertions` for:

- `Be`
- `BeLessThan`
- `BeLessOrEqualTo`
- `BeGreaterThan`
- `BeGreaterOrEqualTo`

---

Some possible items to be discussed:

1. `BeLessThan` and `BeGreaterThan` shows `actual - expected` as difference for consistency with the other assertions. It may make sense to subtract `1` for `BeLessThan` and add `1` for `BeGreaterThan` to show a value that satisfy the condition

2. `BeInRange` and `NotBeInRange` could display the difference between minimum value or maximum value depending on the side of the range if falls (or display both)

3. `BeOneOf` could display the difference of each element in the list of expected values

4. `NotBe`, `BePositive`, `BeNegative`, `BeOfType`, `NotBeOfType`, `Match` wouldn't have any difference to display

### DIFF
--- a/Src/FluentAssertions/AssertionExtensions.cs
+++ b/Src/FluentAssertions/AssertionExtensions.cs
@@ -381,7 +381,7 @@ namespace FluentAssertions
         [Pure]
         public static NumericAssertions<int> Should(this int actualValue)
         {
-            return new NumericAssertions<int>(actualValue);
+            return new IntAssertions(actualValue);
         }
 
         /// <summary>
@@ -391,7 +391,7 @@ namespace FluentAssertions
         [Pure]
         public static NullableNumericAssertions<int> Should(this int? actualValue)
         {
-            return new NullableNumericAssertions<int>(actualValue);
+            return new NullableIntAssertions(actualValue);
         }
 
         /// <summary>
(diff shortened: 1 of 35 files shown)


dennisdoomen:
Looks promising. Some suggested improvements to the failure messages.:

> Expected item to be 2, but found 1. Difference -1

`Expected item to be 2, but found 1 (which differens by -1)`

> "Expected*2, but found <null>. Difference <null>.

In that case, I would not list a difference. 

> BeOneOf could display the difference of each element in the list of expected values

I don't think that's going to be useful. Quite often this one is used to test against enums. 

> BeInRange and NotBeInRange could display the difference between minimum value or maximum value depending on the side of the range if falls (or display both)

I like that. 

> NotBe, BePositive, BeNegative, BeOfType, NotBeOfType, Match wouldn't have any difference to display

Agreed.

jnyrup:
I agree with Dennis' observations.
Just changing "differens" to "differs".

dennisdoomen:
Any progress on this?

augustoproiete:
Thanks for the nudge @dennisdoomen. I'll pick this one back up in the new year. Merry Christmas guys!! 🎅🏻🎄🎁

dennisdoomen:
@augustoproiete did you give up on us?

augustoproiete:
Sorry @dennisdoomen life got in the way. If I ever manage to come back to this PR I'll reopen or submit a new one.

PR #1379, opened 2020-08-16
Improved reporting of properties and field details in the equivalency API

dennisdoomen:
The equivalency assertions will now include the type of the member and whether it involves a field or property.

This required significant refactorings:
* The properties of `IEquivalencyValidationContext` have been combined into the `INode` hierarchy and the two `Reason` and `Tracer` classes
* Moved a lot of logic related to the object graph location into the `Node`, `Field` and `Property` classes.
* The `SelectedMemberInfo` class has been removed, since it main user, `IMemberInfo` has been flattened.
* Several methods that took an `IMemberInfo`, but could also act on other objects than a property or field now take an `IObjectInfo`

Because the final design evolved while doing the refactorings, I had to give up on nice and clean commits. 

* [x] Rename the SelectedMemberInfo hierarchy to `IMember`, Field and Property
* [x] Move the member path and description into `IMember`
* [x] Combine the because and arguments into a reason class
* [x] Move the tracing parts from the context class to a dedicated class
* [x] See if we can remove `RootIsCollection` from `IEquivalencyValidationContext`
* [x] Move `IsRoot` into `IMember`
* [x] Check out the weird usage in `CollectionMemberAssertionRuleDecorator`.
* [x] Why does `IEquivalencyValidationContext` inherit from `IMemberInfo`
* [x] Ensure `INode.Path` has no public `set`ter
* [x] Refactor `EquivalencyValidationContext.AsNestedMember`
* [x] Refactor some of the code that gets or sets the `IMember` properties into more functional methods
* [x] Does `IMember` really need `GetValue` and the access modifier getters?
* [x] Derive `INode.Description` from its state instead of setting it externally
* [x] Figure out why `Tracer.WriteLine` needs to rewrite the path
* [x] Is `INode.Clone` really necessary
* [x] Why do we need `WithoutInitialIndexQualifier`?
* [x] Check that tracing works as expected
* [x] Does the Node hierarchy need GetHashCode/Equals
* [x] Check review comments from Jonas
* [x] Ensure all new members/types are documented
* [x] Add release notes

Follow-up PR
* [ ] Move depth calculation inside ShouldCompareMembersThisDeep into INode
* [ ] Move the `GetExpectationType` calculation into `IEquivalencyContext.ExpectedType`.

PR #1415, opened 2020-10-27
Issue 1363

Serg046:
## IMPORTANT 

Issue #1363

### DIFF
--- a/Src/FluentAssertions/Equivalency/MemberInfoSelectedMemberInfo.cs
+++ b/Src/FluentAssertions/Equivalency/MemberInfoSelectedMemberInfo.cs
@@ -19,6 +19,8 @@ namespace FluentAssertions.Equivalency
 
         public override Type DeclaringType => memberInfo.DeclaringType;
 
+        public override Type ReflectedType => memberInfo.ReflectedType;
+
         protected bool Equals(MemberInfoSelectedMemberInfo other)
         {
             return memberInfo.Equals(other.memberInfo);
--- a/Src/FluentAssertions/Equivalency/SelectedMemberInfo.cs
+++ b/Src/FluentAssertions/Equivalency/SelectedMemberInfo.cs
@@ -45,6 +45,11 @@ namespace FluentAssertions.Equivalency
         public abstract Type DeclaringType { get; }
 
         /// <summary>
+        /// Gets the base class of the class that declares this member.
+        /// </summary>
+        public abstract Type ReflectedType { get; }
+
+        /// <summary>
         /// Gets the access modifier for the getter of this member.
         /// </summary>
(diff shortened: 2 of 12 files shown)


Serg046:
We cannot just change DeclaringType to ReflectedType because of the following test (and Excluding version):
```csharp
[Fact]
public void When_including_a_property_that_is_hidden_in_a_derived_class_it_should_select_the_correct_one()
{
    // Arrange
    var b1 = new ClassThatHidesBaseClassProperty();
    var b2 = new ClassThatHidesBaseClassProperty();

    // Act / Assert
    b1.Should().BeEquivalentTo(b2, config => config.Including(b => b.Property));
}

private class ClassWithGuidProperty
{
    public string Property { get; set; } = Guid.NewGuid().ToString();
}

private class ClassThatHidesBaseClassProperty : ClassWithGuidProperty
{
    public new string[] Property { get; set; }
}
```
So the idea is the same as in the following existing method:
```csharp
public static PropertyInfo FindProperty(this Type type, string propertyName, Type preferredType)
{
    List<PropertyInfo> properties =
        type.GetProperties(PublicMembersFlag)
            .Where(pi => pi.Name == propertyName)
            .ToList();

    return properties.Count > 1
        ? properties.SingleOrDefault(p => p.PropertyType == preferredType)
        : properties.SingleOrDefault();
}
```
Below is what the fix does for the existing test (the first example of the comment)
1. We get the members - `string[] Property` and `string Property`.
2. We group the members with the same name.
3. When the member is unique, it is processed using the same way.
4. Otherwise, we look for the best match (**preferred** member) and process it, other members with that name are excluded from calculations. The best match is when the type of the property and CompileTimeType are the same.

dennisdoomen:
Thanks for this fix. I'm going to have to postpone this for a week or two, because I'm finalizing some [significant refactoring](https://github.com/fluentassertions/fluentassertions/pull/1379) of the internal equivalency API. What would really help us if you would rebase your changes on that PR. I'm not expecting any major changes anymore.

Serg046:
Sure, I will do it a bit later.
But anyway, not sure about the idea of the fix, it is maybe not the best solution here.

dennisdoomen:
> Otherwise, we look for the best match (preferred member) and process it, other members with that name are excluded from calculations. The best match is when the type of the property and CompileTimeType are the same.

This will work for the top-level members of the graph, but as soon as FA dives into the graph, the `CompileTimeType` ends up being the run-time type of the nested members.

Serg046:
Got it, I will take it into account when I will be moving the PR. Anyway, the same idea will work using ReflectedType which was added. I will just need to make it accessible somehow in the place where I need it. Thanks for the note.

PR #1417, opened 2020-10-28
Make IgnoringCyclicReferences working for the types overriding Equals()

### LINKED ISSUE #1370: Combination of ComparingByMember and IgnoreCyclicReferences doesn't work
### Description

If I keep it short: having override `Equals` on objects I have to disable using it with `ComparingByMembers`. But if I do so, the `IgnoreCyclicReferences` seem to stop working.

Here the details.

We have written a lot of NHibernate persistence tests, where we are using FluentAssertions `BeEquivalentOf`.
After adding a recommended for NH Equals-override in the DomainBase class we've got multiple issues with FluentAssertions.
We had to add a Configuration option `ComparingByMembers<DomainBase>()` to prevent FluentAssertions using Equals-override for comparing objects equivalence. And it worked nice for the most of cases, but has introduced another issue - now FluentAssertions seem to ignore the option `IgnoreCyclicReferences()`

Here is the minimal example I could make to reproduce the issue:

```csharp
[TestFixture]
public class FluentAssertionsIssueTest {
    public class Base {
        public Guid Id { get; set; }
        // NHibernate recommends overriding Equals for preventing issues with Sets and while working with multiple sessions, as well as issues with proxies in collections
        public override bool Equals(object obj)  {
            if (ReferenceEquals(this, obj))
                return true;
            return (obj is Base baseObj) && baseObj.Id == Id;
        }
    }

    public class Item : Base {
        public string Title { get; set; }
        public Item Previous { get; set; }
        public Item Next { get; set; }
    }

    [Test]
    public void IgnoreCyclicReferences_Broken()  {
        // imitating NHibernate save in one session
        var first = new Item() {Id = Guid.NewGuid(), Title = "First"};
        var second = new Item() {Id = Guid.NewGuid(), Title = "Second", Previous = first};
        first.Next = second;

        // imitating NHibernate load in another session
        var firstCopy = new Item() {Id = first.Id, Title = first.Title};
        var secondCopy = new Item() {Id = second.Id, Title = second.Title, Previous = firstCopy};
        firstCopy.Next = secondCopy;

        // I'd like to check, that all the properties could be persisted correctly.

        // "The maximum recursion depth was reached" is generated here (IgnoringCyclicReferences doesn't seem to work)
        firstCopy.Should().BeEquivalentTo(first, opt => opt.ComparingByMembers<Base>()
            .IgnoringCyclicReferences(), "they are equivalent");
        first.Title = "Changed";
        //  Removing Equals-override from Base class fixes FA issues, but introduces NHibernate-issues
        //  Removing ComparingByMembers-option removes "The maximum recursion depth was reached", but breaks this check (changed Title is not considered)
        firstCopy.Should().NotBeEquivalentTo(first, opt => opt.ComparingByMembers<Base>()
            .IgnoringCyclicReferences(), "after change");
    }
}
```

### Expected behavior:

Deep object comparison should be done, handling cyclic references correctly. The test should pass

### Actual behavior:

I cannot use deep object comparison on objects with overridden `Equals` and cyclic references. 
The test fails with message:
```
The maximum recursion depth was reached.  
The maximum recursion depth was reached.  
The maximum recursion depth was reached.  
The maximum recursion depth was reached.  

With configuration:
- Compare enums by value
- Ignoring cyclic references

   bei FluentAssertions.Execution.LateBoundTestFramework.Throw(String message)
   bei FluentAssertions.Execution.CollectingAssertionStrategy.ThrowIfAny(IDictionary`2 context)
   bei FluentAssertions.Equivalency.EquivalencyValidator.AssertEquality(EquivalencyValidationContext context)
```

If I comment out `override Equals` everything works in FA, but introduces issues with NHibernate.
If I remove options `ComparingByMembers<Base>()` in both `BeEquivalentTo` / `NotBeEquivalentTo`, then the difference of Title in firstCopy and first is not detected in the `NotBeEquivalentTo`-Check.

### Versions

* 5.10.3, 5.10.2
* Using .NET framework 4.6.1

Serg046:
## IMPORTANT 

Fixes #1370

### DIFF
--- a/Src/FluentAssertions/Common/InternalTypeExtensions.cs
+++ b/Src/FluentAssertions/Common/InternalTypeExtensions.cs
@@ -43,7 +43,7 @@ namespace FluentAssertions.Common
         /// </summary>
         public static bool HasValueSemantics(this Type type)
         {
-            return type.OverridesEquals() &&
+            return type.IsSimpleType() &&
                    !type.IsAnonymousType() && !type.IsTuple() && !IsKeyValuePair(type);
         }
 
--- a/Src/FluentAssertions/Common/TypeExtensions.cs
+++ b/Src/FluentAssertions/Common/TypeExtensions.cs
@@ -143,15 +143,6 @@ namespace FluentAssertions.Common
                     .ToArray();
         }
 
-        public static bool OverridesEquals(this Type type)
-        {
-            MethodInfo method = type
-                .GetMethod("Equals", new[] { typeof(object) });
-
-            return method != null
-                && method.GetBaseDefinition().DeclaringType != method.DeclaringType;
-        }
-
         /// <summary>
         /// Finds a member by its case-sensitive name.
         /// </summary>
@@ -466,6 +457,31 @@ namespace FluentAssertions.Common
             bool IsExactNamespace() => IsNamespacePrefix() && type.Namespace.Length == @namespace.Length;
             bool IsParentNamespace() => IsNamespacePrefix() && type.Namespace[@namespace.Length] == '.';
             bool IsNamespacePrefix() => type.Namespace?.StartsWith(@namespace, StringComparison.Ordinal) == true;
+        }
+
+        /// <summary>
+        /// Returns true if the type is a primitive type, date, decimal, string, or GUID
+        /// </summary>
+        /// <param name="type">Type to be checked</param>
+        /// <returns></returns>
+        public static bool IsSimpleType(this Type type)
+        {
+            if (type == null)
+            {
+                return false;
+            }
+
+            if (type.GetTypeInfo().IsGenericType && type.GetTypeInfo().GetGenericTypeDefinition() == typeof(Nullable<>))
+            {
+                type = Nullable.GetUnderlyingType(type);
+            }
+
+            return type.GetTypeInfo().IsPrimitive
+                   || type == typeof(DateTime)
+                   || type == typeof(string)
+                   || type == typeof(Guid)
+                   || type == typeof(decimal)
+                   || type.GetTypeInfo().IsEnum;
         }
     }
 }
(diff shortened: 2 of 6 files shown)


Serg046:
Implemented what Compare-Net-Objects use. They use a set of ordered comparers where each the comparer has a check if the type is applicable or not. The comparer which will be in use is the first one passing the check.

 This is their [SimpleTypeComparer](https://github.com/GregFinzer/Compare-Net-Objects/blob/master/Compare-NET-Objects/TypeComparers/SimpleTypeComparer.cs) which is ordered right before ClassComparer and this is its [TypeHelper.IsSimpleType](https://github.com/GregFinzer/Compare-Net-Objects/blob/master/Compare-NET-Objects/TypeHelper.cs#L367-L384).

Serg046:
Re-opening this to show you the idea. 3 tests are failing because `System.Type` and `UriBuilder` are not simple types anymore. So the solution could be trying to extend that hardcoded types dictionary in the change. Anyway, even if we don't try to merge something like this, we will still have the issue described by this test (fails on develop):

```csharp
[Fact]
public void Test()
{
    // Arrange
    var valueObject = new ValueObject("value");
    var valueObjectDto = new ValueObjectDto("value");

    // Act
    Action act = () => valueObject.Should().BeEquivalentTo(valueObjectDto);

    // Assert
    act.Should().NotThrow();
}

public class ValueObject
{
    public ValueObject(string value)
    {
        Value = value;
    }

    public string Value { get; }

    public override bool Equals(object obj)
    {
        return ((ValueObject)obj).Value == Value;
    }

    {
        return Value.GetHashCode();
    }
}

public class ValueObjectDto
{
    public ValueObjectDto(string value)
    {
        Value = value;
    }

    public string Value { get; }

    public override bool Equals(object obj)
    {
        return ((ValueObjectDto)obj).Value == Value;
    }

    {
        return Value.GetHashCode();
    }
}
```

dennisdoomen:
I think you're approaching this from the wrong angle. The basic design principle FA should use is that anything that overrides `Equals` should be treated as value types. The bug that #1370 concludes with is that the cyclic reference detector is using a different algorithm.

Serg046:
Yes but if we keep the same approach for value types we will have this issue I described above. That issue is not related to cyclic references at all and looks quite important imo. Maybe I should open a separate issue for that.
Closing the PR again as it got some attention.

PR #1418, opened 2020-10-29
Change AttributeBasedFormatter to allow custom formatter selection ba…

### LINKED ISSUE #309: Allow custom formatters to "kick in" for more specific types
Custom formatters (I am referring to Formatters section in https://github.com/dennisdoomen/fluentassertions/wiki#extensibility) are right now selected by exact match of types:

https://github.com/dennisdoomen/fluentassertions/blob/develop/Src/Core/Formatting/AttributeBasedFormatter.cs#L63

Given #308 I wonder if it would be better to go by `isAssignableFrom()` instead? It would then be possible to implement a _catch all_ custom formatter with `object` as an argument and completely disable the default formatter that builds the object's graph.

Or maybe introduce an interface that formatters would need to implement to give them (the formatters) the ability to respond to `CanHandle()` ?

Serg046:
…sed on the parent type

## IMPORTANT 

Fixes #309


Src/FluentAssertions/Formatting/AttributeBasedFormatter.cs
```diff
@@ -46,9 +46,34 @@ public string Format(object value, FormattingContext context, FormatChild format
...
             Type valueType = value.GetType();
-            MethodInfo formatter = Formatters.FirstOrDefault(m => m.GetParameters().Single().ParameterType == valueType);
+            var possibleFormatters = Formatters
+                .Select(m => new
+                {
+                    Type = m.GetParameters().Single().ParameterType,
+                    Method = m
+                })
+                .Where(f => valueType.IsSameOrInherits(f.Type))
+                .ToDictionary(f => f.Type, f => f.Method);
```
dennisdoomen, line 56: 🤔 Is there a need to order the formatters in some way to get some deterministic order?

Serg046: That would require multiple enumerations and I guess wouldn't improve readability.
One enumeration using do/while looks more efficient and readable to me.

dennisdoomen: Imagine multiple classes implementing the same interface `IFoo`. What would happen if you two formatters, one that accepts a `IFoo` and one, more specialized, that takes an implementation of `IFoo`.

jnyrup: @Serg046 if efficiency is a concern here, could you then cache the `Dictionary` just as `formatters` is?

Serg046: Did it but forced to change one existing test a bit because of cache.

dennisdoomen (approved):

Src/FluentAssertions/Formatting/AttributeBasedFormatter.cs
```diff
@@ -46,11 +47,45 @@ public string Format(object value, FormattingContext context, FormatChild format
...
+            {
+                var possibleFormatters = Formatters
+                    .Select(m => new { Type = m.GetParameters().Single().ParameterType, Method = m })
+                    .GroupBy(f => f.Type)
+                    .Select(g => g.First())
+                    .Where(f => valueType.IsSameOrInherits(f.Type))
+                    .ToDictionary(f => f.Type, f => f.Method);
+
+                formatter = GetFormatter(possibleFormatters, valueType);
+                determinedFormatters.Add(valueType, formatter);
```
jnyrup, line 60: Sorry for being too vague on my thoughts about caching.

As `AttributeBasedFormatter` is instantiated a single time in the static `Formatter.DefaultFormatters` we cannot use `Dictionary<TKey, TValue>` as a cache as `Add` is not thread-safe.

My idea is to change 
```cs
private MethodInfo[] formatters;
``` 
into
```cs
private Dictionary<Type, MethodInfo> formatters;
```

```cs
private Dictionary<Type, MethodInfo> Formatters
{
    get
    {
        HandleValueFormatterDetectionModeChanges();

        return formatters ??= GetDictionary(FindCustomFormatters());
    }
}

// with a better name or merge into FindCustomFormatters
private static Dictionary<Type, MethodInfo> GetDictionary(MethodInfo[] formatters)
{
    return formatters
        .Select(m => new { Type = m.GetParameters().Single().ParameterType, Method = m })
        .GroupBy(f => f.Type)
        .Select(g => g.First())
        .ToDictionary(f => f.Type, f => f.Method);
}
```

```cs
private MethodInfo GetFormatter(object value)
{
    Type valueType = value.GetType();
    return GetFormatter(valueType);
}

private MethodInfo GetFormatter(Type valueType)
{
    do
    {
        if (Formatters.TryGetValue(valueType, out var formatter))
        {
            return formatter;
        }

        valueType = valueType.BaseType;
    }
    while (valueType != null);

    return null;
}
```

Serg046: Thanks @jnyrup, got it.
You miss `IsSameOrInherits` check in your example. How do you suggest including it?

jnyrup: I deliberately ignored `IsSameOrInherits` 😀 
I might be missing something, but I couldn't see it was needed?

Serg046: There was a glitch in my head, I was confused by `possibleFormatters.Count == 1` when we didn't need to analyze type. But I don't think this is really needed. Finally pushed your example.
Just wondering why AttributeBasedFormatter is public.

Tests/FluentAssertions.Specs/Formatting/FormatterSpecs.cs
```diff
@@ -586,22 +586,79 @@ public void When_a_custom_formatter_exists_in_any_loaded_assembly_it_should_over
...
+
+            // Act
+            string result = Formatter.ToString(subject);
+
+            // Assert
+            result.Should().Be("Property is SomeValue", "it should use my custom formatter");
+        }
+
         [Fact]
         public void When_no_custom_formatter_exists_in_the_specified_assembly_it_should_use_the_default()
```
jnyrup, line 644: I'm not sure why this method is changed?

Serg046: Reverted it back, thank you...
It was needed because of old cache.

jnyrup: Thanks for the clarification 👍

jnyrup (approved):
@dennisdoomen Are you fine with the API breakage?

dennisdoomen:
Which breakage are you talking about?

Serg046:
AttributeBasedFormatter is public.
Since we are changing `Formatter` from array to dictionary anyway, I changed the member to private.

jnyrup:
@dennisdoomen 
> Which breakage are you talking about?

Removing `Formatters` from the public type `AttributeBasedFormatter`.
The property has been public since the class was introduced, but it seems more of an implementation detail.

dennisdoomen:
I'm fine with that.

jnyrup:
@Serg046 As a last thing, can I ask you to add a line to `releases.md`?

jnyrup:
@Serg046 Thanks for another nice  contribution and for going through the iterations it took to get to this result.

PR #1421, opened 2020-10-30
Fix Contain, NotContain and OnlyContain to avoid multiple enumeration…

### LINKED ISSUE #327: Linq2Sql entity - default value formatter runs into SQL queries
For beginners it might not be obvious why a simple `myEntityCollection.Should().NotContain(entity)` executes SQL queries and (for bigger tables) takes forever due to the N+1 query problem.

The problem is that the `DefaultValueFormatter` tries to build an output string over many nesting levels of public properties in case `ToString()` is not overloaded (the default case for generated entites in Linq2Sql and afaik also for EntityFramework).

It would probably be best to always use `ToString()` and to enforce that the user explicitly must enable the `DefaultValueFormatter`'s Public-Property-Via-Reflection-ToString behavior.

What are your thoughts on this issue?

BR, D.R.

Serg046:
…s when the condition is false

## IMPORTANT 

Fixes #327

### DIFF
--- a/Src/FluentAssertions/Collections/SelfReferencingCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/SelfReferencingCollectionAssertions.cs
@@ -422,15 +422,16 @@ namespace FluentAssertions.Collections
                     .FailWith("Expected {context:collection} to contain {0}{reason}, but found {1}.", expected, Subject);
             }
 
-            if (!Subject.Contains(expected))
+            var collection = Subject.ConvertOrCastToCollection();
+            if (!collection.Contains(expected))
             {
                 Execute.Assertion
                     .BecauseOf(because, becauseArgs)
-                    .FailWith("Expected {context:collection} {0} to contain {1}{reason}.", Subject, expected);
+                    .FailWith("Expected {context:collection} {0} to contain {1}{reason}.", collection, expected);
             }
 
             return new AndWhichConstraint<TAssertions, T>((TAssertions)this,
-                Subject.Where(
+                collection.Where(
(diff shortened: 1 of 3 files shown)


jnyrup (approved):

jnyrup:
@Serg046 You're unstoppable! 👍 

I assume this problem also occurs for other assertions in `SelfReferencingCollectionAssertions`?
E.g. `HaveCount` which invokes `Count()` and passe on `Subject` to `FailWith`.

Serg046:
@jnyrup, yes, I can imagine that this problem exists in many place. I'd materialize the subject right after `Should()` call somehow so that we don't have the problem at all. But it will be a huge change that should be done separately imo.

Regarding this PR, I am just not sure how many places we should fix here. Because if we now try to find all the places, then add the tests for all of them, and finally decide to materialize the subject on an upper layer, we will just lose our prev. work. Please advise how many places this PR should affect, maybe we just check the whole `SelfReferencingCollectionAssertions` file for now?

dennisdoomen (approved):

jnyrup:
@Serg046 We avoid doing any tranformations on the input and just store it directly on `Subject`, such that e.g. `value.Should().BeSameAs(value)` behaves as expected.

If I understand your proposal, you're suggesting caching the materialized collection on `SelfReferencingCollectionAssertions` it self, to avoid enumerating the input multiple times even for chained assertions.
```cs
myEntityCollection.Should()
    .NotBeEmpty() <-- materialize and cache `Subject`
    .And.Contain(entity) <-- reuse cached collection
    .And.NotContain(otherEntity);  <-- reuse cached collection
```

That approach sounds good to me.

PR #1419, opened 2020-10-30
Add support for asserting on DataSet, DataTable, DataColumn, DataRow

logiclrd:
I tried asserting equivalency on `DataSet` objects and the results were .. dissatisfying. :-)

This PR:

- Adds equivalency steps for `DataSet` and related types.
- Registers these steps in `GetDefaultSteps` in `EquivalencyStepCollection.cs`.
- Adds new overloads of `.Should()` in `AssertionExtensions.cs` drive by new subclasses of `ReferenceTypeAssertions<>`.
- Adds comprehensive unit testing of the new functionality.
- Adds documentation of the new functionality for the web site.
- Updates the ApprovedApi files to include the newly-added API surface.
- Adds mention of the new feature to the release notes on the Releases page.

## IMPORTANT 

* [x] The code complies with the [Coding Guidelines for C#](https://www.csharpcodingguidelines.com/).
* [x] The changes are covered by a new or existing set of unit tests which follow the Arrange-Act-Assert syntax such as is used [in this example](https://github.com/fluentassertions/fluentassertions/blob/daaf35b9b59b622c96d0c034e8972a020b2bee55/Tests/FluentAssertions.Shared.Specs/BasicEquivalencySpecs.cs#L33).
* [x] If the contribution adds a feature or fixes a bug, please update [**the release notes**](https://github.com/fluentassertions/fluentassertions/tree/master/docs), which are published on the [website](https://fluentassertions.com/releases).
* [x] If the contribution changes the public API the changes needs to be included by running `AcceptApiChanges.ps1`/`AcceptApiChanges.sh`.
* [x] If the contribution affects [the documentation](https://github.com/fluentassertions/fluentassertions/tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).

PR #1434, opened 2020-11-17
Performance improvements made in response to profiling

logiclrd:
This PR makes some performance improvements based on profiling with #1419's changes, testing DataTable objects with many records:

- Updated TypeExtensions.cs to short-circuit GetClosedGenericInterfaces if the type isn't an object type.
- Updated AssertionScope.cs to support lazily resolving the context string, storing a Lazy<string> promise instead for as long as needed. Added internal method ResolveContext() to be used by AssertionRuleEquivalencyStep.cs before recursing, so that the rule doesn't get picked up as the context. Updated EquivalencyValidator.cs to set a promise rather than actually computing the Context value on-the-spot every time a new scope is created.
- Updated ContextDataItems.cs to replace DataItems inline rather than removing/re-adding list elements. This means that the list won't always have the most recently added item at the back, but that doesn't seem to be important. Didn't break any tests, anyway.
- Updated CyclicReferenceDetector.cs to use a Dictionary keyed on the target object instead of a List for its object references, so that when the number of objects gets extremely large, performance doesn't suffer due to linear scans. Added an internal Object accessor to ObjectReference.cs.

## IMPORTANT 

* [x] ~If the contribution adds a feature or fixes a bug, please update [**the release notes**](https://github.com/fluentassertions/fluentassertions/tree/master/docs), which are published on the [website](https://fluentassertions.com/releases).~
* [x] ~If the contribution changes the public API the changes needs to be included by running `AcceptApiChanges.ps1`/`AcceptApiChanges.sh`.~
* [x] ~If the contribution affects [the documentation](https://github.com/fluentassertions/fluentassertions/tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).~


logiclrd:
Oops, I picked the branches onto an out-of-date `develop`. Sigh. Fixing.

lg2de:
Would be great to have some measurement on the improvements, e.g. using BenchmarkDotNet with example data before and after these changes.

logiclrd:
That's reasonable. These changes were originally a part of #1419, in which I ran into severe performance issues when trying to work with very large DataTable objects. The two main performance killers were that:

1. Even though it was never meant to, the code is rediscovering the source line of the context every time the context changes, _not_ only when it's about to report a failure. This meant that if the object graph had a million items in it, it was parsing the stack trace and then scanning the source code to find the expression at the line/column in the first non-FluentAssertions frame over and over, exactly the same result each time, a million times.

2. The cyclic reference checking code uses a linear scan for finding matching objects. This causes an O(n^2) performance characteristic, which becomes noticeable with very large object graphs.

I can't remember exactly what DataTable size I did the calculations with, I believe it was 100,000 rows, but with the changes here I observed a 60-fold increase in performance and also confirmed a performance characteristic better than O(n^2).

logiclrd:
The branch behind #1419 includes a profiling fixture based on `DataTable` (which isn't supported properly without those changes). From that profiling fixture:
```
|         Method | RowCount |          Mean |       Error |      StdDev |        Gen 0 |        Gen 1 | Gen 2 |   Allocated |
|--------------- |--------- |--------------:|------------:|------------:|-------------:|-------------:|------:|------------:|
| BeEquivalentTo |       10 |      2.445 ms |   0.0322 ms |   0.0269 ms |     289.0625 |       3.9063 |     - |     1.76 MB |
| BeEquivalentTo |      100 |      7.733 ms |   0.1338 ms |   0.2412 ms |    1007.8125 |      23.4375 |     - |     6.08 MB |
| BeEquivalentTo |     1000 |     55.819 ms |   1.0056 ms |   0.8915 ms |    8111.1111 |     222.2222 |     - |       49 MB |
| BeEquivalentTo |    10000 |    539.920 ms |   5.4939 ms |   4.5877 ms |   79000.0000 |   11000.0000 |     - |   478.04 MB |
| BeEquivalentTo |   100000 |  5,586.120 ms | 105.3597 ms | 112.7337 ms |  793000.0000 |  114000.0000 |     - |  4767.03 MB |
| BeEquivalentTo |  1000000 | 54,524.512 ms | 447.0123 ms | 418.1356 ms | 7953000.0000 | 1137000.0000 |     - | 47793.88 MB |
```
I feel like this could be even faster, but I don't know the codebase well enough to do it. But, this is a heck of a lot better than literally half an hour to compare two large `DataTable` instances. :-P

logiclrd:
I have added a performance benchmark that sets up large objects graphs with a range of sizes and passes them to `.BeEquivalentTo`. I am running it now on the pre-improvements code. I'm not sure how far it'll get, though, because at a graph size of 2,583 objects, it's taking approximately 1.8 _seconds_ per `BeEquivalentTo` call. This benchmark runs up to object graph sizes exceeding 3 million objects. :-P

I'll let it run for a bit and then show what results it has managed to achieve, then I'll run it with the performance changes.

logiclrd:
Old code:
```
N = 16 ; Graph size: 2583 objects

Mean = 1.807 s, StdErr = 0.009 s (0.50%), N = 59, StdDev = 0.070 s
Min = 1.707 s, Q1 = 1.761 s, Median = 1.788 s, Q3 = 1.820 s, Max = 2.059 s
IQR = 0.059 s, LowerFence = 1.672 s, UpperFence = 1.909 s
ConfidenceInterval = [1.775 s; 1.838 s] (CI 99.9%), Margin = 0.031 s (1.74% of Mean)
Skewness = 1.6, Kurtosis = 5.58, MValue = 2

N = 18 ; Graph size: 6764 objects

Mean = 5.152 s, StdErr = 0.026 s (0.51%), N = 19, StdDev = 0.114 s
Min = 4.985 s, Q1 = 5.062 s, Median = 5.140 s, Q3 = 5.220 s, Max = 5.364 s
IQR = 0.158 s, LowerFence = 4.825 s, UpperFence = 5.457 s
ConfidenceInterval = [5.050 s; 5.254 s] (CI 99.9%), Margin = 0.102 s (1.99% of Mean)
Skewness = 0.42, Kurtosis = 1.97, MValue = 2

N = 20 ; Graph size: 17710 objects

WorkloadWarmup   1: 1 op, 16910015000.00 ns, 16.9100 s/op
WorkloadWarmup   2: 1 op, 16080022500.00 ns, 16.0800 s/op
WorkloadWarmup   3: 1 op, 16481630700.00 ns, 16.4816 s/op
WorkloadWarmup   4: 1 op, 16336174700.00 ns, 16.3362 s/op
WorkloadWarmup   5: 1 op, 15959460200.00 ns, 15.9595 s/op
...
```

At these scales, the O(n^2) characteristic of the cyclic reference check isn't terribly obvious because the results are utterly dominated by the repeated computation of the subject (with FluentAssertions does by locating & inspecting the caller's source code). But, when I crunch the numbers further down, it can be seen to be growing noticeably faster than the number of objects.

New code:
```
|         Method |  N |         Mean |      Error |     StdDev |        Gen 0 |       Gen 1 |     Gen 2 |   Allocated |
|--------------- |--- |-------------:|-----------:|-----------:|-------------:|------------:|----------:|------------:|
| BeEquivalentTo | 16 |     52.18 ms |   1.018 ms |   0.952 ms |    8000.0000 |   2000.0000 |  571.4286 |    48.59 MB |
| BeEquivalentTo | 18 |    139.02 ms |   2.726 ms |   3.348 ms |   21250.0000 |   5250.0000 | 1000.0000 |   128.38 MB |
| BeEquivalentTo | 20 |    370.64 ms |   6.778 ms |   6.008 ms |   57000.0000 |  15000.0000 | 1000.0000 |   339.28 MB |
| BeEquivalentTo | 24 |  2,603.46 ms |  50.658 ms |  62.212 ms |  394000.0000 | 100000.0000 | 1000.0000 |  2379.35 MB |
| BeEquivalentTo | 28 | 22,344.94 ms | 442.011 ms | 590.072 ms | 2757000.0000 | 691000.0000 | 1000.0000 | 16608.43 MB |
```

Test sizes:

N = 16 ; Graph size: 2583 objects
N = 18 ; Graph size: 6764 objects
N = 20 ; Graph size: 17710 objects
N = 24 ; Graph size: 121392 objects
N = 28 ; Graph size: 832039 objects

With the new code, the performance proportional to the number of objects is:
```
|  N | Objects | Time/Object |
|----|---------|-------------|
| 16 |    2583 |     20.2 us |
| 18 |    6764 |     20.5 us |
| 20 |   17710 |     20.9 us |
| 24 |  121392 |     21.4 us |
| 28 |  832039 |     26.9 us |
```

The average time/object for the old code, for the duration of testing I let it complete, was:
```
|  N | Objects | Time/Object | Ratio |
|----|---------|-------------|-------|
| 16 |    2583 |    699.6 us |  35x  |
| 18 |    6764 |    761.7 us |  37x  |
| 20 |   17710 |    923.4 us |  44x  |
```

Src/FluentAssertions/Equivalency/AssertionRuleEquivalencyStep.cs
```diff
@@ -83,6 +83,8 @@ private bool ExecuteAssertion(IEquivalencyValidationContext context)
 
             if (subjectIsValidType && expectationIsValidType)
             {
+                AssertionScope.Current.ResolveContext();
```
jnyrup, line 86: I'm struggling to find out why this is necessary.
I have the branch open and see that two tests fails without it, I just can comprehend _why_?
`AssertionScope` is not very friendly to debug/comprehend for _my_ brain.
(I'm not asking you to solve all our problems)

For `When_property_of_other_is_null_the_failure_message_should_not_complain_about_its_type` when it's changed to pass, `DetermineCallerIdentity` is still called twice 😢.

Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -72,17 +73,26 @@ public AssertionScope()
         /// <summary>
         /// Starts a named scope within which multiple assertions can be executed and which will not throw until the scope is disposed.
         /// </summary>
-        public AssertionScope(string context)
+        public AssertionScope(Lazy<string> context)
```
jnyrup, line 76: The `string` constructor should be kept
Just add a new constructor instead.

Src/FluentAssertions/Equivalency/AssertionRuleEquivalencyStep.cs
logiclrd: This is because it is deferring the calculation of the context string, and the calculation depends on the current node. Through this path, the current node changes, so if we wait until after that, we're calculating it with a different node. Calling `ResolveContext` here locks in the computed value for the previous current node value.

Src/FluentAssertions/Execution/AssertionScope.cs
logiclrd: :+1:

Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -134,8 +134,15 @@ public static bool IsEquivalentTo(this IMember property, IMember otherProperty)
                    property.Name == otherProperty.Name;
         }
 
+        private static readonly Type[] EmptyTypeArray = new Type[0];
+
         public static Type[] GetClosedGenericInterfaces(Type type, Type openGenericType)
         {
+            if (Type.GetTypeCode(type) != TypeCode.Object)
+            {
+                return EmptyTypeArray;
```
lg2de, line 143: Why not using `Array.Empty<Type>()`?
I think this is using a singleton, too.

Src/FluentAssertions/Equivalency/AssertionRuleEquivalencyStep.cs
lg2de: @logiclrd , your explaination indicates that this should be documented. If maintainers like Jonas are not able to understand, you should add comments and maybe more specific method name like `BuildContextString`. "Build" is related to the func of the `Lazy` instance which can be "complex building procedure".

@jnyrup , I'm wondering that only 2 tests are failing. This indicates that only 2 of thousand of tests are using the scenario which is described by Jonathan.

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
```diff
@@ -36,7 +36,7 @@ public bool IsCyclicReference(ObjectReference reference, Reason reason = null)
 
             if (reference.IsComplexType)
             {
-                if (orderedReferences.Contains(reference))
+                if (orderedReferences.TryGetValue(reference.Object, out var existingReference) && existingReference.Equals(reference))
```
lg2de, line 39: I think you have implement double comparison. This was not required before and should still not be required.
I think you should simply replace `List` by `HashSet`.

Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -279,12 +289,7 @@ public Continuation FailWith(string message, params Func<object>[] argProviders)
 
         private string GetIdentifier()
         {
-            if (!string.IsNullOrEmpty(Context))
-            {
-                return Context;
-            }
-
-            return CallerIdentifier.DetermineCallerIdentity();
+            return Context?.Value ?? CallerIdentifier.DetermineCallerIdentity();
```
lg2de, line 292: Are you sure that `Context.Value` will never return an empty string?
Before this was expected and handled by using `string.IsNullOrEmpty`.

Src/FluentAssertions/Execution/ContextDataItems.cs
```diff
@@ -42,10 +42,13 @@ public void Add(ContextDataItems contextDataItems)
 
         public void Add(DataItem item)
         {
-            DataItem existingItem = items.SingleOrDefault(i => i.Key == item.Key);
-            if (existingItem != null)
+            for (int i = 0; i < items.Count; i++)
```
lg2de, line 45: Why do you think that `for` is more efficient than `SingleOrDefault`?
I agree that "remove" and "add" is not that good, but `SingleOrDefault` is implemented in efficient way for all containers and even more better to read that the new for loop.

Src/FluentAssertions/Equivalency/AssertionRuleEquivalencyStep.cs
logiclrd: I don't think that's _that_ unusual, because the scenario involved requires that you use a pattern of:
```
result.Should().BeEquivalentTo(
  expectation,
  o => o.Using(c => c.Should().BeSomethingElse());
```
With this pattern, the `.Using()` is re-entering the entire assertion stack from the top again. This is where deferring the calculation becomes a problem, because the existing unit tests indicate that a failure within the `.Using()` should be _reported_ as a failure of the outer `result.Should()`, but the deferred calculation makes it find the `c.Should()` instead.

This can only occur when you have this re-entrant structure, and the vast majority of the unit tests do not.

I will document this. :-)

Src/FluentAssertions/Common/TypeExtensions.cs
logiclrd: Is that available on all the target platforms?

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
logiclrd: That's fair. I was under the mistaken impression that a `Dictionary<object, T>` was doing a different type of comparison, but if it is simply using `.Equals()` and `.GetHashCode()` then this is indeed redundant.

Src/FluentAssertions/Execution/AssertionScope.cs
logiclrd: Well, all the tests pass. *shrug*

I can change this to match the old semantics more exactly.

Src/FluentAssertions/Execution/ContextDataItems.cs
logiclrd: I'm using the index inside the loop body. `SingleOrDefault` doesn't return the index.

Src/FluentAssertions/Common/TypeExtensions.cs
lg2de: I think so, everything since .net framework 4.6
https://docs.microsoft.com/en-us/dotnet/api/system.array.empty

Src/FluentAssertions/Execution/AssertionScope.cs
lg2de: Maybe @jnyrup or @dennisdoomen have more information!?

Src/FluentAssertions/Common/TypeExtensions.cs
logiclrd: :+1:

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
logiclrd: Oh hold on a second. Just looked at this more closely, and the subjects differ.

* When checking the dictionary, it's based on `reference.Object`, i.e. the target of the reference.
* When checking the equality, it's on `reference` itself. This comparison only returns true if the first reference is an earlier point in the chain of that led to the second reference.

This means that it can encounter the same object multiple times in an object graph without issue; each subsequent time it sees the object, it'll find it in the dictionary, but `existingReference.Equals(reference)` will say `false` -- even though it's the same object, the path it took to find `reference` didn't "come from" `existingReference`.

The code includes logic that replaces the dictionary entry when it sees the same object but it wasn't a circular reference. Just thinking about it logically, I'm thinking this may be unnecessary. The caller is using a depth-first search, and so if a circular reference existed, it would _always_ find that before it found unrelated references.

As an aside, when I just took a closer look at `ObjectReference.Equals`, I saw some semantics that seem dubious. The `a.Equals(b)` comparison includes a requirement that `a` be the "parent of" `b`. This would seem to me to break the invariant that `a.Equals(b) == b.Equals(a)`. I will rename `Equals` to something that conveys more clearly what it is doing.

Src/FluentAssertions/Execution/AssertionScope.cs
logiclrd: This might be okay, because `Context` as a `Lazy<string>` is newly-introduced, so it can simply be a convention that it never return an empty string.

Src/FluentAssertions/Execution/ContextDataItems.cs
logiclrd: I can use `FindIndex` for this.

logiclrd:
I am getting unit test failures on `develop` currently. The same tests are failing after my changes here, but hopefully the main line will see fixes onto which this branch can be rebased.

Src/FluentAssertions/Equivalency/AssertionRuleEquivalencyStep.cs
dennisdoomen: What we really want is to start a new nested `AssertionScope` that does not capture the `c.Should()`. Can't we make that more explicit?

Src/FluentAssertions/Execution/AssertionScope.cs
dennisdoomen: I agree that it should never contain an empty string, but I'm not sure we can guarantee that. Somebody could use the `AssertionScope` constructor and pass an empty string. So I think accounting for that is the smart thing to do.

Src/FluentAssertions/Equivalency/AssertionRuleEquivalencyStep.cs
logiclrd: I'm not entirely certain, to be honest -- my familiar with the codebase only started a few weeks ago :-P

Thinking about it, though, it seems that the key to this comes down to how it locates that context. It's looking through the stack trace for the first non-FluentAssertions frame it finds. This change is needed because when the user writes:
```
result.Should().BeEquivalentTo(
  expectation,
  o => o.Using(c => c.Should().BeSomethingElse());
```
...the stack trace starts in user code, then goes into FluentAssertions for `result.Should().BeEquivalentTo()`, then goes _back_ to the user code for `c => c.Should().BeSomethingElse()`, which then goes back into FluentAssertions.

Off the top of my head, initially I was thinking about perhaps making it so that an `AssertionScope` can be given a stack trace to use for resolving the context, so that the re-entrant call actually uses the outer call's stack trace and doesn't see the `c => c.Should().BeSomethingElse()` call at all. But then another thought occurred to me: What if the stack trace were scanned in the opposite order, so that instead of searching for the first frame that isn't in FluentAssertions, we search for the last frame that _is_, and use the frame below that for resolving the context? That would then _always_ use the outermost active call site into FluentAssertions at any point.

logiclrd: Okay, my first stab at this produced tons of test failures, because in the case of `delegate.Should().Throw(..)` we have this exact scenario in reverse.

logiclrd:
This latest commit seems to work. The trickiest part was the `async` machinery, which gums up the works with stack frames that aren't in `FluentAssertions` but do need to be ignored. Because of the fact that `action.Should().Throw(..)` type assertions have this exact same structure, where the stack frames at the bottom of the stack will find `action.Should()` but the message should be finding the subject within the "..", I had to create a mechanism for assertions on delegates to indicate that the CallerIdentifier logic should find the _second_ instance of user code calling into FluentAssertions and not the first. This only activates if the `action.Should()` is the very first call into `FluentAssertions` on the call stack.

Tests/FluentAssertions.Specs/Execution/AssertionScopeSpecs.cs
```diff
@@ -332,7 +332,7 @@ public void When_an_assertion_fails_in_a_named_scope_it_should_use_the_name_as_t
             // Act
             Action act = () =>
             {
-                using var _ = new AssertionScope("foo");
+                using var _ = new AssertionScope(new Lazy<string>(() => "foo"));
```
lg2de, line 335: I think, this can be reverted, now.

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
lg2de: I'm not sure whether I fully understand your explanation.
I still think, `HashSet` should be sufficient. `List.Contains` and `HashSet.Contains` should have similar behavior, but `HashSet` is even faster because using a hash.

You changed the behavior. This would imply you have identified a new bug. Isnt't it?

Tests/FluentAssertions.Specs/Execution/AssertionScopeSpecs.cs
logiclrd: Indeed!

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
logiclrd: Nope, I did not change the behaviour. Comparing if the objects are the same is not the entire test. The dictionary entry is not the object itself, but the `ObjectReference`, and the test on the `ObjectReference`, which was (in my opinion incorrectly) named `Equals` but which I have now renamed, incorporates testing the path at which the object was encountered. For instance, if in the course of walking the object graph, it finds the same object first at `.A.Foo` and then subsequently at `B.Foo`, these objects are the same and will be the same dictionary entries, but they do not constitute a circular reference. But, if it encounters the same object first at `.A.Foo` and then subsequently at `.A.Foo.Bar.Qux.Foo`, then that _is_ a circular reference. The dictionary serves to identify the `ObjectReference` from `.A.Foo` (the previous occurrence) in both cases. The subsequent call to `ObjectReference.IsParentReferenceToSameObjectAs` differentiates these two cases.

This semantic is not new -- my change is only to use a dictionary to find the previous reference. Before my changes, instead of a dictionary it was a `List<ObjectReference>` upon which it did a linear scan.

Src/FluentAssertions/Execution/AssertionScope.cs
logiclrd: It is done. :-)

Src/FluentAssertions/Equivalency/AssertionRuleEquivalencyStep.cs
logiclrd: I've pushed a commit that I think does the trick.

The trickiest part was the `async` machinery, which gums up the works with stack frames that aren't in FluentAssertions but do need to be ignored. Because of the fact that `action.Should().Throw(..)` type assertions have this exact same structure, where the stack frames at the bottom of the stack will find `action.Should()` but the message should be finding the subject within the "..", I had to create a mechanism for assertions on delegates to indicate that the `CallerIdentifier` logic should find the second instance of user code calling into FluentAssertions and not the first. This only activates if the `action.Should()` is the very first call into FluentAssertions on the call stack.

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
logiclrd: (I have come to understand this code better than when I was trying to explain it previously. :-) )

logiclrd: The distinction between scanning the list and doing `IsParentReferenceToSameObjectAs` in the original code is obfuscated by the fact that `IsParentReferenceToSameObjectAs` is implemented by overriding `Equals`. It is being implicitly called by `orderedReferences.Contains`.

The main line's `ObjectReference.Equals` is defined as:
```
return ReferenceEquals(@object, other.@object) && IsParentOf(other);
```

Src/FluentAssertions/Equivalency/AssertionRuleEquivalencyStep.cs
dennisdoomen: > What if the stack trace were scanned in the opposite order, so that instead of searching for the first frame that isn't in FluentAssertions, 

That is exactly what I think might fix the problem.

dennisdoomen:
> I am getting unit test failures on `develop` currently. The same tests are failing after my changes here, but hopefully the main line will see fixes onto which this branch can be rebased.

I've just tried on AppVeyor and locally, but both work without problems. Are you sure? Which one are failing?

logiclrd:
> > I am getting unit test failures on `develop` currently. The same tests are failing after my changes here, but hopefully the main line will see fixes onto which this branch can be rebased.
> 
> I've just tried on AppVeyor and locally, but both work without problems. Are you sure? Which one are failing?

I'm not currently at the workstation where I've been doing my development. I just ran tests on another machine, and `develop` completed without failures. So, false alarm I guess. I'll check into it later when I'm at the other computer.

This "clean" run did show some failures inside my branch, though. Sigh. :-P

logiclrd:
There we go :-)

dennisdoomen (approved):

Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -136,6 +136,11 @@ public static bool IsEquivalentTo(this IMember property, IMember otherProperty)
 
         public static Type[] GetClosedGenericInterfaces(Type type, Type openGenericType)
         {
+            if (Type.GetTypeCode(type) != TypeCode.Object)
```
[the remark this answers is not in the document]
logiclrd: Done.

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
```diff
@@ -36,7 +36,11 @@ public bool IsCyclicReference(ObjectReference reference, Reason reason = null)
 
             if (reference.IsComplexType)
             {
-                if (orderedReferences.Contains(reference))
+                if (!observedReferencesByReferencedObject.TryGetValue(reference.Object, out var existingReference))
```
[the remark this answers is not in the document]
logiclrd: I actually feel like the way `ObjectReference` implements `Equals` is a bit of a code smell. It's breaking the rules of the `Equals` method. The MSDN documentation explicitly states:

> The following statements must be true for all implementations of the `Equals(Object)` method.
> [..]
> * `x.Equals(y)` returns the same value as `y.Equals(x)`.

Because `ObjectReference` also cares about whether `x` is a parent of `y`, by definition it will never return `true` for both `x.Equals(y)` and `y.Equals(x)`.

I'm not convinced it's actually safe to assume that a `HashSet<T>` would work correctly, because it's not actually defined whether the `HashSet<T>` would be using `dataInHashSet.Equals(callerProvidedValue)` or `callerProvidedValue.Equals(dataInHashSet)` -- and in fact it _could_ do both in different places, because of the documented invariant that they are equivalent.

That's why I renamed the method in the proposed changes.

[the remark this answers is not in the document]
logiclrd: I have added a custom `IEqualityComparer` that ensures that the dictionary uses reference comparisons.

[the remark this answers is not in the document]
dennisdoomen: > Because `ObjectReference` also cares about whether `x` is a parent of `y`, by definition it will never return `true` for both `x.Equals(y)` and `y.Equals(x)`.

This is by design. The `ObjectReference` exists to detect that two references point to the same object within the same path.

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
lg2de: The primary intention of this PR is improving performance.
So, to improve performance you should use `HashSet` instead of `List`.
Of you think we have to fix incorrect behavior and you NEED a `Dictionary`, then please provide unit test which fails with `List` and `HashSet` but completes with `Dictionary`.

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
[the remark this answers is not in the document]
logiclrd: I understand that, but it violates a fundamental rule of the .NET Framework. Surely it doesn't decrease confidence in the code to have this method under a different name??

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
logiclrd: I'm kind of astonished by this position, given the intense focus on correctness in this project. The `Object.Equals` instance method is part of a pattern that has very clearly-documented rules, and the current implementation breaks them -- even if you can't find a place in the code currently to _observe_ this causing problems, it is incorrect at a fundamental level. The Framework says, `a.Equals(b) == b.Equals(a)`, and everything within the framework is built on that invariant, and currently, the code is breaking that rule. It's pure chance that `List.Equals` does its comparison in a direction that actually works. Microsoft would not be breaking any rules if they changed the implementation of `List.Equals` so that it asked the supplied object if it was equal to the object in the collection and not vice versa. The direction of this comparison is not contractual _because it doesn't have to be_, because of the rules `Object.Equals` is "guaranteed" to follow.

A `List.Equals` check is conceptually far simpler than a `HashSet.Contains` or `Dictionary.Contains`. This is why, when writing this, switching from `List` to a more complicated type, I felt a great deal of uncertainly relying on undocumented internal behaviour of that new type. Even with `List<T>`, the code is technically incorrect for calling `.Contains`, but could be correct if, instead of calling `.Contains`, FluentAssertions did the loop itself and explicitly called the comparison method. But, at a deeper level, it is fundamentally wrong for the object to be overriding `.Equals` with an implementation that is not symmetrical, and to then be relying on Framework library code written against the documented properties of `.Equals` that aren't being followed... Yes, you have unit tests that say it works in its current form, but I'm quite convinced that it is a serious flaw! I feel certain that if it were something that current code analysis tech were up to, StyleCop would make this a build error.

Do you _really_ want to keep it in this state? Do you want to simply test that the internal implementation of `HashSet` seems to be doing all of its `.Equals` calls in the same direction as `List` and then latch onto that undocumented behaviour? I mean, it's your project, and if that's really what you want, I can change it to do that, but it feels at odds with the fundamental principles of the entire project to me!

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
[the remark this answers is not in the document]
dennisdoomen: Renaming something is fine, but changing the semantics is not.

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
logiclrd: One other way to remedy this, that off the top of my head I _think_ would still serve the purpose, would be to make `ObjectReference.Equals` symmetrical simply by allowing either reference to be a parent of the other. This would mean that the `.Equals` implementation followed the rules, and within CyclicReferenceDetector.cs, it should still behave exactly the same way, because it will never find a parent to a reference _after_ that reference.

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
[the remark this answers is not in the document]
logiclrd: Oh, but how does this change the semantics? I mean, it did before the custom `IEqualityComparer` was put in, but with that in place, doesn't this have exactly the same behaviour as before??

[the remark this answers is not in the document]
dennisdoomen: > I'm not convinced it's actually safe to assume that a `HashSet<T>` would work correctly, because it's not actually defined whether the `HashSet<T>` would be using `dataInHashSet.Equals(callerProvidedValue)` or `callerProvidedValue.Equals(dataInHashSet)` -- and in fact it _could_ do both in different places, because of the documented invariant that they are equivalent.

Yeah, I agree.

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
lg2de: The class is an internal class. **I myself** think it is acceptable that `Equals` (and `GetHashCode`) are implemented as required for the internal use case. @dennisdoomen  and @jnyrup may decide whether the strict implementation is required or not.

Further I think that we can trust in the current implementation (of `List` and `HashSet`). Microsoft is very conservative with changing behavior.

Please note in your explanation that `HashSet` and `Dictionary` are currently do exactly what you have implemented manually:
1. Get hash code of `ObjectReference.object`. The implementation of `ObjectReference.GetHashCode` is returning hash code of object.
2. Check for all objects with (potentially) same hash code using `Equals`.

The implementation **may** fail if we will have two instances of `ObjectReference` with same object and different parents. I do not know if this can ever happen. It seems so far it has never happen.
If so, we have a bug we can validate with a new unit test.

logiclrd: I'm pretty sure that can and does come up when you have multiple references to an object that are not cyclic.
```
public class Test
{
  public string A;
  public string B;
}
```
Consider an instance where `A` and `B` point at the same `string` object. In this case, when comparing `obj.A`, it gets added to the `CyclicReferenceDetector` with a path of `[A]`. Then, later, when comparing `obj.B`, it goes back to the `CyclicReferenceDetector` with an `ObjectReference` for the same object, but this time the path is `[B]`. The dictionary will contain the previous reference with path `[A]`, but `IsParentReferenceToSameObjectAs` will return `false`, because `[A]` is not a parent of `[B]`. I'm pretty sure this is doing exactly what it is supposed to do.

With the old `List` implementation, the list would end up with both `ObjectReference` instances in it. With the new implementation, it can only refer to one of them at a time, since it's using the object itself as the dictionary key, but it seems to me that it really doesn't matter which one it is comparing, because:

* If there _is_ a cyclic reference, it will have already found it during the depth-first recursion into the object.
* If there _isn't_ a cyclic reference, then it doesn't matter what path it gets to the object by, it's never going to find a cyclic reference.

As for your assertion that what's happening internally in `HashSet` or `Dictionary` is _exactly_ the same, well, it's not -- or rather, it's not _guaranteed_ to be, and the crux of it is that with the current implementation, it makes a big difference whether the internal implementation is using `dataInHashSet.Equals(callerProvidedValue)` or `callerProvidedValue.Equals(dataInHashSet)` -- and it _shouldn't_ make a difference, per the contractual definition of `object.Equals`. As I've written it, the dictionary look-up is doing a strict object reference equality check on the _target_ of the `ObjectReference` to find the dictionary entry (this leverages `RuntimeHelpers.GetHashCode` to ensure that it isn't using any object's own override of `GetHashCode`), and then is calling `objectReference.IsParentReferenceToSameObjectAs` on the result of that as a separate step.

logiclrd: I have been exploring the possibility of returning it to an `Equals` based implementation by making `Equals` behave in a symmetrical manner. I've run into a snag, though; in order to add a unit test of this, `FluentAssertions.Specs` needs to be able to see the internals of `FluentAssertions`, but `FluentAssertions` is signed and `FluentAssertions.Specs` is not. So, sign `FluentAssertions.Specs`, right? Well, `FluentAssertions.Specs` uses a NuGet package called `Chill` (@dennisdoomen I see you're a contributor on this too?), and `Chill` isn't signed. The requirement that signed assemblies only expose their `internal`s to other signed assemblies has a bit of a ripple effect. :-/

logiclrd: For what it's worth, using a `HashSet<ObjectReference>` with `ObjectReference` providing a symmetrical `Equals` is passing all tests. The only snag is that currently, to make it testable, `ObjectReference` becomes a `public` type.

logiclrd: Another thought occurs to me. Is there any situation where cyclic reference detection is being used that _isn't_ a depth-first scan of the object graph? If not, then instead of an ever-increasing collection of `ObjectReference` objects, maybe this could be achieved by means of essentially a stack of objects, so that objects are only in the collection for the time during which they are relevant.

Each time the tree walking code starts work on an object, it gets an `IDisposable` object from `CyclicReferenceDetector`. The creation of this object adds the subject to a `HashSet<object>(new ReferenceEqualsComparer())`, and then when it gets `Dispose`d, the object is removed from the `HashSet`. Instead of having to worry about paths at all, then, the check for cyclic references consists entirely of checking if an object is already in the `HashSet` before setting up a new scope for it.

Pseudocode:
```
public class CyclicReferenceDetector
{
  class ReferenceScope : IDisposable
  {
    CyclicReferenceDetector _owner;
    object _subject;

    public ReferenceScope(CyclicReferenceDetector owner, object subject)
    {
      _owner = owner;
      _subject = subject;
      _owner.EnterScope(subject);
    }

    public void Dispose()
    {
      _owner?.ExitScope(_subject);
      _owner = null;
    }
  }

  public IDisposable BeginScope(object subject)
    => new ReferenceScope(this, subject);

  HashSet<object> _objectsInScope = new HashSet<object>(new ReferenceEqualsComparer());

  void EnterScope(object subject)
  {
    if (!_objectsInScope.Add(subject))
      throw new Exception("Cyclic reference detected");
  }

  void ExitScope(object subject)
  {
    _objectsInScope.Remove(subject);
  }
}

...

void ObjectsAreEquivalent(object result, object expectation)
{
  using (var scope = _cyclicReferenceDetector.BeginScope(result))
  {
    foreach (var member in GetMembersForComparison(result))
    {
      object resultChild = member.GetValue(result);
      object expectationChild = member.GetValue(expectation);

      if (IsComplexType(resultChild) && IsComplexType(expectationChild))
        ObjectsAreEquivalent(resultChild, expectationChild);
      else
        ValuesAreEquivalent(resultChild, expectationChild);
    }
  }
}
```

Obviously it would need to be a bit more involved in order to give a meaningful error that can show you the path at which the cycle occurred, but can you think of a reason why this approach wouldn't work? It would mean that the `CyclicReferenceDetector` did not build up an ever-growing collection of objects that never shrank.

dennisdoomen: Gee guys. I can barely catch up with all these long and in-depth discussions ;-)

Looking at the current code, I don't think the order matters because the parent condition check happens on the dotted path. However, [5 years ago],(https://github.com/fluentassertions/fluentassertions/commit/030b14c299c49c8b6047c2d79946c7e655a8496c), I did fix some unclear bug by switching from a `HashSet` to a simple list. Unfortunately, I did not add a test for this. Not sure why. So it _seems_ that the order _is_ important after all. 

But if we can't find any proof for that, then an implementation that first tries to find the `ObjectReference` based on a reference equality on the subject under test, and _then_ uses the `Equals` of `ObjectReference` may work indeed. I also agree with the dubious implementation of that `Equals`, so maybe renaming it to a more self-describing method makes sense.

lg2de: > an implementation that first tries to find the `ObjectReference` based on a reference equality on the subject under test, and _then_ uses the `Equals` of `ObjectReference`

This is what the (current) implementation of `HashSet.Contains` together with `ObjectReference` does.

dennisdoomen: Yes, I think that should be fine.

logiclrd: @lg2de It isn't, though, because the actual call to `.Equals` is taking place inside opaque framework code.

I am investigating making `.Equals` symmetric, though, which would obviate the entire issue. That's actually stuck on the `Chill` dependency downstream not being strong-named, though. :-P

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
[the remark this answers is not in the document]
logiclrd: Okay, I believe this should be resolved to everyone's satisfaction now :-)

* ObjectReference.cs now uses `Equals` again, but this is okay because `Equals` is now symmetric. This is unit tested.
* CyclicReferenceDetector.cs can then confidently use `HashSet<ObjectReference>`. All existing tests pass.

I had thus far been amending my commit to incorporate PR review, having observed this elsewhere in the repository, but in this case the changes required for this are in a separate commit on the branch so they can be easily inspected independently.

Src/FluentAssertions/Equivalency/CyclicReferenceDetector.cs
logiclrd: Okay, I believe this should be resolved to everyone's satisfaction now :-)

* ObjectReference.cs now uses `Equals` again, but this is okay because `Equals` is now symmetric. This is unit tested.
* CyclicReferenceDetector.cs can then confidently use `HashSet<ObjectReference>`. All existing tests pass.

I had thus far been amending my commit to incorporate PR review, having observed this elsewhere in the repository, but in this case the changes required for this are in a separate commit on the branch so they can be easily inspected independently.

dennisdoomen: It does change the behavior of the original `ObjectReference`, but IMHO it does not matter whether two objects are each other's parent or child.

dennisdoomen:
I'm afraid to ask, but because of all the review comment and rework, it looks like we need to recombine the relevant changes into dedicated commits.

logiclrd:
You'd like to see it squashed into a single commit? Or should separate along logical subdivisions, if I can find them?

dennisdoomen:
What I would do is to regroup the relevant changes into a new set of commits. See my [blog post](https://www.continuousimprover.com/2020/03/keep-source-control-history-clean.html) if you need some inspiration for that.

logiclrd:
How does this look? (b7a1fe7 as compared to 954d109)

dennisdoomen:
> How does this look? ([b7a1fe7](https://github.com/fluentassertions/fluentassertions/commit/b7a1fe7398b60c7e14978cd1555ad4aeee8c18cc) as compared to [954d109](https://github.com/fluentassertions/fluentassertions/commit/954d109660ee3b5b80f60d166081153ba16a2f0c))

Looks nice. Only the signing of the specs project could have been separate, but that's me nitpicking.

logiclrd:
> Looks nice. Only the signing of the specs project could have been separate, but that's me nitpicking.

I saw it as being sort intrinsically a "part of" the other changes -- a change with no motivation of its own, driven by the need to unit test an internal detail. Each of the commits achieves an outcome, and that would be a commit that doesn't actually achieve an outcome of its own, but merely reworks things to allow an actual change.

I could split it though if you think it would make a difference :-)

dennisdoomen:
Nah, don't bother. Let's wait for the review from @jnyrup

jnyrup (approved):

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -260,7 +260,21 @@ private static async Task<Exception> InvokeWithInterceptionAsync(Func<Task> acti
         {
             try
             {
-                await action();
+                // For the duration of this nested invocation, configure CallerIdentifier
+                // to match the contents of the subject rather than our own call site.
+                //
+                //   Func<Task> action = async () => await subject.Should().BeSomething();
+                //   await action.Should().Throw<Exception>();
```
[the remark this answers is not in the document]
logiclrd: I have reworked this this change into the relevant commit.

Issue #1353, opened 2020-07-16
Some assertions are not including variable names in assertion message


mdentremont:
### Description

When using fluent assertions for things like objects & lists, assertion failures will typically log the names of the variables. This doesn't appear to happen when comparing `DateTime`s, `Guid`s, objects, etc.

It is very useful to have the variables names for context, as when the assertion message doesn't have these it is difficultto figure out which assertion failed when looking at test output for "AssertionScope" blocks (and for non-assertion scope blocks, you can only find the failing assertion via the line numbers in the stack).

Here is what the output looks like for one of the assertions which includes variables names:
```csharp
var z = "SSSS";
z.Should().Contain("BBB");

> Expected z "SSSS" to contain "bbb".
```

### Complete minimal example reproducing the issue

```csharp
var x = DateTime.UtcNow;
var y = DateTime.UtcNow.AddSeconds(1);

x.Should().Be(y);
```

### Expected behavior:

Assertion failure should be something like: `Expected x to be <2020-07-15 16:51:43.1621939>, but was <2020-07-15 16:51:43.5991913>.`

### Actual behavior:

`Expected date and time to be <2020-07-15 16:51:43.1621939>, but found <2020-07-15 16:51:43.5991913>.`

### Versions

* FluentAssertions 5.10.2
* .NET framework 4.8

jnyrup:
When I run your snippet, I get the following failure message
```
Expected x to be <2020-07-16 15:26:37.7544405>, but found <2020-07-16 15:26:36.7544405>.
```

Is your test project compiled in debug mode?
https://fluentassertions.com/introduction#subject-identification

mdentremont:
@jnyrup Interesting - I'm building in debug mode, and I'm assuming the symbols are present because my stacktrace has line numbers:
```
Expected date and time to be <2020-07-16 15:34:09.4916424>, but found <2020-07-15 15:34:09.4916424>.
   at FluentAssertions.Execution.LateBoundTestFramework.Throw(String message) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\LateBoundTestFramework.cs:line 16
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\AssertionScope.cs:line 225
[...]

PR #1440, opened 2020-11-30
Improve XML assertion messages

lg2de:
I was triggered by reading #1353 that most XML assertion messages does not include variable name too.

So I've started reworking these assertions to
* include variable name
* provide some useful context information.

While reviewing the texts in the unit tests I found some messages which are misleading because the XmlValidator was following the subject instead of the expectation.

You may want to review the two commits separately.

I'll add changelog when the PR number is available...
Should these two commits result into two statements in the changelog?

### DIFF
--- a/Src/FluentAssertions/Xml/Equivalency/XmlReaderValidator.cs
+++ b/Src/FluentAssertions/Xml/Equivalency/XmlReaderValidator.cs
@@ -32,7 +32,7 @@ namespace FluentAssertions.Xml.Equivalency
 
             if (!shouldBeEquivalent && failure is null)
             {
-                assertion.FailWith("Did not expect Xml to be equivalent{reason}, but it is.");
+                assertion.FailWith("Did not expect {context:subject} to be equivalent{reason}, but it is.");
             }
         }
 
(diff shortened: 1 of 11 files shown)


dennisdoomen:
> So I've started reworking these assertions to

Great initiative.

> Should these two commits result into two statements in the changelog?

Don't think so. Also, the 2nd commit seems to contain unrelated changes. In other words, you've changed the expected output in the specs, but there are no other changes that are the cause of this.

lg2de:
The second commit contains change of test and change using `expectationIterator` instead of `subjectIterator`.

dennisdoomen:
> The second commit contains change of test and change using `expectationIterator` instead of `subjectIterator`.

Yes, so it's weird that you had to change a test even though nothing changed in that commit. So I fail to understand why you were asking about the commits.

lg2de:
The commits are doing what their commit message reports:
1. add variable name and some additional information
2. change kind of XML validation to get better message

Regarding changelog this could be two changes or a single one.

dennisdoomen:
We must be on the wrong channel ;-)

The 2nd commit updates `XElementAssertionSpecs.cs`, but there's nothing in that commit that seems to cause a change in the assertion method. So I don't understand why that change is in the 2nd commit. If it was caused by the 1st commit, that change should belong there. Just a matter of a [fixup commit](https://www.continuousimprover.com/2020/03/keep-source-control-history-clean.html).

jnyrup:
Thanks for giving the xml assertions and specs some needed love.

I think the two commits are so closely related, that the single release note covers them both - and that's fine by me.

jnyrup (approved):

dennisdoomen (approved):

PR #1460, opened 2021-01-09
Make new `CloneDefaults` less assuming

jnyrup:
The new `CloneDefaults` added in #1419 _assumes_ that `TOptions` has a constructor that takes an `EquivalencyAssertionOptions`.
Instead of assuming the existence of such a constructor, inject a predicate to make it more type safe.

Two other benefits:
* The compiler now easily sees that `DataEquivalencyAssertionOptions` is constructed.
* It's faster to construct and inject a predicate.

```
|            Method |       Mean |    Error |   StdDev | Ratio |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------------ |-----------:|---------:|---------:|------:|-------:|-------:|------:|----------:|
|     CloneDefaults | 1,437.0 ns | 27.29 ns | 29.20 ns |  1.00 | 0.4349 | 0.0038 |     - |   2.68 KB |
| CloneDefaultsFunc |   708.4 ns |  5.20 ns |  4.86 ns |  0.49 | 0.3729 | 0.0038 |     - |    2.3 KB |
```

### DIFF
--- a/Src/FluentAssertions/AssertionOptions.cs
+++ b/Src/FluentAssertions/AssertionOptions.cs
@@ -21,12 +21,12 @@ namespace FluentAssertions
             return new EquivalencyAssertionOptions<T>(defaults);
         }
 
-        public static TOptions CloneDefaults<T, TOptions>()
+        internal static TOptions CloneDefaults<T, TOptions>(Func<EquivalencyAssertionOptions, TOptions> predicate)
             where TOptions : EquivalencyAssertionOptions<T>
         {
-            return (TOptions)Activator.CreateInstance(
-                typeof(TOptions),
-                defaults);
+            Guard.ThrowIfArgumentIsNull(predicate, nameof(predicate));
+
+            return predicate(defaults);
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


Src/FluentAssertions/AssertionOptions.cs
```diff
@@ -21,12 +21,12 @@ public static EquivalencyAssertionOptions<T> CloneDefaults<T>()
             return new EquivalencyAssertionOptions<T>(defaults);
         }
 
-        public static TOptions CloneDefaults<T, TOptions>()
+        public static TOptions CloneDefaults<T, TOptions>(Func<EquivalencyAssertionOptions, TOptions> predicate)
```
dennisdoomen, line 24: Should we make this method `internal`?

dennisdoomen (approved):

Src/FluentAssertions/AssertionOptions.cs
jnyrup: Yeah, let's do that for now.

PR #1463, opened 2021-01-09
Bumped most dependencies to the latest and greatest

dennisdoomen:
Skipped the FxCop assemblies because bumping them causes deprecation message to appear. See also https://docs.microsoft.com/en-us/visualstudio/code-quality/migrate-from-fxcop-analyzers-to-net-analyzers?view=vs-2019

### DIFF
--- a/Src/FluentAssertions/FluentAssertions.csproj
+++ b/Src/FluentAssertions/FluentAssertions.csproj
@@ -40,7 +40,7 @@
     <AdditionalFiles Include="BannedSymbols.txt" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="JetBrains.Annotations" Version="2020.1.0" PrivateAssets="all" />
+    <PackageReference Include="JetBrains.Annotations" Version="2020.3.0" PrivateAssets="all" />
     <None Include="..\FluentAssertions.png" Pack="true" Visible="false" PackagePath="" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
--- a/Tests/FluentAssertions.Specs/FluentAssertions.Specs.csproj
+++ b/Tests/FluentAssertions.Specs/FluentAssertions.Specs.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>net47;netcoreapp2.0;netcoreapp2.1;netcoreapp3.0</TargetFrameworks>
@@ -29,7 +29,6 @@
   </PropertyGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net47'">
-    <PackageReference Include="System.Collections.Immutable" Version="1.4.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
@@ -37,7 +36,6 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.0'">
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="[16.2.0]" />
     <PackageReference Include="xunit.runner.visualstudio" Version="[2.4.1]">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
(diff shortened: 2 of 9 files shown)


Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -58,7 +58,7 @@
     <Compile Remove="Common/ConfigurationStoreExceptionInterceptor.cs" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.0" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
```
jnyrup, line 61: Why bumping this dependency?

Tests/FluentAssertions.Specs/FluentAssertions.Specs.csproj
```diff
@@ -29,23 +29,21 @@
...
   <ItemGroup Condition=" '$(TargetFramework)' == 'net47'">
-    <PackageReference Include="System.Collections.Immutable" Version="1.4.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.0'">
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="[16.2.0]" />
```
jnyrup, line 40: Why removing this?

Tests/FluentAssertions.Specs/FluentAssertions.Specs.csproj
```diff
@@ -55,21 +53,22 @@
...
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="Xunit.StaFact" Version="1.0.37" />
     <PackageReference Include="System.Data.DataSetExtensions" Version="4.5.0" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.0'">
-    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="3.3.1">
+    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.164">
```
jnyrup, line 63: I upgraded StyleCop.Analyzers 4 lines below.
and the beta.321 fixed the false warnings on `new()`

maybe a wrongly solved merge/rebase conflict?

Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -80,15 +80,15 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="3.3.1">
```
jnyrup, line 83: Due to a merge/rebase conflict that these two dependencies are switched?

jnyrup (changes requested):

Src/FluentAssertions/FluentAssertions.csproj
dennisdoomen: Because there's a newer version?

Tests/FluentAssertions.Specs/FluentAssertions.Specs.csproj
dennisdoomen: Apparently it's no longer needed. It was removed automatically by the IDE.

Src/FluentAssertions/FluentAssertions.csproj
jnyrup: Unless we as a library need a new version, I think we should let that be up to the consumer.

dennisdoomen: Fair point.

Tests/FluentAssertions.Specs/FluentAssertions.Specs.csproj
jnyrup: Ahh, it's because an older version comes with `xunit.runner.visualstudio`
image
I can't tell if it makes a difference for us, there's a gazillion changes between 15.0.0 and 16.2.0
https://github.com/Microsoft/vstest-docs/blob/master/docs/releases.md

jnyrup (approved):

Tests/FluentAssertions.Specs/FluentAssertions.Specs.csproj
jnyrup: I tried checking out the branch
* the `Microsoft.NET.Test.Sdk` dlls are not included in the netcoreapp2.0 bin folder.
* I could run the tests from VS
* I could run the test from CLI

So I'm fine with deleting it.

PR #1466, opened 2021-01-10
Add WithParameterName

### LINKED ISSUE #1453: [Request] Add 'WithParamName' validation extension for 'ArgumentException' and derived exception types
### Description

I'd like to have a `WithParamName(string)` (or similarly named) validation extension whenever an `ArgumentException` or any of its derived types is used in a validation.

Currently, we only have a `WithMessage(string)` extension.

In all my unit tests of public APIs, I always have tests checking that the expected argument exceptions are thrown, and that they are thrown for the correct arguments. This is important for methods where multiple arguments are present and more than one of them cannot be `null`. This requires me to write verbose code to check for the argument name. I believe an extension would make the code a bit simpler for this common case.

### Expected behavior:

```csharp
performingAction.Should().Throw<ArgumentNullException>().WithParamName("parameter");
```

This should work seamlessly with other methods like `WithMessage`:
```csharp
performingAction.Should().Throw<ArgumentNullException>()
    .WithParamName("parameter")
    .WithMessage("my exception message");
```

### Actual behavior:

```csharp
performingAction.Should().Throw<ArgumentNullException>().Which.ParamName.Should().Be("parameter");
```

### Versions

5.10.3

* Which .NET runtime and version are you targeting?
Net 5.0

### Additional Information

Alternatively, `WithMessage` could be dropped in favor of a different syntax, like:

```csharp
performingAction.Should().Throw<ArgumentNullException>().Which.Should()
    .TargetParameter("parameter")
    .And
    .HaveMessage("my exception message");
```

This is more verbose, but probably considerably easier to extend (but has the potentially weird 2 `Should` calls).

jnyrup:
Add `WithParameterName` extension to assert on the `ParamName` property on `ArgumentException`.

This fixes #1453

### DIFF
--- a/Src/FluentAssertions/AssertionExtensions.cs
+++ b/Src/FluentAssertions/AssertionExtensions.cs
@@ -882,6 +882,51 @@ namespace FluentAssertions
             return (await task).WithInnerException<TInnerException>(because, becauseArgs);
         }
 
+        /// <summary>
+        /// Asserts that the thrown exception has a parameter which name matches <paramref name="paramName" />.
+        /// </summary>
+        /// <param name="parent">The <see cref="ExceptionAssertions{TException}"/> containing the thrown exception.</param>
+        /// <param name="paramName">The expected name of the parameter</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
+        /// </param>
+        public static ExceptionAssertions<TException> WithParameterName<TException>(
+            this ExceptionAssertions<TException> parent,
+            string paramName,
+            string because = "",
+            params object[] becauseArgs)
+            where TException : ArgumentException
+        {
+            parent.Which.ParamName.Should().Be(paramName, because, becauseArgs);
+            return parent;
+        }
+
+        /// <summary>
+        /// Asserts that the thrown exception has a parameter which name matches <paramref name="paramName" />.
+        /// </summary>
+        /// <param name="task">The <see cref="ExceptionAssertions{TException}"/> containing the thrown exception.</param>
+        /// <param name="paramName">The expected name of the parameter</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
+        /// </param>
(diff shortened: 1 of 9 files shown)


dennisdoomen (approved):

Src/FluentAssertions/AssertionExtensions.cs
```diff
@@ -882,6 +882,51 @@ public static TTo As<TTo>(this object subject)
...
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
+        /// </param>
+        public static ExceptionAssertions<TException> WithParameterName<TException>(
+            this ExceptionAssertions<TException> parent,
+            string paramName,
```
julealgon, line 899: I know this has already been merged, but I'd suggest adding argument validation (!= null) for `paramName` here to improve the stacktrace in case people misuse the method.

Src/FluentAssertions/AssertionExtensions.cs
```diff
@@ -882,6 +882,51 @@ public static TTo As<TTo>(this object subject)
...
+            string because = "",
+            params object[] becauseArgs)
+            where TException : ArgumentException
+        {
+            parent.Which.ParamName.Should().Be(paramName, because, becauseArgs);
```
julealgon, line 904: Question: won't the final error message here be a bit cryptic to an outside consumer? Wouldn't it be better to create a custom validation instead of relying on the standard `Should().Be` in this case?

Src/FluentAssertions/AssertionExtensions.cs
```diff
@@ -882,6 +882,51 @@ public static TTo As<TTo>(this object subject)
...
+        /// <param name="task">The <see cref="ExceptionAssertions{TException}"/> containing the thrown exception.</param>
+        /// <param name="paramName">The expected name of the parameter</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
+        /// </param>
+        public static async Task<ExceptionAssertions<TException>> WithParameterName<TException>(
```
julealgon, line 920: Shouldn't this one have the `Async` suffix?

jnyrup: The current approach in this class is that they don't have `Async` suffix.

Src/FluentAssertions/AssertionExtensions.cs
jnyrup: Good point, I'll change that.

Src/FluentAssertions/AssertionExtensions.cs
jnyrup: I didn't think about that case, but `ArgumentException` allows a `null` `ParamName`.

Maybe I'm misunderstanding you, can you give an example on "misuse"?

PR #1468, opened 2021-01-11
Use WithParameterName in our own tests

jnyrup:
Let's use this new extension method ourselves.


dennisdoomen (approved):
Maybe we should keep one example of `Which.ParamName`?

jnyrup:
There a 20+ tests that exercises that you can call `.Which.Message`.

PR #1473, opened 2021-01-18
Fix unit test namespaces according to the file structure

lg2de:
Having a look on the TODO window found that "low hanging fruit".
Because only unit tests are affected, I guess no changelog entry is required.

### DIFF
--- a/Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
@@ -5,10 +5,11 @@ using FluentAssertions.Extensions;
 #if NET47
 using FluentAssertions.Specs.Common;
 #endif
+using FluentAssertions.Specs.Exceptions;
 using Xunit;
 using Xunit.Sdk;
 
-namespace FluentAssertions.Specs
+namespace FluentAssertions.Specs.Specialized
 {
     [SuppressMessage("ReSharper", "ExpressionIsAlwaysNull")]
     public class TaskOfTAssertionSpecs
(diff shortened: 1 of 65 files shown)


Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
```diff
@@ -2,13 +2,11 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Threading.Tasks;
 using FluentAssertions.Extensions;
-#if NET47
```
jnyrup, line 5: I'm pretty sure this is still used for `Divide` and `Multiply` extension methods.

jnyrup (approved):

Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
lg2de: Sorry for that!
Wondering why this was not seen before in the local build?
I trusted to much in Rider...

jnyrup: No worries, I have *almost* removed it several times.

lg2de: How about moving it to top level namespace (in Specs) to prevent this problem in future?

jnyrup: I've added a note to myself about this.
It's been a bit back and forth with that class
https://github.com/fluentassertions/fluentassertions/pull/1376/files
https://github.com/fluentassertions/fluentassertions/pull/1424/files

dennisdoomen (approved):

Issue #1340, opened 2020-05-31
Capture design guidelines, principles and patterns


dennisdoomen:
* High-level principles such as written down in https://github.com/fluentassertions/fluentassertions/issues/1281#issuecomment-636414698
* Behavioral patterns such as how to deal with `null` arguments
* When to return `AndWhichConstraint` vs `AndConstraint`.
* When to use `FailWith` vs throwing a normal .NET exception type
* Prefer "Did not expect something to be" over "Expected something not to be"
* Decisions (like a poor mans Architecture Design Log) such as why we target certain frameworks 
* When using `AndWhichConstraint<..., T>` the `T` element must be fetched in way resilient to that `FailWith` don't throw immediately when encapsulated in a custom `AssertionScope`.
* Assertion class over non-sealed classes or interfaces should probably be generic in type to avoid loosing type information.
  E.g instead of `MyClassAssertions` then `MyClassAssertions<T> where T : MyClass`
* Be aware that an assertion might be wrapped in an `AssertionScope` so `FailWith` does not halts execution and extra precautions must be taken to avoid e.g. en `NullReferenceException` after verifying that `Subject is not null`
* Don't use `predicate.Body` to format a predicate in a failure message. Just pass the predicate to `FailWith` and let the `PredicateLambdaExpressionValueFormatter` handle formatting it properly
* Naming and grouping guidelines (based on [this post](https://www.continuousimprover.com/2023/03/test-naming.html))
  * Group tests for the same API using a nested class such as in [NullableNumericSpecs](https://github.com/fluentassertions/fluentassertions/blob/develop/Tests/FluentAssertions.Specs/Numeric/NullableNumericAssertionSpecs.cs#L210) so you don't have to repeat the API name in the name of the test
  * Avoid the use of `When` and `Should` in test names and use concise names like `Exclusion_of_missing_members_works_with_mapping`
* When writing parameterized tests prefer using `[InlineData]`, unless:
  * the values are not compile-time constants, or
  * there is a _good_ reason to share the parameters between multiple tests.
* Don't use `Should().NotThrow` in the assert. 
  
### Code style
* String formatting should use Invariant culture when formatting non-strings.
* Lines should not be wider than 130 characters.
* Prefer `is null` and `is not null` over `!=/== null`

dennisdoomen:
Moved to https://github.com/fluentassertions/fluentassertions/wiki/Design-Guidelines

PR #1475, opened 2021-01-18
More use of Invariant culture

jnyrup:
While working on another PR I had used `Convert.ChangeType(object, Type)` instead of `Convert.ChangeType(object, Type, IFormatProvider)` but I didn't get a build error from any analyzer.

It is handled by CA1305 which I [initially](https://github.com/fluentassertions/fluentassertions/pull/1283/files#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5bR130) disabled - as I recall due to being too noisy at that point in time.

Enabling it helped find an inconsistency where `becauseArgs` were formatted using `CurrentCulture` instead of `InvariantCulture`.

I've added some tests in `StringComparisonSpecs.cs`, but I omitted testing the addition of `Invariant` in `AutoConversionStep`, `DictionaryEquivalencyStep` and `EnumerableEquivalencyValidator`.
Partly because I found it difficult to write a test of good quality.

### DIFF
--- a/Src/FluentAssertions/AndWhichConstraint.cs
+++ b/Src/FluentAssertions/AndWhichConstraint.cs
@@ -43,10 +43,8 @@ namespace FluentAssertions
                     matchedElements.Select(
                         ele => "\t" + Formatter.ToString(ele)));
 
-                string message = string.Format(
-                    "More than one object found.  FluentAssertions cannot determine which object is meant.  Found objects:{0}{1}",
-                    Environment.NewLine,
-                    foundObjects);
+                string message = "More than one object found.  FluentAssertions cannot determine which object is meant."
+                              + $"  Found objects:{Environment.NewLine}{foundObjects}";
 
                 Services.ThrowException(message);
             }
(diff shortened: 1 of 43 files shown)


Src/FluentAssertions/AndWhichConstraint.cs
```diff
@@ -43,10 +43,8 @@ public AndWhichConstraint(TParentConstraint parentConstraint, IEnumerable<TMatch
                     matchedElements.Select(
                         ele => "\t" + Formatter.ToString(ele)));
 
-                string message = string.Format(
-                    "More than one object found.  FluentAssertions cannot determine which object is meant.  Found objects:{0}{1}",
-                    Environment.NewLine,
-                    foundObjects);
+                string message = "More than one object found.  FluentAssertions cannot determine which object is meant."
```
dennisdoomen, line 46: Would have been nice if these changes were kept in a separate commit. The PR changes now mix the invariant changes and other things.

dennisdoomen (approved):
Can we add the expected guidelines to https://github.com/fluentassertions/fluentassertions/issues/1340?

Src/FluentAssertions/AndWhichConstraint.cs
jnyrup: It is related, as it otherwise triggered `string.Format` without providing an `IFormatProvider`.

I can split the commit up into two with:
* the cases that rightfully needed to ignore culture, and
* those where the changes are to please the analyzer. (e.g. when all arguments to be formatted are already strings)

dennisdoomen: Never mind.

PR #1478, opened 2021-01-20
Add WithResult to TaskCompletionSourceAssertion

lg2de:
While further development of tests using async assertions (async methods and TaskCompletionSource) I found that asserting the result is not as fluent as possible.
Therefore I provide new idea to assert the result.

If you agree I'll complete with
* extention method for Task<AndWhichConstraint<GenericAsyncFunctionAssertions<TResult>, TResult>>
* documentation, release notes, etc.
* api validation update

### DIFF
--- a/Src/FluentAssertions/AsyncAssertionsExtensions.cs
+++ b/Src/FluentAssertions/AsyncAssertionsExtensions.cs
@@ -1,0 +1,55 @@
+using System.Threading.Tasks;
+using FluentAssertions.Specialized;
+
+namespace FluentAssertions
+{
+    public static class AsyncAssertionsExtensions
+    {
+#pragma warning disable AV1755 // "Name of async method ... should end with Async"; Async suffix is too noisy in fluent API
+        /// <summary>
+        /// Asserts that the completed <see cref="Task{TResult}"/> provides the specified result.
+        /// </summary>
+        /// <param name="task">The <see cref="GenericAsyncFunctionAssertions{T}"/> containing the <see cref="Task{TResult}"/>.</param>
+        /// <param name="expected">The expected value.</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
+        /// </param>
+        public static async Task<AndWhichConstraint<GenericAsyncFunctionAssertions<T>, T>> WithResult<T>(
+            this Task<AndWhichConstraint<GenericAsyncFunctionAssertions<T>, T>> task,
+            T expected, string because = "", params object[] becauseArgs)
+        {
+            var andWhichConstraint = await task;
+            var subject = andWhichConstraint.Subject;
+            subject.Should().Be(expected, because, becauseArgs);
+
+            return andWhichConstraint;
+        }
+
+        /// <summary>
+        /// Asserts that the completed <see cref="TaskCompletionSource{TResult}"/> provides the specified result.
+        /// </summary>
+        /// <param name="task">The <see cref="TaskCompletionSourceAssertions{T}"/> containing the <see cref="TaskCompletionSource{TResult}"/>.</param>
+        /// <param name="expected">The expected value.</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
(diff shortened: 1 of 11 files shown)


dennisdoomen (approved):
Please also update the release notes

Src/FluentAssertions/AsyncAssertionsExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <param name="expected">The expected value.</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
+        /// </param>
+        /// <typeparam name="T"></typeparam>
+        public static async Task WithResult<T>(this Task<AndWhichConstraint<TaskCompletionSourceAssertions<T>, T>> task,
```
jnyrup, line 22: Should this return a `Task<AndWhichConstraint<TaskCompletionSourceAssertions<T>, T>>`?

lg2de: This should be possible.
What is the example (test) you would like to build?

jnyrup: My thought was to allow further chaining.
I don't have a concrete example at hand.

lg2de: While trying to implement that I found that with the option to chain these assertions I run into same problem I had before implementing this extension: I need to await the result before chaining.

I feel we need a new `AsyncAndContraint` or similar to chain the await of the `Task`s, isn't it?
For now I stopped to implement this option.

Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
```diff
@@ -54,6 +55,53 @@ public async Task When_task_completes_fast_async_it_should_succeed()
...
+
+                (await funcSubject.Should(timer).CompleteWithinAsync(100.Milliseconds()))
+                    .Which.Should().Be(42);
+            };
+
+            taskFactory.SetResult(99);
+            timer.Complete();
+
+            // Assert
+            // TODO message currently shows "Expected (await funcSubject to be...", but should be "Expected funcSubject to be...",
```
lg2de, line 78: While having deep look on the generated failure message I found the following current result:
`Expected (await funcSubject to be...`
The brace should be removed, isn't it @dennisdoomen , @jnyrup ?
Should we remove `await` too?

What should be the correct messge?
`Expected await funcSubject to be...` or `Expected funcSubject to be...`

This should be tracked separately!?

jnyrup, line 78: I haven't used it myself yet, but I think decorating `WithResult` with [`[CustomAssertion]`](https://fluentassertions.com/introduction#subject-identification) can solve this.

eNeRGy164: You could replace `await` with 'awaitable' to make it more fluent and keep it distinct from a non-awaited execution?

dennisdoomen: I like the suggestion @eNeRGy164 made.

lg2de: The `WithResult` extension is not affected.
The previous one is affected because of `(await foo).Which.Should().Be(...)`.

lg2de: @eNeRGy164, you mean the generated error message should contain "awaitable" instead of "await"!?

eNeRGy164: Yes, so to pick your example

`Expected awaitable funcSubject to be...`

lg2de: I've moved this issue to #1488.

Src/FluentAssertions/AsyncAssertionsExtensions.cs
lg2de: Finally I changed my mind...
Even the explicit use of the result (with `(await ...).And...`) is not fluent api, but returning the constraint allows other extension methods, I think.

jnyrup (approved):

dennisdoomen:
@lg2de can you please rebase the PR so we can merge it?

PR #1375, opened 2020-08-12
Move [Not]HaveFlag to new EnumAssertions

jnyrup:
Motivations:
* To get rid of `HaveFlag` from `ObjectAssertions` as most objects are not enums
* Additional compile time safety

I've moved `[Not]HaveFlag` to the new `EnumAssertions` which derives from the generic `ObjectAssertions` introduced in #1371.

Optimally we would want to create
```cs
Should<TEnum>(this TEnum @enum)
    where TEnum : struct, Enum
    => new EnumAssertions(@enum);
```
but we have a tradition of not introducing overloads of `Should` directly generic in `T`.

Current state of develop branch
```cs
// Uses ObjectAssertions
TestEnum.One.Should().HaveFlag(TestEnum.One);
TestEnum.One.Should().HaveFlag(OtherEnum.First); // compiles but fails at runtime
TestEnum.One.Should().HaveFlag(null); // compiles but fails at runtime
((TestEnum?)TestEnum.One).Should().HaveFlag(TestEnum.One);
((TestEnum?)TestEnum.One).Should().HaveFlag(OtherEnum.First); // compiles but fails at runtime
((TestEnum?)TestEnum.One).Should().HaveFlag(null); // compiles but fails at runtime
```

With this PR
```cs
// Uses EnumAssertions
TestEnum.One.Should().HaveFlag(TestEnum.One);
TestEnum.One.Should().HaveFlag(OtherEnum.First); // compiles but fails at runtime
//TestEnum.One.Should().HaveFlag((Enum)null); // CS0453: The type 'Enum' must be a non-nullable value type in order to use it as parameter 'T'
((TestEnum?)TestEnum.One).Should().HaveFlag(TestEnum.One);
((TestEnum?)TestEnum.One).Should().HaveFlag(OtherEnum.First); // compiles but fails at runtime
//((TestEnum?)TestEnum.One).Should().HaveFlag((Enum)null); // CS0453: The type 'Enum' must be a non-nullable value type in order to use it as parameter 'T'
```

For *additional* compile-time safety, you can also create a dedicated `EnumAssertions` for your favorite `enum`.

```cs
public class TestEnumAssertions : EnumAssertions<TestEnum, TestEnumAssertions>
{
    public TestEnumAssertions(TestEnum subject)
        : base(subject)
    {
    }
}

public static class Ext
{
    public static TestEnumAssertions Should(this TestEnum testEnum) => 
        new TestEnumAssertions(testEnum);
}
```

```cs
[...]

PR #1479, opened 2021-01-23
Rework EnumAssertions

### LINKED ISSUE #1204: Comparing `enum` and `int`
### Description
Before #1202 we had an asymmetry in how we compared an `enum` and an `int`, as it worked in one direction, but not the other. 

```csharp
enum MyEnum
{
    One = 1
}

[Fact]
public void when_comparing_an_enum_to_a_numeric_it_should_TODO()
{
    MyEnum.One.As<object>().Should().Be(1);
	// passes
}

[Fact]
public void when_comparing_a_numeric_to_an_enum_it_should_TODO()
{
    1.As<object>().Should().Be(MyEnum.One);
	// fails
}
```

After #1202, none of them passes.
There were no tests covering these cases.
Should we allow comparing `enum`s with numeric types?

### LINKED ISSUE #1403: Enum comparison does not work as expected
### Description

According to the documentation, with the standard method Should().Be(), the Enum comparison should work the same as the .NET Implementation Enum.Equals() method. 

However, with the following code this does not work for me. 

```csharp
public enum MyEnumResult
{
    Successfully,
    Failed
}

public enum MySecondEnumResult
{
    Successfully,
    AnotherValue,
    Failed
}

[Test]
public void MyTest()
{
    MyEnumResult result = _sut.Get(_id);

    //This works as expected
    result.Should().Be(MyEnumResult.Successfully);
    //This works also, but should not
    result.Should().Be(MySecondEnumResult.Successfully);
}
```

Although the two enums are not of the same type, the test does not fail. Apparently only the underlying value is tested, but not the type.

### Versions

* Fluent Assertions version is 5.10.3
* .NET framework 4.7.2

jnyrup:
New `EnumAssertions` to follow up #1375 and to solve #1204

`Be()` now expects an expectation of the same type of enum, which was an issue that gave false negatives, e.g. #1403.

More lightweight/relevant IntelliSense as it doesn't inherit assertion methods from `ReferenceAssertions`.

Before:
image

After:
image

I've excluded `[Not]BeEquivalentTo(object)` as all cases that I can think of are now covered by:
* `[Not]HaveSameValueAs`
* `[Not]HaveSameNameAs`
* `[Not]HaveValue(decimal)`

One question that remains for me, it whether it is too confusing that `HaveValue` is both used in:
* `NullableEnumAssertions` to assert if it is not null, and in
* `EnumAssertions` to assert on the underlying numeric value.

cc: @lg2de

This fixes #1204 
This fixes #1403

### DIFF
--- a/Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
+++ b/Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
@@ -15,10 +15,9 @@ namespace FluentAssertions.Equivalency
         /// </summary>
         public bool CanHandle(IEquivalencyValidationContext context, IEquivalencyAssertionOptions config)
         {
-            Type subjectType = config.GetExpectationType(context.RuntimeType, context.CompileTimeType);
+            Type expectationType = config.GetExpectationType(context.RuntimeType, context.CompileTimeType);
 
-            return (subjectType?.IsEnum == true) ||
-                   (context.Expectation?.GetType().IsEnum == true);
+            return expectationType.IsEnum;
         }
 
         /// <summary>
@@ -96,9 +95,9 @@ namespace FluentAssertions.Equivalency
             if (o.GetType().IsEnum)
             {
                 string typePart = o.GetType().Name;
-                string namePart = Enum.GetName(o.GetType(), o);
+                string namePart = o.ToString().Replace(", ", "|", StringComparison.Ordinal);
                 string valuePart = v.Value.ToString(CultureInfo.InvariantCulture);
-                return $"{typePart}.{namePart}({valuePart})";
+                return $"{typePart}.{namePart} {{{{value: {valuePart}}}}}";
             }
 
             return v.Value.ToString(CultureInfo.InvariantCulture);
--- a/Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
+++ b/Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
@@ -33,6 +33,16 @@ namespace FluentAssertions.Equivalency
         public bool Handle(IEquivalencyValidationContext context, IEquivalencyValidator parent,
             IEquivalencyAssertionOptions config)
         {
+            Execute.Assertion
+                .ForCondition(context.Subject?.GetType().IsEnum == true)
+                .FailWith(() =>
+                {
+                    decimal? expectationsUnderlyingValue = ExtractDecimal(context.Expectation);
+                    string expectationName = GetDisplayNameForEnumComparison(context.Expectation, expectationsUnderlyingValue);
+
+                    return new FailReason($"Expected {{context:enum}} to be equivalent to {expectationName}{{reason}}, but found {{0}}.", context.Subject);
+                });
+
             switch (config.EnumEquivalencyHandling)
             {
                 case EnumEquivalencyHandling.ByValue:
@@ -68,7 +78,7 @@ namespace FluentAssertions.Equivalency
 
         private static void HandleByName(IEquivalencyValidationContext context)
         {
-            string subject = context.Subject?.ToString();
+            string subject = context.Subject.ToString();
             string expected = context.Expectation.ToString();
 
             Execute.Assertion
@@ -89,18 +99,13 @@ namespace FluentAssertions.Equivalency
         {
             if (o is null || v is null)
             {
-                return "null";
+                return "<null>";
             }
 
-            if (o.GetType().IsEnum)
-            {
-                string typePart = o.GetType().Name;
-                string namePart = o.ToString().Replace(", ", "|", StringComparison.Ordinal);
-                string valuePart = v.Value.ToString(CultureInfo.InvariantCulture);
-                return $"{typePart}.{namePart} {{{{value: {valuePart}}}}}";
-            }
-
-            return v.Value.ToString(CultureInfo.InvariantCulture);
+            string typePart = o.GetType().Name;
+            string namePart = o.ToString().Replace(", ", "|", StringComparison.Ordinal);
+            string valuePart = v.Value.ToString(CultureInfo.InvariantCulture);
+            return $"{typePart}.{namePart} {{{{value: {valuePart}}}}}";
         }
 
         private static decimal? ExtractDecimal(object o)
--- a/Src/FluentAssertions/Primitives/EnumAssertions.cs
+++ b/Src/FluentAssertions/Primitives/EnumAssertions.cs
@@ -1,14 +1,16 @@
 ﻿using System;
+using System.Globalization;
 using FluentAssertions.Execution;
 
 namespace FluentAssertions.Primitives
 {
     /// <summary>
-    /// Contains a number of methods to assert that an <see cref="Enum"/> is in the expected state.
+    /// Contains a number of methods to assert that a <typeparamref name="TEnum"/> is in the expected state.
     /// </summary>
-    public class EnumAssertions : EnumAssertions<Enum, EnumAssertions>
+    public class EnumAssertions<TEnum> : EnumAssertions<TEnum, EnumAssertions<TEnum>>
+        where TEnum : struct, Enum
     {
-        public EnumAssertions(Enum subject)
+        public EnumAssertions(TEnum subject)
             : base(subject)
         {
         }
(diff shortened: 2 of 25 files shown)


jnyrup:
I'd especially like an extra eye on the negative statements, if they handle nulls as expected.
E.g. how `null.Should().NotHaveFlag()` should behave.

docs/_pages/upgradingtov6.md
```diff
@@ -7,3 +7,62 @@ sidebar:
...
+```
+
+In Fluent Assertions v6 enums are now handled by `EnumAssertions` which should cover the same use cases and even more, but without the old bugs and unexpected surprises.
+Technically nullable enums are handled by `NullableEnumAssertions` which in addition to the assertions mentioned below have `[Not]BeNull` and `[Not]HaveValue`.
+`Be` and `HaveFlag` now calls directly into `Enum.Equals` and `Enum.HasFlag`, as you would expect, and any try to compare an enum with another type of enum or integer leads to a compilation error.
+If you want to compare two enums of different types, you can use `HaveSameValueAs` or `HaveSameNameAs` depending on how _you_ define equality for different enums.
+Lastly, if you want to verify than an enum has a specific integral value, you can use `HaveValue`.
+
+When comparing object graphs with enum members, we have constrained when we consider them to be equivalent.
+an enum is now only considered to be equivalent to another enum, but you can control whether they should equal by name or by value.
```
dennisdoomen, line 55: 🙃 Sentence should start with a capital.

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
```diff
@@ -90,18 +99,19 @@ private static string GetDisplayNameForEnumComparison(object o, decimal? v)
...
             if (o is null || v is null)
             {
-                return "null";
+                return "<null>";
             }
 
-            if (o.GetType().IsEnum)
+            string typePart = o.GetType().Name;
+            string namePart = o.ToString();
+            if (namePart.Contains(",", StringComparison.Ordinal))
```
dennisdoomen, line 107: ❓ Where does this `,` come from?

Src/FluentAssertions/Primitives/EnumAssertions.cs
```diff
@@ -68,22 +294,23 @@ public EnumAssertions(TEnum subject)
...
-                .Then
-                .ForCondition(!Subject.HasFlag(unexpectedFlag))
-                .FailWith("Did not expect the enum to have flag {0}{reason}.", unexpectedFlag);
+                .ForCondition(SubjectInternal?.HasFlag(unexpectedFlag) != true)
+                .FailWith("Expected the enum to not have flag {0}{reason}.", unexpectedFlag);
 
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
+
+        private static decimal GetValue<T>(T @enum)
```
dennisdoomen, line 308: 🙃 I'm not that fond of this particular usage of an expression-bodied method.

Src/FluentAssertions/Primitives/NullableEnumAssertions.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// Asserts that a nullable <typeparamref name="TEnum"/> value is not <c>null</c>.
+        /// </summary>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])"/> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndWhichConstraint<TAssertions, TEnum> HaveValue(string because = "", params object[] becauseArgs)
```
dennisdoomen, line 42: 🤔 I noticed that we have `HaveValue` and `NotBeNull`. I guess this is just for convenience as well as consisteny with other similar cases, right?

docs/_pages/objectgraphs.md
```diff
@@ -183,6 +183,9 @@ An option to compare an `Enum` only by name is also available, using the followi
 orderDto.Should().BeEquivalentTo(expectation, options => options.ComparingEnumsByName());
 ```
 
+Note that even though an enum's underlying value equals a numeric value or the enum's name equals some string value, we do not consider those to be equivalent.
+In other words, enums are only considered to be equivalent to other enums, but you can control whether they should equal by name or by value.
```
dennisdoomen, line 187: > In other words, enums are only considered to be equivalent to other enums,

🔧  ...of the same type or another

docs/_pages/releases.md
```diff
@@ -12,6 +12,7 @@ sidebar:
 **What's New**
 * Added `WithParameterName` extension to ease asserting on the parameter name for a thrown `ArgumentException` - [#1466](https://github.com/fluentassertions/fluentassertions/pull/1466).
 * Added `NotCompleteWithinAsync` to `TaskCompletionSourceAssertions` - [#1474](https://github.com/fluentassertions/fluentassertions/pull/1474).
+* Major overhaul on how enums are handled, see the [Migration Guide](/upgradingtov6) for more details - [#1479](https://github.com/fluentassertions/fluentassertions/pull/1479).
```
dennisdoomen, line 15: 🤔 Isn't this a breaking change?

docs/_pages/upgradingtov6.md
```diff
@@ -7,3 +7,62 @@ sidebar:
   nav: "sidebar"
 ---
 
+## Enums ##
```
dennisdoomen, line 10: 👍 You really did a great job here.

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
jnyrup: It's for flags enums. `(MyEnum.One | MyEnum.Two).ToString()` is `"One, Two"`

docs/_pages/releases.md
jnyrup: It includes both new assertions and breaking changes, but I thought putting in both places was maybe too much. Do you think it fits better under breaking changes?

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
jnyrup: `Enum.GetName` returns null for flags enums.
```cs
[Flags] enum MyEnum { One = 1, Two = 2 }
Enum.GetName(typeof(MyEnum), MyEnum.One|MyEnum.Two)
// null
```

So this was my attempt format such flags enums, but I admit that I was never fully satisfied with it.

See e.g. this test in the PR
```cs
[Fact]
public void When_enum_does_not_have_specified_flag_it_should_fail_with_a_descriptive_message()
{
    // Arrange
    TestEnum someObject = TestEnum.One | TestEnum.Two;

    // Act
    Action act = () => someObject.Should().HaveFlag(TestEnum.Three, "we want to test the failure {0}", "message");

    // Assert
        .WithMessage("Expected*have flag TestEnum.Three(4)*because we want to test the failure message*but found TestEnum.{One, Two}(3).");
}
```

Src/FluentAssertions/Primitives/NullableEnumAssertions.cs
jnyrup: Yes, I included both for consistency with other `NullableSomethingAssertions`.

docs/_pages/releases.md
jnyrup: Added the new assertions to "What's new" and the link to upgrade notes in "Breaking changes"

Src/FluentAssertions/Equivalency/EnumEqualityStep.cs
dennisdoomen: What about:

`Expected*have flag TestEnum.Three {value: 4}*because we want to test the failure message*but found TestEnum.One|Two {value: 3}.`

Reasoning:
* Usually we use curly braces to denote internals
* Flags enums are combined using the pipe symbol

jnyrup: Looks much better!
I've changed the formatting to your suggestions.

dennisdoomen (approved):

PR #1490, opened 2021-02-04
Remove "Async" postfix from all assertions on asynchronous operations

lg2de:
As discussed on slack we should remove `Async` postfix from all assertions to get better **fluent** api.

### DIFF
--- a/Src/FluentAssertions/ExceptionAssertionsExtensions.cs
+++ b/Src/FluentAssertions/ExceptionAssertionsExtensions.cs
@@ -4,12 +4,12 @@ using System.Threading.Tasks;
 using FluentAssertions.Execution;
 using FluentAssertions.Specialized;
 
+#pragma warning disable AV1755 // "Name of async method ... should end with Async"; Async suffix is too noisy in fluent API
+
 namespace FluentAssertions
 {
     public static class ExceptionAssertionsExtensions
     {
-#pragma warning disable AV1755 // "Name of async method ... should end with Async"; Async suffix is too noisy in fluent API
-
         /// <summary>
         /// Asserts that the thrown exception has a message that matches <paramref name="expectedWildcardPattern" />.
         /// </summary>
@@ -128,7 +128,5 @@ namespace FluentAssertions
         {
             return (await task).WithParameterName(paramName, because, becauseArgs);
         }
-
-#pragma warning restore AV1755
     }
 }
--- a/Src/FluentAssertions/ExceptionAssertionsExtensions.cs
+++ b/Src/FluentAssertions/ExceptionAssertionsExtensions.cs
@@ -4,8 +4,6 @@ using System.Threading.Tasks;
 using FluentAssertions.Execution;
 using FluentAssertions.Specialized;
 
-#pragma warning disable AV1755 // "Name of async method ... should end with Async"; Async suffix is too noisy in fluent API
-
 namespace FluentAssertions
 {
     public static class ExceptionAssertionsExtensions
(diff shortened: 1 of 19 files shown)


Src/FluentAssertions/ExceptionAssertionsExtensions.cs
```diff
@@ -4,12 +4,12 @@
 using FluentAssertions.Execution;
 using FluentAssertions.Specialized;
 
+#pragma warning disable AV1755 // "Name of async method ... should end with Async"; Async suffix is too noisy in fluent API
```
dennisdoomen, line 7: I think it's better to just disable that rule globally.

jnyrup:
A case to take into consideration before eventually merging this.

Together with the removal of blocking `Throw` from async assertions, upgrading from v5 to v6 changes the behavior of `Throw` from sync to async, but everything will continue to compile.
If `Foo` now throws an exception the assertion won't fail, as it's not awaited.

```cs
[TestClass]
public class UnitTest1
{
    public Task Foo() => Task.FromException(new Exception());

    [TestMethod]
    public void MyTest()
    {
        this.Invoking(e => e.Foo()).Should().NotThrow();
    }
}
```

I'm not ultimately against this change, right now I just think the potential downsides might be greater than the upsides.
Not all our users will meticulously read the release notes or write their tests using the red-green methodology.

lg2de:
Yes, you are right, @jnyrup. :(
Hope we find a good solution therefore. Let's suspend this for some time.

dennisdoomen:
> Together with the removal of blocking `Throw` from async assertions, upgrading from v5 to v6 changes the behavior of `Throw` from sync to async, but everything will continue to compile.

Man, how did you figure that out?

jnyrup:
> Man, how did you figure that out?

After reading these two blog posts by Eric Lippert and Jon skeet, it's no longer a matter of _if_ changes to a public surface breaks [something](https://xkcd.com/1172/), but _how_ they break it and how discoverable the breakage is.
https://ericlippert.com/2012/01/09/every-public-change-is-a-breaking-change/
https://codeblog.jonskeet.uk/2018/04/13/backward-compatibility-and-overloading/

Add to that a gut feeling that tends to warn me if I might me missing out one something and need to think more.

Src/FluentAssertions/ExceptionAssertionsExtensions.cs
lg2de: done

dennisdoomen:
Given the concerns we identified with potential breaking changes, we decided to keep the current API as is.

lg2de:
Not changing this will result in inconsistent (non fluent) API.
Please let me propose a solution before you close the issue.

jnyrup:
We're _just_ about to tag beta1, which means a stop for breaking changes, so if you have another proposal you have be _really_ quick.

lg2de:
The implementation of my proposal will take some time. So, let's discuss the idea to decide whether this can become a solution.

Implementing a source code analyzer could verify the usage of the relevant methods.
Currently there is the [warning CS4014](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs4014). But this applies only to usages within async methods.
An additional warning would point the user to the breaking change.
To prevent from side effects the analyzer can/should focus on the FA methods.

BTW: xunit is also coming with an [analyzer](https://github.com/xunit/xunit.analyzers).

What do you think? Would this be a solution?

If yes, I recommend to take this PR with the current changes into beta-1.
The analyzer could be an update without changing the API again.

dennisdoomen:
Hi @lg2de. We really appreciate the effort you are willing to put in this, but we've concluded that all of this work is not worth the time and extra maintenance work that it will bring with it. I hope you understand.

lg2de:
I do not agree, because now we dot have a real fluent API.
Some asynchronous methods are named with suffix, others are not.
I am saddened that a consistent FLUENT API is not your primary goal.

dennisdoomen:
It is, but we also have to consider the impact on existing users, the discoverability of things and the maintenance it is going to cost us. Like every software projects, it's a trade-off.

Issue #33, opened 2013-11-03
Fails when comparing objects if one property is nullable


dennisdoomen:
See [StackOverflow](http://stackoverflow.com/questions/19714080/fluentassertions-fails-when-comparing-objects-if-one-property-is-nullable) topic.

PR #1494, opened 2021-02-13
Improve `Using<TMember>`

### LINKED ISSUE #799: Changing the behavior of WhenTypeIs for value types
Currently `WhenTypeIs` checks the runtime type of the expectation.
This can cause a `NullReferenceException` in `CreateFromEquivalencyValidationContext` in the following example.

```c#
var obj1 = new { date = (DateTime?)null };
var obj2 = new { date = (DateTime?)default(DateTime) };
var actual = new[] { obj1, obj2 };
var expected = new[] { obj2, obj1 };

actual.Should().BeEquivalentTo(expected, options => options
	.Using<DateTime>(ctx => ctx.Subject.Should().BeCloseTo(ctx.Expectation, 1.Seconds()))
	.WhenTypeIs<DateTime>());
```

As the runtime type of `default(DateTime)` is `DateTime`, `WhenTypeIs` will match.
`CreateFromEquivalencyValidationContext` will then cast the `null` subject to a `DateTime` causing a `NullReferenceException`.

Another case is when a struct implements an interface, and `CreateFromEquivalencyValidationContext` cast a null into the struct type.

I propose that we change the logic of `WhenTypeIs` to consider whether the `RuntimeType` or `CompileTimeType` is a value type.

I have implemented the proposal [here](
https://github.com/jnyrup/fluentassertions/commit/cd1227f52bee262933a3a98b5b5b893baa4eee4).
It is not a compile-breaking change, but if you relied on `WhenTypeIs<DateTime>` to match `DateTime?` it will break your code.
As there aren't *that* many tests of `WhenTypeIs` I wanted to start a discussion on the intended/expected behavior before submitting a PR:

This was inspired by the problem presented in #778 .

jnyrup:
Restrict `WhenTypeIs<TMemberType>` to types assignable to `TMember`.
This should catch `NullReferenceException` or `InvalidCastException` in `CreateFromEquivalencyValidationContext` at compile-time.
image

The snippet below might have worked, but it assumes that the subject and expectation are both non-null.
```cs
.Using<int>()
.WhenTypeIs<int?>()
```

Instead, you now have to make this assumption more explicit by using `Nullable<int>.Value` 
```cs
Using<int?>(e => e.Subject.Value.Should().Be(e.Subject.Value))
```

Make `WhenTypeIs<TMemberType>` optional unless when chaining further.
For the most cases `TMemberType` is exactly `TMember` and in those cases `WhereTypeIs<>()` is noisy.
By adding an implicit cast operator we can avoid that.
The exception being when we need more chaining after `Using<>`, see the added `When_overriding_with_custom_assertion_it_should_be_chainable`.

Avoid trying to cast null to non-nullable type.
This handles when subject or expectation is null, but `Using<int>` would either:
* cast `null` to `int` leading to a NullReferenceException`, or
* use `default(int)` when `Expectation` is `null`, which would make `null` and `0` equal.

As far as I can read from #33 the changes in `When_a_nullable_property_is_overriden_with_a_custom_assertion_it_should_use_it` are fine, as the original bug report was about `using<long?>` didn't match a non-null `long?`

This fixes #799

### DIFF
--- a/Src/FluentAssertions/Equivalency/AssertionContext.cs
+++ b/Src/FluentAssertions/Equivalency/AssertionContext.cs
@@ -24,12 +24,10 @@ namespace FluentAssertions.Equivalency
 
         internal static AssertionContext<TSubject> CreateFromEquivalencyValidationContext(IEquivalencyValidationContext context)
         {
-            TSubject expectation = (context.Expectation is not null) ? (TSubject)context.Expectation : default;
-
             return new AssertionContext<TSubject>(
                 context.CurrentNode,
                 (TSubject)context.Subject,
-                expectation,
+                (TSubject)context.Expectation,
                 context.Reason.FormattedMessage,
                 context.Reason.Arguments);
         }
--- a/Src/FluentAssertions/Equivalency/AssertionRuleEquivalencyStep.cs
+++ b/Src/FluentAssertions/Equivalency/AssertionRuleEquivalencyStep.cs
@@ -83,12 +83,19 @@ namespace FluentAssertions.Equivalency
 
             if (subjectIsValidType && expectationIsValidType)
             {
+                if ((subjectIsNull || expectationIsNull) && !CanBeNull<TSubject>())
+                {
+                    return false;
+                }
+
                 assertion(AssertionContext<TSubject>.CreateFromEquivalencyValidationContext(context));
                 return true;
             }
 
             return false;
         }
+
+        private static bool CanBeNull<T>() => !typeof(T).IsValueType || Nullable.GetUnderlyingType(typeof(T)) is not null;
 
         /// <summary>
         /// Returns a string that represents the current object.
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
@@ -698,6 +698,7 @@ namespace FluentAssertions.Equivalency
             /// <typeparamref name="TMemberType" />
             /// </summary>
             public TSelf WhenTypeIs<TMemberType>()
+                where TMemberType : TMember
             {
                 When(info => info.RuntimeType.IsSameOrInherits(typeof(TMemberType)));
                 return options;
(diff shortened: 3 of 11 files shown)


jnyrup:
Missing release notes.

Do you think this requires a section in the upgrading guide?

dennisdoomen (approved):

dennisdoomen:
> Do you think this requires a section in the upgrading guide?

I think so. It took me a couple of times before I understood what this PR was trying to do.

jnyrup:
Hmm... wondering if the implicit cast operator creates more confusion than value?
With this PR one can omit `WhenTypeIs<T>()` but _only_ when `Using<T>()` is the last part of the chain.

dennisdoomen:
> Hmm... wondering if the implicit cast operator creates more confusion than value?
> With this PR one can omit `WhenTypeIs<T>()` but _only_ when `Using<T>()` is the last part of the chain.

But it _does_ simplify the use of this syntax.

jnyrup:
Yes, it simplifies the usage, when having `using` at the last term.

To give a more concrete example of the potential confusion.

Say I start out with this snippet and want to exclude a property `ExcludeMe`
```cs
subject.Should().BeEquivalentTo(expected, opt => opt.
    .Using<int>(...)
```

Adding `Excluding()` directly after `Using<int>()` does not compile, as `Excluding` is not a member of `Restriction` and the implicit conversion has not kicked in yet.
```cs
subject.Should().BeEquivalentTo(expected, opt => opt.
    .Using<int>(...)
    .Excluding(e => e.ExcludeMe)
```

To make it work I have to either put `Excluding` before `Using`
```cs
subject.Should().BeEquivalentTo(expected, opt => opt.
    .Excluding(e => e.ExcludeMe)
    .Using<int>(...)
```

or add `WhenTypeIs` as one has to do right now.
```cs
subject.Should().BeEquivalentTo(expected, opt => opt.
    .Using<int>(...)
    .WhenTypeIs<int>()
    .Excluding(e => e.ExcludeMe)
```

dennisdoomen:
I see your point. In the current implementation, it's immediately clear what your options are since Intellisense knows what can be chained at any point in time.

jnyrup:
> I see your point. In the current implementation, it's immediately clear what your options are since Intellisense knows what can be chained at any point in time.

Was that comment towards keeping or leaving out the implicit conversion operator?

dennisdoomen:
> Was that comment towards keeping or leaving out the implicit conversion operator?

Merely that I got your doubts about the whole thing. But I don't have a strong opinion on this.

PR #1496, opened 2021-02-21
Simplify signature of `[Not]Contain(strings)`

jnyrup:
Since the introduction of `StringCollectionAssertions` in #42 `Contain(IEnumerable<string>)` have had an extra `object becauseArg` argument.

I can only speculate that it had to do with overload resolution to avoid breaking changes.


dennisdoomen (approved):
I have no clue either ;-)

Issue #1480, opened 2021-01-24
Idea for new collection assertion


vanya-lebedev:
Hello, fluentassertions community, 

I would like to propose a new overload of the BeEquivalentTo method for collection assertions. Here's an example:

```C#
plannedOrders.Should().BeEquivalentTo(
    order => order.Type == OrderType.Purchase && order.Quantity == 5 && order.Date == tomorrow,
    order => order.Type == OrderType.Transfer && order.SourceLocation == distributionCenter && order.Quantity == 3 && order.Date == today,
    order => order.Type == OrderType.Transfer && order.SourceLocation == factory && order.Quantity == 5 && order.Date == today.AddDays(10)
);

```
Basically, the API should be similar to the existing collection.Should().BeEquivalentTo() API but instead of providing a collection of elements as input we would instead provide a collection of predicates that elements of the collection must match. Assertion should pass if it is possible to have a 1-1 mapping between elements of the collection and the predicates.

The proposed API is useful in scenarios when the collection has elements with many properties but only few of these are insteresting in the scope of a given test scenario.

Could you please let me know if there is interest in the community in this feature? I would be happy to contribute.

Best regards,
Vanya

jnyrup:
Hi Vanya.

This sounds like the existing `SatisfyRespectively` method.

Here is the example from our [documentation](https://fluentassertions.com/collections/).

```cs
var collection = new []
{
    new { Id = 1, Name = "John", Attributes = new string[] { } },
    new { Id = 2, Name = "Jane", Attributes = new string[] {"attr"} }
};
collection.Should().SatisfyRespectively(
    first =>
    {
        first.Id.Should().Be(1);
        first.Name.Should().StartWith("J");
        first.Attributes.Should().NotBeNull();
    },
    second =>
    {
        second.Id.Should().Be(2);
        second.Name.Should().EndWith("e");
        second.Attributes.Should().NotBeEmpty();
    });
```

A difference between `SatisfyRespectively` and what I understand from your proposal, it that `SatisfyRespectively` requires the subject and list of predicates to be in the same order, while `BeEquivalentTo` per default allows unordered 1:1 equivalency mappings.

You mention the desire for the proposed extension of `BeEquivalentTo` is because your element type has many properties, but only a small subset is relevant for the test.
Another way you might archive what you're seeking might be to use anonymous objects, as the expectation type decides what properties to compare.

```cs
var expected = new[]
{
[...]

Issue #1486, opened 2021-01-31
BeEquivalentTo is sensitive to input ordering


jnyrup:
### Description

Our algorithm to verify if two collections are equivalent when disregarding ordering, i.e. the default behavior, is sensitive to the order of inputs.

In the example below we proceed as follows:
* First match `subject[0]` to `expected[0]` and unmark them both for further consideration.
* The only remaining `subject[1]` and `expected[1]` do not match.
* We conclude that that `subject` and `expected` are not equivalent.

If we reverse the order of `expected`, the test passes.

A simple brute-forcing algorithm to solve this would be
```
for each permutation in permutations(expected)
  if subject is equivalent to permutation with strict ordering
	return
	
throw
```

... but that increases the complexity from O(n^2) to an unacceptable O(n!*n).

From my reading this morning I think this _might_ be modelled as the [Stable marriage with indifference](https://en.wikipedia.org/wiki/Stable_marriage_with_indifference) with two modifications:
* Only keep equivalent persons in each preference list.
* All remaining persons in each preference list are equally favored.

### Complete minimal example reproducing the issue

```csharp
var subject = new object[]
{
    new { Foo = 1, Bar = 2 },
    new { Foo = 1, Bar = 3 }
};

var expected = new object[]
{
    new { Foo = 1 },
    new { Foo = 1, Bar = 2 }
};

subject.Should().BeEquivalentTo(expected, opt => opt.WithoutStrictOrdering());
```

### Expected behavior:

Test should pass

### Actual behavior:

```
Expected property subject[1].Bar (of type int) to be 2, but found 3.
```

dennisdoomen:
I believe that right now, we already compare each item in subject with each item in expected.

vanya-lebedev:
As part of the pull request below I have implemented a simplified version of the Ford-Fulkerson algorithm for the maximum matching problem. I think that it can be reused for this issue as well. Complexity is O(N x E) where N is number of elements and E is total number of possible matches between elements and assertions. 

https://github.com/fluentassertions/fluentassertions/pull/1500

jnyrup:
Fiddled a bit with this, and replacing the current approach with a _correct_ implementation was straight forward.
```cs
private void AssertElementGraphEquivalencyWithLooseOrdering<T>(object[] subjects, T[] expectations)
{
    var predicatesList = expectations.Select<T, Expression<Func<object, bool>>>((e, i) => (object s) => TryToMatch(s, e, i));
[...]

PR #1500, opened 2021-02-27
Adds SelfReferencingCollectionAssertions.Satisfy method which is not sensitive to the order of elements and accepts predicate expressions.

vanya-lebedev:
Introduce a version of the `SelfReferencingCollectionAssertions.SatisfyRespecitvely` method which is not sensitive to the order of elements and accepts predicate expressions. 

Example of an assertion:
```CSharp
plannedOrders.Should().Satisfy(
    order => order.Type == OrderType.Purchase && order.Quantity == 5 && order.Date == tomorrow,
    order => order.Type == OrderType.Transfer && order.SourceLocation == distributionCenter && order.Quantity == 3 && order.Date == today,
    order => order.Type == OrderType.Transfer && order.SourceLocation == factory && order.Quantity == 5 && order.Date == today.AddDays(10)
);
```

Example of a failure message:
```
Expected collection to satisfy all predicates because we want to test formatting (args), but:
    
    The following predicates did not have matching elements:
    
    (_.Text == "two") AndAlso (_.Number == 2)
    
    The following elements did not match any predicate:
    
    FluentAssertions.Specs.Collections.GenericCollectionAssertionsSpecs+SomeClass
    {
       Number = 1
       Text = "one"
    }
    
    FluentAssertions.Specs.Collections.GenericCollectionAssertionsSpecs+SomeClass
    {
       Number = 3
       Text = "two"
    }
    
    FluentAssertions.Specs.Collections.GenericCollectionAssertionsSpecs+SomeClass
    {
       Number = 3
       Text = "three"
    }
```

The discussion about this API is tracked here:
https://github.com/fluentassertions/fluentassertions/issues/1480

To enable the necessary functionality three major changes were introduced:
- The `Satisfy` API was added to the `SelfReferencingCollectionAssertions` class. 
- The `MaximumMatchingSolver` class was added to encapsulate the simplified version of the Ford-Fulkerson algorithm to find matching between the asserted predicates and the elements of the collection. 
- The `PredicateLambdaExpressionValueFormatter` class was introduced to produce more human-readable descriptions of predicates in assertion failure messages.

## IMPORTANT 


dennisdoomen:
That's an impressive piece of work. Thanks for that.

docs/_pages/collections.md
```diff
@@ -168,3 +168,16 @@ collection.Should().SatisfyRespectively(
...
+```csharp
+var collection = new []
+{
+    new { Id = 1, Name = "John", Attributes = new string[] { } },
+    new { Id = 2, Name = "Jane", Attributes = new string[] {"attr"} }
+};
+
+collection.Should().Satisfy(
+    _ => _.Id == 2 && _.Name == "Jane" && _.Attributes == null,
+    _ => _.Id == 1 && _.Name == "John" && _.Attributes?.Length > 0);
```
jnyrup, line 182: This will not compile as `?.` cannot be used in expression trees.

Src/FluentAssertions/Formatting/PredicateLambdaExpressionValueFormatter.cs
```diff
@@ -0,0 +1,127 @@
...
+    /// The <see cref="PredicateLambdaExpressionValueFormatter" /> is responsible for formatting
+    /// boolean lambda expressions.
+    /// </summary>
+    public class PredicateLambdaExpressionValueFormatter : IValueFormatter
+    {
+        public bool CanHandle(object value) => value is LambdaExpression lambdaExpression && lambdaExpression.ReturnType == typeof(bool);
+
+        public string Format(object value, FormattingContext context, FormatChild formatChild)
+        {
+            var lambdaExpression = value as LambdaExpression;
```
jnyrup, line 21:
```suggestion
            var lambdaExpression = (LambdaExpression)value;
```

Src/FluentAssertions/Formatting/PredicateLambdaExpressionValueFormatter.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// </summary>
+        internal class AndOperatorChainExtractor : ExpressionVisitor
+        {
+            public List<Expression> AndChain { get; } = new List<Expression>();
+
+            public override Expression Visit(Expression node)
+            {
+                if (node.NodeType == ExpressionType.AndAlso)
+                {
+                    var binaryExpression = node as BinaryExpression;
```
jnyrup, line 114:
```suggestion
                    var binaryExpression = (BinaryExpression)node;
```

Src/FluentAssertions/Formatting/PredicateLambdaExpressionValueFormatter.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// <summary>
+        /// Expression visitor which can replace constant sub-expressions with constant values.
+        /// </summary>
+        internal class ConstantSubExpressionReductionVisitor : ExpressionVisitor
+        {
+            public override Expression Visit(Expression node)
+            {
+                if (ExpressionIsConstant(node))
+                {
+                    return Expression.Constant(Expression.Lambda(node).Compile().DynamicInvoke());
```
jnyrup, line 88:
```suggestion
                    return Expression.Constant(Expression.Lambda(node).Compile().DynamicInvoke());
```

Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingSolution.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Text;
```
jnyrup, line 5:
```suggestion
```

Src/FluentAssertions/Formatting/PredicateLambdaExpressionValueFormatter.cs
```diff
@@ -0,0 +1,127 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Runtime.Serialization;
+using System.Text;
```
jnyrup, line 7:
```suggestion
using System.Collections.Generic;
using System.Linq;
using System.Linq.Expressions;
```

Src/FluentAssertions/Formatting/PredicateLambdaExpressionValueFormatter.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// <summary>
+        /// Expression visitor which can detect whether the expression depends on parameters.
+        /// </summary>
+        internal class ParameterDetector : ExpressionVisitor
+        {
+            public bool HasParameters { get; private set; }
+
+            public override Expression Visit(Expression node)
+            {
+                // As soon as at least one parameter was found in the expression tree we should stop iterating (this is achieved by not callling base.Visit).
```
jnyrup, line 68:
```suggestion
                // As soon as at least one parameter was found in the expression tree we should stop iterating (this is achieved by not calling base.Visit).
```

Src/FluentAssertions/Formatting/PredicateLambdaExpressionValueFormatter.cs
```diff
@@ -0,0 +1,127 @@
...
+        private static Expression ReduceConstantSubExpressions(Expression expression)
+        {
+            return new ConstantSubExpressionReductionVisitor().Visit(expression);
+        }
+
+        /// <summary>
+        /// This step simplifies the lambda expression by removing unnecessary parentheses for root level chain of AND operators.
+        /// E.g. (_.Text == "two") AndAlso (_.Number == 3) AndAlso (_.OtherText == "foo")
+        /// Instead of ((_.Text == "two") AndAlso (_.Number == 3)) AndAlso (_.OtherText == "foo")
+        /// This simplification is only implemented for the chain of AND operators beacuse this is the most common predicate scenario.
```
jnyrup, line 49:
```suggestion
        /// This simplification is only implemented for the chain of AND operators because this is the most common predicate scenario.
```

docs/_pages/collections.md
```diff
@@ -168,3 +168,16 @@ collection.Should().SatisfyRespectively(
         second.Attributes.Should().NotBeEmpty();
     });
 ```
+
+If you need to perform individual assertions on all elements of a collection without setting expectation about the order of elements:
+```csharp
+var collection = new []
+{
+    new { Id = 1, Name = "John", Attributes = new string[] { } },
+    new { Id = 2, Name = "Jane", Attributes = new string[] {"attr"} }
```
jnyrup, line 177:
```suggestion
    new { Id = 2, Name = "Jane", Attributes = new string[] { "attr" } }
```

Src/FluentAssertions/Collections/SelfReferencingCollectionAssertions.cs
```diff
@@ -768,6 +769,61 @@ public AndConstraint<TAssertions> SatisfyRespectively(IEnumerable<Action<T>> exp
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
+        public AndConstraint<TAssertions> Satisfy(params Expression<Func<T, bool>>[] predicates)
```
jnyrup, line 772: The two overloads of `Satisfy` needs a summary.

jnyrup (commented):
Really nice work!
Only minor comments from here.

vanya-lebedev:
Dennis, Jonas, thanks a lot for your support!

Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionsSpecs.cs
```diff
@@ -1931,5 +1931,237 @@ public void When_inspectors_count_does_not_equal_asserting_collection_length_it_
...
+            var collection = new int[] { 1, 2 };
+
+            // Act
+            Action act = () => collection.Should().Satisfy(
+                element => element == 3,
+                element => element == 1,
+                element => element == 2);
+
+            // Assert
+            act.Should().Throw<XunitException>();
```
dennisdoomen, line 1982: 🤔  I would at least assert a part of the exception message (using wildcards) to ensure it the missing predicate is reported.

Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionsSpecs.cs
```diff
@@ -1931,5 +1931,237 @@ public void When_inspectors_count_does_not_equal_asserting_collection_length_it_
             act.Should().Throw<XunitException>().WithMessage("*because we want to test the failure*");
         }
         #endregion
+
+        #region Satisfy (Predicates)
+
+        [Fact]
+        public void When_collection_element_at_each_position_matches_predicate_at_same_position_Satisfy_should_not_throw()
```
dennisdoomen, line 1938: 🙃 Normally we try to avoid using the literal name of an assertion method in the test name so to keep it functional, but for those new tests, I'm struggling myself to keep them clear enough.

Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionsSpecs.cs
```diff
@@ -1931,5 +1931,237 @@ public void When_inspectors_count_does_not_equal_asserting_collection_length_it_
...
+                {
+                    element => element.Text == fourText && element.Number == 4,
+                    element => element.Text == twoText && element.Number == 2,
+                },
+                because: "we want to test formatting ({0})",
+                becauseArgs: "args");
+
+            // Assert
+            act.Should().Throw<XunitException>().WithMessage(
+@"Expected collection to satisfy all predicates because we want to test formatting (args), but:
```
dennisdoomen, line 2067: 🤔 The full verification of the output should be diverted to the `PredicateLambdaExpressionValueFormatterSpecs`

docs/_pages/releases.md
```diff
@@ -12,6 +12,7 @@ sidebar:
 **What's New**
 * Added `WithParameterName` extension to ease asserting on the parameter name for a thrown `ArgumentException` - [#1466](https://github.com/fluentassertions/fluentassertions/pull/1466).
 * Added `NotCompleteWithinAsync` to `TaskCompletionSourceAssertions` - [#1474](https://github.com/fluentassertions/fluentassertions/pull/1474).
+* Added `Satisfy` to `SelfReferencingCollectionAssertions` - [#1500](https://github.com/fluentassertions/fluentassertions/pull/1500).
```
dennisdoomen, line 15: 🤔 I noticed the previous example is wrong too, but those release notes should be as functional as possible. So instead of referring to `SelfReferencingCollectionAssertions`, I would you use a phrase like:

> Added `Satisfy` to be able to compare a collection with a set of predicates in any order.

dennisdoomen (changes requested):
**Comments
* 👍 Impressive work. You've kind of changed my mind about the importance of algorithms,
* 🔧 I think it reads better if you replace the term `NotMatched` with `Unmatched` everywhere.
* 🔧 Since the namespace is already called `MaximumMatching`, I would repeat that word on every class and method.
* 🔧 I think the code will be much easier to understand if you would wrap the predicate and its index in a first-class class such as `Predicate`, and then have it override `Equals`. This will reduce the amount of fiddling around with indexes and using those to find the actual predicate. And this probably also applies to the element collection and its indexes.

**Legend for the emojis**
• ❓ I have a concrete question. There is nothing here to be changed yet.
• ❌ Here is an actual problem. Please change this according to my suggestion. It might be a bug or a strict code style fix.
• 🔧 Here is an idea/suggestion. I think it would improve things, but feel free to disagree :)
• 🙃 This is a nitpick. Feel free to change of leave it.
• 🤔 This might be wrong. Let’s talk about it, maybe we can come up with a better solution
🤡 Here is a problem independent from your work. It would be nice if you would be willing to include it

Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingProblem.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using FluentAssertions.Collections.MaximumMatching;
```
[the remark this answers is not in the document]
jn-foreflight: The algorithm is called [Ford-Fulkerson algorithm](https://en.wikipedia.org/wiki/Ford%E2%80%93Fulkerson_algorithm) and it can be used to solve the Maximum _Flow_ Problem.
The Maximum Matching Problem can be modelled as a Maximum Flow Problem.

We want to find a _perfect_ matching in a bipartite graph, i.e. a one-one-one mapping between all predicates and items.
This can be solved by finding a maximum matching and if no predicates or items are unmatched, it is perfect.

Maximum and maximal are two different things in [matching theory](https://en.wikipedia.org/wiki/Matching_(graph_theory)#Definitions).
In plain words, a _maximal_ matching is some valid solution, where a _maximum_ matching is a solution that is no less worse than any other solution.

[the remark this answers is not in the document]
jnyrup: The algorithm is called [Ford-Fulkerson algorithm](https://en.wikipedia.org/wiki/Ford%E2%80%93Fulkerson_algorithm) and it can be used to solve the Maximum _Flow_ Problem.
The Maximum Matching Problem can be modelled as a Maximum Flow Problem.

We want to find a _perfect_ matching in a bipartite graph, i.e. a one-one-one mapping between all predicates and items.
This can be solved by finding a maximum matching and if no predicates or items are unmatched, it is perfect.

Maximum and maximal are two different things in [matching theory](https://en.wikipedia.org/wiki/Matching_(graph_theory)#Definitions).
In plain words, a _maximal_ matching is some valid solution, where a _maximum_ matching is a solution that is no less worse than any other solution.

Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingSolver.cs
```diff
@@ -0,0 +1,148 @@
...
+                        var finalAssignment = new Assignment { Predicate = unassignedPredicate, Element = element };
+                        return breadthFirstSearchTracker.GetAssignmentChain(finalAssignment);
+                    }
+                }
+            }
+
+            return Enumerable.Empty<Assignment>();
+        }
+
+        private struct Assignment
```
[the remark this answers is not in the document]
jnyrup: This comes from the literature where the problem is also known as the [Assignment Problem](https://en.wikipedia.org/wiki/Assignment_problem). where tasks are assigned to agents.
By not using _match_ it is more easily distinguished from _matching_.

Src/FluentAssertions/Collections/SelfReferencingCollectionAssertions.cs
```diff
@@ -768,6 +769,91 @@ public AndConstraint<TAssertions> SatisfyRespectively(IEnumerable<Action<T>> exp
...
+                .Then
+                .ForCondition(Subject.Any())
+                .FailWith("but collection is empty.")
+                .Then
+                .ClearExpectation();
+
+            var elements = Subject.ConvertOrCastToList();
+            var maximumMatchingSolution = new MaximumMatchingProblem<T>(predicatesList, elements).Solve();
+
+            if (maximumMatchingSolution.NotMatchedPredicatesExist || maximumMatchingSolution.NotMatchedElementsExist)
```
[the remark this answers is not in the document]
jnyrup: Could be useful when we get around to use this algorithm for #1486

Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionsSpecs.cs
jnyrup: I believe the assertion of the missing predicate is handled in `When_Satisfy_fails_then_failure_message_must_contain_predicates_without_matching_elements_and_elements_without_matching_predicates`.

jnyrup:
>     * 🔧 I think the code will be much easier to understand if you would wrap the predicate and its index in a first-class class such as `Predicate`, and then have it override `Equals`. This will reduce the amount of fiddling around with indexes and using those to find the actual predicate. And this probably also applies to the element collection and its indexes.

A middle way could maybe be to make two small structs to wrap the raw `int`s making it more visible what they represent.
```cs
struct PredicateIndex : IEquatable<PredicateIndex>
{
    private int value;
    ...
}

struct ElementIndex : IEquatable<ElementIndex >
{
    private int value;
    ...
}
```

Then
```cs
private readonly Dictionary<int, int> matchedElementIndexesByPredicateIndex;
```

could be

```cs
private readonly Dictionary<PredicateIndex, ElementIndex> matchedElementIndexes;
```

Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingSolver.cs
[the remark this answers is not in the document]
dennisdoomen: Then that should be explained on the type itself. If you're not versed in this domain (like me), the name does not mean anything.

Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingProblem.cs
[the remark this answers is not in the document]
dennisdoomen: I learned something new today 😜

Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingProblem.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// The <see cref="MaximumMatchingProblem{TElement}"/> class defines input for the maximum matching problem.
+    /// The input is a list of predicates and a list of elements.
+    /// The goal of the problem is to find such mapping between predicates and elements that would maximize number of matches.
+    /// A predicate can be mapped with only one element.
+    /// An element can be mapped with only one predicate.
+    /// </summary>
+    /// <typeparam name="TElement">The type of elements which must be matched with predicates.</typeparam>
+    internal class MaximumMatchingProblem<TElement>
+    {
+        private readonly Dictionary<int, List<int>> matchingElementsByPredicate = new Dictionary<int, List<int>>();
```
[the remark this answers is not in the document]
vanya-lebedev: Thanks a lot for the valuable feedback, Dennis! I'm on it :)

I'm not 100% sure about the suggestion below:
_Since the namespace is already called MaximumMatching, I would repeat that word on every class and method._
Removing the MaximumMatching prefix from class names would leave us with classes called Problem, Solution and Solver. These names are very generic. It would be hard to understand what the calling code is doing. Examples from well-known libraries where prefix is used: SqlConnection instead of Connection, CosmosClient instead of Client. I think keeping the prefix would be the right thing to do. But pls let me know if you'd still rather rename.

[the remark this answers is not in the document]
dennisdoomen: It was a suggestion anyway, so it's up to you. And since this code is `internal` anyway, we can always rename it afterwards if we change our minds again.

Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionsSpecs.cs
vanya-lebedev: Yeah. I had the same challenge :) There are several popular test naming conventions https://dzone.com/articles/7-popular-unit-test-naming. I think that the ones having the test method in the name work best for the GenericCollectionAssertionsSpecs class which tests several different APIs. Another option could be to have separate class per API under test.
Anyways, if you have a suggestion how to change this - I will be happy to implement.

Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionsSpecs.cs
vanya-lebedev: I added the validation. Won't harm for sure ;)

Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionsSpecs.cs
dennisdoomen: > There are several popular test naming conventions https://dzone.com/articles/7-popular-unit-test-naming

Indeed. We've been using 6, but I've also used 5 on a client code base. I definitely stay away from 1 and 2. 

> Another option could be to have separate class per API under test.

This is what I've been thinking about as well.

Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingProblem.cs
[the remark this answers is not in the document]
vanya-lebedev: I hope I have addressed the comments in the latest iteration. Also, I know that the matching problem is adding a lot of complexity to the project. Apologize for it. I tried making it as readable as possible and any suggestions to make it better are very welcome.

Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionsSpecs.cs
vanya-lebedev: I moved the Satisfy tests to a separate class

Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionsSpecs.cs
vanya-lebedev: I think this particular test is still valuable as a form of documentation - so we can see the full output. But I have simplified it a bit in the latest iteration and moved formatting-specific validation to a separate test class. Pls let me know if you'd like more scenarios to be tested

vanya-lebedev:
Pardon. I didn't realize that I was pushing iterations without squashing my local commits. I'm used to working with Azure Dev Ops rather than Git Hub and it does squashing automatically.

Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingSolver.cs
```diff
@@ -0,0 +1,148 @@
...
+namespace FluentAssertions.Collections
+{
+    /// <summary>
+    /// The <see cref="MaximumMatchingSolver"/> class encapsulates the algorithm
+    /// for solving the maximum matching problem (see <see cref="MaximumMatchingProblem{TElement}"/>).<br />
+    /// A simplified variation of the Ford-Fulkerson algorithm is used for solving the problem. <br />
+    /// </summary>
+    internal class MaximumMatchingSolver
+    {
+        public static Dictionary<int, int> FindMaximumMatching<T>(MaximumMatchingProblem<T> maximumMatchingProblem)
```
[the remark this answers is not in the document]
vanya-lebedev: I hope the new iteration addresses the concern

Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingSolver.cs
```diff
@@ -0,0 +1,148 @@
...
+        {
+            var assignments = new AssignmentCollection();
+
+            foreach (var predicate in maximumMatchingProblem.AllPredicateIndices)
```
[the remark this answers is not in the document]
vanya-lebedev: I hope this is addressed with the introduction of the Predicate and Element classes

Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingSolver.cs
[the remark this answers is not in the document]
vanya-lebedev: I guess flipping the name to "Match" and "MatchCollection" will make it more self-explantory?

Src/FluentAssertions/Collections/SelfReferencingCollectionAssertions.cs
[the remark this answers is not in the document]
vanya-lebedev: Maybe let's do it as part of tackling #1486? That's when we'll know the code reuse scenario and requirements precisely.

docs/_pages/collections.md
```diff
@@ -168,3 +168,16 @@ collection.Should().SatisfyRespectively(
...
+```csharp
+var collection = new []
+{
+    new { Id = 1, Name = "John", Attributes = new string[] { } },
+    new { Id = 2, Name = "Jane", Attributes = new string[] { "attr" } }
+};
+
+collection.Should().Satisfy(
+    _ => _.Id == 2 && _.Name == "Jane" && _.Attributes == null,
+    _ => _.Id == 1 && _.Name == "John" && _.Attributes != null && _.Attributes.Length > 0);
```
jnyrup, line 182:
```suggestion
    e => e.Id == 2 && e.Name == "Jane" && e.Attributes == null,
    e => e.Id == 1 && e.Name == "John" && e.Attributes != null && e.Attributes.Length > 0);
```

Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingSolver.cs
```diff
@@ -0,0 +1,166 @@
...
+            public Predicate<TValue> GetMatchedPredicate(Element<TValue> element)
+            {
+                return matchesByElement[element].Predicate;
+            }                
+
+            public bool Contains(Element<TValue> element) => matchesByElement.ContainsKey(element);
+
+            public IEnumerator<Match> GetEnumerator() => matchesByElement.Values.GetEnumerator();
+
+            System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator() => matchesByElement.Values.GetEnumerator();
```
jnyrup, line 118: Please import the namespace instead of the fully qualified name

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.approved.txt
```diff
@@ -491,6 +491,8 @@ namespace FluentAssertions.Collections
         public FluentAssertions.AndWhichConstraint<TAssertions, T> NotContain(T unexpected, string because = "", params object[] becauseArgs) { }
         public FluentAssertions.AndConstraint<TAssertions> NotHaveCount(int unexpected, string because = "", params object[] becauseArgs) { }
         public FluentAssertions.AndConstraint<TAssertions> OnlyContain(System.Linq.Expressions.Expression<System.Func<T, bool>> predicate, string because = "", params object[] becauseArgs) { }
+        public FluentAssertions.AndConstraint<TAssertions> Satisfy(params System.Linq.Expressions.Expression<>[] predicates) { }
+        public FluentAssertions.AndConstraint<TAssertions> Satisfy(System.Linq.Expressions.Expression<>[] predicates, string because = "", params object[] becauseArgs) { }
```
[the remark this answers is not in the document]
vanya-lebedev: Fixed. They are green now

docs/_pages/collections.md
```diff
@@ -168,3 +168,16 @@ collection.Should().SatisfyRespectively(
...
+```csharp
+var collection = new []
+{
+    new { Id = 1, Name = "John", Attributes = new string[] { } },
+    new { Id = 2, Name = "Jane", Attributes = new string[] { "attr" } }
+};
+
+collection.Should().Satisfy(
+    _ => e.Id == 2 && e.Name == "Jane" && e.Attributes == null,
+    _ => e.Id == 1 && e.Name == "John" && e.Attributes != null && e.Attributes.Length > 0);
```
jnyrup, line 182:
```suggestion
    e => e.Id == 2 && e.Name == "Jane" && e.Attributes == null,
    e => e.Id == 1 && e.Name == "John" && e.Attributes != null && e.Attributes.Length > 0);
```

Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingSolver.cs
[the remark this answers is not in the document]
dennisdoomen: Yep.

dennisdoomen (approved):

Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingSolver.cs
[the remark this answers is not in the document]
jnyrup: I'm fine with `Match`.

jnyrup (approved):

PR #1505, opened 2021-03-06
Replace ApprovalTests with Verify

jnyrup:
By suggestion from @matkoch this PR replaces [ApprovalTests](https://github.com/approvals/ApprovalTests.Net) with [Verify](https://github.com/VerifyTests/Verify/).

I added a custom comparer such that the failure message only contains the _changed_ lines instead of all ~25,000 lines.

I disabled the automatic opening of a diff tool as that always annoyed me massively (I've just now found out about the environment variable [`DiffEngine_Disabled`](https://github.com/VerifyTests/DiffEngine#disable-for-a-machineprocess)).
I'd like to decide myself _when_ I want to look at the diffs instead of 5 new tabs opening like another auto-play add.
...and when I do look at the diffs I mostly use `git diff` from a console.

I've checked that both `AcceptApiChanges.ps1` and `AcceptApiChanges.sh` still works.


dennisdoomen (approved):

dennisdoomen:
> I disabled the automatic opening of a diff tool as that always annoyed me massively

Yeah, I kind of agree with that. I never got it work in such a way that it does not affect my development work.

PR #1515, opened 2021-04-03
AssertionScope - allow reportables which are only calculated on failure

anthony-c-martin:
Discussed [here](https://github.com/fluentassertions/fluentassertions/discussions/1509) - I thought I'd contribute a potential solution.

My personal use case is that I've got some costly calculations to display user-friendly failure info, and I'd like to avoid doing this for the 99% case where the tests run successfully.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -170,7 +170,7 @@ namespace FluentAssertions.Execution
         /// In addition to the numbered <see cref="string.Format(string,object[])"/>-style placeholders, messages may contain a few
         /// specialized placeholders as well. For instance, {reason} will be replaced with the reason of the assertion as passed
         /// to <see cref="BecauseOf(FluentAssertions.Execution.Reason)"/>. Other named placeholders will be replaced with the <see cref="Current"/> scope data
-        /// passed through <see cref="AddNonReportable"/> and <see cref="AddReportable"/>. Finally, a description of the
+        /// passed through <see cref="AddNonReportable"/> and <see cref="AddReportable(string,string)"/>. Finally, a description of the
         /// current subject can be passed through the {context:description} placeholder. This is used in the message if no
         /// explicit context is specified through the <see cref="AssertionScope"/> constructor.
         /// Note that only 10 <paramref name="args"/> are supported in combination with a {reason}.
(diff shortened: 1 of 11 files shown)


Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -328,6 +340,15 @@ public void AddReportable(string key, string value)
             contextData.Add(new ContextDataItems.DataItem(key, value, reportable: true, requiresFormatting: false));
         }
 
+        /// <summary>
+        /// Adds some information to the assertion scope that will be included in the message
+        /// that is emitted if an assertion fails. The value is only calculated on failure.
+        /// </summary>
+        public void AddDeferredReportable(string key, Func<string> valueFunc)
```
dennisdoomen, line 347: Why not make it an overload of  `AddReportable`?

anthony-c-martin: Good point, let me fix that.

dennisdoomen (commented):
Would you be willing to add the corresponding release note changes and possibly some docs as explained in the [Pull Request template](https://github.com/fluentassertions/fluentassertions/blob/develop/.github/PULL_REQUEST_TEMPLATE.md)?

anthony-c-martin:
> Would you be willing to add the corresponding release note changes and possibly some docs as explained in the [Pull Request template](https://github.com/fluentassertions/fluentassertions/blob/develop/.github/PULL_REQUEST_TEMPLATE.md)?

Absolutely! I've added to the release notes for 6.0.0. Please let me know if I missed anything there.

I don't see any reference to `AddReportable` in the current docs but can certainly take a stab at adding it in.

dennisdoomen (approved):

Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -29,6 +29,18 @@ public sealed class AssertionScope : IAssertionScope
         private string fallbackIdentifier = "object";
         private bool? succeeded;
 
+        private sealed class DeferredReportable
+        {
+            private readonly Lazy<string> lazyValue;
```
jnyrup, line 34: Is there a known case where `valueFunc` is invoked more than once?

jnyrup (approved):

Src/FluentAssertions/Execution/AssertionScope.cs
dennisdoomen: We're only calling `ContextDateItems.GetReportable` when the scope is disposed, and only if it's the outer scope, so I think we're fine.

dennisdoomen:
You need to update the approval files.

dennisdoomen:
Thank you for your contribution.

PR #1522, opened 2021-04-11
Create SourceLink-Enabled and Deterministic nupkg

jnyrup:
Enable SourceLink and Deterministic builds:
* Two more green badges in Nuget Package Explorer
* Size reduction (1.50 MB -> 1.14 MB) since pdb files are no longer embedded in nupkg, but only downloaded on demand.

Possible downsides/caveats:
* Only nuget.org has a symbol server (this might not be true anymore), this is only a problem if we want to switch/add provider.
* You can't step through FA while being offline (unless the symbols are cached somewhere locally)
* Your IDE is too old to have support for symbols servers (before Visual Studio 2017 version 15.9)
* Your IDE isn't set up to download symbols from the nuget symbol server (before Visual Studio 2019 version 16.1)

The limitations of snupkg in contrast to pdb

> * They do not currently support Windows PDBs (generated by VC++, or for managed projects that set build property `DebugType` to `full`)
> * They require the library to be built by newer C#/VB compiler (Visual Studio 2017 Update 9).
> * The consumer of the package also needs Visual Studio 2017 Update 9 debugger.
> * Not supported by [Azure DevOps Artifacts](https://azure.microsoft.com/en-us/services/devops/artifacts) service.

https://github.com/dotnet/sourcelink#alternative-pdb-distribution

A thing that talks in favor of snupkg is that there in an active [issue](https://github.com/dotnet/sdk/issues/1458) with applications targeting .net core 3.0 and symbols from pdb files.
I tested this with Newtonsoft.Json in VS to see when I could step into the source of `JsonConvert`.
12.0.1 and 12.0.2 both have sourcelink, but the former includes symbols in pdb files and the latter publishes a snupkg.

|               | 12.0.1/pdb | 12.0.2/snupkg |
|---------------|:----------:|:-------------:|
| netcoreapp2.1 |     OK     |       OK      |
| netcoreapp3.1 |    FAIL    |       OK      |

Links:
* https://github.com/dotnet/sourcelink
* https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink
* https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/nuget#symbol-packages

Before
image

After
image

### DIFF
--- a/Build/Build.cs
+++ b/Build/Build.cs
@@ -24,9 +24,7 @@ class Build : NukeBuild
        - Microsoft VSCode           https://nuke.build/vscode
     */
 
-    public static int Main() => Execute<Build>(
-        x => x.UnitTests,
-        x => x.Pack);
+    public static int Main() => Execute<Build>(x => x.Pack);
 
     [Solution(GenerateProjects = true)] readonly Solution Solution;
     [GitVersion] readonly GitVersion GitVersion;
@@ -35,13 +33,13 @@ class Build : NukeBuild
     AbsolutePath ArtifactsDirectory => RootDirectory / "Artifacts";
 
     Target Clean => _ => _
-        .Before(Restore)
         .Executes(() =>
         {
             EnsureCleanDirectory(ArtifactsDirectory);
         });
 
     Target Restore => _ => _
+        .DependsOn(Clean)
         .Executes(() =>
         {
             DotNetRestore(s => s
@@ -119,6 +117,7 @@ class Build : NukeBuild
                 .SetProject(Solution.Core.FluentAssertions)
                 .SetOutputDirectory(ArtifactsDirectory)
                 .SetConfiguration("Release")
+                .EnableContinuousIntegrationBuild() // Necessary for deterministic builds
                 .SetVersion(GitVersion.NuGetVersionV2));
         });
 }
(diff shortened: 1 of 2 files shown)


dennisdoomen:
Hmm, shipping without PDBs also means you don't get files and line numbers when somebody runs into an FA internal error. I personally don't think loosing that is worth reducing the size of the package.

jnyrup:
aha, I hadn't thought about the line numbers of FA.
I agree in keeping pdb files.

Here's some tests about whether line numbers are included in the failure message

|               | 12.0.1/pdb | 12.0.2/snupkg |
|---------------|:----------:|:-------------:|
| net7          |    FAIL    |       FAIL    |
| netcoreapp2.1 |     OK     |       FAIL    |
| netcoreapp3.1 |    FAIL    |       FAIL    |


Applying the workaround of https://github.com/dotnet/sdk/issues/1458#issuecomment-695119194
|               | 12.0.1/pdb | 12.0.2/snupkg |
|---------------|:----------:|:-------------:|
| net7          |     OK     |       FAIL    |
| netcoreapp2.1 |     OK     |       FAIL    |
| netcoreapp3.1 |     OK     |       FAIL    |


If the workaround above is applied I can step also into Newtonsoft.Json with netcoreapp3.1 and pdb files.

dennisdoomen:
> Here's some tests about whether line numbers are included in the failure message

What are you trying to say with this?

jnyrup:
Somehow I wasn't aware that the pdb files were necessary for providing line numbers. 
I wanted to experiment if snupkg could provide line numbers, but it doesn't seem so.

dennisdoomen (approved):

PR #1526, opened 2021-04-18
Allow using lambdas for asserting the order of elements in collections

### LINKED ISSUE #998: Ordering assertions in CollectionAssertions should use IComparer, not IComparer<object>
### Description
The ordering assertions in `CollectionAssertions` expect an `IComparer<object>`, although the declared type of the items in the collection does not need to be object. This makes it hard to use standard comparers such as `StringComparer` and comparers derived from `Comparer<T>`.

The following does not compile:

```cs
string[] actual = { "a", "b", "c" };
actual.Should().BeInAscendingOrder(StringComparer.CurrentCultureIgnoreCase);
```

### Expected behavior:

* You should be able to use a comparer that implements `IComparer<T>` for a generic collection with items of type `T`.
* You should be able to use a comparer that implements `IComparer` for a non-generic collection.

### Actual behavior:

Does not compile.

### Versions

* Which version of Fluent Assertions are you using? v5.6.0
* Which .NET runtime and version are you targeting? .NET framework 4.6.1

dennisdoomen:
Added `BeInAscendingOrder` and `BeInDescendingOrder` to `string` collections taking either an `IComparer<string>` or a lambda expression 

Fixes #998


dennisdoomen:
Rebased on #1529

Src/FluentAssertions/Collections/StringCollectionAssertions.cs
```diff
@@ -290,5 +290,49 @@ private bool NotContainsMatch(string wildcardPattern)
                 return !scope.Discard().Any();
             });
         }
+
+        /// <summary>
+        /// Expects the current collection of strings to have all elements in ascending order. Elements are compared
+        /// using the given <see cref="IComparer{T}" /> implementation.
```
jnyrup, line 296: Should the `IComparer` part be left out or changed to refer to the provided lambda expression?

Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionOfStringSpecs.cs
```diff
@@ -1946,5 +1946,192 @@ public void When_string_collection_does_not_satisfy_all_inspectors_it_should_thr
...
+        }
+
+        [Fact]
+        public void When_strings_are_in_ascending_order_according_to_a_custom_lambda_it_should_succeed()
+        {
+            // Arrange
+            string[] strings = { "alpha", "beta", "theta" };
+
+            // Act
+            Action act = () => strings.Should().BeInAscendingOrder((sut, exp) => sut.Reverse().First().CompareTo(exp.Reverse().First()));
```
jnyrup, line 2015: Out of curiosity, why do you write `Reverse().First()` instead of `Last()`?

jnyrup (approved):
I never get tired of seeing the capabilities of this library, where you can add a new functionality by adding two one-liners.

For completeness `NotBeInAscendingOrder` and `NotBeInDescendingOrder` should also be augmented with the new overloads.

Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionOfStringSpecs.cs
dennisdoomen: Because I forgot about `Last` 😜

jnyrup (approved):

jnyrup:
Now I think of it...

Why don't we just add `BeInAscendingOrder(Func<T, T, int>, string, params object[])` and friends to `GenericCollectionAssertions` instead?
Then `StringCollectionAssertions` gets it for free through inheritance.

dennisdoomen:
> Why don't we just add `BeInAscendingOrder(Func<T, T, int>, string, params object[])` and friends to `GenericCollectionAssertions` instead?

Why didn't I think of this myself...

docs/_pages/collections.md
```diff
@@ -80,7 +80,11 @@ IEnumerable otherCollection = new[] { 1, 2, 5, 8, 1 };
 IEnumerable anotherCollection = new[] { 10, 20, 50, 80, 10 };
 collection.Should().IntersectWith(otherCollection);
 collection.Should().NotIntersectWith(anotherCollection);
+```
+
+Asserting that a collection contains items in a certain order is as easy as using one of the several overloads of `BeInAscendingOrder` or `BeInDescendingOrder`. The default overload will use the default `Comparer` for the specified type, but overloads also exist that take an `IComparer<T>`, a property expression to sort by an object's property, or for `string` collections, a lambda expression.
```
jnyrup, line 85: Lambda expression is applicable not only to string collection.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -424,6 +424,28 @@ public AndConstraint<SubsequentOrderingAssertions<T>> BeInAscendingOrder(string
             return BeInAscendingOrder(Comparer<T>.Default, because, becauseArgs);
         }
 
+        /// <summary>
+        /// Expects the current collection of strings to have all elements in ascending order. Elements are compared
```
jnyrup, line 428:
```suggestion
        /// Expects the current collection to have all elements in ascending order. Elements are compared
```

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -424,6 +424,28 @@ public AndConstraint<SubsequentOrderingAssertions<T>> BeInAscendingOrder(string
...
+        /// using the given lambda expression.
+        /// </summary>
+        /// <param name="comparison">
+        /// A lambda expression that should be used to determine the expected ordering between two strings.
```
jnyrup, line 432: "two strings", here and three more places.

docs/_pages/collections.md
```diff
@@ -80,7 +80,11 @@ IEnumerable otherCollection = new[] { 1, 2, 5, 8, 1 };
 IEnumerable anotherCollection = new[] { 10, 20, 50, 80, 10 };
 collection.Should().IntersectWith(otherCollection);
 collection.Should().NotIntersectWith(anotherCollection);
+```
+
+Asserting that a collection contains items in a certain order is as easy as using one of the several overloads of `BeInAscendingOrder` or `BeInDescendingOrder`. The default overload will use the default `Comparer` for the specified type, but overloads also exist that take an `IComparer<T>`, a property expression to sort by an object's property, or a lambda expression to avoid the need for `ICompare<T>` implementations.
```
jnyrup, line 85: ’ICompare<T>’

jnyrup (approved):

jnyrup (approved):
:shipit:

m-ringler:
Thanks for improving this!

Issue #1346, opened 2020-06-10
Using Should().Equal() on two List<object> fails even though list contents are identical


GeeWee:
### Description

Using `Should().Equal()` to compare two `List<object>` always fails, even though the items in the list are equal. I have reproduced below with strings, and confirmed it is also true  with ints.

### Complete minimal example reproducing the issue
A reproducible example below. The exact same test but the type of the list goes
from `List<string>` to `List<object>`
```csharp
// Two identical strings - should succeed
var strings = new List<string>{"foo"};
var identicalStrings = new List<string>{"foo"};
strings.Should().BeEquivalentTo(identicalStrings); // Passes
strings.Should().Equal(identicalStrings); // Passes

// Same test, but here the specified type is List<object>
var objects = new List<object>{"bar"};
var identicalObjects = new List<object>{"bar"};
objects.Should().BeEquivalentTo(identicalObjects); // Passes
objects.Should().Equal(identicalObjects); // This fails with
// Expected objects to be equal to {{"bar"}}, but {"bar"} differs at index 0.
```

Full stack trace:
```
 Error Message:
   Expected objects to be equal to {{"bar"}}, but {"bar"} differs at index 0.
  Stack Trace:
     at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message)
   at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message)
   at FluentAssertions.Execution.DefaultAssertionStrategy.HandleFailure(String message)
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
   at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args)
   at FluentAssertions.Execution.GivenSelector`1.FailWith(String message, Object[] args)
   at FluentAssertions.Execution.GivenSelector`1.FailWith(String message, Func`2[] args)
   at FluentAssertions.Execution.GivenSelectorExtensions.AssertCollectionsHaveSameItems[TActual,TExpected](GivenSelector`1 givenSelector, ICollection`1 expected, Func`3 findIndex)
   at FluentAssertions.Collections.CollectionAssertions`2.AssertSubjectEquality[TActual,TExpected](IEnumerable expectation, Func`3 equalityComparison, String because, Object[] becauseArgs)
[...]

Issue #1036, opened 2019-04-30
BeEquivalentTo and NotBeEquivalentTo consistency


djmeijer:
### Description

I may have found some weird cases when using the BeEquivalentTo and NotBeEquivalentTo functions. To put it short:

1. Using compile time type to check equivalence seems a little bit weird.
2. BeEquivalentTo is not mutually exclusive with NotBeEquivalentTo on IEnumerable.
3. NotBeEquivalentTo is not implemented on an object where BeEquivalentTo is.
4. BeEquivalentTo on an object is not commutative.

I like to hear some other views and opinions on this.

### Complete minimal example reproducing the issue

The test cases 1 to 6 represent my line of thought.

```csharp
using System.Collections.Generic;
using FluentAssertions;
using FluentAssertions.Execution;
using NUnit.Framework;

namespace FluentAssertionTests
{
  public abstract class Car
  {
    protected Car(string licensePlate)
    {
      LicensePlate = licensePlate;
    }

    public string LicensePlate { get; }
  }

  public class Peugeot : Car
  {
    public Peugeot(string licensePlate, string peugeotNumber)
      : base(licensePlate)
    {
      PeugeotNumber = peugeotNumber;
    }

    public string PeugeotNumber { get; }
  }

  [TestFixture]
  public class Tests
  {
    private Car _expected;
    private IEnumerable<Car> _expectedIEnumerable;
    private Car _actual;
    private IEnumerable<Car> _actualIEnumerable;

    /// <summary>
    ///   I expect all the statements in this scope to fail, but they all succeed.
    /// </summary>
    [Test]
    public void TestCase1()
    {
      using(new AssertionScope())
      {
        _expectedIEnumerable.Should().BeEquivalentTo(_actualIEnumerable);
        new[] { _expected }.Should().BeEquivalentTo(_actualIEnumerable);
        _expected.Should().BeEquivalentTo(_actual);
      }
    }

    /// <summary>
    ///   All the statements fail, as I expect.
    /// </summary>
    [Test]
    public void TestCase2()
    {
      using(new AssertionScope())
      {
        new[] { _expected }.Should().BeEquivalentTo(new[] { _actual });
        _expectedIEnumerable.Should().BeEquivalentTo(new[] { _actual });
      }
    }

    /// <summary>
    ///   In order to let TestCase1 fail, the RespectingRuntimeTypes option has to be enabled. Can this be the default instead
    ///   of compile time type?
    /// </summary>
    [Test]
[...]

PR #1529, opened 2021-04-25
Remove support for non-generic collections

jnyrup:
This PR removes support for non-generic collections.

Merged these three classes into `GenericCollectionAssertions`
* `CollectionAssertions`
* `NonGenericCollectionAssertions`
* `SelfReferencingCollectionAssertions`

`StringCollectionAssertions` now derives from `GenericCollectionAssertions`, which gives the following assertions for free
* `[Not]Contain(IEnumerable<string>)`
* `[Not]ContainInOrder(params string[])`
* `[Not]ContainInOrder(IEnumerable<string>)`

Deleted overloads:
* `Should(this IEnumerable)` - described in the migration guide
  * Users can use `actual.Cast<T>()` or `actual.Cast<object>()`.
* `BeEquivalentTo(params object[] expectations)` - described in the migration guide
  * Use `BeEquivalentTo<TExpectation>(IEnumerable<TExpectation> expectations)` to avoid ambiguous cases.
* `[Not]Contain(IEnumerable<T> expectedItemsList, params T[] additionalExpectedItems)`
  * Users can concat their elements, this opens up for _finally_ resolving #806.

Deleted overloads where a generic variant exists:
* `[Not]BeEquivalentTo(IEnumerable expectation)` => `[Not]BeEquivalentTo<TExpectation>(IEnumerable<TExpectation> expectation)`
* `[Not]Equal(params object[])`
* `[Not]Contain(object expected)`
* `[Not]ContainInOrder(IEnumerable expected)`
* `[Not]ContainInOrder(params object[] expected)`
* `[Not]Contain(IEnumerable expected)`
* `[Not]BeIn{Ascending,Descending}Order(IComparer<object>)`

Overloads now generic in `T`:
* `[Not]Equal(IEnumerable)`
* `EndWith(object element)`
* `StartWith(object element)`
* `HaveElementAt(int index, object element)`
* `HaveElementPreceding(object successor, object expectation)`
* `HaveElementSucceeding(object predecessor, object expectation)`
* `[Not]IntersectWith(IEnumerable otherCollection)`
* `[Not]BeSubsetOf(IEnumerable expectedSuperset)`

Changes in return types:
* `BeIn{Ascending,Descending}Order()` now returns `AndConstraint<SubsequentOrderingAssertions<T>>`.
* `ContainEquivalentOf` now returns `AndWhichConstraint<TAssertions, T>`.
* `HaveElementAt` now returns `AndWhichConstraint<TAssertions, T>`.

Added:
* `{ContinuedAssertionScope,IAssertionScope}.FailWith(string message)`
* `{ContinuedAssertionScope,IAssertionScope}.FailWith(string message, params Func<object>[] argProviders)`

Functional changes:
* Declared types are now respected, as we always use a generic overload of `BeEquivalentTo`
  * `When_the_expectation_is_an_array_of_interface_type_it_should_respect_declared_types`
  * `When_the_expectation_has_fewer_dimensions_than_a_multi_dimensional_subject_it_should_fail`

Tests:
* Removed tests of `Contain(IEnumerable<T> expectedItemsList, params T[] additionalExpectedItems)`.
* `When_the_expectation_has_fewer_dimensions_than_a_one_dimensional_subject_it_should_succeed` no longer compiles, as multidimentional arrays do not implement `IEnumerable<T>`.
* `When_the_number_of_dimensions_of_the_arrays_are_not_the_same_it_should_throw` changed to still reach `MultiDimensionalArrayEquivalencyStep.HaveSameRank`.
* A few tests now has its expected elements wrapped in `new[] {}` as the `params` overload of `BeEquivalentTo` was removed.

Related issues:
#903
#1139
#1036
#1346


Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.cs
```diff
@@ -1289,7 +1262,7 @@ public void When_asserting_identical_collections_to_be_equal_it_should_enumerate
             actual.Should().Equal(expected);
 
             // Assert
-            actual.GetEnumeratorCallCount.Should().Be(1);
+            actual.GetEnumeratorCallCount.Should().BeLessOrEqualTo(1);
```
dennisdoomen, line 1265: Why did you need this?

docs/_pages/upgradingtov6.md
```diff
@@ -160,4 +161,54 @@ public void Format(object value, FormattedObjectGraph formattedGraph, Formatting
...
+Case `(3)` works as expected, as `List<T>` is not implicitly convertible to `T[]` and the compiler picks `BeEquivalentTo(IEnumerable<I>)`, which retains the compile time information about the expectation being objects of type `I`.
+A case which led to some confusion among our users was that `(2)` changed behavior when adding assertion options, as shown in `(4)`, as `BeEquivalentTo(IEnumerable<T>, Func<EquivalencyAssertionOptions<I>>)` was now picked because `I[]` is implicitly covertible to `IEnumerable<I>`.
+
+You might ask if we could not just have changed the overload signature from `params object[]` to `params T[]` to solve the type problem while keeping the convient overload.
+The answer is no (as far as we know), as the compiler prefers resolving e.g. `IEnumerable<T>` to `params IEnumerable<T>[]` over `IEnumerable<T>`.
+
+For Fluent Assertions 6.0 the implications of this are that `(1)` no longer compiles, as `BeEquivalentTo` takes an `IEnumerable<T>`, but `(2)` now works as expected.
+
+// performance boxing
```
dennisdoomen, line 213: What did you intend to do here?

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.cs
jnyrup: Historically #438 added materialization of `Subject` to avoid multiple enumerations.

in master `ConvertOrCastToCollection<object>` always did an enumeration of `Subject` (as `CountingGenericCollection<int>` doesn't implement `ICollection<object>`) using `CountingGenericCollection.GetEnumerator`. 
That gave the expected `GetEnumeratorCallCount` with value `1`.
```cs
NotEqual(IEnumerable unexpected
...
ICollection<object> actualItems = Subject.ConvertOrCastToCollection<object>();
```

In this PR `ConvertOrCastToCollection` now just casts `CountingGenericCollection<int>` to `ICollection<int>` and avoids enumeration using `CountingGenericCollection.GetEnumerator`.
Instead uses `CountingGenericCollection.CopyTo`, which does not increment `GetEnumeratorCallCount`, to make a defensive copy of `Subject`.

The other two enumerations of `actualItems` are `SequenceEquals` and maybe `EnumerableValueFormatter.Format` if failing.

I think the tests might be misleading, but it depends on our usage of materialization.
As I understand it, we materialize to avoid enumerating _lazy_ enumerables multiple times, but accept multiple enumerations of e.g. `ICollection<T>` - this is at least how `ConvertOrCastToCollection` and `ConvertOrCastToList` are implemented.

Should `When_asserting_identical_collections_to_be_equal_it_should_enumerate_the_subject_only_once` and `When_asserting_identical_collections_to_not_be_equal_it_should_enumerate_the_subject_only_once`
 rather be using a `CountingGenericLazyEnumerable`?

docs/_pages/upgradingtov6.md
jnyrup: That's a left-over note, I'll remove it.
I thought about writing about how this give fewer allocations due to boxing, but I ended deciding it wasn't relevant.

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.cs
dennisdoomen: I just ran that test and it returns `1` for `GetEnumeratorCallCount`, which is what I was expecting.

dennisdoomen:
Hmm, looks like this PR fixes part of what I was trying to fix in #1526

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.cs
jnyrup: I must have edited `When_asserting_identical_collections_to_be_equal_it_should_enumerate_the_subject_only_once` and then changing something else... 🤦 

If you run `When_asserting_identical_collections_to_not_be_equal_it_should_enumerate_the_subject_only_once` it is `0`.

dennisdoomen (approved):

Tests/FluentAssertions.Specs/Equivalency/CollectionEquivalencySpecs.cs
```diff
@@ -1301,21 +1301,6 @@ public void When_asserting_equivalence_of_collections_and_configured_to_use_runt
...
 
-        [Fact]
-        public void When_injecting_a_null_config_to_non_generic_overload_it_should_throw()
-        {
-            // Arrange
-            ICollection collection1 = null;
-            ICollection collection2 = null;
-
-            // Act
-            Action act = () => collection1.Should().BeEquivalentTo(collection2, config: null);
```
jnyrup, line 1326: This used to call `BeEquivalent(IEnumerable, Func<...>)`, but that overload is now removed and it resolves to `BeEquivalent(object, Func<...>)` which we have other tests for.

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.cs
```diff
@@ -24,16 +24,6 @@ public void When_collection_is_expected_to_be_null_and_it_is_it_should_not_throw
...
         }
 
-        [Fact]
-        public void When_a_custom_enumerable_implementation_is_expected_not_to_be_null_and_it_is_it_should_not_throw()
-        {
-            // Arrange
-            var enumerable = new CustomEnumerable();
-
-            // Act / Assert
-            enumerable.Should().NotBeNull();
```
jnyrup, line 34: This now resolves to `ObjectAssertions.NotBeNull` which we have other tests for

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -2917,7 +2916,7 @@ protected static IEnumerable<TExpectation> RepeatAsManyAs<TExpectation>(TExpecta
                 .ClearExpectation();
         }
 
-        protected void AssertSubjectEquality<TActual, TExpectation>(IEnumerable<TExpectation> expectation, Func<TActual, TExpectation, bool> equalityComparison,
+        protected void AssertSubjectEquality<TExpectation>(IEnumerable<TExpectation> expectation, Func<T, TExpectation, bool> equalityComparison,
```
jnyrup, line 2919: The generics was not needed as `TActual` is now equal to `T` and `Subject` is already of type `IEnumerable<T>`

jnyrup (commented):
Just want to double check if you're fine with changing the tests from `CountingGenericCollection` to `CountingGenericEnumerable`.

dennisdoomen:
> Just want to double check if you're fine with changing the tests from `CountingGenericCollection` to `CountingGenericEnumerable`.

Why wouldn't I be? The only reason we need those to ensure we're not accidentally enumerating the enumerable twice

Issue #245, opened 2015-03-17
ShouldAllBeEquivalentTo with Including still compares all properties


MichaelLogutov:
The recent version broken a lot of our tests due to the fact that ShouldAllBeEquivalentTo with Including still compares all properties and not only that specified in equivalency options:

    public class MyClass
    {
        public string A { get; set; }
        public string B { get; set; }
    }

    [TestMethod]
    public void Test ()
    {
        // arrange
        var result = new[] { new MyClass { A = "aaa", B = "bbb" } };

        // assert
        var expected = new MyClass
                        {
                            A = "aaa",
                            B = "ccc"
                        };

        result.ShouldAllBeEquivalentTo (new[] { expected },
                                        options => options.Including (x => x.A));
    }

Got:
```
Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException: Expected item[0].B to be "ccc", but "bbb" differs near "bbb" (index 0).

With configuration:
- Use declared types and members
- Compare enums by value
- Select member Bug.Tests+MyClass.A
- Match member by name (or throw)
- Be strict about the order of items in byte arrays
- FluentAssertions.Equivalency.ShouldAllBeEquivalentToHelper+CollectionMemberOrderingRuleDecorator
```

dennisdoomen:
Thanks for reporting that. I'll try to look at that tonight. Sounds like a pretty serious bug. We must be missing a unit test if that is true.

dennisdoomen:
Might be caused by the same reasons that caused #194.

dennisdoomen:
Yes, we're missing a couple of unit tests...

dennisdoomen:
@vossad01 Do you remember the purpose of [this construction](https://github.com/dennisdoomen/fluentassertions/blame/develop/FluentAssertions.Core/Equivalency/EquivalencyAssertionOptionsBase.cs#L552)?

vossad01:
I would need to dig into the code to give a more precise answer, which I can to tonight if you need.

The just of it was because of ordering of the rules.  Previously, the rules were added immediately which required some methods to clear the collection or rules or remove specific ones.  As I introduced more ways to interact with those rules, the clearing logic seemed a little odd and wanted to impose a invocation order dependency on the methods.

The boolean is used to defer inserting of the standard rules until all other configuration is completed so the configuration did not need to concern itself with preserving them.  Strangely, in searching that file it appears the boolean is no longer used?

[...]

PR #1540, opened 2021-05-02
Removed an old decorator that did not anything anymore

dennisdoomen:
I'm not 100% why we needed the `CollectionMemberAssertionRuleDecorator` in the past, but surely it's not doing anything anymore. Discovered this while working on #1539


jnyrup (approved):
For reference:
It was reported in #245
That bug was fixed in 2028a0602506cd9ea732d74b9378d065b48c26ad with the comment
> Refactored the internal state that keeps track of whether or not to add the default all-including selection rules into a more explicit selection rule property. This was needed because it didn't copy that internal flag when constructing the equivalency options for a collection's member type. Fixes #245.

`When_a_specific_property_is_included_it_should_ignore_the_rest_of_the_properties` was included to verify the fix and that test still exists.

Src/FluentAssertions/Equivalency/CollectionMemberAssertionOptionsDecorator.cs
```diff
@@ -44,7 +44,7 @@ public OrderingRuleCollection OrderingRules
 
         public IEnumerable<IEquivalencyStep> UserEquivalencySteps
         {
-            get { return inner.UserEquivalencySteps.Select(step => new CollectionMemberAssertionRuleDecorator(step)).ToArray(); }
+            get { return inner.UserEquivalencySteps.ToArray(); }
```
jnyrup, line 47: Maybe remove the `.ToArray()`

PR #1542, opened 2021-05-05
Add CreateXmlElement and CreateXmlDocument

jnyrup:
Every time I see the `new XmlDocument()` + `XmlDocument.LoadXml()` combo I wonder why `static XmlDocument XmlDocument.LoadFromXml(string xml)` does not exist in BCL.

### DIFF
--- a/Tests/FluentAssertions.Specs/Xml/XmlElementAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Xml/XmlElementAssertionSpecs.cs
@@ -17,12 +17,8 @@ namespace FluentAssertions.Specs.Xml
             // XmlElementAssertions inherits XmlNodeAssertions.
 
             // Arrange
-            var xmlDoc = new XmlDocument();
-            xmlDoc.LoadXml("<user>grega</user>");
-            var element = xmlDoc.DocumentElement;
-            var expectedDoc = new XmlDocument();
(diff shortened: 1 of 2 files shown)


dennisdoomen (approved):
Not a big fan of helper methods like that....

jnyrup:
I recall you dislike (too) DRY but use test data builders and favour a test being readable from top to bottom without jumping backwards.
My intention with this was to fix a tiny gap in linearity.

This is _effectively_ a single purpose test data builder, but could be rewritten a full blown one.

```
var xmlDocument = new XmlDocumentBuilder().WithXml("...").Build();
```

The continuous improver in me is very interested in where you see the distinction between a helper method and a test data builder.

I thought we used helper methods more places in FA, but looking it up, it seems you have been good at keeping those out.
I'm fine with dropping this as I don't want to introduce an announce to you to solve my own.

dennisdoomen:
IMHO, in this case, putting the helper at the top would have improved the situation a bit. But I have to admit it's mostly a taste thing. I tend to believe that Test Data Builders capture the intent a bit better. Also, helpers tend to grow beyond their initial purpose with all kinds of optional parameters.

jnyrup:
Fully agree on your point of view of all helpers starting out with clear focus, but tend to grow ugly.
I don't want this to be a step stone to introduce helpers.

PR #1546, opened 2021-05-08
Upmerge changes from master

dennisdoomen:
I forgot to regularly merge changes from `master` to `develop`.


Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -21,7 +21,7 @@
       Supports the unit test frameworks MSTest2, NUnit3, XUnit2, MSpec, and NSpec3.
     </PackageDescription>
     <PackageProjectUrl>https://www.fluentassertions.com</PackageProjectUrl>
-    <RepositoryUrl>https://github.com/fluentassertions/fluentassertions.git</RepositoryUrl>
+    <RepositoryUrl>https://github.com/fluentassertions/fluentassertions</RepositoryUrl>
```
jnyrup, line 24: hmm...
https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets has the `.git` extension.

README.md
```diff
@@ -4,8 +4,13 @@
 [![](https://img.shields.io/librariesio/dependents/nuget/FluentAssertions.svg?label=dependent%20libraries)](https://libraries.io/nuget/FluentAssertions)
 ![](https://img.shields.io/badge/release%20strategy-githubflow-orange.svg)
 
-# What is this and why do I need this?
-See https://www.fluentassertions.com for background information, usage documentation, an extensibility guide, support information and more tips & tricks.
+# About this project
+A very extensive set of extension methods that allow you to more naturally specify the expected outcome of a TDD or BDD-style unit tests. Targets .NET Framework 4.5 and 4.7, as well as .NET Core 2.0, .NET Core 2.1, .NET Core 3.0, .NET Standard 1.3, 1.6 2.0 and 2.1.
```
jnyrup, line 8: note: We do don't target all those TFM in develop

docs/_pages/about.md
```diff
@@ -61,26 +61,11 @@ We heavily practice Test Driven Development and one of the promises TDD makes is
...
-* [Adam Voss](https://github.com/adamvoss)
-* [Jonas Nyrup](https://github.com/jnyrup)
-* [Artur Krajweski](https://github.com/krajek)
-
-The [Fluent Assertions logo](./logo/fluent_assertions.svg) was created by
-
-* [IUsername](https://github.com/IUsername)
-
-If you have any comments or suggestions, please let us know via [twitter](https://twitter.com/search?q=fluentassertions&src=typd), through the [issues](https://github.com/dennisdoomen/FluentAssertions/issues) page, or through [StackOverflow](http://stackoverflow.com/questions/tagged/fluent-assertions).
+My name is [Dennis Doomen](https://twitter.com/ddoomen) and I work for [Aviva Solutions](https://www.avivasolutions.nl/) in The Netherlands. I maintain a [blog](https://www.continuousimprover.com/) on my everlasting quest for knowledge that significantly improves the way you build your key systems in an agile world. Fluent Assertions is one of those aspects of that. Since a year two, [Jonas Nyrup](https://github.com/jnyrup) has joined to project and help bring it forward.
```
jnyrup, line 64: "Since a year two" -> "Since 2018"
(Had to look up our Slack conversations to see when I formally joined)

Src/FluentAssertions/FluentAssertions.csproj
dennisdoomen: What are you trying to say?

README.md
dennisdoomen: That's what you get by blindly upmerging things. Apparently this wasn't even mentioned in the read-me on `develop`.

Src/FluentAssertions/FluentAssertions.csproj
jnyrup: That I'm not sure why we would want to remove the `.git` suffix from the url, as the documentation for that attribute uses it.

dennisdoomen: Ah, I get it. I think the `.git` extension is deprecated. If you go to a URL with that postfix, you'll get redirected to the one without it.

jnyrup: To actually answer your question, I didn't knew if it made any difference.
But digging into to it, 
[MSBuild](https://github.com/dotnet/msbuild/blob/47be9829672e2c207626419971aa1cba8f445409/Directory.Build.props#L23) doesn't use the `.git` extension, so I'm fine removing it.

jnyrup (approved):

PR #1549, opened 2021-05-09
Fix sticky footer overlapping content

commonquail:
The sticky footer overlaps and conceals content on nearly every page.

There are [several ways to make sticky footers][1]. Replace the current one with a grid variant. It's not strictly necessary to fix the footer's position in the grid as long as the DOM has the desired element order.

[1]: https://css-tricks.com/couple-takes-sticky-footer/

This makes

before

look more like

after

I've checked

* wide screens
* narrow screens
* short content
* long content
* search box (which I'm not sure does anything...?)

### DIFF
--- a/docs/_sass/minimal-mistakes/_footer.scss
+++ b/docs/_sass/minimal-mistakes/_footer.scss
@@ -9,11 +9,6 @@
   margin-right: 0;
   width: 100%;
   clear: both;
-  /* sticky footer fix start */
-  position: absolute;
-  bottom: 0;
-  height: auto;
-  /* sticky footer fix end */
   margin-top: 3em;
   color: $muted-text-color;
   -webkit-animation: $intro-transition;
--- a/docs/_sass/minimal-mistakes/_page.scss
+++ b/docs/_sass/minimal-mistakes/_page.scss
@@ -20,6 +20,22 @@
   }
 }
 
+body {
+  display: -webkit-box;
+  display: -ms-flexbox;
+  display: flex;
+  min-height: 100vh;
+  -webkit-box-orient: vertical;
+  -webkit-box-direction: normal;
+  -ms-flex-direction: column;
+          flex-direction: column;
+}
+
+.initial-content,
+.search-content {
+  flex: 1;
+}
+
 .page {
   @include breakpoint($large) {
     float: right;
--- a/docs/_sass/minimal-mistakes/_page.scss
+++ b/docs/_sass/minimal-mistakes/_page.scss
@@ -33,7 +33,7 @@ body {
 
 .initial-content,
 .search-content {
-  flex: 1;
+  flex: 1 0 auto;
 }
 
 .page {


dennisdoomen:
That's a nice fix. Could you please retarget to `master`? That's the one that the site is build from.

commonquail:
That should do it.

dennisdoomen (approved):

jnyrup:
Aren't we modifying the theme files directly?
Skimming the changelog it seems https://github.com/mmistakes/minimal-mistakes/pull/2289 might fix the bug.

dennisdoomen:
@commonquail would you be willing to take care of that?

commonquail:
@dennisdoomen I can copy the upstream changes directly (and confirm that it solves this problem). Would it be smarter to move to the Gemfile managed installation or remote theme instead, though?

dennisdoomen:
> Would it be smarter to move to the Gemfile managed installation or remote theme instead, though?

I don't think that would be possible. I applied quite some changes to files that probably originate from the original theme.

jnyrup:
I just tried comparing 4.11.1 from https://rubygems.org/downloads/minimal-mistakes-jekyll-4.11.1.gem with our version.

The changed files are
```
Files original/_data/navigation.yml and FA/_data/navigation.yml differ
Files original/_includes/analytics-providers/google-universal.html and FA/_includes/analytics-providers/google-universal.html differ
Files original/_includes/feature_row and FA/_includes/feature_row differ
Files original/_includes/footer.html and FA/_includes/footer.html differ
Files original/_includes/head/custom.html and FA/_includes/head/custom.html differ
Files original/_includes/head.html and FA/_includes/head.html differ
Files original/_includes/masthead.html and FA/_includes/masthead.html differ
Files original/_includes/scripts.html and FA/_includes/scripts.html differ
Files original/_layouts/single.html and FA/_layouts/single.html differ
Files original/_sass/minimal-mistakes/_archive.scss and FA/_sass/minimal-mistakes/_archive.scss differ
Files original/_sass/minimal-mistakes/_base.scss and FA/_sass/minimal-mistakes/_base.scss differ
Files original/_sass/minimal-mistakes/_buttons.scss and FA/_sass/minimal-mistakes/_buttons.scss differ
Files original/_sass/minimal-mistakes/_masthead.scss and FA/_sass/minimal-mistakes/_masthead.scss differ
Files original/_sass/minimal-mistakes/_page.scss and FA/_sass/minimal-mistakes/_page.scss differ
Files original/_sass/minimal-mistakes/_search.scss and FA/_sass/minimal-mistakes/_search.scss differ
Files original/_sass/minimal-mistakes/_variables.scss and FA/_sass/minimal-mistakes/_variables.scss differ
Files original/assets/css/main.scss and FA/assets/css/main.scss differ
Files original/assets/js/_main.js and FA/assets/js/_main.js differ
Files original/assets/js/main.min.js and FA/assets/js/main.min.js differ
Files original/assets/js/plugins/jquery.greedy-navigation.js and FA/assets/js/plugins/jquery.greedy-navigation.js differ
```

When excluding `navigation.yml` and `jquery.greedy-navigation.js` it's around 155 lines of changes.

```
diff -rwy --suppress-common-lines --exclude={*.min.{css,js},*.yml,*navigation.js} --width=320 original/ FA/ | grep -v 'Only in ' | less
```

I have no idea if we can properly override/hook in without modifying the original files, but it sure would be preferable if we ever need to update the page in hurry for whatever reason.
Alternatively it would be nice if the changes to the files were just in separate commits so they effectively are patches.
If that second approach is also infeasible I'm fine with modifying the files directly.

commonquail:
> Alternatively it would be nice if the changes to the files were just in separate commits so they effectively are patches.

2 of the 3 upstream commits for this issue apply cleanly with `git am --exclude='*min.js' --directory=docs <patch>`. The third one had its change applied by hand some time ago. This is probably workable for several other small upstream changes.

Here's what it looks like now (discoloration is a picture artifact):

sticky-footer-fix

dennisdoomen:
> I have no idea if we can properly override/hook in without modifying the original files, but it sure would be preferable if we ever need to update the page in hurry for whatever reason.
> Alternatively it would be nice if the changes to the files were just in separate commits so they effectively are patches.
> If that second approach is also infeasible I'm fine with modifying the files directly.

I think that could end up becoming quite some work. The mistake I made in the past was to start from a copy of the site instead of the Gem file.

jnyrup (approved):

jnyrup:
@commonquail Thanks for spotting and fixing this.

dennisdoomen:
I'm with @jnyrup here. Thank you very much.

PR #1556, opened 2021-05-11
Add max line length in editor config

eNeRGy164:
As mentioned in #1555, lines should not exceed 130 characters.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,6 +26,7 @@ indent_size = 4
 [*.cs]
 indent_style = space
 indent_size = 4
+max_line_length = 130
 
 # New line preferences
 csharp_new_line_before_open_brace = all


jnyrup:
What tools use that handle and do they emit a warning or an error?

For VS I have now installed "Editor Guidelines" which uses the value to print a red line.
image

jnyrup (approved):

eNeRGy164:
> What tools use that handle and do they emit a warning or an error?
> 

https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#max_line_length
I've none of these tools

> For VS I have now installed "Editor Guidelines" which uses the value to print a red line.
> image

I installed the _Indent Guides_ extension, same effect.

PR #1559, opened 2021-05-14
Robust methodinfo assertions

eNeRGy164:
More coverage of #1039

## IMPORTANT 

### DIFF
--- a/Tests/FluentAssertions.Specs/Types/MethodBaseAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Types/MethodBaseAssertionSpecs.cs
@@ -72,6 +72,21 @@ namespace FluentAssertions.Specs.Types
                     "Expected the return type of method to be *.String *failure message*, but methodInfo is <null>.");
         }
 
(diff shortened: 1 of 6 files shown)


Tests/FluentAssertions.Specs/Types/MethodBaseAssertionSpecs.cs
```diff
@@ -57,6 +56,37 @@ public void When_asserting_a_void_method_returns_string_it_fails_with_a_useful_m
                              "error message, but it is \"System.Void\".");
         }
 
+        [Fact]
+        public void When_subject_is_null_return_should_fail()
```
dennisdoomen, line 60: We really need to improve this naming convention or the grouping of tests, because using the literal name in the test name is what I want to avoid. 

It _would_ be fine if it said: When subject is null, `return` should fail. But without the markup, it's difficult to read as a sentence.

dennisdoomen (approved):

jnyrup (approved):

PR #1563, opened 2021-05-15
Robust TypeSelector Assertions

eNeRGy164:
More coverage of #1039

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Common/Guard.cs
+++ b/Src/FluentAssertions/Common/Guard.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace FluentAssertions.Common
 {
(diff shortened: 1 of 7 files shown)


dennisdoomen (approved):

jnyrup:
I have the same concerns as discussed in #1565.

eNeRGy164:
Will update this one after #1565 & #1569 are merged.

eNeRGy164:
This PR has shrunken quite a bit :)

Src/FluentAssertions/Common/Guard.cs
```diff
@@ -54,6 +56,22 @@ public static void ThrowIfArgumentIsOutOfRange<T>([ValidatedNotNull] T value, st
             }
         }
 
+        public static void ThrowIfArgumentContainsNull<T>([ValidatedNotNull] IEnumerable<T> values, string paramName)
```
jnyrup, line 59: I don't think the attribute should be applied to these methods or the enum ones. 

The purpose of them is to aid some one analyzer in verifying that parameters in public methods are null checked.

As I recall the usage of the attribute is "when this method returns (without excepting), the variable bound to the decorated parameter is verified to be non null".

eNeRGy164: Ah ok, I'll clean them up.

jnyrup (approved):

PR #1564, opened 2021-05-15
Robust XML assertions

eNeRGy164:
More coverage of #1039

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Xml/Equivalency/XmlReaderValidator.cs
+++ b/Src/FluentAssertions/Xml/Equivalency/XmlReaderValidator.cs
@@ -9,16 +9,18 @@ namespace FluentAssertions.Xml.Equivalency
     internal class XmlReaderValidator
     {
         private readonly AssertionScope assertion;
-        private readonly XmlIterator subjectIterator;
-        private readonly XmlIterator expectationIterator;
+        private readonly XmlReader subjectReader;
+        private readonly XmlReader expectationReader;
+        private XmlIterator subjectIterator;
+        private XmlIterator expectationIterator;
         private Node currentNode = Node.CreateRoot();
 
         public XmlReaderValidator(XmlReader subjectReader, XmlReader expectationReader, string because, object[] becauseArgs)
         {
             assertion = Execute.Assertion.BecauseOf(because, becauseArgs);
 
-            subjectIterator = new XmlIterator(subjectReader);
-            expectationIterator = new XmlIterator(expectationReader);
+            this.subjectReader = subjectReader;
+            this.expectationReader = expectationReader;
         }
 
         public void Validate(bool shouldBeEquivalent)
@@ -38,6 +40,21 @@ namespace FluentAssertions.Xml.Equivalency
 
         private Failure Validate()
         {
+            if (subjectReader is null && expectationReader is null)
+            {
+                return null;
+            }
+
+            Failure failure = ValidateAgainstNulls();
+
+            if (failure is not null)
+            {
+                return failure;
+            }
+
+            subjectIterator = new XmlIterator(subjectReader);
+            expectationIterator = new XmlIterator(expectationReader);
+
             while (!subjectIterator.IsEndOfDocument && !expectationIterator.IsEndOfDocument)
             {
                 if (subjectIterator.NodeType != expectationIterator.NodeType)
@@ -53,7 +70,7 @@ namespace FluentAssertions.Xml.Equivalency
                         currentNode.GetXPath());
                 }
 
-                Failure failure = null;
+                failure = null;
 
 #pragma warning disable IDE0010 // The default case handles the many missing cases
                 switch (expectationIterator.NodeType)
@@ -205,6 +222,18 @@ namespace FluentAssertions.Xml.Equivalency
                 return new Failure(
                     "Expected content to be {0} in {context:subject} at {1}{reason}, but found {2}.",
                     expected, currentNode.GetXPath(), subject);
+            }
+
+            return null;
+        }
+
+        private Failure ValidateAgainstNulls()
+        {
+            if ((expectationReader is null) != (subjectReader is null))
+            {
+                return new Failure(
+                    "Expected {context:subject} to be equivalent to {0}{reason}, but found {1}.",
+                    subjectReader, expectationReader);
             }
 
             return null;
--- a/Src/FluentAssertions/Xml/XAttributeAssertions.cs
+++ b/Src/FluentAssertions/Xml/XAttributeAssertions.cs
@@ -33,7 +33,7 @@ namespace FluentAssertions.Xml
         public AndConstraint<XAttributeAssertions> Be(XAttribute expected, string because = "", params object[] becauseArgs)
         {
             Execute.Assertion
-                .ForCondition(Subject.Name == expected.Name && Subject.Value == expected.Value)
+                .ForCondition(Subject?.Name == expected?.Name && Subject?.Value == expected?.Value)
                 .BecauseOf(because, becauseArgs)
                 .FailWith("Expected {context} to be {0}{reason}, but found {1}.", expected, Subject);
 
@@ -55,7 +55,7 @@ namespace FluentAssertions.Xml
         public AndConstraint<XAttributeAssertions> NotBe(XAttribute unexpected, string because = "", params object[] becauseArgs)
         {
             Execute.Assertion
-                .ForCondition(!(Subject.Name == unexpected.Name && Subject.Value == unexpected.Value))
+                .ForCondition(!(Subject?.Name == unexpected?.Name && Subject?.Value == unexpected?.Value))
                 .BecauseOf(because, becauseArgs)
                 .FailWith("Did not expect {context} to be {0}{reason}.", unexpected);
 
@@ -75,11 +75,19 @@ namespace FluentAssertions.Xml
         /// </param>
         public AndConstraint<XAttributeAssertions> HaveValue(string expected, string because = "", params object[] becauseArgs)
         {
-            Execute.Assertion
-                .ForCondition(Subject.Value == expected)
+            bool success = Execute.Assertion
                 .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context} \"{0}\" to have value {1}{reason}, but found {2}.",
-                    Subject.Name, expected, Subject.Value);
+                .ForCondition(Subject is not null)
+                .FailWith("Expected the attribute to have value {0}{reason}, but {context:member} is <null>.", expected);
+
+            if (success)
+            {
+                Execute.Assertion
+                    .ForCondition(Subject.Value == expected)
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith("Expected {context} \"{0}\" to have value {1}{reason}, but found {2}.",
+                        Subject.Name, expected, Subject.Value);
+            }
 
             return new AndConstraint<XAttributeAssertions>(this);
         }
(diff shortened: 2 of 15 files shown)


Src/FluentAssertions/Xml/XAttributeAssertions.cs
```diff
@@ -30,12 +32,23 @@ public XAttributeAssertions(XAttribute attribute)
         /// <param name="becauseArgs">
         /// Zero or more objects to format using the placeholders in <paramref name="because" />.
         /// </param>
+        /// <exception cref="ArgumentNullException"><paramref name="expected"/> is <c>null</c>.</exception>
         public AndConstraint<XAttributeAssertions> Be(XAttribute expected, string because = "", params object[] becauseArgs)
         {
-            Execute.Assertion
-                .ForCondition(Subject.Name == expected.Name && Subject.Value == expected.Value)
+            Guard.ThrowIfArgumentIsNull(expected, nameof(expected));
```
dennisdoomen, line 38: 🤔 I wonder about this one. What if you're comparing two references to an `XAttribute` and both are `null`. Shouldn't this assertion succeed? This is also how `StringAssertions` work.

eNeRGy164: Yeah, that might be a good exception.

jnyrup: And the same applies for `[Not]BeEquivalentTo`

Tests/FluentAssertions.Specs/Xml/XAttributeAssertionSpecs.cs
```diff
@@ -40,6 +40,35 @@ public void When_asserting_an_xml_attribute_is_equal_to_a_different_xml_attribut
...
+                theAttribute.Should().Be(new XAttribute("name", "value"), "we want to test the failure {0}", "message");
+
+            // Assert
+            act.Should().Throw<XunitException>()
+                .WithMessage(
+                    "Expected the attribute to be name=\"value\" *failure message*, but theAttribute is <null>.");
+        }
+
+        [Fact]
+        public void When_asserting_àttribute_should_be_equal_to_null_it_should_throw()
```
lg2de, line 59: You added (accidentally?) an `à` instead of `a`.

Tests/FluentAssertions.Specs/Xml/XAttributeAssertionSpecs.cs
```diff
@@ -87,6 +116,35 @@ public void When_asserting_an_xml_attribute_is_not_equal_to_the_same_xml_attribu
...
+                theAttribute.Should().NotBe(new XAttribute("name", "value"), "we want to test the failure {0}", "message");
+
+            // Assert
+            act.Should().Throw<XunitException>()
+                .WithMessage(
+                    "Did not expect the attribute to be name=\"value\" *failure message*, but theAttribute is <null>.");
+        }
+
+        [Fact]
+        public void When_asserting_àttribute_should_be_different_to_null_it_should_throw()
```
lg2de, line 135: Another `à`.

Tests/FluentAssertions.Specs/Xml/XAttributeAssertionSpecs.cs
eNeRGy164: Yeah, must have hit it while coding in bed :)

Src/FluentAssertions/Xml/XAttributeAssertions.cs
eNeRGy164: It was more work then expected.
Also added more tests to cover these scenarios.

Src/FluentAssertions/Xml/Equivalency/XmlReaderValidator.cs
```diff
@@ -209,5 +221,17 @@ private Failure ValidateText()
 
             return null;
         }
+
+        private Failure ValidateAgainstNulls()
+        {
+            if ((expectationIterator is null) ^ (subjectIterator is null))
```
jnyrup, line 227: When trying this out with
```cs
var document = null as XDocument;
var otherDocument = XDocument.Parse("<parent><child /></parent>");
document.Should().BeEquivalentTo(otherDocument);
```

I get a somewhat unhelpful failure message.
```
Expected document to be equivalent to <null>, but found "".
```

Instead of adding checks in `XmlReaderValidator`, I would prefer if we null checked at call-site.
It's more lines of code, but it avoids adding complexity to `XmlReaderValidator`.

Something like this
```cs
public AndConstraint<XDocumentAssertions> BeEquivalentTo(XDocument expected, string because = "", params object[] becauseArgs)
{
    if (Subject is null && expected is null)
    {
        return new AndConstraint<XDocumentAssertions>(this);
    }

        .ForCondition(Subject is not null && expected is not null)
        .FailWith(() => new FailReason("Expected {context:subject} to be equivalent to {0}{reason}, but found {1}.",
            Subject?.ToString(), expected?.ToString()));

    if (success)
    {
        using (XmlReader subjectReader = Subject.CreateReader())
        using (XmlReader otherReader = expected.CreateReader())
        {
            var xmlReaderValidator = new XmlReaderValidator(subjectReader, otherReader, because, becauseArgs);
            xmlReaderValidator.Validate(shouldBeEquivalent: true);
        }
    }

    return new AndConstraint<XDocumentAssertions>(this);
}
```

The code above gives this failure message.
```
Expected document to be equivalent to <null>, but found "<parent>
      <child />
    </parent>".
```

eNeRGy164: Why would `XmlReaderValidator` not handle null values on its own? I took this code from `StringEqualityValidator` (`StringValidator`) where it is responsible for the null checks.

jnyrup: As `StringValidator` also does it that way I'm fine with it and your solution solves the problem 👍 
It's internal so we can always improve later if we want to.

eNeRGy164: The correct content is now also part of the error message.

jnyrup (approved):

Src/FluentAssertions/Xml/Equivalency/XmlReaderValidator.cs
```diff
@@ -209,5 +226,27 @@ private Failure ValidateText()
 
             return null;
         }
+
+        private Failure ValidateAgainstNulls()
+        {
+            if ((expectationReader is null) ^ (subjectReader is null))
```
dennisdoomen, line 232: 🤔 I think it would be better to use more ordinary `&&`, `||` and negation instead of using xor. I had to look up this operator to remind myself what it did again.

Src/FluentAssertions/Xml/Equivalency/XmlReaderValidator.cs
```diff
@@ -209,5 +226,27 @@ private Failure ValidateText()
...
+            {
+                return new Failure(
+                    "Expected {context:subject} to be equivalent to {0}{reason}, but found {1}.",
+                    GetReaderContent(subjectReader), GetReaderContent(expectationReader));
+            }
+
+            return null;
+
+            static string GetReaderContent(XmlReader reader)
```
dennisdoomen, line 241: 🤔 Isn't it cleaner to create a little `IValueFormatter` for XML readers?

Src/FluentAssertions/Xml/Equivalency/XmlReaderValidator.cs
```diff
@@ -38,6 +40,21 @@ public void Validate(bool shouldBeEquivalent)
 
         private Failure Validate()
         {
+            if (subjectReader is null && expectationReader is null)
```
dennisdoomen, line 43: 🤔 It's kind of weird that this block and `ValidateAgainstNull` are both testing against `null`. Shouldn't this be combined?

Src/FluentAssertions/Xml/Equivalency/XmlReaderValidator.cs
eNeRGy164: 🤡 I Just copied it from https://github.com/fluentassertions/fluentassertions/blob/develop/Src/FluentAssertions/Primitives/StringValidator.cs

Src/FluentAssertions/Xml/Equivalency/XmlReaderValidator.cs
eNeRGy164: 🤡 I Just copied it from https://github.com/fluentassertions/fluentassertions/blob/develop/Src/FluentAssertions/Primitives/StringValidator.cs

Src/FluentAssertions/Xml/Equivalency/XmlReaderValidator.cs
eNeRGy164: Ah, didn't think of that

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -123,35 +131,43 @@ public void When_asserting_a_xml_document_is_not_equal_to_the_same_xml_document_
         }
 
         [Fact]
-        [SuppressMessage("ReSharper", "ExpressionIsAlwaysNull")]
```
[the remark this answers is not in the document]
eNeRGy164: I'm not sure what to do here as I have no Rider to confirm if the behavior is stall present.
Visual Studio is actively saying the attribute is redundant, and I also would expect this message then on every unit test with `null`'s and this is the only place in the past weeks where I encountered these attributes.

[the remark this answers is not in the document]
jnyrup: I have some other analyzers to adjust, so I'll take of this as well.
So the PR is fine.

Src/FluentAssertions/Xml/Equivalency/XmlReaderValidator.cs
jnyrup: It's only 20 months since you looked it up last time 🙃 
https://github.com/fluentassertions/fluentassertions/pull/1141#discussion_r326902258

Would changing it to 
```cs
if ((expectationReader is null) != (subjectReader is null))
```
be more readable to you?
It's an equivalent way to expresses "one, but not both".

dennisdoomen: 20 months?? I don't even remember what I did last week ;-). 

But yes, much more expressive.

jnyrup: FYI, I've changed the other occurences of `^` to `!=` in #1574.

eNeRGy164: Changed

Src/FluentAssertions/Xml/Equivalency/XmlReaderValidator.cs
eNeRGy164: You can't combine it, as it is a tri-state:

1. Both `null` -> Stop validation without failure
2. One `null` -> Stop validation with failure
3. None `null` -> Continue validation

dennisdoomen: Fair enough

dennisdoomen (approved):

PR #1565, opened 2021-05-16
Update PropertyInfoSelector assertions

eNeRGy164:
More coverage of #1039

_All these null reference exception paths.. already made me doubt TDD practices since a training on 27 Februari 2010 in Doorn_ 😉 

## IMPORTANT 


Tests/FluentAssertions.Specs/Types/PropertyInfoSelectorAssertionSpecs.cs
```diff
@@ -57,6 +58,26 @@ public void
                    "String FluentAssertions*ClassWithNonVirtualPublicProperties.ProtectedNonVirtualProperty");
         }
 
+        [Fact]
+        public void When_property_info_selector_is_null_then_be_virtual_should_fail()
+        {
+            // Arrange
+            PropertyInfoSelector propertyInfoSelector = null;
```
dennisdoomen, line 65: I don't think we should be directly testing against the internal classes. The `PropertyInfoSelector` is not a "public" contract to consumers of our library. And if you go through the real API, you should never reach this path. 

This applies to the majority if not the entire PR

dennisdoomen (commented):
I wonder whether it was worth the trouble. Especially since you're now directly testing the internal

Tests/FluentAssertions.Specs/Types/PropertyInfoSelectorAssertionSpecs.cs
jnyrup: I agree on this.
How about we just add null guards to these two methods and get back to the Cava?

```cs
public static PropertyInfoSelectorAssertions Should(this PropertyInfoSelector propertyInfoSelector)
public PropertyInfoSelectorAssertions(params PropertyInfo[] properties)
```

eNeRGy164: Also putting the discussion from #1563 here.

Don't hold back if things should be different, I really don't mind changing things.

However, I did get at least a null reference exception or other internal exception for every test I added. I did not add any code that was not fixing such a path.

But, I think the TypeSelector is way more public then you maybe know.
It's part of `TypeExtensions` and `TypeEnumerableExtensions`. So basically on every `Assembly` and `Type`. as the extensions are part of the `FluentAssertions` root namespace.

You can currently construct it in 3 problematic ways:

```csharp
new TypeSelector((Type)null)
new TypeSelector((Type[])null)
new TypeSelector(new Type[] { null })
```

The first one has the same behavior as the third one.

The first and third will result in broken methods that expect every item in the `types` list to be a Type.
The second one will break every method accessing the `types` list itself.

Other then the constructor, this also happens with:

```csharp
((Assembly)null).Types
((Type)null).Types
((Type[])null).Types
```

and of course `((TypeSelector)null).Should()` breaks with a null reference exception.

We could make the constructor of TypeSelector `internal`. It does not break anything other then the public API. 
Making the types list initialized by default would also prevent some exceptions.
Just null checking then adding items to the list is possible.
Both in a relaxed matter (adding null(s) result in an empty or filtered list) or more restrictive manner, throwing argument exceptions when a null is detected.

Not having public constructors and having Guards at all the extension methods should create a moat that will prevent a lot of the other paths that are now possible with null reference exceptions.

Much of this applies the same to `PropertyInfoSelector` and `MethodInfoSelector` as they also have these kinds of extensions on `Type` and `TypeSelector`.

jnyrup: No doubt you can get a `NullReferenceException` if you really want one.

I'm always surprised in how people use FA, so I can't reject that someone would write something that could call `Types` with a null value.
E.g.
```cs
AllTypes.From(...).FirstOrDefault().Types().Should().BeSealed();
```
or
```cs
new MyClass().GetResult().Properties().Should().BeWritable();

class MyClass
{
    public Type GetResult() => null;
}
```

I think the extension methods should have guards instead of passing a null value along.

I'm not sure if we should make the constructors internal, at least I use them directly in a test helper at work.
```
public static PropertyInfoSelector Properties(this IEnumerable<Type> types) => new PropertyInfoSelector(types)
```

eNeRGy164: I tested it, and it compiles and all test run.
I think this is because of the `InternalsVisibleToAttribute`

But I will look if I can build some fences with Guards

dennisdoomen: The problem is that we also try to be open to folks that want to extend our library. And then exposing the public constructors make sense. But we shouldn't use those constructors directly in our tests. We should only use the public API as we think people should be using the library.

eNeRGy164: New approach.

I think this closes all reasonable paths.

Src/FluentAssertions/Types/PropertyInfoSelector.cs
```diff
@@ -18,18 +18,23 @@ public class PropertyInfoSelector : IEnumerable<PropertyInfo>
...
         /// Initializes a new instance of the <see cref="PropertyInfoSelector"/> class.
         /// </summary>
         /// <param name="types">The types from which to select properties.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="types"/> is <c>null</c>.</exception>
         public PropertyInfoSelector(IEnumerable<Type> types)
         {
-            selectedProperties = types.SelectMany(t => t
+            Guard.ThrowIfArgumentIsNull(types, nameof(types));
+
+            selectedProperties = types.Where(t => t is not null).SelectMany(t => t
```
jnyrup, line 37: Is the `Where(t => t is not null)` intended to defensively guard against calling the other ctor with a null type?
If yes, then that constructor could instead be re-written to
```cs
public PropertyInfoSelector(Type type)
    : this(new[] { type ?? throw new ArgumentNullException(nameof(type)) })
```

jnyrup (commented):
Just a single question, otherwise I like this approach 👍

Src/FluentAssertions/Types/PropertyInfoSelector.cs
eNeRGy164: No, it was intended to be defensive against

```csharp
new PropertyInfoSelector(new { typeof(int), null, typeof(string) });
```

jnyrup: I see...
I don't think we should filter off invalid values like in that example.

As a user I would rather have an Exception thrown in my face here than FA hiding my bug.
I've spent way to many hours in a debugger hunting nulls stemming from methods returning null when getting invalid data.

eNeRGy164: Made it more restrictive

jnyrup (approved):

Tests/FluentAssertions.Specs/Types/PropertyInfoSelectorSpecs.cs
```diff
@@ -1,14 +1,56 @@
...
     public class PropertyInfoSelectorSpecs
     {
+        [Fact]
+        public void When_property_info_selector_is_created_with_a_null_type_it_should_throw()
+        {
+            // Arrange
+            PropertyInfoSelector propertyInfoSelector;
+
+            // Act
+            Action act = () => propertyInfoSelector = new PropertyInfoSelector((Type)null);
```
dennisdoomen, line 19: I'm still uncomfortable directly using the "internal" types in tests, but we have no clear guidance on what we should and should not do.

dennisdoomen (approved):

jnyrup (approved):

PR #1567, opened 2021-05-19
Make StreamAssertions null-safe to being wrapped in an AssertionScope

jnyrup:
When an assertion on a stream was wrapped in an `AssertionScope` we would continue executing after a null check failed and get an NRE.

We could consider always augmenting our null tests with an enclosing `AssertionScope` to verify the null-safeness.
That's why I hijacked the original null tests for stream assertions.

cc: @eNeRGy164

### DIFF
--- a/Src/FluentAssertions/Streams/BufferedStreamAssertions.cs
+++ b/Src/FluentAssertions/Streams/BufferedStreamAssertions.cs
@@ -41,16 +41,20 @@ namespace FluentAssertions.Streams
         /// </param>
         public AndConstraint<TAssertions> HaveBufferSize(int expected, string because = "", params object[] becauseArgs)
         {
-            Execute.Assertion
+            bool success = Execute.Assertion
                 .BecauseOf(because, becauseArgs)
-                .WithExpectation("Expected the buffer size of {context:stream} to be {0}{reason}, ", expected)
                 .ForCondition(Subject is not null)
-                .FailWith("but found a <null> reference.")
-                .Then
-                .ForCondition(Subject.BufferSize == expected)
-                .FailWith("but it was {0}.", Subject.BufferSize)
-                .Then
-                .ClearExpectation();
+                .FailWith("Expected the buffer size of {context:stream} to be {0}{reason}, but found a <null> reference.",
+                    expected);
+
+            if (success)
+            {
+                Execute.Assertion
+                    .BecauseOf(because, becauseArgs)
+                    .ForCondition(Subject.BufferSize == expected)
+                    .FailWith("Expected the buffer size of {context:stream} to be {0}{reason}, but it was {1}.",
+                        expected, Subject.BufferSize);
+            }
 
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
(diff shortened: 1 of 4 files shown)


jnyrup:
Hmm... The subsequent predicates are still invoked 🤔 
See #1568

jnyrup (commented):
Two considerations.

> If a class derived from Stream does not support seeking, calls to Length, SetLength, Position, and Seek may throw a NotSupportedException.

https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.canseek?view=net-5.0#remarks

This means that we might be checking an implementation details only applicable to a subset of `Stream`s.
Though I haven't found an example in BCL where it does not hold.

Not all stream supports `Length` and `Position`.
E.g. `BrotliStream` always throws a `NotSupportedException`.

So instead we might want to just wrap reading `Position` in a try-catch and see what happens.

dennisdoomen:
> So instead we might want to just wrap reading `Position` in a try-catch and see what happens.

But if the underlying stream doesn't support Length and Position, shouldn't we re-throw that exception and return something useful?

dennisdoomen (approved):

jnyrup:
> But if the underlying stream doesn't support Length and Position, shouldn't we re-throw that exception and return something useful?

Do you mean _catch_ the exception and return something useful?

```cs
    .FailWith("Expected the position of {context:stream} to be {0}{reason}, but found a <null> reference.",
        expected);

if (success)
{
    try
    {
        long position = Subject.Position;

        Execute.Assertion
            .ForCondition(position == expected)
            .FailWith("Expected the position of {context:stream} to be {0}{reason}, but it was {1}.",
                expected, position);
    }
    catch (Exception exception)
    {
        Execute.Assertion
            .FailWith("Expected the position of {context:stream} to be {0}{reason}, but it failed with:{1}{2}",
                expected, Environment.NewLine, exception.Message);
    }
}

```

jnyrup:
Since I changed the PR since the approval, I'd like a re-approval of the latest changes before merging.

dennisdoomen:
> Do you mean _catch_ the exception and return something useful?

I meant catching the specific exception that you talked about when you try to access a method or property that throws `NotImplementedException`. But now I'm wondering whether we should just let the framework throw the exception. Maybe it shouldn't be FA's responsibility to handle those cases.

jnyrup:
[`Stream.Position`](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.position?view=net-5.0) is documented to throw `IOException`, `NotSupportedException` or `ObjectDisposedException`.

To avoid relying too much on implementation/documentation details, I think we should either:
* catch _all_ exceptions and provide a failure message, or
* catch _no_ exceptions and just let the exception bubble up.

For reference, these three methods also catches exceptions to provide a failure message.
* `BeBinarySerializable`
* `BeDataContractSerializable`
* `BeXmlSerializable`

Using the first approach allows to continuing on the assertion scope.

```cs
using var stream = new NonSeekableStream();

stream.Should().HavePosition(10, "we want to test the failure {0}", "message");
stream.Should().NotBeWritable();
```

with try/catch
```
Expected the position of stream to be 10L because we want to test the failure message, but it failed with:"
""GetPositionExceptionMessage"
Expected stream not to be writable, but it was.
```

without
```
System.NotSupportedException : GetPositionExceptionMessage
```

dennisdoomen:
Then I think we _should_ rely on the documented exception types. Otherwise we end up swallowing other problems.

jnyrup:
Fine. 
I'll change it to catch the three documented exceptions.

PR #1569, opened 2021-05-19
Guard MethodInfoSelector(Assertions) against null reference exceptions

eNeRGy164:
More coverage of #1039

Follows the changes made in #1565

Will need a rebase after #1565 is merged.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Types/MethodInfoSelector.cs
+++ b/Src/FluentAssertions/Types/MethodInfoSelector.cs
@@ -18,6 +18,7 @@ namespace FluentAssertions.Types
         /// Initializes a new instance of the <see cref="MethodInfoSelector"/> class.
         /// </summary>
         /// <param name="type">The type from which to select methods.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="type"/> is <c>null</c>.</exception>
         public MethodInfoSelector(Type type)
             : this(new[] { type })
         {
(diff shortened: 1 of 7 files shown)


Src/FluentAssertions/Types/MethodInfoSelector.cs
```diff
@@ -27,8 +28,16 @@ public MethodInfoSelector(Type type)
         /// Initializes a new instance of the <see cref="MethodInfoSelector"/> class.
         /// </summary>
         /// <param name="types">The types from which to select methods.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="types"/> is <c>null</c>.</exception>
         public MethodInfoSelector(IEnumerable<Type> types)
         {
+            Guard.ThrowIfArgumentIsNull(types, nameof(types));
+
+            if (types.Any(t => t is null))
```
dennisdoomen, line 36: 🔧 Maybe add a dedicated `Guard` for that?

dennisdoomen (approved):

Src/FluentAssertions/Types/MethodInfoSelector.cs
eNeRGy164: Maybe do that in a separate PR, as it is also applicable for `PropertyInfoSelector`

jnyrup (approved):

PR #1571, opened 2021-05-21
Add first-class support for C# 9.0 records

### LINKED ISSUE #1451: Detailed message for graph comparison of C# 9 records
Hi!
When used with regular classes, comparing two different objects yields a message detailing what exactly is different, like in my case:
```
Expected member WorkingHoursJson to be <null>, but found 
"[{"dayOfWeek": "friday", "latestTime": 120, "earliestTime": 60}]"
```
When I make that class a record, those details are lost, and both objects are just dumped into message:
`Expected createdBusiness to be BusinessDto { Id = 1, Name = Halvorson - Orn, CreatedAt = 12.09.2020 13:47:54, UpdatedAt = 12.09.2020 13:47:54, EnableA...`
...and so on. Is there a way to get detailed messages for records?
Thanks!

EDIT: found ComparingByMembers option, this solves the problem. But I think it would be better to compare records by members by default.

dennisdoomen:
* By default, records are compared by their members
* This can be overridden by `ComparingRecordsByValue` or `ComparingByValue<T>`
* Also improved the reporting on how types with members are compared.

Fixes #1451

### DIFF
--- a/Src/FluentAssertions/Common/TypeExtensions.cs
+++ b/Src/FluentAssertions/Common/TypeExtensions.cs
@@ -580,6 +580,16 @@ namespace FluentAssertions.Common
             return hasCompilerGeneratedAttribute;
         }
 
+        public static bool IsRecord(this Type type)
+        {
+            return type.GetMethod("<Clone>$") is not null &&
+                type.GetTypeInfo()
+                    .DeclaredProperties
+                    .FirstOrDefault(p => p.Name == "EqualityContract")?
+                    .GetMethod?
+                    .GetCustomAttribute(typeof(CompilerGeneratedAttribute)) is not null;
+        }
+
         private static bool IsKeyValuePair(Type type)
         {
             return type.IsGenericType && type.GetGenericTypeDefinition() == typeof(KeyValuePair<,>);
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
@@ -57,6 +57,7 @@ namespace FluentAssertions.Equivalency
         private bool includeProperties;
 
         private bool includeFields;
+        private bool compareRecordsByValue;
 
         #endregion
 
@@ -79,6 +80,7 @@ namespace FluentAssertions.Equivalency
             useRuntimeTyping = defaults.UseRuntimeTyping;
             includeProperties = defaults.IncludeProperties;
             includeFields = defaults.IncludeFields;
+            compareRecordsByValue = defaults.CompareRecordsByValue;
 
             ConversionSelector = defaults.ConversionSelector.Clone();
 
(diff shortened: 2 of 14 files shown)


Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -580,6 +580,16 @@ private static bool IsAnonymousType(this Type type)
...
         }
 
+        public static bool IsRecord(this Type type)
+        {
+            return type.GetMethod("<Clone>$") != null &&
+                type.GetTypeInfo()
+                    .DeclaredProperties
+                    .FirstOrDefault(p => p.Name == "EqualityContract")?
+                    .GetMethod?
+                    .GetCustomAttribute(typeof(CompilerGeneratedAttribute)) != null;
```
jnyrup, line 590: nit, we've changed all instances of `!= null` to `is not null`

Tests/FluentAssertions.Specs/Equivalency/BasicEquivalencySpecs.cs
```diff
@@ -3515,6 +3524,138 @@ public void When_a_tuple_is_compared_it_should_compare_its_components()
...
+            };
+
+            actual.Should().BeEquivalentTo(expected, o => o
+                .ComparingRecordsByValue()
+                .ComparingRecordsByMembers());
+        }
+
+        private record MyRecord
+        {
+            public string StringProperty;
```
jnyrup, line 3652: Why is `StringProperty` a Pascal cased field?

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
```diff
@@ -180,6 +184,10 @@ EqualityStrategy IEquivalencyAssertionOptions.GetEqualityStrategy(Type type)
             {
                 strategy = EqualityStrategy.ForceEquals;
             }
+            else if (type.IsRecord())
```
jnyrup, line 187: Just sharing thoughts.
If any users report a slowdown on `BeEquivalentTo`, caching of `IsRecord` might be a place to look.

jnyrup (approved):

Tests/FluentAssertions.Specs/Equivalency/BasicEquivalencySpecs.cs
dennisdoomen: Because I made it a field later on. Renamed it. But it's still public, so the casing is correct.

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
dennisdoomen: I still wanted to do a performance test on the entire test suite. Will add it to the 6.0 release list.

rizi:
@dennisdoomen what does CompareByMembers do and what does CompareByValue do, what's the difference between them when talking about records?
Br

dennisdoomen:
@rizi see https://github.com/fluentassertions/fluentassertions/blob/release-6.0/docs/_pages/objectgraphs.md#value-types

rizi:
> @rizi see https://github.com/fluentassertions/fluentassertions/blob/release-6.0/docs/_pages/objectgraphs.md#value-types

Thank you very much for the fast response, the docs are excellent and now the naming makes absolutely sense.
Br

PR #1572, opened 2021-05-22
Added native support for XDocument, XElement and XAttribute properties and fields to BeEquivalentTo

### LINKED ISSUE #1535: BeEquivalentTo on a graph should compare XElements using deep comparisons
The problem is that `BeEquivalentTo` on an `XDocument` does a deep XML comparison. But if you use `BeEquivalentTo` on an object that has an `XDocument` property, it falls back on equality/reference checks.

dennisdoomen:
`BeEquivalentTo` now supports properties and fields of type `XDocument`, `XElement` and `XAttribute`. This will make graphs with such type behave the same way as calling `BeEquivalentTo` on an `XElement` directly. For `XAttribute`, we use its `Be` method. 

* [x] ~~Documentation~~ _Can't think of a good place to mention this_
* [x] Release Notes

Fixed #1535


Src/FluentAssertions/Equivalency/XDocumentEquivalencyStep.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace FluentAssertions.Equivalency
+{
+    public class XDocumentEquivalencyStep : EquivalencyStep<XDocument>
+    {
+        protected override EquivalencyResult OnHandle(Comparands comparands, IEquivalencyValidationContext context, IEquivalencyValidator nestedValidator)
+        {
+            var subject = (XDocument)comparands.Subject;
+            var expectation = (XDocument)comparands.Expectation;
+
+            subject.Should().BeEquivalentTo(expectation, context.Reason.FormattedMessage, context.Reason.Arguments);
```
jnyrup, line 12: I haven't thought through if this is problematic, but I observe that this does not take an `EquivalencyAssertionOptions`.
So it cannot be customized unless setting the overrides globally.

dennisdoomen: There's nothing to customize right now. See https://github.com/fluentassertions/fluentassertions/blob/develop/Src/FluentAssertions/Xml/XDocumentAssertions.cs#L83

jnyrup (approved):

Issue #744, opened 2018-01-27
Redundant BindingFlags


jnyrup:
From `TypeExtensions.cs`

```csharp
private const BindingFlags PublicMembersFlag =
BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance;

private const BindingFlags AllMembersFlag =
PublicMembersFlag | BindingFlags.NonPublic | BindingFlags.Static;
```

Note how `BindingFlags.NonPublic` is included in both constants.

Is that just redundant?

dennisdoomen:
No, that can be removed.

jnyrup:
Come again?

dennisdoomen:
I mean the duplicate `BindingFlags.NonPublic`

jnyrup:
I just wondered since 

```csharp
private static IEnumerable<PropertyInfo> GetPublicProperties(Type type)
{
    return type.GetProperties(PublicMembersFlag);
}
```

uses `PublicMembersFlag`, which includes `BindingFlags.NonPublic`, and hence `GetPublicProperties` will also return private properties.

dennisdoomen:
Sure. Just remove it.

Issue #899, opened 2018-08-21
Smarter formatting


jnyrup:
The `DefaultFormatter` recursively prints all non-private members of an object graph to a limit of 5 levels.
#889 showed that with a limit of 15 this could cause an OOM.

```csharp
var col = new HtmlNodeCollection(new HtmlNode(HtmlNodeType.Element, new HtmlDocument(), 0));
col.Add(new HtmlNode(HtmlNodeType.Element, new HtmlDocument(), 0));
col.Should().BeNullOrEmpty();
```

When I run the test, the failure message contains both the internal field `_ownerdocument` and the public property `OwnerDocument`.
OwnerDocument is just returning _ownerdocument.
https://github.com/zzzprojects/html-agility-pack/blob/master/src/HtmlAgilityPack.Shared/HtmlNode.cs

Can we somehow limit the formatter to only select internal members if their `DeclaringType` is marked with `InternalsVisibleTo` the assembly containing the unit test?
Maybe the `AppDomain` is of any help?

I don't think we should exclude internal members in general, as people tend to use internal access modifier to expose the members to unit tests.

dennisdoomen:
I think `BeEquivalentTo` is also including ìnternal` members because of the same shared `GetNonPrivateMembers` method and should be changed as well.

dennisdoomen:
I suggest to add some acceptance criteria:
* All include `public` properties and fields 
* Include only `internal` properties and fields if FA was explicitly set-up to see those fields
* Don't include the same object twice

jnyrup:
If that's the case, yes.
The formatting should reflect the members we compare

dennisdoomen:
Isn't this solved by the #1014?

jnyrup:
#1469 prevents it from throwing an OOM, so the _main_ problem is solved ✅ 

There are still related things that would be nice to improve when formatting complex object graphs, but I don't think any of those are a blocker for shipping 6.0.

Many of the properties on `HtmlNode` are simply forwarding an internal field, so it would be nice to exclude fields and/or internal members from the failure message.
That helps hiding the unimportant and emphasizing the important.

A workaround is to add a custom formatter for `HtmlNode` and exclude the unwanted members.
If you want to do that for multiple types, it may be nice with built-in handles to control that, e.g. on the `FormattingOptions`.


When using `ExcludingFields`, fields could be excluded from the failure message
```cs
[...]

PR #1575, opened 2021-05-24
Optionally include internal members in equivalency assertions

### LINKED ISSUE #1205: BeEquivalentTo does not exclude internal members
`BeEquivalentTo` fails if an `internal` property has a different value.

The [documentation](https://fluentassertions.com/objectgraphs/), section "Including properties and/or fields" states:

> Barring other configuration, Fluent Assertions will include all **public** properties and fields

However, it also seems to take into account all `internal` properties and fields.

Who is wrong? The documentation or the implementation? What is `BeEquivalentTo` supposed to do?

In the following code sample, PrivateMembersShouldNotBeCompared succeeds but InternalMembersShouldNotBeCompared fails on the compare of property `InternalProperty`.
```C#
using FluentAssertions;
using Microsoft.VisualStudio.TestTools.UnitTesting;

namespace UnitTestProject
{
    [TestClass]
    public class TestBeEquivalentTo
    {
        class TestSubject
        {
            private string PrivateProperty { get; }
            internal string InternalProperty { get; }
            public string PublicProperty { get; }

            public TestSubject(string privateString, string internalString, string publicString)
            {
                PrivateProperty = privateString;
                InternalProperty = internalString;
                PublicProperty = publicString;
            }
        }

        [TestMethod]
        public void PrivateMembersShouldNotBeCompared()
        {
            TestSubject first = new TestSubject("A", "A", "A");
            TestSubject second = new TestSubject("B", "A", "A");

            first.Should().BeEquivalentTo(second);
        }

        [TestMethod]
        public void InternalMembersShouldNotBeCompared()
        {
            TestSubject first = new TestSubject("A", "A", "A");
            TestSubject second = new TestSubject("B", "B", "A");

            first.Should().BeEquivalentTo(second);
        }
    }
}
```

dennisdoomen:
`BeEquivalentTo` will no longer include `internal` properties and fields, unless `IncludingInternalProperties` or `IncludingInternalFields` is used

Fixes #1205, #744

* [x] Release notes

### DIFF
--- a/Src/FluentAssertions/Common/TypeExtensions.cs
+++ b/Src/FluentAssertions/Common/TypeExtensions.cs
@@ -10,11 +10,14 @@ namespace FluentAssertions.Common
 {
     internal static class TypeExtensions
     {
-        private const BindingFlags PublicMembersFlag =
+        private const BindingFlags PublicInstanceMembersFlag =
+            BindingFlags.Public | BindingFlags.Instance;
+
+        private const BindingFlags AllInstanceMembersFlag =
             BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance;
 
-        private const BindingFlags AllMembersFlag =
-            PublicMembersFlag | BindingFlags.NonPublic | BindingFlags.Static;
+        private const BindingFlags AllStaticAndInstanceMembersFlag =
+            PublicInstanceMembersFlag | BindingFlags.NonPublic | BindingFlags.Static;
 
         public static bool IsDecoratedWith<TAttribute>(this Type type)
             where TAttribute : Attribute
@@ -167,7 +170,7 @@ namespace FluentAssertions.Common
         public static PropertyInfo FindProperty(this Type type, string propertyName, Type preferredType)
         {
             List<PropertyInfo> properties =
-                type.GetProperties(PublicMembersFlag)
+                type.GetProperties(AllInstanceMembersFlag)
                     .Where(pi => pi.Name == propertyName)
                     .ToList();
 
@@ -185,7 +188,7 @@ namespace FluentAssertions.Common
         public static FieldInfo FindField(this Type type, string fieldName, Type preferredType)
         {
             List<FieldInfo> properties =
-                type.GetFields(PublicMembersFlag)
+                type.GetFields(AllInstanceMembersFlag)
                     .Where(pi => pi.Name == fieldName)
                     .ToList();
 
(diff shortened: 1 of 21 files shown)


Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -194,46 +197,62 @@ public static FieldInfo FindField(this Type type, string fieldName, Type preferr
...
-        public static IEnumerable<PropertyInfo> GetNonPrivateProperties(this Type typeToReflect,
-            IEnumerable<string> filter = null)
+        public static IEnumerable<PropertyInfo> GetNonPrivateProperties(this Type typeToReflect, MemberVisibility visibility)
         {
             IEnumerable<PropertyInfo> query =
-                from propertyInfo in GetPropertiesFromHierarchy(typeToReflect)
+                from propertyInfo in GetPropertiesFromHierarchy(typeToReflect, visibility)
                 where HasNonPrivateGetter(propertyInfo)
                 where !propertyInfo.IsIndexer()
-                where filter is null || filter.Contains(propertyInfo.Name)
```
jnyrup, line 212: Was a `filter` never provided by any caller?
Is the functionality handled elsewhere since it was originally added?

Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -194,46 +197,62 @@ public static FieldInfo FindField(this Type type, string fieldName, Type preferr
...
+        private static IEnumerable<PropertyInfo> GetPropertiesFromHierarchy(Type typeToReflect, MemberVisibility memberVisibility)
+        {
+            bool includeInternals = memberVisibility.HasFlag(MemberVisibility.Internal);
+
+            return GetMembersFromHierarchy(typeToReflect, type =>
+            {
+                return type
+                    .GetProperties(AllInstanceMembersFlag)
+                    .Where(property => property.GetMethod?.IsPrivate == false)
+                    .Where(property => includeInternals || property.GetMethod?.IsAssembly == false)
```
jnyrup, line 228: How does this handle properties marked as `private protected` or `protected internal`?

Src/FluentAssertions/Formatting/DefaultValueFormatter.cs
```diff
@@ -58,7 +59,7 @@ public void Format(object value, FormattedObjectGraph formattedGraph, Formatting
         /// <remarks>The default is all non-private members.</remarks>
         protected virtual MemberInfo[] GetMembers(Type type)
         {
-            return type.GetNonPrivateMembers().ToArray();
+            return type.GetNonPrivateMembers(MemberVisibility.Public).ToArray();
```
jnyrup, line 62: Does this mean that if two objects only differ in an internal member, the formatted objects only contains the identical public members?

Src/FluentAssertions/Common/TypeExtensions.cs
dennisdoomen: I tried to look back until 4.x and it wasn't even used then.

Src/FluentAssertions/Formatting/DefaultValueFormatter.cs
dennisdoomen: Yep. Good point. Maybe we should include the internal `members` as well? Making it configurable seems a bit much to me.

jnyrup: Without having any data to back me up, I think it might become too noisy for the default case if we include internal members.
E.g. for #889 one of the problems was that [HtmlNode](https://github.com/zzzprojects/html-agility-pack/blob/master/src/HtmlAgilityPack.Shared/HtmlNode.cs) has +20 internal backing fields.
Even with #899 when excluding already processed references, the formatted output has a lot of `avoiding circular references`.

If the failure message is clear enough about the missing/non-matching internal members, I would go with that for now and only print the public ones.

Src/FluentAssertions/Common/TypeExtensions.cs
dennisdoomen: AFAIK `private protected` is not relevant here, but `protect internal` was. I've added extra tests to cover this scenario.

Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -284,12 +303,12 @@ private static Type[] GetInterfaces(Type type)
 
         private static PropertyInfo[] GetPublicProperties(Type type)
```
jnyrup, line 304: `GetPublicProperties` and `GetPublicFields` are now unused.

Src/FluentAssertions/Common/TypeExtensions.cs
jnyrup: I'm wondering a bit about what that means to protected members.

We ignore `protected` members
unless it's `protected internal`
unless it `private protected`?

dennisdoomen:
> I'm wondering a bit about what that means to protected members.
> We ignore protected members
> unless it's protected internal
> unless it private protected?

Why do you care about `protected` members? They are not `public`, nor `internal`.

jnyrup:
I do not care about `protected`, I just mixed them up.
It made sense to me after seeing it visualized as a Venn Diagram like https://stackoverflow.com/a/22958035/1087627

jnyrup (approved):

PR #1576, opened 2021-05-24
Split Type assertions in separate files

eNeRGy164:
As discussed in #1551

* Put [Not] in filename to make clear both version are in the same file
* Kept regions to make group positive and negative (not) methods
* Kept generic variants with the non-generic methods
* Except for the main file, added `<content>` instead of `<summary>`  to make sure StyleCop rules were passing, but any documentation (unlikely) would come from a single source.
* Discovered the `NotHavePropertyOfT` specs are missing

## IMPORTANT 


dennisdoomen:
I like the way the files are split, but I'm not sure I like the `[Not]` in the name. [Option 1](https://github.com/fluentassertions/fluentassertions/discussions/1551#discussioncomment-720622) did not have that specific element. What do you think @jnyrup? Do you think it adds value?

eNeRGy164:
I can change it of course.

When organizing, I was thinking how logical it is for anybody looking at the code independent of IDE features that both the positive and negative version are in the same file.

I was imagining several ways (that are allowed by the filesystem).
I settled on `[Not]` as that is also the notation used in the documentation/release notes.

So, the file containing: `HaveImplicitConversionOperator`, `HaveImplicitConversionOperator<T>`, `NotHaveImplicitConversionOperator` and `NotHaveImplicitConversionOperator<T>` is called...

```plain
TypeAssertionSpecs.HaveImplicitConversionOperator.cs 
TypeAssertionSpecs.[Not]HaveImplicitConversionOperator.cs 
TypeAssertionSpecs.Not-HaveImplicitConversionOperator.cs 
TypeAssertionSpecs.Not.HaveImplicitConversionOperator.cs 
TypeAssertionSpecs.Not_HaveImplicitConversionOperator.cs 
```

dennisdoomen:
Yes, I understand, but I don't quite like it myself.

jnyrup:
I would prefer if we just left out the `[Not]`. 
Other than that I like the changes here 👍

eNeRGy164:
I'll change it. Downside is that the **Be** comes before **cs**
image

jnyrup (approved):

dennisdoomen (approved):

PR #1581, opened 2021-05-25
Rename WhichValue to WhoseValue

gliljas:
I realize this has sort of been discussed in #577 and #1126, but without any real conclusion, which I think is unfortunate. However, in the case of WhichValue (on a dictionary), there can never be any ambiguity and "Which" will always be the wrong word. It could be argued that it should rather be something like "WhoseCorrespondingValue", since the Value does not belong to the Key, but that's a bit long.


dennisdoomen (approved):

jnyrup:
Do we have some documentation examples that also needs to be updated?

jnyrup (approved):

dennisdoomen:
> Do we have some documentation examples that also needs to be updated?

Did a quick scan, but apparently we never documented this feature.

PR #1587, opened 2021-05-28
A succeeding condition should not be evaluated if the first one failed

### LINKED ISSUE #1568: ForCondition continues when wrapped in AssertionScope
### Description

The documentation of `GivenSelector<T>.ForCondition` says
> The condition will not be evaluated if the prior assertion failed [...]

When wrapped in an `AssertionScope` that is no longer the case.

From `AssertionScope`
> Starts an unnamed scope within which multiple assertions can be executed 
> and which will not throw until the scope is disposed.

Maybe it's just an unfortunate reality, but it prevents writing guards to prevent e.g. NullReferenceExceptions in a fluent way.

Looks related to #1325

```csharp
string subject = null;
Execute.Assertion
    .Given(() => subject)
    .ForCondition(s => s is not null)
    .FailWith("but is was null")
    .Then
    .ForCondition(s => s.GetHashCode() == 42) // this predicate is invoked
    .FailWith("it should be 42");
```

### Expected behavior:

The second predicate should not be invoked

### Actual behavior:

The predicate was invoked.

### Versions

* Fluent Assertions `develop`@ e9c847c39d280350d21b606b18302fe2e8deac0b

dennisdoomen:
In a chained assertion API call, a second call to ForCondition should not even evaluate its lambda when the previous assertion failed.

Fixes #1568

### DIFF
--- a/Src/FluentAssertions/Execution/GivenSelector.cs
+++ b/Src/FluentAssertions/Execution/GivenSelector.cs
@@ -11,8 +11,9 @@ namespace FluentAssertions.Execution
     public class GivenSelector<T>
     {
         private readonly AssertionScope predecessor;
-        private readonly bool continueAsserting;
         private readonly T subject;
+
+        private bool continueAsserting;
 
         internal GivenSelector(Func<T> selector, AssertionScope predecessor, bool continueAsserting)
         {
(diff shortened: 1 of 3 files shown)


jnyrup (approved):

eNeRGy164:
So now we can change all those `if` statements again? 🤣

jnyrup:
> So now we can remove all those `if` statements again?

Only those using a `GivenSelector`.

For the majority, in order to remove the `if`s, we would have to first rewrite them to use a `GivenSelector`.

PR #1591, opened 2021-05-30
Fix build warning

eNeRGy164:
During a build, the Benchmark project generates warnings about `System.Data.DataSetExtensions`.
Adding a condition should fix this warning.


jnyrup (approved):
Thanks. 
Now I can remove that from my todo

PR #1603, opened 2021-06-09
A collection of similarly typed key-value pairs should be equivalent to a dictionary

### LINKED ISSUE #1598: A collection of key value pairs is no longer equivalent to a dictionary
### Description

[Description of the issue]

```csharp
[Fact]
public void When_a_collection_of_key_value_pairs_is_equivalent_to_the_dictionary_it_should_succeed()
{
    // Arrange
    var collection = new List<KeyValuePair<string, int>> { new KeyValuePair<string, int>("hi", 1) };

    // Act / Assert
    Action act = () => collection.Should().BeEquivalentTo(new Dictionary<string, int>()
    {
        { "hi", 2 }
    });

    act.Should().Throw<XunitException>().WithMessage("Expected item[0].Value to be 2, but found 1.*");
}
```

### Expected behavior:

The test passes

### Actual behavior:
```
Expected collection (of type Dictionary<string, int>) to be a System.Collections.Generic.IDictionary`2[System.String,System.Int32],
 but found a System.Collections.Generic.List`1[System.Collections.Generic.KeyValuePair`2[System.String,System.Int32]].
```
### Versions

6.0 Beta 1

dennisdoomen:
Because of some accidental design choices, in previous releases, when the expectation is a dictionary, it was picked up by the `GenericEnumerableEquivalencyStep`. This resulted in an `IDictionary<TKey, TValue>` being treated as a collection of `KeyValuePair<TKey, TValue>`. In 6.0, some of the semantics of the equivalency steps changed in such a way that a dictionary is now picked up by the `GenericEquivalencyStep`. This is expected, but had the side-effect that when a dictionary is expected, it really expects a dictionary. 

This PR changes `GenericEnumerableEquivalencyStep` so it tries to convert a collection of key-value pairs back to a dictionary and then does the equivalency check.

Fixes #1598

### DIFF
--- a/Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
@@ -11,8 +11,8 @@ namespace FluentAssertions.Equivalency.Steps
     {
 #pragma warning disable SA1110 // Allow opening parenthesis on new line to reduce line length
         private static readonly MethodInfo AssertSameLengthMethod =
-            new Func<IDictionary<object, object>, IDictionary<object, object>, bool>
-                (AssertSameLength).GetMethodInfo().GetGenericMethodDefinition();
+            new Func<IDictionary<object, object>, IDictionary<object, object>, bool>(AssertSameLength).GetMethodInfo()
+                .GetGenericMethodDefinition();
 
         private static readonly MethodInfo AssertDictionaryEquivalenceMethod =
             new Action<EquivalencyValidationContext, IEquivalencyValidator, IEquivalencyAssertionOptions,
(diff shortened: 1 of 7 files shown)


Tests/FluentAssertions.Specs/Equivalency/DictionaryEquivalencySpecs.cs
```diff
@@ -448,6 +448,21 @@ public void When_a_dictionary_property_is_detected_it_should_ignore_the_order_of
...
+        public void When_a_collection_of_key_value_pairs_is_equivalent_to_the_dictionary_it_should_succeed()
+        {
+            // Arrange
+            var collection = new List<KeyValuePair<string, int>> { new("hi", 1) };
+
+            // Act / Assert
+            Action act = () => collection.Should().BeEquivalentTo(new Dictionary<string, int>()
+            {
+                { "hi", 2 }
+            }, x => x.WithTracing());
```
jnyrup, line 461: is `, x => x.WithTracing()` part of the test?

Tests/FluentAssertions.Specs/Equivalency/DictionaryEquivalencySpecs.cs
```diff
@@ -647,7 +662,7 @@ public void When_the_subjects_key_type_is_not_compatible_with_the_expected_key_t
 
             // Assert
             act.Should().Throw<XunitException>()
-                .WithMessage("*expectation*keys*Int32*compatible types*IDictionary`2[System.String,System.String]*");
+                .WithMessage("Expected the expectation (of type Dictionary<int, string>) to be a dictionary or collection of key-value pairs that is keyed to type System.Int32*");
```
jnyrup, line 665: `expectation` and `actual` seems flipped in the `act`.

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
```diff
@@ -24,40 +24,33 @@ public class GenericDictionaryEquivalencyStep : IEquivalencyStep
             IEquivalencyValidator nestedValidator)
         {
             Type expectationType = comparands.GetExpectedType(context.Options);
-
-            if (comparands.Expectation is null || !GetIDictionaryInterfaces(expectationType).Any())
+            bool isDictionary = DictionaryInterfaceInfo.TryGetFrom(expectationType, "expectation", out var expectedDictionary);
+            if (comparands.Expectation != null && isDictionary)
```
jnyrup, line 28: Consider only calling `DictionaryInterfaceInfo.TryGetFrom` if `comparands.Expectation != null`

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
```diff
@@ -74,95 +67,46 @@ private static bool AssertExpectationIsNotNull(object subject, object expectatio
...
-
-            AssertionScope.Current.FailWith(
-                "{context:Expectation} implements multiple dictionary types.  "
-                + $"It is not known which type should be use for equivalence.{Environment.NewLine}"
-                + $"The following IDictionary interfaces are implemented: {string.Join(", ", (IEnumerable<Type>)interfaces)}");
-
-            return false;
-        }
-
-        private static bool AssertIsCompatiblyTypedDictionary(Type expectedType, object subject)
```
jnyrup, line 94: There's a now outdated comment on line 118 mentioning `AssertIsCompatiblyTypedDictionary`

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
```diff
@@ -74,95 +67,46 @@ private static bool AssertExpectationIsNotNull(object subject, object expectatio
...
-            if (suitableDictionaryInterfaces.Length > 1)
+            if (!isDictionary)
             {
-                // SMELL: Code could be written to handle this better, but is it really worth the effort?
-                AssertionScope.Current.FailWith(
-                    "The subject implements multiple IDictionary interfaces. ");
+                if (expectedDictionary.TryConvertFrom(comparands.Subject, out var convertedSubject))
+                {
+                    comparands.Subject = convertedSubject;
+                    DictionaryInterfaceInfo.TryGetFrom(comparands.Subject.GetType(), "subject", out actualDictionary);
```
jnyrup, line 81: double check: is the return value of `TryGetFrom` unimportant?

jnyrup (commented):
Can you give a high level description of what's changed.
I have a bit of a hard time separating the restructuring from the change in logic.

Have you tried my suggestion from
https://github.com/fluentassertions/fluentassertions/issues/1598#issuecomment-858369350 ?

dennisdoomen:
> Can you give a high level description of what's changed.

Sorry. Forgot about that one. Will do.

> Have you tried my suggestion from #1598 (comment) ?

Not really. The old implementation was kind of an accidental side-effect. Since the expectation is a dictionary, I decided to implement this where it belongs. Now it's completely symmetric. 

> I have a bit of a hard time separating the restructuring from the change in logic.

I understand, but I don't think I could have avoided that. The idea of extracting some of the logic happened while I was trying to implement the algorithm.

PR #1606, opened 2021-06-10
Make xml documentation more consistent for AssertionScope and friends

### LINKED ISSUE #1605: Inconsistent description on AssertionScope methods
### Description

Although the methods share name and signature, the XML Documentation on `AssertionScope`, `ContinuedAssertionScope` and `GivenSelector<T>` is vastly different.
As a consumer of FA I would expect the text to be consistent.

Look at the documentation of every `ForCondition` or `FailWith` in the following example:
(for example, hover over them in Visual Studio)

```csharp
Execute.Assertion
    .ForCondition(true)
    .FailWith("First assertion")
    .Then
    .ForCondition(true)
    .FailWith("Second assertion")
    .Then
    .Given(() => true)
    .ForCondition(subject => true)
    .FailWith("Third assertion");
```

### Expected behavior:

Consistent documentation.
Some differences might be valid, but in general the text should be the same.

### Actual behavior:

Every method has a different text

**ForCondition**
1. image
1. image
1. image

**FailWith**
1. image
1. image
1. image

I will create a PR to improve this.

eNeRGy164:
* Used `<inhertdoc/>` where possible
* Added paragraphs to make long text more readable
* Added emphasis for certain words
* Shortened references where possible

- [x] Release notes

Close #1605

### DIFF
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -11,8 +11,8 @@ namespace FluentAssertions.Execution
     /// Represents an implicit or explicit scope within which multiple assertions can be collected.
     /// </summary>
     /// <remarks>
-    /// This class is supposed to have a very short life time and is not safe to be used in assertion that cross thread-boundaries such as when
-    /// using <c>async</c> or <c>await</c>.
+    /// This class is supposed to have a very short life time and is not safe to be used in assertion that cross thread-boundaries
+    /// such as when using <c>async</c> or <c>await</c>.
     /// </remarks>
     public sealed class AssertionScope : IAssertionScope
     {
@@ -84,7 +84,8 @@ namespace FluentAssertions.Execution
         }
 
         /// <summary>
-        /// Starts a named scope within which multiple assertions can be executed and which will not throw until the scope is disposed.
+        /// Starts a named scope within which multiple assertions can be executed
+        /// and which will not throw until the scope is disposed.
         /// </summary>
         public AssertionScope(string context)
             : this()
@@ -96,7 +97,8 @@ namespace FluentAssertions.Execution
         }
 
         /// <summary>
-        /// Starts a named scope within which multiple assertions can be executed and which will not throw until the scope is disposed.
+        /// Starts a named scope within which multiple assertions can be executed
+        /// and which will not throw until the scope is disposed.
         /// </summary>
         public AssertionScope(Lazy<string> context)
             : this()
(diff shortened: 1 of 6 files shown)


Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -122,12 +124,7 @@ public static AssertionScope Current
             private set => SetCurrentAssertionScope(value);
         }
 
-        /// <summary>
-        /// Forces the formatters that support it to add the necessary line breaks.
-        /// </summary>
-        /// <remarks>
-        /// This is just shorthand for modifying the <see cref="FormattingOptions"/> property.
-        /// </remarks>
+        /// <inheritdoc cref="IAssertionScope.UsingLineBreaks"/>
```
dennisdoomen, line 127: I wonder if we don't use `inheritdoc`, would you still get the right Intellisense if you use FA as a binary. Because with Rider, while working with the FA code base, it just shows me the docs from the interface.

eNeRGy164: I think that is a Rider trick then, as `UsingLineBreaks` is not inheriting from the interface itself. It's only references by an explicit implementation of the interface (bottom of the file)

In Visual Studio it doesn't show up without the `inheritdoc`.

dennisdoomen (approved):

Src/FluentAssertions/Execution/AssertionScope.cs
eNeRGy164: For methods that inherit from the interface, it can be left out as it will inherit implicitly, but it results in a SA message about public elements that should be documented.
So, other then making the behavior and intent more explicit, it also removes some SA messages.

docs/_pages/releases.md
```diff
@@ -16,6 +16,7 @@ Changes since 6.0.0 Beta 1
 
 **Fixes**
 * Added parameter checking for `Be(string)` for GuidAssertions - [#1597](https://github.com/fluentassertions/fluentassertions/pull/1597).
+* Improved consistency of XML documentation on `AssertionScope`, `ContinuedAssertionScope`, and `GivenSelector<T>` methods. - [#1605](https://github.com/fluentassertions/fluentassertions/pull/1605).
```
jnyrup, line 19:
```suggestion
* Improved consistency of XML documentation on `AssertionScope`, `ContinuedAssertionScope`, and `GivenSelector<T>` methods. - [#1606](https://github.com/fluentassertions/fluentassertions/pull/1606).
```

jnyrup (approved):

PR #1609, opened 2021-06-10
Add new BeExactly assertions for DateTimeOffset

gliljas:
Not quite ready yet. More a basis for discussion.

The value formatting was one issue. I don't think the offset should be hidden, even if it's zero. It's no less significant than a positive or negative offset.

The assertion could perhaps be made with a Then step and a "...but the offsets differ" message.

### DIFF
--- a/Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
+++ b/Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
@@ -21,6 +21,7 @@ namespace FluentAssertions.Formatting
         public void Format(object value, FormattedObjectGraph formattedGraph, FormattingContext context, FormatChild formatChild)
         {
             DateTimeOffset dateTimeOffset;
+            bool significantOffset = false;
 
             if (value is DateTime dateTime)
             {
@@ -29,6 +30,7 @@ namespace FluentAssertions.Formatting
             else
             {
                 dateTimeOffset = (DateTimeOffset)value;
+                significantOffset = true;
             }
 
             formattedGraph.AddFragment("<");
@@ -70,11 +72,18 @@ namespace FluentAssertions.Formatting
                 formattedGraph.AddFragment(" +");
                 formatChild("offset", dateTimeOffset.Offset, formattedGraph);
             }
-
-            if (dateTimeOffset.Offset < TimeSpan.Zero)
+            else if (dateTimeOffset.Offset < TimeSpan.Zero)
             {
                 formattedGraph.AddFragment(" ");
                 formatChild("offset", dateTimeOffset.Offset, formattedGraph);
+            }
+            else if (significantOffset && (hasDate || hasTime))
+            {
+                formattedGraph.AddFragment(" +0h");
+            }
+            else
+            {
+                // No offset added, since it was deemed unnecessary
             }
 
             if (!hasDate && !hasTime)
--- a/Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
+++ b/Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
@@ -45,7 +45,7 @@ namespace FluentAssertions.Primitives
         public DateTimeOffset? Subject { get; }
 
         /// <summary>
-        /// Asserts that the current <see cref="DateTimeOffset"/> is exactly equal to the <paramref name="expected"/> value.
+        /// Asserts that the current <see cref="DateTimeOffset"/> represents the same point in time as the <paramref name="expected"/> value.
         /// </summary>
         /// <param name="expected">The expected value</param>
         /// <param name="because">
@@ -59,16 +59,21 @@ namespace FluentAssertions.Primitives
             params object[] becauseArgs)
         {
             Execute.Assertion
-                .ForCondition(Subject == expected)
                 .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context:the date and time} to be {0}{reason}, but it was {1}.",
-                    expected, Subject);
+                .WithExpectation("Expected {context:the date and time} to represent the same point in time as {0}{reason}, ", expected)
+                .ForCondition(Subject.HasValue)
+                .FailWith("but found a <null> DateTimeOffset.")
+                .Then
+                .ForCondition(Subject == expected)
+                .FailWith("but {0} does not.", Subject)
+                .Then
+                .ClearExpectation();
 
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
         /// <summary>
-        /// Asserts that the current <see cref="DateTimeOffset"/> is exactly equal to the <paramref name="expected"/> value.
+        /// Asserts that the current <see cref="DateTimeOffset"/> represents the same point in time as the <paramref name="expected"/> value.
         /// </summary>
         /// <param name="expected">The expected value</param>
         /// <param name="because">
@@ -81,17 +86,32 @@ namespace FluentAssertions.Primitives
         public AndConstraint<TAssertions> Be(DateTimeOffset? expected, string because = "",
             params object[] becauseArgs)
         {
-            Execute.Assertion
-                .ForCondition(Subject == expected)
-                .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context:the date and time} to be {0}{reason}, but it was {1}.",
-                    expected, Subject);
+            if (!expected.HasValue)
+            {
+                Execute.Assertion
+                   .BecauseOf(because, becauseArgs)
+                   .ForCondition(!Subject.HasValue)
+                   .FailWith("Expected {context:the date and time} to be <null>{reason}, but it was {0}.", Subject);
+            }
+            else
+            {
+                Execute.Assertion
+                    .BecauseOf(because, becauseArgs)
+                    .WithExpectation("Expected {context:the date and time} to represent the same point in time as {0}{reason}, ", expected)
+                    .ForCondition(Subject.HasValue)
+                    .FailWith("but found a <null> DateTimeOffset.")
+                    .Then
+                    .ForCondition(Subject == expected)
+                    .FailWith("but {0} does not.", Subject)
+                    .Then
+                    .ClearExpectation();
+            }
 
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
         /// <summary>
-        /// Asserts that the current <see cref="DateTimeOffset"/> is not equal to the <paramref name="unexpected"/> value.
+        /// Asserts that the current <see cref="DateTimeOffset"/> does not represent the same point in time as the <paramref name="unexpected"/> value.
         /// </summary>
         /// <param name="unexpected">The unexpected value</param>
         /// <param name="because">
@@ -107,13 +127,13 @@ namespace FluentAssertions.Primitives
             Execute.Assertion
                 .ForCondition(Subject != unexpected)
                 .BecauseOf(because, becauseArgs)
-                .FailWith("Did not expect {context:the date and time} to be {0}{reason}, but it was.", unexpected);
+                .FailWith("Did not expect {context:the date and time} to represent the same point in time as {0}{reason}, but it did.", unexpected);
 
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
         /// <summary>
-        /// Asserts that the current <see cref="DateTimeOffset"/> is not equal to the <paramref name="unexpected"/> value.
+        /// Asserts that the current <see cref="DateTimeOffset"/> does not represent the same point in time as the <paramref name="unexpected"/> value.
         /// </summary>
         /// <param name="unexpected">The unexpected value</param>
         /// <param name="because">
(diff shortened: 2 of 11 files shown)


Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
```diff
@@ -21,6 +21,7 @@ public bool CanHandle(object value)
         public void Format(object value, FormattedObjectGraph formattedGraph, FormattingContext context, FormatChild formatChild)
         {
             DateTimeOffset dateTimeOffset;
+            bool significantOffset = false;
```
dennisdoomen, line 24: Can you try to move those changes in a separate commit so we can easily see why this change was needed?

Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
```diff
@@ -70,12 +72,15 @@ public void Format(object value, FormattedObjectGraph formattedGraph, Formatting
...
-            if (dateTimeOffset.Offset < TimeSpan.Zero)
+            else if (dateTimeOffset.Offset < TimeSpan.Zero)
             {
                 formattedGraph.AddFragment(" ");
                 formatChild("offset", dateTimeOffset.Offset, formattedGraph);
             }
+            else if (significantOffset && (hasDate || hasTime))
+            {
+                formattedGraph.AddFragment(" UTC");
+            }
```
dennisdoomen, line 83: 🤔 What happens if the last `else if` is not happening?

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
```diff
@@ -68,7 +68,7 @@ public DateTimeOffsetAssertions(DateTimeOffset? value)
         }
 
         /// <summary>
-        /// Asserts that the current <see cref="DateTimeOffset"/> is exactly equal to the <paramref name="expected"/> value.
+        /// Asserts that the current <see cref="DateTimeOffset"/> represents the same point in time as the <paramref name="expected"/> value.
```
dennisdoomen, line 71: 🤔 Shouldn't we also update the failure message to cover this important distinction?

Tests/FluentAssertions.Specs/Primitives/DateTimeOffsetAssertionSpecs.cs
```diff
@@ -364,6 +364,189 @@ public void
 
         #endregion
 
+        #region (Not) BeExactly
+
+        [Fact]
+        public void Should_succeed_when_asserting_datetimeoffset_value_is_exactly_equal_to_the_same_value()
```
dennisdoomen, line 370: 🔧 No need to mention `datetimeoffset` in there since it's clear from the context. Also, we prefer to keep literal names of code elements out of the test name.

Tests/FluentAssertions.Specs/Primitives/DateTimeOffsetAssertionSpecs.cs
```diff
@@ -364,6 +364,189 @@ public void
...
+        {
+            // Arrange
+            DateTimeOffset dateTime = new DateTime(2016, 06, 04);
+            DateTimeOffset sameDateTime = new DateTime(2016, 06, 04);
+
+            // Act
+            Action act = () => dateTime.Should().BeExactly(sameDateTime);
+
+            // Assert
+            act.Should().NotThrow();
```
dennisdoomen, line 380: 🔧 If the assertion is not expected to throw anything, you don't need to explicitly use `NotThrow` (anymore).

Tests/FluentAssertions.Specs/Primitives/DateTimeOffsetAssertionSpecs.cs
gliljas: Gladly. I just copied the Be/NotBe tests, which has the misleading "numeric" name, and chanhed it. Will fix.

Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
gliljas: That would leave us in the original path. No offset appended/prepended.

Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
gliljas: The formatter changes? I'll do anything you ask for, as soon as I understand what it is :)

My rationale for the formatting change was:

1. Without the offset mentioned, even if it's TimeSpan.Zero, the assertion messages can become cryptic. "Exptected 2021-06-01 +1h, but it was 2021-06-01"
2. I'm just guessing, but I think it's actually more common to deal with DateTimeOffsets which have a non-zero offset, so the change shouldn't be that big of a deal.
3. The offset is skipped for plain DateTime. I thought that would make sense.

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
gliljas: Done. However, the messages became a bit wordy. Perhaps too much so.

Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
dennisdoomen: > My rationale for the formatting change was:

@jnyrup what do you think? Makes sense to me.

> The formatter changes? I'll do anything you ask for, as soon as I understand what it is :)

To split your changes into two commits. One with the formatter changes (with the rational in the title and description), and one with the new `DateTimeOffset` assertions. See also https://www.continuousimprover.com/2020/03/keep-source-control-history-clean.html

Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
dennisdoomen: I would add an `else` with a comment explaining that. Avoids unnecessary frowns later on.

Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
gliljas: Also, using "UTC" as the offset was just a choice that seemed OK at the time. UTC is not an offset, it's a time zone. There are other time zones which have a zero offset, so something like "Z", "+0h" or even "±0h" could make more sense.

dennisdoomen: I think `+0h` would be clearer.

gliljas: Agreed. Done. I will fix the commits according to you liking if and when we conclude the matter. The documentation will need some additions too.

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
dennisdoomen: Maybe, but it'll make it very clear why the test failed.

Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
```diff
@@ -70,12 +72,19 @@ public void Format(object value, FormattedObjectGraph formattedGraph, Formatting
                 formattedGraph.AddFragment(" +");
                 formatChild("offset", dateTimeOffset.Offset, formattedGraph);
             }
-
-            if (dateTimeOffset.Offset < TimeSpan.Zero)
+            else if (dateTimeOffset.Offset < TimeSpan.Zero)
             {
                 formattedGraph.AddFragment(" ");
```
jnyrup, line 77: Shouldn't this have a minus sign?
```suggestion
                formattedGraph.AddFragment(" -");
```

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
```diff
@@ -59,7 +59,117 @@ public DateTimeOffsetAssertions(DateTimeOffset? value)
             params object[] becauseArgs)
         {
             Execute.Assertion
+                .BecauseOf(because, becauseArgs)
+                .WithExpectation("Expected {context:the date and time} to represent the same point in time as {0}{reason}, ", expected)
+                .ForCondition(Subject.HasValue)
+                .FailWith("but found a <null> DateTimeOffset.")
+                .Then
                 .ForCondition(Subject == expected)
+                .FailWith("but its value {0} doesn't.", Subject)
```
jnyrup, line 68:
```suggestion
                .FailWith("but {0} does not.", Subject)
```

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
```diff
@@ -101,11 +211,11 @@ public DateTimeOffsetAssertions(DateTimeOffset? value)
         /// <param name="becauseArgs">
         /// Zero or more objects to format using the placeholders in <paramref name="because" />.
         /// </param>
-        public AndConstraint<TAssertions> NotBe(DateTimeOffset unexpected, string because = "",
+        public AndConstraint<TAssertions> NotBeExactly(DateTimeOffset unexpected, string because = "",
```
jnyrup, line 214: Should the summary be updated?

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
```diff
@@ -78,11 +188,11 @@ public DateTimeOffsetAssertions(DateTimeOffset? value)
         /// <param name="becauseArgs">
         /// Zero or more objects to format using the placeholders in <paramref name="because" />.
         /// </param>
-        public AndConstraint<TAssertions> Be(DateTimeOffset? expected, string because = "",
+        public AndConstraint<TAssertions> BeExactly(DateTimeOffset? expected, string because = "",
             params object[] becauseArgs)
         {
             Execute.Assertion
-                .ForCondition(Subject == expected)
+                .ForCondition((Subject == null && expected == null) || (Subject != null && expected != null && Subject.Value.EqualsExact(expected.Value)))
```
jnyrup, line 195: We strive to keep lines at atmost 130 characters wide.
```suggestion
                .ForCondition((Subject == null && expected == null) 
                    || (Subject != null && expected != null && Subject.Value.EqualsExact(expected.Value)))
```

Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
gliljas: I was confused by that one too, but indeed, it's added by the formatter.

Yes, there's test coverage.

dennisdoomen (approved):

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
```diff
@@ -80,18 +85,55 @@ public DateTimeOffsetAssertions(DateTimeOffset? value)
...
+            else
+            {
+                Execute.Assertion
+                    .BecauseOf(because, becauseArgs)
+                    .WithExpectation("Expected {context:the date and time} to represent the same point in time as {0}{reason}, ", expected)
+                    .ForCondition(Subject.HasValue)
+                    .FailWith("but found a <null> DateTimeOffset.", expected)
+                    .Then
+                    .ForCondition(Subject == expected)
+                    .FailWith("but {1} does not.", expected, Subject)
```
jnyrup, line 105:
```suggestion
                    .FailWith("but {0} does not.", Subject)
```

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
```diff
@@ -101,12 +143,102 @@ public DateTimeOffsetAssertions(DateTimeOffset? value)
...
+        public AndConstraint<TAssertions> BeExactly(DateTimeOffset expected, string because = "",
+            params object[] becauseArgs)
+        {
+            Execute.Assertion
+                .BecauseOf(because, becauseArgs)
+                .WithExpectation("Expected {context:the date and time} to be exactly {0}{reason}, ", expected)
+                .ForCondition(Subject.HasValue)
+                .FailWith("but found a <null> DateTimeOffset.")
+                .Then
+                .ForCondition(Subject.Value.EqualsExact(expected) == true)
```
jnyrup, line 177: I haven't checked the code out, but I would believe `== true` to be superfluous.
```suggestion
                .ForCondition(Subject.Value.EqualsExact(expected))
```

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
```diff
@@ -101,12 +143,102 @@ public DateTimeOffsetAssertions(DateTimeOffset? value)
...
+            else
+            {
+                Execute.Assertion
+                    .BecauseOf(because, becauseArgs)
+                    .WithExpectation("Expected {context:the date and time} to be exactly {0}{reason}, ", expected)
+                    .ForCondition(Subject.HasValue)
+                    .FailWith("but found a <null> DateTimeOffset.", expected)
+                    .Then
+                    .ForCondition(Subject.Value.EqualsExact(expected.Value))
+                    .FailWith("but it was {0}.", expected, Subject)
```
jnyrup, line 216:
```suggestion
                    .FailWith("but it was {0}.", Subject)
```

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
```diff
@@ -101,12 +143,102 @@ public DateTimeOffsetAssertions(DateTimeOffset? value)
...
+                   .ForCondition(!Subject.HasValue)
+                   .FailWith("Expected {context:the date and time} to be <null>{reason}, but it was {0}.", Subject);
+            }
+            else
+            {
+                Execute.Assertion
+                    .BecauseOf(because, becauseArgs)
+                    .WithExpectation("Expected {context:the date and time} to be exactly {0}{reason}, ", expected)
+                    .ForCondition(Subject.HasValue)
+                    .FailWith("but found a <null> DateTimeOffset.", expected)
```
jnyrup, line 213:
```suggestion
                    .FailWith("but found a <null> DateTimeOffset.")
```

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
```diff
@@ -80,18 +85,55 @@ public DateTimeOffsetAssertions(DateTimeOffset? value)
...
+                   .ForCondition(!Subject.HasValue)
+                   .FailWith("Expected {context:the date and time} to be <null>{reason}, but it was {0}.", Subject);
+            }
+            else
+            {
+                Execute.Assertion
+                    .BecauseOf(because, becauseArgs)
+                    .WithExpectation("Expected {context:the date and time} to represent the same point in time as {0}{reason}, ", expected)
+                    .ForCondition(Subject.HasValue)
+                    .FailWith("but found a <null> DateTimeOffset.", expected)
```
jnyrup, line 102:
```suggestion
                    .FailWith("but found a <null> DateTimeOffset.")
```

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
```diff
@@ -101,12 +143,102 @@ public DateTimeOffsetAssertions(DateTimeOffset? value)
...
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> NotBeExactly(DateTimeOffset unexpected, string because = "",
+            params object[] becauseArgs)
+        {
+            Execute.Assertion
+                .ForCondition(Subject?.EqualsExact(unexpected) != true)
+                .BecauseOf(because, becauseArgs)
                 .FailWith("Did not expect {context:the date and time} to be {0}{reason}, but it was.", unexpected);
```
jnyrup, line 242:
```suggestion
                .FailWith("Did not expect {context:the date and time} to be exactly {0}{reason}, but it was.", unexpected);
```

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
```diff
@@ -123,11 +255,11 @@ public DateTimeOffsetAssertions(DateTimeOffset? value)
...
         /// </param>
-        public AndConstraint<TAssertions> NotBe(DateTimeOffset? unexpected, string because = "",
+        public AndConstraint<TAssertions> NotBeExactly(DateTimeOffset? unexpected, string because = "",
             params object[] becauseArgs)
         {
             Execute.Assertion
-                .ForCondition(Subject != unexpected)
+                .ForCondition(!((Subject == null && unexpected == null) || (Subject != null && unexpected != null && Subject.Value.EqualsExact(unexpected.Value))))
                 .BecauseOf(because, becauseArgs)
                 .FailWith("Did not expect {context:the date and time} to be {0}{reason}, but it was.", unexpected);
```
jnyrup, line 264:
```suggestion
                .FailWith("Did not expect {context:the date and time} to be exactly {0}{reason}, but it was.", unexpected);
```

jnyrup (changes requested):
Aside from the failing tests, two other step are necessary to finish up this PR
* Run `AcceptChanges.ps1/sh` to update the accepted public methods.
* Add a line to the release notes.

Both steps are explained in https://github.com/fluentassertions/fluentassertions/blob/release-6.0/CONTRIBUTING.md, but ping me if anything is unclear or causing trouble.

jnyrup (approved):

PR #1619, opened 2021-06-28
Homogenize predicate formatting

gliljas:
Many predicate assertions uses `predicate.Body` to format their messages, thereby circumventing the formatting offered by `PredicateLambdaExpressionValueFormatter`.  More specifically, it's the part which resolves constant values which is sorely missing, e.g when using `collection.Should().Contain(x => x.Name == expectedName && x.Value == expectedValue)`.

This could be fixed by adding a new formatter or modifying the existing, but I think it makes more sense to provide the full lambda to the formatter.

Possible additions would be:

1. Include the lambda parameter part in the output. 
2. ...or removing it when possible, outputting `Name == "Expected"` instead of `a.Name == "Expected"`
3. ...and/or renaming the lambda parameter, e.g so that its always named "item", when used in a collection assertion.

There's one test, `When_single_item_contains_string_interpolation_it_should_format_brackets_properly`, which asserted a specific "Format" output. It can't do that now, but I'm not sure it makes much sense to keep the formatting call as is. This was discussed in #1404. Either way, I think it makes sense to centralize the predicate formatting and deal with any special cases in the formatter.

### DIFF
--- a/Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionsSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionsSpecs.cs
@@ -318,14 +318,15 @@ namespace FluentAssertions.Specs.Collections
         {
             // Arrange
             IEnumerable<int> collection = Enumerable.Empty<int>();
-            Expression<Func<int, bool>> expression = item => item == 2;
+            var targetValue = 2;
+            Expression<Func<int, bool>> expression = item => item == targetValue;
 
             // Act
             Action act = () => collection.Should().ContainSingle(expression);
 
             // Assert
             string expectedMessage =
-                $"Expected collection to contain a single item matching {expression.Body}, but the collection is empty.";
+                $"Expected collection to contain a single item matching (item == 2), but the collection is empty.";
 
             act.Should().Throw<XunitException>().WithMessage(expectedMessage);
         }
@@ -335,14 +336,15 @@ namespace FluentAssertions.Specs.Collections
         {
             // Arrange
             const IEnumerable<int> collection = null;
-            Expression<Func<int, bool>> expression = item => item == 2;
+            var targetValue = 2;
+            Expression<Func<int, bool>> expression = item => item == targetValue;
 
             // Act
             Action act = () => collection.Should().ContainSingle(expression);
 
             // Assert
             string expectedMessage =
-                $"Expected collection to contain a single item matching {expression.Body}, but found <null>.";
+                $"Expected collection to contain a single item matching (item == 2), but found <null>.";
 
             act.Should().Throw<XunitException>().WithMessage(expectedMessage);
         }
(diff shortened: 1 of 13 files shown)


dennisdoomen (approved):
Thanks for this contribution. Would you mind updating the `release.md` as well?

jnyrup (approved):
quickly grep'ing for `.Body` there seems to be a few more cases in `GenericCollectionAssertions.cs` and a single one in `ExceptionAssertion.cs`.
Were those intentionally skipped?

gliljas:
> Were those intentionally skipped?

Not the one in ExceptionsAssertion.cs. The other ones were not predicates, in that they did not return bool, so they wouldn't match the formatter anyway. 

I've fixed that now. Note that the result is that parentheses now encloses the formatted output in these cases.

docs/_pages/releases.md
```diff
@@ -11,6 +11,7 @@ sidebar:
 Changes since 6.0.0 Alpha 2
 
 **What's New**
+* Homegenized assertion message formatting of predicate expressions - [#1619](https://github.com/fluentassertions/fluentassertions/pull/1619).
```
jnyrup, line 14:
```suggestion
* Homogenized assertion message formatting of predicate expressions - [#1619](https://github.com/fluentassertions/fluentassertions/pull/1619).
```

Tests/FluentAssertions.Specs/Collections/GenericCollectionAssertionsSpecs.cs
```diff
@@ -318,14 +318,15 @@ public void When_asserting_an_empty_collection_contains_a_single_item_matching_a
         {
             // Arrange
             IEnumerable<int> collection = Enumerable.Empty<int>();
-            Expression<Func<int, bool>> expression = item => item == 2;
+            var targetValue = 2;
```
jnyrup, line 321: Why extracting `2` to a variable?

gliljas: Basically, to assert that the lambda output formatting is invoked properly. Previously, having this syntax would have rendered the message "Expected collection to contain a single item matching (item == value(FluentAssertions.Specs.Collections.GenericCollectionAssertionsSpecs+<>c__DisplayClass21_0).targetValue)"

jnyrup: It seems that `When_first_level_properties_are_tested_for_equality_against_constant_expressions_then_output_should_contain_values_of_constant_expressions` is responsible for asserting that captured constants in expressions are properly reduced, and not a concern these tests need to exercise.

But I do like that you're replacing the dynamic part of `expectedMessage` 👍

gliljas: `When_first_level_properties_are_tested_for_equality_against_constant_expressions_then_output_should_contain_values_of_constant_expressions` does indeed test the formatting, but the previous use of `.Body` resulted in `PredicateLambdaExpressionValueFormatter` not being used at all. I guess, in a sense, it turns the test into more of an integration test, but that's the case for all tests asserting exception messages.

The dynamic part + the use of an expression with a constant `(item => item == 2)` is what hid the problem from view. That was my reasoning behind this change. If you which me to reset it to `item => item == 2` I will gladly do so.

jnyrup: If you change it to
```cs
Expression<Func<int, bool>> expression = item => item == 2;
```
and
```cs
"Expected collection to contain a single item matching (item => item == 2), but the collection is empty."
```

that's perfectly fine with me 👍 

You could add another test that asserts on how the message is formatted, but I don't have a strong opinion about that.

PR #1620, opened 2021-07-03
Faking time because timing executions can be flaky on AppVeyor

### LINKED ISSUE #1085: Abstract time away from ExecutionTime

dennisdoomen:
On AppVeyor builds, we have observed some flakiness where timing a simple `Thread.Sleep(200)` can result in up to 400 milliseconds of execution time. For those specific unit tests, we have decided to replace `Stopwatch` usage with a fake timer abstraction for those specific cases. For the other execution timing tests, I've decided to keep using the real `Stopwatch`.

Fixes #1085


jnyrup:
Wohoo!

PR #1631, opened 2021-07-30
Support covariant and inherited property exclusion and inclusion in BeEquivalentTo

### LINKED ISSUE #1562: Excluding not functioning on overridden covariant property
### Description

Excluding not functioning on overridden covariant property.

See https://github.com/fluentassertions/fluentassertions/issues/1077#issuecomment-830863819 for original context.

```csharp
using FluentAssertions;

public class BaseProp
{
    public string BaseVal { get; set; }
}

public class DerivedProp : BaseProp
{
    public string DerivedVal { get; set; }
}

public abstract class Base
{
    public abstract BaseProp Prop { get; }
}

public sealed class Derived : Base
{
    public override DerivedProp Prop { get; }
    public string OtherProp { get; set; }
    
    public Derived(DerivedProp prop)
    {
        Prop = prop;
    }
}

public class Program
{
    public static void Main()
    {
        var a = new Derived(new DerivedProp { DerivedVal = "a", BaseVal = "a_base" })
        {
            OtherProp = "other"
        };
        var b = new Derived(new DerivedProp { DerivedVal = "b", BaseVal = "b_base" })
        {
	        OtherProp = "other"
        };
        
        a.Should().BeEquivalentTo(b, opts => opts
                                            .ComparingByMembers<Derived>()
                                            .Excluding(d => d.Prop));
    }
}
```

### Expected behavior:

The `.Should().BeEquivalentTo` call should result in no test failure exception being thrown.

### Actual behavior:

```
Unhandled exception. FluentAssertions.Execution.AssertionFailedException: Expected member Prop.BaseVal to be "b_base", but "a_base" differs near "a_b" (index 0).

With configuration:
- Compare enums by value
- Exclude member root.Prop

   at FluentAssertions.Execution.FallbackTestFramework.Throw(String message) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\FallbackTestFramework.cs:line 18
   at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\TestFrameworkProvider.cs:line 40
   at FluentAssertions.Execution.CollectingAssertionStrategy.ThrowIfAny(IDictionary`2 context) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Execution\CollectingAssertionStrategy.cs:line 44
   at FluentAssertions.Equivalency.EquivalencyValidator.AssertEquality(EquivalencyValidationContext context) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Equivalency\EquivalencyValidator.cs:line 43
   at FluentAssertions.Primitives.ObjectAssertions.BeEquivalentTo[TExpectation](TExpectation expectation, Func`2 config, String because, Object[] becauseArgs) in C:\projects\fluentassertions-vf06b\Src\FluentAssertions\Primitives\ObjectAssertions.cs:line 125
   at Program.Main()
Command terminated by signal 6
```

### Versions

* FluentAssertions 5.10.3
* .NET 5

dennisdoomen:
`Excluding()` or `Including()` covariant properties in a `BeEquivalentTo` call didn't work. To reproduce this, I had to include .NET 5 to the frameworks the unit test project targets. Also, covariant properties behave completely different from normal properties. In fact, from a Reflection point of view, it looks like a covariant property is a completely different property, a bit like `new` properties. See also this [SO thread](https://stackoverflow.com/questions/68592010/how-to-find-the-base-class-property-of-a-covariant-property-override-in-c-sharp?noredirect=1#comment121223522_68592010). 

Then while looking at #1363, I discovered another problem. Simply calling `GetProperties` on a hierarchy that includes covariant or "new" properties will also include the hidden properties, causing weird side-effects later on. The new implementation traverses the class hierarchy and will only return the fields and properties that are really visible from a certain type.

Fixes #1562 and #1363

### DIFF
--- a/Src/FluentAssertions/Common/MemberPath.cs
+++ b/Src/FluentAssertions/Common/MemberPath.cs
@@ -1,5 +1,7 @@
 using System;
 using System.Linq;
+using System.Reflection;
+using FluentAssertions.Equivalency;
 
 namespace FluentAssertions.Common
 {
@@ -9,13 +11,20 @@ namespace FluentAssertions.Common
     /// </summary>
     internal class MemberPath
     {
-        private readonly Type declaringType;
         private readonly string dottedPath;
+        private readonly Type reflectedType;
+        private readonly Type declaringType;
 
         private string[] segments;
 
-        public MemberPath(Type declaringType, string dottedPath)
+        public MemberPath(IMember member, string parentPath)
+            : this(member.ReflectedType, member.DeclaringType, parentPath.Combine(member.Name))
         {
+        }
+
+        public MemberPath(Type reflectedType, Type declaringType, string dottedPath)
+        {
+            this.reflectedType = reflectedType;
             this.declaringType = declaringType;
             this.dottedPath = dottedPath;
         }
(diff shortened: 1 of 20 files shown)


Src/FluentAssertions/Common/MemberPath.cs
```diff
@@ -9,56 +11,61 @@ namespace FluentAssertions.Common
...
+        {
+        }
 
-        public MemberPath(Type declaringType, string dottedPath)
+        public MemberPath(Type reflectedType, Type declaringType, string dottedPath)
         {
+            this.reflectedType = reflectedType;
             this.declaringType = declaringType;
             this.dottedPath = dottedPath;
+            segments = dottedPath.Split(new[] { '.', '[', ']' }, StringSplitOptions.RemoveEmptyEntries);
```
jnyrup, line 29: I seem to recall that `segments` was lazily initialised because not all instances of `MemberPath` needed it to be calculated and it stood out in some profiling.
But my memory may be playing tricks on me.

Tests/FluentAssertions.Specs/Equivalency/BasicEquivalencySpecs.cs
```diff
@@ -2083,8 +2083,112 @@ internal class ClassWithoutProperty
...
+
+            public int Value2 { get; set; }
+        }
+
+        public class DerivedClassImplementingInterface : BaseProvidingSamePropertiesAsInterface, IInterfaceWithTwoProperties
+        {
+        }
+
+        [Fact]
+        public void Exclude()
```
jnyrup, line 2155: Can we have better test names than `Exclude` and `Include`?

Src/FluentAssertions/Common/MemberPath.cs
dennisdoomen: You're right. I need to put that back. I had some intermediate changes that required me to get the segments from the constructor, but I no longer need that.

Tests/FluentAssertions.Specs/Equivalency/BasicEquivalencySpecs.cs
```diff
@@ -2083,8 +2083,112 @@ internal class ClassWithoutProperty
...
+            {
+                OtherProp = "other"
+            };
+
+            var expectation = new DerivedWithCovariantOverride(new DerivedWithProperty { DerivedProperty = "b", BaseProperty = "b_base" })
+            {
+                OtherProp = "other"
+            };
+
+            actual.Should().BeEquivalentTo(expectation, opts => opts
```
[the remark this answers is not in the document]
dennisdoomen: I'll add your scenarios.

Tests/FluentAssertions.Specs/Equivalency/BasicEquivalencySpecs.cs
dennisdoomen: Forgot to rename that. Will do.

jnyrup (approved):

PR #1435, opened 2020-11-21
Some assertions are not including variable names in assertion message

Serg046:
## IMPORTANT 

* [x] The code complies with the [Coding Guidelines for C#](https://www.csharpcodingguidelines.com/).
* [x] The changes are covered by a new or existing set of unit tests which follow the Arrange-Act-Assert syntax such as is used [in this example](https://github.com/fluentassertions/fluentassertions/blob/daaf35b9b59b622c96d0c034e8972a020b2bee55/Tests/FluentAssertions.Shared.Specs/BasicEquivalencySpecs.cs#L33).
* [x] If the contribution adds a feature or fixes a bug, please update [**the release notes**](https://github.com/fluentassertions/fluentassertions/tree/master/docs), which are published on the [website](https://fluentassertions.com/releases).
* [x] If the contribution changes the public API the changes needs to be included by running `AcceptApiChanges.ps1`/`AcceptApiChanges.sh`.
* [ ] If the contribution affects [the documentation](https://github.com/fluentassertions/fluentassertions/tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).

Fixes #1353

PR #1639, opened 2021-08-10
Merge missed improvements from develop

dennisdoomen:
Brings #1435 and #1619 on the 6.0 branch.

### DIFF
--- a/Src/FluentAssertions/CallerIdentification/AwaitParsingStrategy.cs
+++ b/Src/FluentAssertions/CallerIdentification/AwaitParsingStrategy.cs
@@ -1,0 +1,46 @@
+﻿using System.Text;
+
+namespace FluentAssertions.CallerIdentification
+{
+    internal class AwaitParsingStrategy : IParsingStrategy
+    {
+        private const string KeywordToSkip = "await";
+
+        public ParsingState Parse(char symbol, StringBuilder statement)
+        {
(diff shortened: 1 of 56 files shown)


Src/FluentAssertions/CallerIdentification/AwaitParsingStrategy.cs
```diff
@@ -0,0 +1,39 @@
...
+                var leftIndex = statement.Length - 1;
+                var rightIndex = KeywordToSkip.Length - 1;
+
+                for (var offset = 0; offset < KeywordToSkip.Length; offset++)
+                {
+                    if (statement[leftIndex - offset] != KeywordToSkip[rightIndex - offset])
+                    {
+                        return ParsingState.InProgress;
+                    }
+                }
```
jnyrup, line 22: I had to read this a couple of times to realise that it's a `Stringbuilder` variant of `EndsWith`.

jnyrup (approved):

Src/FluentAssertions/CallerIdentification/AwaitParsingStrategy.cs
dennisdoomen: Yeah, copied it from the `ShouldCallParsingStrategy` and then tried to see if I could use spans or the new ranged indexes. But I don't think it helps a lot.

PR #1640, opened 2021-08-11
Add GitHub build action

eNeRGy164:
Add GitHub as a build environment.

I'm not sure of additional (and maybe conditional steps) that the AppVeyor environment contains.

Example: https://github.com/eNeRGy164/fluentassertions/actions/runs/1119600839

**Todo for the maintainers**:
* [ ] Assign correct triggers
* [ ] Add additional/conditional steps

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,44 @@
+name: Fluent Assertions - Pull Requests
+
+on:
+  push:
+    branches: [ develop, release-6.0 ]
+  pull_request:
+    branches: [ develop, release-6.0 ]
+
+jobs:
+  build:
+
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    
+    - name: Setup .NET 5
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 5.0.x
+    
+    - name: Setup .NET 3.1
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 3.1.x
+        
+    - name: Setup .NET 2.1
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 2.1.x
+    
+    - name: Install NUKE
+      run: |
+        dotnet tool install --global Nuke.GlobalTool
+        
+    - name: Run NUKE
+      run: nuke
+
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v2
+      with:
+        path: ./Artifacts/*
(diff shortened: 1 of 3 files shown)


eNeRGy164:
Didn't expect them to race each other immediately 😛 

image

eNeRGy164:
I see the naming of the NuGet package is different.
AppVeyor: `FluentAssertions.6.1.0-github-actions0001.nupkg`
GitHub: `FluentAssertions.6.1.0-pullrequest1640-0010.nupkg`

dennisdoomen:
* How would this work with other branches like `release-*`, `support-*`, `hotfix-*`
*

eNeRGy164:
I'm not expert at all, but some toughts:

> * How would this work with other branches like `release-*`, `support-*`, `hotfix-*`

Make it general, for example, every push or pr:
```yaml
on: [push, pull_request]
```

or 

```yaml
on:
  push:
    branches:    
      - master
      - 'release-*'
      - 'support-*'
      - 'hotfix-*'
```

> The branches, branches-ignore, tags, and tags-ignore keywords accept glob patterns that use characters like *, **, +, ?, ! and others to match more than one branch or tag name.

> * Can we trigger a "master" branch on tags only?

I not sure, maybe something like:
```yaml
on:  
  push:
    branches: [ 'master' ]
    tags: [ '*' ]
```

But I don't know if you can combine these triggers with the others, or a separate workflow would be easier

> * How can I configure an API key so that tagged build can directly push to Nuget?

Add it as a secret to the project and refence it.
<https://docs.github.com/en/actions/reference/encrypted-secrets>

For example
```yaml
- name: Push NuGet
  if: ${{ github.ref == 'refs/heads/main' }}
  run: dotnet nuget push ./src/PlantUml.Builder/bin/Release/*.nupkg --source https://api.nuget.org/v3/index.json --api-key ${{secrets.NUGET_API_KEY}} --skip-duplicate
```
See <https://github.com/eNeRGy164/PlantUml.Builder/blob/main/.github/workflows/continuous-integration-workflow.yml#L68>

My sources:
- <https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#about-yaml-syntax-for-workflows>
- <https://docs.github.com/en/actions/reference/events-that-trigger-workflows>
- <https://futurestud.io/tutorials/github-actions-run-a-workflow-when-creating-a-tag>
- <https://docs.github.com/en/actions/reference/encrypted-secrets>

dennisdoomen:
@jnyrup I suggest to merge this so we can start getting some early feedback. Maybe we can even disable the existing AppVeyor PR build

jnyrup (approved):
Woohoo!

PR #1642, opened 2021-08-11
Hide protected member to avoid confusions for folks  upgrading to v6

dennisdoomen:
In v6, there's no `Should().Throw()` that handles `async` code because you need to use `Should().ThrowAsync()` for that. However, while dogfooding, I discovered that existing calls to `Throw()` will raise a `Cannot access a protected member here`. By renaming the protected members of `DelegateAssertionsBase`, it'll be clearer that a different method must be used.


jnyrup (approved):

jnyrup:
I would have expected api approval tests to fail?

dennisdoomen:
> I would have expected api approval tests to fail?

Good point. The build did not even trigger. And on my local machine, the tests didn't fail. Anyway, after re-running everything, it did flag them as breaking changes.

PR #1646, opened 2021-08-17
Added a missing breaking change to the release notes

### LINKED ISSUE #1645: Ambiguous invocation when using BeEquivalentTo with Excluding with <6.0
This code brings compiler exception "ambiguous invocation" when using 6.0.0 version, but works with lower versions.

```cs
private class Test1
{
}

private class Test2
{
}

[Fact]
public async Task Test_Case()
{
    var test1 = new Test1();
    var test2 = new Test2();

    //Work around
    //Expression<Func<IMemberInfo, bool>> expression = info => info.Name == "Level.Level.Text";
    //test1.Should().BeEquivalentTo(test2, opts => opts.Excluding(expression));

    test1.Should().BeEquivalentTo(test2,
        opts => opts.Excluding(ctx => ctx.SelectedMemberPath == "Level.Level.Text"));
}
```

dennisdoomen:
Fixes #1645

### DIFF
--- a/docs/_pages/releases.md
+++ b/docs/_pages/releases.md
@@ -140,6 +140,8 @@ sidebar:
 * Renamed `EquivalencyStepCollection` to `EquivalencyPlan` and `AssertionOptions.EquivalencyCollection` to `EquivalencyPlan` - [#1539](https://github.com/fluentassertions/fluentassertions/pull/1539).
 * `BeEquivalentTo` will no longer include `internal` properties and fields, unless `IncludingInternalProperties` or `IncludingInternalFields` is used - [#1575](https://github.com/fluentassertions/fluentassertions/pull/1575).
 * Renamed `protected` methods on `DelegateAssertionsBase` such as `Throw` and `NotThrow` to avoid confusion - [#1642](https://github.com/fluentassertions/fluentassertions/pull/1642)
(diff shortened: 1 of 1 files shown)


docs/_pages/releases.md
```diff
@@ -151,6 +151,7 @@ sidebar:
 * The `IEquivalencyValidationContext` has undergone significant refactorings where its properties have been moved into the `INode` hierarchy and the two `Reason` and `Tracer` classes - [#1379](https://github.com/fluentassertions/fluentassertions/pull/1379)
 * The `IMemberMatchingRule`, `IMemberSelectionRule` and `IOrderingRule` have been changed to replace their dependency on `SelectedMemberInfo` to `INode` and its derivatives `IMember` - [#1379](https://github.com/fluentassertions/fluentassertions/pull/1379)
 * The `SelectedMemberInfo` class has been removed, since it main user, `IMemberInfo` has been flattened - [#1379](https://github.com/fluentassertions/fluentassertions/pull/1379)
+* `IMemberInfo.SelectedMemberInfo`, `SelectedMemberPath` and `SelectedMemberDescription` have been replaced by `Name`, `Type`, `DeclaringType` and `Path` - [#1379](https://github.com/fluentassertions/fluentassertions/pull/1379)
```
jnyrup, line 154: I'm not sure I understand what properties have been replaced with what.
E.g. the three old properties have been replaced by four new.

dennisdoomen: Rephrased it. Hopefully it's better now.

jnyrup (approved):

PR #1649, opened 2021-08-17
Throw when asserting on AndConstraint

### LINKED ISSUE #1617: Assertion of type (BeOfType<>) after And fails
### Description

Assertion of type after `And` fails

```csharp
using Microsoft.VisualStudio.TestTools.UnitTesting;
using System;

namespace FluentAssertions
{
    [TestClass]
    public class FluentAssertionTest
    {
        private Exception exception = new ArgumentNullException();

        [TestMethod]
        public void This_should_work_but_it_does_not()
        {
            // TestMethod fails: Expected type to be System.ArgumentNullException, 
            // but found FluentAssertions.Primitives.ObjectAssertions.
            exception.Should().NotBeNull().And.Should().BeOfType<ArgumentNullException>();
        }

        [TestMethod]
        public void This_works()
        {
            exception.Should().NotBeNull();
            exception.Should().BeOfType<ArgumentNullException>();
        }
    }
}

```

### Expected behavior:

Both test method to work

### Actual behavior:

Test method with `And` (This_should_work_but_it_does_not) is not working

### Versions

FluentAssertions version: 5.10.3 (and 6.0.0-beta0001)
.NET Core 3.1

eNeRGy164:
This is an attempt to help people when they write `.Should()` directly after `.And` by throwing an exception.

As classes with assertions are not decorated with something to recognize (e.g. a shared base class or interface) I use reflection to get all the return types for `Should` methods on the `AssertionExtensions` class.

Had to rewrite unit tests that were using `ObjectAssertions` to test cross-assembly type comparison as they would also fail with the new exception.

Should fix #1617

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/AssertionExtensions.cs
+++ b/Src/FluentAssertions/AssertionExtensions.cs
@@ -846,5 +846,108 @@ namespace FluentAssertions
         {
             return subject is TTo to ? to : default;
         }
+
+        #region Prevent chaining on AndConstraint
+
+        /// <inheritdoc cref="Should(ExecutionTimeAssertions)" />
+        [Obsolete("You are asserting the 'AndConstraint' itself. Remove the 'Should()' method directly following 'And'", error: true)]
+        public static void Should<TSubject, TAssertions>(this ReferenceTypeAssertions<TSubject, TAssertions> _)
+            where TAssertions : ReferenceTypeAssertions<TSubject, TAssertions>
+        {
+            InvalidShouldCall();
+        }
+
+        /// <inheritdoc cref="Should(ExecutionTimeAssertions)" />
+        [Obsolete("You are asserting the 'AndConstraint' itself. Remove the 'Should()' method directly following 'And'", error: true)]
+        public static void Should<TAssertions>(this BooleanAssertions<TAssertions> _)
+            where TAssertions : BooleanAssertions<TAssertions>
+        {
+            InvalidShouldCall();
+        }
+
+        /// <inheritdoc cref="Should(ExecutionTimeAssertions)" />
+        [Obsolete("You are asserting the 'AndConstraint' itself. Remove the 'Should()' method directly following 'And'", error: true)]
+        public static void Should<TAssertions>(this DateTimeAssertions<TAssertions> _)
+            where TAssertions : DateTimeAssertions<TAssertions>
+        {
+            InvalidShouldCall();
+        }
+
+        /// <inheritdoc cref="Should(ExecutionTimeAssertions)" />
+        [Obsolete("You are asserting the 'AndConstraint' itself. Remove the 'Should()' method directly following 'And'", error: true)]
+        public static void Should<TAssertions>(this DateTimeOffsetAssertions<TAssertions> _)
+            where TAssertions : DateTimeOffsetAssertions<TAssertions>
+        {
+            InvalidShouldCall();
+        }
+
+        /// <summary>
+        /// You are asserting the <see cref="AndConstraint{T}"/> itself. Remove the <c>Should()</c> method directly following <c>And</c>.
+        /// </summary>
+        [Obsolete("You are asserting the 'AndConstraint' itself. Remove the 'Should()' method directly following 'And'", error: true)]
+        public static void Should(this ExecutionTimeAssertions _)
+        {
+            InvalidShouldCall();
+        }
+
+        /// <inheritdoc cref="Should(ExecutionTimeAssertions)" />
+        [Obsolete("You are asserting the 'AndConstraint' itself. Remove the 'Should()' method directly following 'And'", error: true)]
+        public static void Should<TAssertions>(this GuidAssertions<TAssertions> _)
+            where TAssertions : GuidAssertions<TAssertions>
+        {
+            InvalidShouldCall();
+        }
+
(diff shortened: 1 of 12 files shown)


dennisdoomen:
The release notes are missing, but now I'm in doubt whether we should retain the `develop` branch. If we add the release notes, it'll become available on the website right away.

lg2de:
If you write "Unreleased" instead of "6.1.0" the Changelog is correct.

https://keepachangelog.com/

dennisdoomen:
> If you write "Unreleased" instead of "6.1.0" the Changelog is correct.

Oh, I like that.

dennisdoomen (approved):

jnyrup:
I've thought of an alternative approach where we instead add honeypot overloads of `Should`.

Pros:
* It's very explicit
* Does not have any runtime impact on `Should(object)`

Cons:
* Changes to API

I could find 12 base assertions, that should cover all cases:
* BooleanAssertions
* DateTimeAssertions
* DateTimeOffsetAssertions
* ExecutionTimeAssertions
* GuidAssertions
* MethodInfoSelectorAssertions
* NumericAssertions
* PropertyInfoSelectorAssertions
* ReferenceTypeAssertions
* SimpleTimeSpanAssertions
* TaskCompletionSourceAssertions
* TypeSelectorAssertions

```cs
/// <summary>
/// You are asserting the 'AndConstaint' itself. You probably want to remove the 'Should()' method directly following 'And'.
/// </summary>
public void Should(ReferenceTypeAssertions<TSubject, TAssertions>)
    where TAssertions : ReferenceTypeAssertions<TSubject, TAssertions>
{
    InvalidShouldCall();
}

/// <summary>
/// You are asserting the 'AndConstaint' itself. You probably want to remove the 'Should()' method directly following 'And'.
/// </summary>
public void Should(BooleanAssertions<TAssertions>)
    where TAssertions : BooleanAssertions<TAssertions>
{
    InvalidShouldCall();
}

... 10 more ...

private static void InvalidShouldCall() => 
    throw new InvalidOperationException("You are asserting the 'AndConstaint' itself. You probably want to remove the 'Should()' method directly following 'And'.");
```

dennisdoomen:
> I've thought of an alternative approach where we instead add honeypot overloads of `Should`.

I like both approaches equally. The advantage of this "honeypot" `Should` is that you can also mark it with an `[Obsolete]` attribute that provides immediate feedback.

jnyrup:
Another benefit!
It also prevents you from trying to chain further, as it returns void instead of `ObjectAssertions`.

dennisdoomen:
> It also prevents you from trying to chain further, as it returns void instead of `ObjectAssertions`.

Fair enough. @eNeRGy164 what do you think?

lg2de:
May be another approach: Instead of pervert the call of `Should` it could return the `AndConstraint`.

eNeRGy164:
> > It also prevents you from trying to chain further, as it returns void instead of `ObjectAssertions`.
> 
> Fair enough. @eNeRGy164 what do you think?

I didn't go on this route as it changed the API (as mentioned by @jnyrup) and will require attention for authors in the future if they add new assertion classes.
However, I also expected this would require 60 new methods, but as Jonas points out the number is limited due to inheritance.

So, I will modify the code to this.

I think the remark about the return value by @lg2de is also interesting. `[Obsolete]` will produce a warning or error, but we could chain it anyway and leave it up for the user to correct their mistake. It could also be an Analyzer for that matter.

You could argue that making the method `Void` would be a breaking change in the API.

lg2de:
Technical there are breaking changes. But there are no functional breaking changes, except there are users explicitly asserting on `AndConstraint` or similar.

Adding `[Obsolete]` gets breaking change for users with "warning as error" enabled.

eNeRGy164:
> Adding `[Obsolete]` gets breaking change for users with "warning as error" enabled.

Because of the `, true)` in the attribute it is an error, not a warning. So, it becomes uncompilable.
Therefore I also can't create a unit test anymore for this.

dennisdoomen:
We could debate is that we're just fixing a false positive for them.

lg2de:
@dennisdoomen How about support chaining instead of preventing? 
Using second Should() is still a fluent statement, isn't it?

dennisdoomen:
> How about support chaining instead of preventing?
> Using second Should() is still a fluent statement, isn't it?

Not sure what that would look like. Extension methods on all concrete versions of `AndConstraint<T>`?

jnyrup:
```cs
public static TAssertions Should<TAssertions>(this BooleanAssertions<TAssertions> @this)
    where TAssertions : BooleanAssertions<TAssertions>
{
    return (TAssertions)@this;
}
```

Would allow one to write
```cs
true.Should().BeTrue().And.Should().BeFalse();
```

As the second `Should()` is a no-op, you can also write
```cs
true.Should().BeTrue().And.Should().Should().Should().BeFalse();
```

dennisdoomen:
Nice. Just tried it.

eNeRGy164:
So, which direction do you want to go? The prevention or the NOOP chaining?

dennisdoomen:
I'm fine with the direction we took. @jnyrup needs to approve the PR.

jnyrup (approved):

PR #1654, opened 2021-08-19
updated collection.Should().BeEquivalentTo to not use params

ursenzler:
change affects only the documentation


jnyrup (approved):

dennisdoomen:
Thank you for your contribution.

PR #1658, opened 2021-08-20
Add extension method `WithInnerExceptionExactly` for async exception assertions

### LINKED ISSUE #1400: Add missing extensions for ThrowAsync
Hello, thanks for the great project!

This is identical to #1052 - Add `async`-support for `WithMessage()`
Which added an extension on the Task<..Assertion> itself for WithMessage

Can I submit a PR for the other 3 ExceptionAssertion methods, so that parentheses around the whole await expression are not required?

Namely,
-WithInnerException
-WithInnerExceptionExactly
-Where

LennartKoot:
Adds extension method for the remaining method `WithInnerExceptionExactly` for `ExceptionAssertions<T>` when using `ThrowAsync`.

Closes #1400

### DIFF
--- a/Src/FluentAssertions/ExceptionAssertionsExtensions.cs
+++ b/Src/FluentAssertions/ExceptionAssertionsExtensions.cs
@@ -81,6 +81,29 @@ namespace FluentAssertions
         }
 
         /// <summary>
+        /// Asserts that the thrown exception contains an inner exception of the exact type <typeparamref name="TInnerException" /> (and not a derived exception type).
+        /// </summary>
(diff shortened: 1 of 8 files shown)


LennartKoot:
Hmm, after second thought, the `And` and `Which` extensions method are not really useful as you still have to wrap it in an `await` to get the actual exception. Will drop that change.

Please let me know if I am missing a different approach to provide `And` and `Which` chaining directly after `ThrowAsync`. I'll happily implement it.

dennisdoomen (approved):

jnyrup (approved):

PR #1662, opened 2021-08-22
Finalize the Github Actions migration

dennisdoomen:
## IMPORTANT 

Finishes the work that @eNeRGy164 started to build all PRs, branches `develop` and `release-*` and to ignore PRs only affecting the `docs` folder. Not sure if the Nuget push will work for tags, but we'll know when we get there.


jnyrup (approved):
Nice!

.github/workflows/build.yml
```diff
@@ -42,3 +45,10 @@ jobs:
       uses: actions/upload-artifact@v2
       with:
         path: ./Artifacts/*
+        
+    - name: Push to Nuget
+      run: dotnet nuget push ./artifacts/*.nupkg -k {ApiKey} --skip-duplicate --no-symbols true
+      if: contains(github.ref, 'refs/tags/')
```
lg2de, line 51: Do you think that this will really work?
I'm missing the trigger to build tags.

dennisdoomen: You might be on to something.

Issue #1652, opened 2021-08-19
Migration guide to v6 is missing  breaking change Throw/ThrowAsync


myshon:
I'm upgrading FluentAssertions to v6 (good job!). In the previous version v5, I had a lot of tests like that 

```csharp
Func<Task> action = async () => await repository.CreateFile();
action .Should().Throw<FileAlreadyExists>();
```

However, after the upgrade I encountered builds errors, these tests need to be fixed 

```csharp
Func<Task> action = async () => await repository.CreateFile();
await action .Should().Throw<FileAlreadyExists>();
```

Of course, it's pretty easy to change. I only wonder, why this breaking change it's not mentioned in the migration guide https://fluentassertions.com/upgradingtov6 
It applies also to ThrowExactly, NotThrow etc.

jnyrup:
Hi Szymon
Thanks for the kind words, we're also really glad to have it out.

Regarding your question, I think we simply forgot as we only started the migration guide later in the development process of v6.
Dennis did though mention it in his [blog post](https://www.continuousimprover.com/2021/08/fluent-assertions-v6.html) about v6.

But I agree that the migration guide should also mention the removal of blocking `Throw` for async delegates.

dennisdoomen:
Added a small section on this to the migration guide.

Issue #337, opened 2016-01-06
IEventRecorder .WithArgs<T> returns unfiltered IEventRecorder - feature / room for improvement?


joergbattermann:
I just noticed an somewhat unexpected behaviour testing events for one of my classes:

Given the following test

```c#
[Fact]
public void AddRangeRaisesPropertyChangedEventForItemIndexer()
{
	// given
	using (var observableDictionary = new ObservableDictionary<int, string>())
	{
		observableDictionary.ThresholdAmountWhenItemChangesAreNotifiedAsReset = 0;
		observableDictionary.MonitorEvents();

		var items = new List<KeyValuePair<int, string>>()
		{
			new KeyValuePair<int, string>(1, "One"),
			new KeyValuePair<int, string>(2, "Two"),
			new KeyValuePair<int, string>(3, "Three"),
		};

		// when
		observableDictionary.AddRange(items);

		// then
		observableDictionary
			.ShouldRaise(nameof(observableDictionary.PropertyChanged))
			.WithSender(observableDictionary)
			.WithArgs<PropertyChangedEventArgs>(args => args.PropertyName == "Item[]")
			.Should().HaveCount(1);
	}
}
```

I get a failing test because Count (the very last line) was expected to be 1, but it is 2.

Now the thing is - exactly one event is being raised for `Item[]`, the second one is for a different property (.Count).

Looking at the source of [PropertyChangedSourceExtensions.cs](https://github.com/dennisdoomen/fluentassertions/blob/e240c8c8d02160fd69e22d7590901df429f86bd3/Src/Core/PropertyChangedSourceExtensions.cs#L164) it becomes apparent that currently the IEventRecorder instance is returned back _as-is_ and not a filtered-down version using the supplied predicate.

I was expecting that every .With*() method here checks for the condition and if it passes, it returns the filtered eventRecorder back so one can actually fluently chain up checks.

I.e.:
- `.ShouldRaise(nameof(observableDictionary.PropertyChanged))` checks for any events for the given name, fails if none are raised or if they are, returns a filtered EventRecorder back only containing those events
- `.WithSender(observableDictionary)` then checks whether those ("PropertyChanged") events have the given sender, fails if not OR if they do, return a further filtered EventRecorder back - only containing "PropertyChanged" events back with the given sender
[...]

PR #1663, opened 2021-08-23
Added more upgrade info to fix the breaking changes in v6

dennisdoomen:
#337
#1652


docs/_pages/upgradingtov6.md
```diff
@@ -218,3 +218,35 @@ You might ask if we could not just have changed the overload signature from `par
...
+As part of fixing [this PR](https://github.com/fluentassertions/fluentassertions/issues/337) we had to change the event assertions work. Now, event constraining extensions like `WithArgs` and `WithSender` will return only the events that match the constraints. This means that this will no longer work:
+
+```csharp
+foo
+    .ShouldRaise("SomeEvent")
+         .WithArgs<string>(args => args == "some event payload")
+         .WithArgs<string>(args => args == "other payload");
+```
+
+In 6.0, you need to rewrite this as:
```
jnyrup, line 242: I think(?) this should be "rewrite this to" or "write this as"

jnyrup (approved):

PR #1666, opened 2021-08-23
Rearrange all (generic) collection tests in separate files

eNeRGy164:
* Combines _Collection Specs_ & _Generic Collection Specs_ as there is no difference anymore
* Splits specs over separate files
* Removes 24 duplicate unit tests

Confirmed no changes in code coverage before & after change (94,82% for the net5.0 project)


jnyrup (approved):
Thanks!

dennisdoomen (approved):

PR #1673, opened 2021-08-27
Change names of "OrEqualTo" assertion methods

gliljas:
The names of the GreaterOrEqualsTo methods (and similar) is a bit off, but more importantly they don't follow the established terminology.

### DIFF
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
@@ -1276,7 +1277,7 @@ namespace FluentAssertions.Collections
         }
 
         /// <summary>
-        /// Asserts that the number of items in the collection is greater or equal to the supplied <paramref name="expected" /> amount.
+        /// Asserts that the number of items in the collection is greater than or equal to the supplied <paramref name="expected" /> amount.
         /// </summary>
         /// <param name="expected">The number to which the actual number items in the collection will be compared.</param>
         /// <param name="because">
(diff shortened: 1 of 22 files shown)


lg2de:
Many projects have activated "Warnung as Error". So, `Obsolete` will cause build errors. Therefore it should not be used here.
I think it is OK to add `EditorBrowsable`.
In Version 7 the old version can be removed. @dennisdoomen should we create an issue to collect all work shifted to V7? Or do you like to use "milestones"?

@gliljas, I think you should rebase to fix the conflicts.

dennisdoomen:
Yes, a milestone is fine. But we're not going to start on that anytime soon. Probably in two years or so.

jnyrup:
I'm fine with the new overloads and only hiding the current ones.
Let's not add Obsolete - it would make me a bit sad if that scares/hinders people from upgrading to v6.

The release notes should have a line about the new overloads.

gliljas:
Pushed to verify rebase. Will fix verification and release notes.

dennisdoomen:
> and only hiding the current ones.

Wait, what do you mean with that? Because the latest commit removes them, which would be a breaking change.

gliljas:
> > and only hiding the current ones.
> 
> Wait, what do you mean with that? Because the latest commit removes them, which would be a breaking change.

I think he meant the EditorBrowsable thing, making use of the current methods somewhat discouraged. The current ones are still there, as far as I know/intended.

dennisdoomen:
`EditorBrowsable` was only used for UI designers. Why is it relevant here?

jnyrup:
Adding that attribute will hide the current/old overload from intellisense suggestions in VS

dennisdoomen:
> Adding that attribute will hide the current/old overload from intellisense suggestions in VS

Nice. I didn't know that. Never too old to learn something new.

dennisdoomen (approved):

jnyrup (approved):

PR #1678, opened 2021-09-05
Add test to find missing guarding Should overloads

jnyrup:
Following up on #1649 this PR adds a test to verify that we don't forget to add new guarding `Should` overloads.
In the list I created in #1649 I only included overloads from `AssertionExtensions` but the `Should` for `EnumAssertions` is defined in `EnumAssertionsExtensions`.
This new test caught that oversight.


dennisdoomen:
Great idea!

dennisdoomen (approved):

PR #1708, opened 2021-10-16
IgnoreCyclicReferences in BeEquivalentTo now works with ComparingByMembers

### LINKED ISSUE #1370: Combination of ComparingByMember and IgnoreCyclicReferences doesn't work

dennisdoomen:
When an object graph implementing value semantics (by overriding `Equals`) contains cyclic references, the `IgnoringCyclicReferences` option did not work. 

Fixes #1370


dennisdoomen:
> While correctness is more important that speed, I'm curious what effect deletion of the cache has.

I thought about that, but the `GetEqualityStrategy` method uses its own cache. The only concern I had is that the `IsRecord` method might be expensive as well, but looking at the build time, it doesn't seem to have such an effect. I'll run a profile session on my big machine when I have time.

PR #1724, opened 2021-10-31
Add test of clearing cache when calling `ComparingByMembers`

jnyrup:
In addition I noted that it doesn't seem necessary to clear the cache on calling `ComparingRecordsByValue`/`ComparingRecordsByMembers` as:
* `compareRecordsByValue` is initialized to `defaults.CompareRecordsByValue` and
* `GetEqualityStrategy` does not call recursively to determine strategy when `type.IsRecord()`.

### DIFF
--- a/Src/FluentAssertions/Common/TypeExtensions.cs
+++ b/Src/FluentAssertions/Common/TypeExtensions.cs
@@ -217,7 +217,7 @@ namespace FluentAssertions.Common
 
         public static IEnumerable<PropertyInfo> GetNonPrivateProperties(this Type typeToReflect, MemberVisibility visibility)
         {
-            return NonPrivatePropertiesCache.GetOrAdd((typeToReflect, visibility), key =>
+            return NonPrivatePropertiesCache.GetOrAdd((typeToReflect, visibility), static key =>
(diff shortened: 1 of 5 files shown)


dennisdoomen (approved):
Nice work. And I didn't know about the `static` modifier on lambdas.

lg2de:
Could you please add a description here why the `static` modifier is helpful or even required?
For me there is no relation between the title, description and this change.

jnyrup:
The [`static`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/static-anonymous-functions) modifier is not required.
It doesn't change the meaning of the program, but enforces that we per by default cannot capture state for the static shared caches.
Adding `static` was what helped me in https://github.com/fluentassertions/fluentassertions/pull/1719#issuecomment-950184975 to see  that we captured state in the `valueFactory` lambda and hence had a bug if we didn't clear the cache.

The changes in this PR is about making the non-capturing pitfall more obvious.
Either by enforcing it with the `static` modifier or by adding a comment about it.

I agree that the title and description could be more explanatory.

PR #1725, opened 2021-11-01
Adding `ThatAreAsync()` and `ThatAreNotAsync()` to `MethodInfoSelector`

chvollm:
## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Types/MethodInfoSelector.cs
+++ b/Src/FluentAssertions/Types/MethodInfoSelector.cs
@@ -134,6 +134,24 @@ namespace FluentAssertions.Types
         }
 
         /// <summary>
+        /// Only return methods that are async. 
+        /// </summary>
+        public MethodInfoSelector ThatAreAsync()
+        {
+            selectedMethods = selectedMethods.Where(method => method.IsAsync());
+            return this;
+        }
+
+        /// <summary>
+        /// Only return methods that are not async. 
+        /// </summary>
+        public MethodInfoSelector ThatAreNotAsync()
+        {
+            selectedMethods = selectedMethods.Where(method => !method.IsAsync());
+            return this;
+        }
+
+        /// <summary>
         /// Select return types of the methods
         /// </summary>
         public TypeSelector ReturnTypes()
--- a/Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Reflection;
+using System.Threading.Tasks;
 using FluentAssertions.Types;
 using Internal.Main.Test;
 using Xunit;
@@ -273,6 +274,32 @@ namespace FluentAssertions.Specs.Types
         }
 
         [Fact]
+        public void When_selecting_methods_that_are_async_it_should_only_return_the_applicable_methods()
+        {
+            // Arrange
+            Type type = typeof(TestClassForMethodSelectorWithAsync);
+
+            // Act
+            IEnumerable<MethodInfo> methods = type.Methods().ThatAreAsync().ToArray();
+
+            // Assert
+            methods.Should().ContainSingle();
+        }
+
+        [Fact]
+        public void When_selecting_methods_that_are_not_async_it_should_only_return_the_applicable_methods()
+        {
+            // Arrange
+            Type type = typeof(TestClassForMethodSelector);
+
+            // Act
+            MethodInfo[] methods = type.Methods().ThatAreNotAsync().ToArray();
+
+            // Assert
+            methods.Length.Should().Be(7);
+        }
+
+        [Fact]
         public void When_selecting_methods_not_decorated_with_or_inheriting_a_noninheritable_attribute_it_should_only_return_the_applicable_methods()
         {
             // Arrange
@@ -365,6 +392,14 @@ namespace FluentAssertions.Specs.Types
     internal class TestClassForMethodSelectorWithNonInheritableAttributeDerived : TestClassForMethodSelectorWithNonInheritableAttribute
     {
         public override void PublicVirtualVoidMethodWithAttribute() { }
+    }
+
+    internal class TestClassForMethodSelectorWithAsync
+    {
+        public async virtual void PublicVirtualVoidAsyncMethod()
+        {
+            await Task.FromResult(42);
+        }
     }
 
     internal class TestClassForMethodReturnTypesSelector
--- a/Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
@@ -283,8 +283,8 @@ namespace FluentAssertions.Specs.Types
             MethodInfo[] methods = type.Methods().ThatAreAsync().ToArray();
 
             // Assert
-            methods.Should().ContainSingle();
-            methods[0].Name.Should().Be("PublicVirtualVoidAsyncMethod");
+            methods.Should().ContainSingle()
+                .Which.Name.Should().Be("PublicVirtualVoidAsyncMethod");
         }
 
         [Fact]
@@ -297,8 +297,8 @@ namespace FluentAssertions.Specs.Types
             MethodInfo[] methods = type.Methods().ThatAreNotAsync().ToArray();
 
             // Assert
-            methods.Should().ContainSingle();
-            methods[0].Name.Should().Be("PublicVirtualVoidNotAsyncMethod");
+            methods.Should().ContainSingle()
+                .Which.Name.Should().Be("PublicVirtualVoidNotAsyncMethod");
         }
 
         [Fact]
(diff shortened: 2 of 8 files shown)


Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
```diff
@@ -272,6 +273,32 @@ public void When_selecting_methods_not_decorated_with_a_noninheritable_attribute
...
+        public void When_selecting_methods_that_are_not_async_it_should_only_return_the_applicable_methods()
+        {
+            // Arrange
+            Type type = typeof(TestClassForMethodSelector);
+
+            // Act
+            MethodInfo[] methods = type.Methods().ThatAreNotAsync().ToArray();
+
+            // Assert
+            methods.Length.Should().Be(7);
```
lg2de, line 299: `methods.Should().HaveCount(7);`

Src/FluentAssertions/Types/MethodInfoSelector.cs
```diff
@@ -133,6 +133,24 @@ public MethodInfoSelector ThatAreNotDecoratedWithOrInherit<TAttribute>()
...
+        public MethodInfoSelector ThatAreAsync()
+        {
+            selectedMethods = selectedMethods.Where(method => method.IsAsync());
+            return this;
+        }
+
+        /// <summary>
+        /// Only return methods that are not async. 
+        /// </summary>
+        public MethodInfoSelector ThatAreNotAsync()
```
lg2de, line 148: How about `ThatAreSync`?
Maybe you could also add a description for your motivation for these changes.

chvollm: `Sync` is the keyword that every developer knows. It also fits into the current naming pattern. I would in the future suggest the same when it comes to e.g. static. Then we would have `ThatAreNotStatic` instead of `ThatAreInstanceMethods`.

docs/_pages/releases.md
```diff
@@ -10,6 +10,7 @@ sidebar:
 ## Unreleased
 
 ### What's New
+* Adding `ThatAreAsync()` and `ThatAreNotAsync()` to `MethodInfoSelector` [#645](https://github.com/fluentassertions/fluentassertions/pull/1725)
```
dennisdoomen, line 13: `MethodInfoSelector` is an internal class not really visible to consumers, so maybe we can keep it a bit more functional, like:

> Adding `ThatAreAsync()` and `ThatAreNotAsync()` for filtering in method assertions"

Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
```diff
@@ -272,6 +273,32 @@ public void When_selecting_methods_not_decorated_with_a_noninheritable_attribute
             methods.Should().ContainSingle();
         }
 
+        [Fact]
+        public void When_selecting_methods_that_are_async_it_should_only_return_the_applicable_methods()
+        {
+            // Arrange
+            Type type = typeof(TestClassForMethodSelectorWithAsync);
```
dennisdoomen, line 280: I recommend renaming this to `TestClassWithSingleAsyncMethod`

Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
dennisdoomen: The `7` seems to be a bit of magic number, especially because `TestClassForMethodSelector` is not immediately following the test. Maybe the test itself should use Reflection to calculate the number of expected non-async methods.

dennisdoomen (changes requested):

Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
jnyrup, line 280: As this class does not have any non-async methods, this test does not verify that `ThatAreAsync` excludes non-async methods.

How about adding a non-async method to the class and chain the `ContainSingle` with an assertion that the single method is the expected one?

You could then do almost the same in the test below to verify `ThatAreNotAsync`.

Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
```diff
@@ -367,6 +394,14 @@ internal class TestClassForMethodSelectorWithNonInheritableAttributeDerived : Te
         public override void PublicVirtualVoidMethodWithAttribute() { }
     }
 
+    internal class TestClassForMethodSelectorWithAsync
+    {
+        public async virtual void PublicVirtualVoidAsyncMethod()
+        {
+            await Task.FromResult(42);
+        }
```
jnyrup, line 402: Like in `ClassWithAllMethodsAsync`, this could be written as 
```cs
public async Task PublicAsyncDoNothing()
{
    await Task.Yield();
}
```

E.g. the `virtual` is irrelevant for this test.

docs/_pages/releases.md
jnyrup, line 13: `[#645]` -> `[#1725]`
separating `-` between description and link

jnyrup (changes requested):
Thanks for adding this, only some smaller suggestions from my side.

docs/_pages/releases.md
chvollm: Sorry, I'm not sure I understand this correctly. Do you want 2 changes? Changing the number from 645 to 1725 and adding a `-` between description and link?

Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
chvollm: I'm changing the new class to have one method for each.

docs/_pages/releases.md
chvollm: Done

Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
chvollm: Thanks. I was looking for a way to have minimal code for an async method. :-)

Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
chvollm: I have changed the class accordingly.

docs/_pages/releases.md
jnyrup: Yes 👍 
I'll try to be less note-form in my review.

chvollm: I'm unsure if you normally link to an issue or always to the PR, but I've done the change you requested.

Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
```diff
@@ -272,6 +273,34 @@ public void When_selecting_methods_not_decorated_with_a_noninheritable_attribute
...
+        {
+            // Arrange
+            Type type = typeof(TestClassForMethodSelectorWithAsyncAndNonAsyncMethod);
+
+            // Act
+            MethodInfo[] methods = type.Methods().ThatAreAsync().ToArray();
+
+            // Assert
+            methods.Should().ContainSingle();
+            methods[0].Name.Should().Be("PublicVirtualVoidAsyncMethod");
```
jnyrup, line 287:
```cs
methods.Should().ContainSingle()
    .Which.Name.Should().Be("PublicVirtualVoidAsyncMethod");
```

Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
```diff
@@ -367,6 +396,16 @@ internal class TestClassForMethodSelectorWithNonInheritableAttributeDerived : Te
...
     }
 
+    internal class TestClassForMethodSelectorWithAsyncAndNonAsyncMethod
+    {
+        public async virtual void PublicVirtualVoidAsyncMethod()
+        {
+            await Task.Yield();
+        }
+
+        public virtual void PublicVirtualVoidNotAsyncMethod() { }
```
jnyrup, line 406:
```suggestion
        public async Task PublicAsyncMethod()
        {
            await Task.Yield();
        }

        public void PublicNotAsyncMethod() { }
```

docs/_pages/releases.md
jnyrup: You could argue for both, but as I know all links in the Fluent Assertions release notes are to PRs.

chvollm: 👍

Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
chvollm: I didn't see the "Commit suggestion" button, nice feature.

dennisdoomen (approved):

chvollm:
@jnyrup Anything else you need for this PR?

Tests/FluentAssertions.Specs/Types/MethodInfoSelectorSpecs.cs
```diff
@@ -367,6 +396,16 @@ internal class TestClassForMethodSelectorWithNonInheritableAttributeDerived : Te
...
 
+    internal class TestClassForMethodSelectorWithAsyncAndNonAsyncMethod
+    {
+        public async void PublicVirtualVoidAsyncMethod()
+        {
+            await Task.Yield();
+        }
+
+        public void PublicVirtualVoidNotAsyncMethod() { }
+    }
```
jnyrup, line 407: Sorry to nag you about this, but as `virtual` is not relevant for testing async vs non-async I'd like to get rid of it.
```cs
internal class TestClassForMethodSelectorWithAsyncAndNonAsyncMethod
{
    public async Task PublicAsyncMethod()
    {
        await Task.Yield();
    }

    public Task PublicNotAsyncMethod() => Task.Completed;
}
```

chvollm: Oh, in the method name! Sorry!

dennisdoomen:
🚀

Issue #1115, opened 2019-08-17
Support c# 8 nullable reference


iron9light:
Just found this https://github.com/dotnet/standard/pull/1359 .
Maybe this lib can do similar things for c# 8 nullable reference.

jnyrup:
This is indeed an interesting area, which I'm following closely.
For anyone else interested [here](https://devblogs.microsoft.com/dotnet/try-out-nullable-reference-types/) a blog describing the features.

If we add this, it should align with #1074.
As these `Attribute`s will be available from netcoreapp3.0, we would need to think about which frameworks we target.
Implementing #902 would e.g. require netcoreapp2.1 or netstandard2.1.

Related issues about nullability:
#1058
#1039

iron9light:
I don't know the implementation details.
But I do know that: netstandard 2.0 will be the main target framework for most live libraries and projects since the .Net Framework will not support netstandard 2.1 .
And many people will use 8.0 with netstandard 2.0, not everyone will use Resharper, it's not free.
Support nullable reference should be a must-have (if it's possible). `obj.Should().NotBeNull()` is too common in the test code.

jnyrup:
Be aware that these attributes will also *not* be available in .netstandard 2.0.
See Philip Carters [comment](https://devblogs.microsoft.com/dotnet/try-out-nullable-reference-types/#comments-2649).

iron9light:
> Hey Jonas, The same as with .NET Framework. Async enumerables, Index and Range support, Default Implementations, and Nullable Annotations in first-party libraries.

Do you mean this? There're some BCL libraries, e.g. https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces . Maybe there will be one for those attributes.

jnyrup:
This comment:
> ... It's currently a bit tricky when multitargeting. We've updated the specification for .NET Standard 2.1 to include all of the more advanced attributes so that you can share that code across .NET Core and Xamarin. But the .NET Standard 2.0 spec won't be updated, so if you use that then you won't have access to these types. Similarly, since they aren't on .NET Framework, any specific target for net4X will not see these types.

If I we're to add the attributes, I would probably wait until there is more information available about how to use them in multi-targeting libraries.

dennisdoomen:
And apparently some folks managed to get it to work with .NET Standard. See https://twitter.com/ddoomen/status/1160171490081419271

danielmpetrov:
[...]

PR #1737, opened 2021-11-09
Introduce assertions for StatusCode of HttpResponseMessage

mu88:
## IMPORTANT 


Tests/FluentAssertions.Specs/Primitives/HttpResponseMessageAssertionSpecs.cs
```diff
@@ -0,0 +1,261 @@
...
+
+        [Theory]
+        [MemberData(nameof(GetClientErrorStatusCodes))]
+        public void Should_succeed_when_asserting_statuscode_is_client_error(HttpStatusCode statusCodeOfResponse)
+        {
+            // Arrange
+            var testee = new HttpResponseMessage(statusCodeOfResponse);
+
+            // Act / Assert
+            testee.Should().BeClientError();
```
dennisdoomen, line 42: 🤔 Considering that the error is a property of the response message, it feels to me that `HaveClientError` would be a more natural way of asserting this. Same with the other overloads that take an error.

Tests/FluentAssertions.Specs/Primitives/HttpResponseMessageAssertionSpecs.cs
```diff
@@ -0,0 +1,261 @@
...
+using System.Net.Http;
+using Xunit;
+using Xunit.Sdk;
+
+namespace FluentAssertions.Specs.Primitives
+{
+    public class HttpResponseMessageAssertionSpecs
+    {
+        [Theory]
+        [MemberData(nameof(GetSuccessStatusCodes))]
```
dennisdoomen, line 13: 🤔 I don't think it's very useful to run this test for every possible value. Just taking some sensible values, including edge cases would be much easier to understand. This also avoids the need to have a helper method, which I personally feel obfuscates the test purpose. Also applies to the other tests.

Tests/FluentAssertions.Specs/Primitives/HttpResponseMessageAssertionSpecs.cs
```diff
@@ -0,0 +1,261 @@
...
+                new HttpResponseMessage(HttpStatusCode.Gone).Should().BeSuccessful("because we want to test the failure {0}", "message");
+
+            // Assert
+            action
+                .Should().Throw<XunitException>()
+                .WithMessage("Expected HttpStatusCode to be successful (2xx) because we want to test the failure message, but found HttpStatusCode.Gone {value: 410}.");
+        }
+
+        [Fact]
+        public void Should_fail_when_asserting_statuscode_error_is_redirection()
```
dennisdoomen, line 95: 🔧 I suggest to group the success and failure cases for the same methods together.

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
```diff
@@ -0,0 +1,181 @@
...
+    {
+        public HttpResponseMessageAssertions(HttpResponseMessage value) => Subject = value;
+
+        /// <summary>
+        /// Gets the object which value is being asserted.
+        /// </summary>
+        public HttpResponseMessage Subject { get; }
+
+        /// <summary>
+        /// Asserts that the <see cref="HttpStatusCode"/> is successful.
```
dennisdoomen, line 36: 🔧 I suggest to add the error number to the docs. Same for the other methods.

Tests/FluentAssertions.Specs/Primitives/HttpResponseMessageAssertionSpecs.cs
```diff
@@ -0,0 +1,261 @@
...
+using Xunit;
+using Xunit.Sdk;
+
+namespace FluentAssertions.Specs.Primitives
+{
+    public class HttpResponseMessageAssertionSpecs
+    {
+        [Theory]
+        [MemberData(nameof(GetSuccessStatusCodes))]
+        public void Should_succeed_when_asserting_statuscode_is_successful(HttpStatusCode statusCodeOfResponse)
```
dennisdoomen, line 14: 🙃 I'm nitpicking here, but `statuscode` consists of two separate words. I would also remove the `asserting` word to make the functional names more concise

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
```diff
@@ -0,0 +1,181 @@
...
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> BeSuccessful(string because = "", params object[] becauseArgs)
+        {
+            Execute.Assertion
+                .ForCondition(Subject.IsSuccessStatusCode)
```
dennisdoomen, line 48: 🤔 What happens if `Subject` is `null`?

docs/_pages/httpresponsemessages.md
```diff
@@ -0,0 +1,28 @@
+---
+title: HttpResponseMessages
```
dennisdoomen, line 2: ❌ This doesn't work, unless you also update https://github.com/fluentassertions/fluentassertions/blob/master/docs/_data/navigation.yml

dennisdoomen (changes requested):
Thank you for the great contribution.

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
mu88: What about introducing _Nullable Reference Types_?

mu88:
Sorry about the build failure 🙄

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
dennisdoomen: Unfortunately that's [much more complicated](https://github.com/fluentassertions/fluentassertions/issues/1115) than we though. For now, we've settled on specific errors.

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
```diff
@@ -0,0 +1,191 @@
...
+    /// </summary>
+    [DebuggerNonUserCode]
+    public class HttpResponseMessageAssertions<TAssertions>
+        where TAssertions : HttpResponseMessageAssertions<TAssertions>
+    {
+        /// <summary>Initializes a new instance of the <see cref="HttpResponseMessageAssertions{TAssertions}"/> class.</summary>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="value"/> is <c>null</c></exception>
+        public HttpResponseMessageAssertions(HttpResponseMessage value)
+        {
+            Guard.ThrowIfArgumentIsNull(value, nameof(value));
```
dennisdoomen, line 36: ❌ We still need to properly document the [design guidelines](https://github.com/fluentassertions/fluentassertions/issues/1340), but we use guards to capture misuse of our library. However, a `null` message is something that FA should be prepared for. In those cases, we use a proper `FailWith` construct, preferably per assertion method.

docs/_data/navigation.yml
```diff
@@ -73,6 +73,8 @@ sidebar:
         url: /xml
       - title: Execution Time
         url: /executiontime
+      - title: HttpResponseMessage
```
dennisdoomen, line 76: 🙃 I'd prefer something like `HTTP response messages`

docs/_pages/releases.md
```diff
@@ -11,6 +11,7 @@ sidebar:
 
 ### What's New
 * Adding `ThatAreAsync()` and `ThatAreNotAsync()` for filtering in method assertions - [#1725](https://github.com/fluentassertions/fluentassertions/pull/1725)
+* Adding new assertions for the `HttpStatusCode` of a `HttpResponseMessage` - [#1698](https://github.com/fluentassertions/fluentassertions/discussions/1698)
```
dennisdoomen, line 14: ❌ Could you please point the link to the PR, not the issue.

dennisdoomen (changes requested):
Almost there. Just some minor changes.

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// </param>
+        public AndConstraint<TAssertions> BeRedirection(string because = "", params object[] becauseArgs)
+        {
+            Execute.Assertion
+                .ForCondition(Subject is not null)
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected HttpStatusCode to be redirection (3xx){reason}, but HttpResponseMessage was <null>.");
+
+            Execute.Assertion
+                .ForCondition((int)Subject.StatusCode >= 300 && (int)Subject.StatusCode <= 399)
```
jnyrup, line 78: Suggestion
```cs
.ForCondition((int)Subject.StatusCode is >= 300 and <= 399)
```

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> HaveStatusCode(HttpStatusCode expected, string because = "", params object[] becauseArgs)
+        {
+            var success = Execute.Assertion
+                .ForCondition(Subject is not null)
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected HttpStatusCode to be {0}{reason}, but HttpResponseMessage was <null>.", expected);
+
+            if (success)
```
jnyrup, line 178: This guard check should also be added to the other assertions.
It avoid an NRE when an assertion is wrapped in an `AssertionScope`, which defers failure until the disposal of the `AssertionScope`.

```cs
HttpResponseMessage subject = null;
subject.Should().BeSuccessful();

```

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
```diff
@@ -0,0 +1,222 @@
...
+            : base(value)
+        {
+        }
+    }
+
+    /// <summary>
+    /// Contains a number of methods to assert that a <see cref="HttpResponseMessage"/> is in the expected state.
+    /// </summary>
+    [DebuggerNonUserCode]
+    public class HttpResponseMessageAssertions<TAssertions>
```
jnyrup, line 25: This should probably either derive from `ReferenceTypeAssertions` or `ObjectAssertions`.
Deriving from `ObjectAssertions` will give the least invasive changes, as `HttpResponseMessage` currently should-resolves to `ObjectAssertions`.

Tests/Benchmarks/Benchmarks.csproj
```diff
@@ -16,5 +16,6 @@
   </ItemGroup>
   <ItemGroup>
     <Reference Include="System.Data.DataSetExtensions" Condition="'$(TargetFramework)' == 'net472'" />
+    <Reference Include="System.Net.Http" Condition="'$(TargetFramework)' == 'net472'" />
```
jnyrup, line 19: Why is this needed?

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
dennisdoomen: I love C# 9.0 ;-)

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> BeSuccessful(string because = "", params object[] becauseArgs)
+        {
+            Execute.Assertion
```
dennisdoomen, line 47: ❌ If the call to `BeSuccessful` was encapsulated by an `AssertionScope`, the `FailWith` will just queue an error message on the current scope, but not throw an actual exception. You either need to chain the two assertions using `Then` or account for the boolean return value of `FailWith`. Same for the other tests.

dennisdoomen (changes requested):

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
mu88: Yes me too! They just have to be integrated into my muscle memory ;)

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
mu88: Thanks for the explanation!

Tests/Benchmarks/Benchmarks.csproj
mu88: With the "new" .NET (.NET Standard, .NET Core, .NET ≥ 5.0), `System.Net.Http` has become part of .NET's _Base Class Libraries_. Therefore the assembly is automatically and implicitly referenced in all newer projectes.

But with .NET Framework, this is not the case. The assembly is always present in the GAC (because the .NET Framework installs it), but it has to be referenced manually.

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
```diff
@@ -0,0 +1,234 @@
...
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> HaveServerError(string because = "", params object[] becauseArgs)
+        {
+            var success = Execute.Assertion
+                .ForCondition(Subject is not null)
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected HttpStatusCode to be server error (5xx){reason}, but HttpResponseMessage was <null>.");
+
+            if (success)
```
dennisdoomen, line 161: 🔧 You can also do this by chaining those calls together. See https://github.com/fluentassertions/fluentassertions/blob/42f0f8f001cc8f291ae6d3c6efac1bf0da34100f/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs#L115

dennisdoomen (approved):

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
mu88: Would you prefer that pattern?

dennisdoomen: With a 🔧 I mean it's a suggestion. Up to you if you still want to do it. I've already approved the PR

mu88:
Do you need anything from my side to close this PR, @dennisdoomen ? Or will you resolve the conflicts when creating the next release?

dennisdoomen:
I'm waiting for the approval of @jnyrup

jnyrup:
Sorry for the delay.
I wanted to fiddle a bit with this to fully understand why a project reference to `System.Net.Http` is needed as I would hate to force an extra dependency on consumers of this library.
I should have time this week to test.

dennisdoomen:
And I thought you were supposed to be off your computer this week ;-)

> Yuck!

It's [really](https://github.com/Microsoft/dotnet/blob/master/releases/net472/KnownIssues/613745%20-%20Single-name%20references%20are%20removed%20by%20the%20SDK%20when%20targeting%204.7.2.md) ugly... 

> If we want to make sure that one can upgrade without having to manually adding a reference to `System.Net.Http` we could conditionally add a `PackageReference` for net47 and cross our fingers that it doesn't break anything else for consumers.

I've tried creating a .NET 4.7.0, .NET 4.7.2 and a .NET 6 project. For the first two, I had to add an assembly reference to refer to an `HttpResponseMessage`. For the latter, it was just there. 

I'm fine to take the plunge and add that conditional package reference.

dennisdoomen:
@jnyrup do you stick with your opinion?

jnyrup:
Let's conditionally add `System.Net.Http` as a nuget dependency to the net47 TFM.
```cs
<PackageReference Include="System.Net.Http" Version="4.3.4" />
```

As the last [link](https://github.com/Microsoft/dotnet/blob/master/releases/net472/KnownIssues/613745%20-%20Single-name%20references%20are%20removed%20by%20the%20SDK%20when%20targeting%204.7.2.md) mentions there's a workaround and even though the issue doesn't seem to have been updated accordingly a newer version of MSBuild should have a proper fix.

dennisdoomen:
@mu88 can you move this further?

mu88:
@dennisdoomen Sure! I'll do this next week after the holidays

mu88:
@dennisdoomen & @jnyrup : good for you this way?

docs/_pages/releases.md
```diff
@@ -4,15 +4,27 @@ permalink: /releases/
...
 classes: wide
 sidebar:
-  nav: "sidebar"
+nav: "sidebar"
 ---
 
 ## Unreleased
 
+### What's New
+
```
dennisdoomen, line 13: 🤔 Did you forget to add the release notes?

dennisdoomen (approved):

docs/_pages/releases.md
```diff
@@ -4,15 +4,27 @@ permalink: /releases/
 layout: single
 classes: wide
 sidebar:
-  nav: "sidebar"
+nav: "sidebar"
```
jnyrup, line 7: Was this change intentional?
It seems to change the generated html

Tests/FluentAssertions.Specs/Primitives/HttpResponseMessageAssertionSpecs.cs
```diff
@@ -0,0 +1,442 @@
...
+        {
+            // Act
+            Action action = () =>
+                new HttpResponseMessage(HttpStatusCode.Gone).Should().BeSuccessful("because we want to test the failure {0}", "message");
+
+            // Assert
+            action
+                .Should().Throw<XunitException>()
+                .WithMessage("Expected HttpStatusCode to be successful (2xx) because we want to test the failure message, but found HttpStatusCode.Gone {value: 410}.");
+        }
```
jnyrup, line 47: I know we're not consistent about this in the tests, but let's combine these two tests to a single one.
Verifying that `because` and `becauseArgs` are included in the failure message is part of the behavior when failing.

docs/_pages/httpresponsemessages.md
```diff
@@ -0,0 +1,28 @@
...
+permalink: /httpresponsemessages/
+layout: single
+classes: wide
+sidebar:
+nav: "sidebar"
```
jnyrup, line 7: If this should be consistent with other markdown files.
```suggestion
  nav: "sidebar"
```

jnyrup (commented):
Mostly informational, but I stumbled upon this the other day.

https://github.com/balanikas/FluentAssertions.Http
https://www.nuget.org/packages/fluentassertions.http

docs/_pages/releases.md
mu88: Nope, stupid mistake from my side, sorry about that

docs/_pages/httpresponsemessages.md
mu88: Was related to the other mistake

Tests/FluentAssertions.Specs/Primitives/HttpResponseMessageAssertionSpecs.cs
mu88: But combining only these two tests makes it imho less consistent 🙂 wouldn't it be better to combine the other tests as well?

jnyrup: Bad phrasing from my side.
Yes, let's combine each pair of `Throw<XunitException>()` and `Throw<XunitException>().WithMessage()` tests.

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
```diff
@@ -0,0 +1,234 @@
...
+                    .ForCondition(IsServerError())
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith("Expected HttpStatusCode to be server error (5xx){reason}, but found {0}.", Subject.StatusCode);
+            }
+
+            return new AndConstraint<TAssertions>((TAssertions)this);
+        }
+
+        /// <summary>
+        /// Asserts that the value is equal to the specified <paramref name="expected"/> value.
```
jnyrup, line 173:
```suggestion
        /// Asserts that the <see cref="HttpStatusCode"/> is equal to the specified <paramref name="expected"/> value.
```

Src/FluentAssertions/Primitives/HttpResponseMessageAssertions.cs
```diff
@@ -0,0 +1,234 @@
...
+                    .ForCondition(Subject!.StatusCode == expected)
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith("Expected HttpStatusCode to be {0}{reason}, but found {1}.", expected, Subject.StatusCode);
+            }
+
+            return new AndConstraint<TAssertions>((TAssertions)this);
+        }
+
+        /// <summary>
+        /// Asserts that the value is not equal to the specified <paramref name="unexpected"/> value.
```
jnyrup, line 202:
```suggestion
        /// Asserts that the <see cref="HttpStatusCode"/> is not equal to the specified <paramref name="unexpected"/> value.
```

docs/_pages/releases.md
```diff
@@ -11,6 +11,7 @@ sidebar:
 
 ### What's New
 * Adding `ThatAreAsync()` and `ThatAreNotAsync()` for filtering in method assertions - [#1725](https://github.com/fluentassertions/fluentassertions/pull/1725)
+* Adding new assertions for the `HttpStatusCode` of a `HttpResponseMessage` - [#1737](https://github.com/fluentassertions/fluentassertions/pull/1737)
```
jnyrup, line 14:
```suggestion
* Adding new assertions for the `HttpStatusCode` of an `HttpResponseMessage` - [#1737](https://github.com/fluentassertions/fluentassertions/pull/1737)
```

jnyrup (commented):
Just a few cosmetic changes before smashing the merge button.

jnyrup (approved):

dennisdoomen:
Please rebase the PR on the latest `master`.

docs/_pages/releases.md
```diff
@@ -11,6 +11,7 @@ sidebar:
 
 ### What's New
 * Adding `ThatAreStatic()` and `ThatAreNotStatic()` for filtering in method assertions - [#1740](https://github.com/fluentassertions/fluentassertions/pull/1740)
+* Adding new assertions for the `HttpStatusCode` of an `HttpResponseMessage` - [#1737](https://github.com/fluentassertions/fluentassertions/pull/1737)
```
lg2de, line 14: How about adding a hint for the new reference (in net47) to `System.Net.Http`?

dennisdoomen: You mean in the release notes?

lg2de: Yes. For 6.X.Y, there will not be any migration guide, isn't it?

dennisdoomen: No, because there's nothing to migrate.

lg2de: Then this would be the only place to inform the user.
This is not a breaking change but in an environment of restricted nuget access (like my company's package filter) it would be good to know before.

dennisdoomen: Hmm, that's kind of edgy case. Are you saying you have to get approval to take a dependency on Microsoft packages? If so, feel free to propose a PR that captures this.

PR #1740, opened 2021-11-10
Adds ThatAreStatic and ThatAreNotStatic to MethodInfoSelector

chvollm:
## IMPORTANT 

I'm unsure if this change will affect existing users. There is a change in https://github.com/fluentassertions/fluentassertions/blob/586035f3613e46ab84d5d4ec515103e9be065075/Src/FluentAssertions/Types/MethodInfoSelector.cs#L38 to include static methods. Is this rather a breaking change?

### DIFF
--- a/docs/_pages/releases.md
+++ b/docs/_pages/releases.md
@@ -9,9 +9,6 @@ sidebar:
 
 ## Unreleased
 
-**NOTE**
-From this release on `MethodInfoSelector` is now including static methods!
-
 ### What's New
 
 ### Fixes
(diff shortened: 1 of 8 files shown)


jnyrup:
When I opened #645 I didn't consider that `MethodInfoSelector` only works on instance methods.
Since no users have asked for these APIs I do not think we should change the behavior - at least not now.

chvollm:
Fine. So what do you do in this case? Leave the PR open until we get version 7.0.0?

dennisdoomen:
> When I opened #645 I didn't consider that `MethodInfoSelector` only works on instance methods.
> Since no users have asked for these APIs I do not think we should change the behavior - at least not now.

You could rationalize this by treating this as a bug. We never intended to exclude static members.

dennisdoomen:
> You could rationalize this by treating this as a bug. We never intended to exclude static members.

What do you think @jnyrup ?

jnyrup:
I'm fine with the changes, but the release notes should have an explicit note about `MethodInfoSelector` now including static methods.

dennisdoomen:
Agreed.

dennisdoomen:
@chvollm are you willing to follow through?

chvollm:
@dennisdoomen Sure. What am I expected to do? Update the files to resolve the conflicts? Or more? I'm still new to this.

dennisdoomen:
> I'm fine with the changes, but the release notes should have an explicit note about `MethodInfoSelector` now including static methods.

This. And we still have to complete the review.

jnyrup (approved):

chvollm:
@dennisdoomen I've changed the release notes. Tell me what you think.

docs/_pages/releases.md
```diff
@@ -9,8 +9,12 @@ sidebar:
 
 ## Unreleased
 
+**NOTE**
+From this release on `MethodInfoSelector` is now including static methods!
```
dennisdoomen, line 13: Please phrase it as a fix and keep it functional. Nobody knows what `MethodInfoSelector` is. What does it mean to a consumer of this library? And also please add the same PR information.

chvollm: OK, have a look now.

dennisdoomen (approved):

dennisdoomen:
Great. Now the only thing you still need to do is to rebase your changes on `master`

chvollm:
@dennisdoomen  @jnyrup  After a few hickups with the rebase, looking good now.

dennisdoomen:
Great work man. Thank you for this contribution.

dennisdoomen:
You're right. I missed that as well. I've manually updated the release notes for you. So no worries.

PR #1742, opened 2021-11-11
Allow mapping properties and/or fields with different names

### LINKED ISSUE #535: Should().BeEquivalentTo that allows mapping
Suppose that I have an entity class `Order` and a DTO class `OrderDto` like this :
```csharp
class Order
{
  public long Id { get; set; }
  public DateTime Date { get; set; }
  public int State { get; set; }
}

class OrderDto
{
  public long Id { get; set; }
  public DateTime Date { get; set; }
  public StateDto State { get; set; }
}

class StateDto
{
  public int Code { get; set; }
  public String Text { get; set; }
}
```

Now, I have a test where I want to verify that a retrieved DTO is equivalent to existing entity.
```csharp
Order existing;
OrderDto actual;

// Arrange
existing = ....;

// Act
actual = ....;

// Assert
actual.ShouldBeEquivalentTo(existing);
```

This assertion fails because there is no `State.Code` and `State.Text` properties in the `Order` class.
So I changed my assertion to :
```csharp
actual.ShouldBeEquivalentTo(existing, o => o.ExcludingMissingMembers());
```
Now it works but it does not assert that the `State.Code` nested property of the `OrderDto` class is equal to the `State` property of the `Order` class.
So my test is not complete.
What would be the best way to assert the state code of the retrieved DTO ? (I have other similar properties in the DTO class).

dennisdoomen:
Closes #535 

* [x] Release notes
* [x] Documentation examples
* [x] XML docs on `WithMapping` overloads

### DIFF
--- a/Src/FluentAssertions/Common/MemberPath.cs
+++ b/Src/FluentAssertions/Common/MemberPath.cs
@@ -1,6 +1,6 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
-using System.Reflection;
 using FluentAssertions.Equivalency;
 
 namespace FluentAssertions.Common
@@ -23,9 +23,18 @@ namespace FluentAssertions.Common
         }
 
         public MemberPath(Type reflectedType, Type declaringType, string dottedPath)
+            : this(dottedPath)
         {
             this.reflectedType = reflectedType;
             this.declaringType = declaringType;
+        }
+
+        public MemberPath(string dottedPath)
+        {
+            Guard.ThrowIfArgumentIsNullOrEmpty(
+                dottedPath, nameof(dottedPath),
+                "A member path cannot be null or empty");
+
             this.dottedPath = dottedPath;
         }
 
@@ -40,7 +49,7 @@ namespace FluentAssertions.Common
 
         public bool IsSameAs(MemberPath candidate)
         {
-            if ((declaringType == candidate.declaringType) || declaringType.IsAssignableFrom(candidate.reflectedType))
+            if ((declaringType == candidate.declaringType) || declaringType?.IsAssignableFrom(candidate.reflectedType) == true)
             {
                 string[] candidateSegments = candidate.Segments;
 
@@ -66,7 +75,33 @@ namespace FluentAssertions.Common
                    && candidateSegments.SequenceEqual(Segments.Take(candidateSegments.Length));
         }
 
+        /// <summary>
+        /// Determines whether the current path is the same as <paramref name="path"/> when ignoring any specific indexes.
+        /// </summary>
+        public bool IsEquivalentTo(string path)
+        {
+            return path.WithoutSpecificCollectionIndices() == dottedPath.WithoutSpecificCollectionIndices();
+        }
+
+        public bool HasSameParentAs(MemberPath path)
+        {
+            return Segments.Length == path.Segments.Length
+                   && GetParentSegments().SequenceEqual(path.GetParentSegments());
+        }
+
+        private IEnumerable<string> GetParentSegments() => Segments.Take(Segments.Length - 1);
+
+        /// <summary>
+        /// Gets a value indicating whether the current path contains an indexer like `[1]` instead of `[]`.
+        /// </summary>
+        public bool GetContainsSpecificCollectionIndex() => dottedPath.ContainsSpecificCollectionIndex();
+
         private string[] Segments => segments ??= dottedPath.Split(new[] { '.', '[', ']' }, StringSplitOptions.RemoveEmptyEntries);
+
+        /// <summary>
+        /// Returns the name of the member the current path points to without its parent path.
+        /// </summary>
+        public string MemberName => Segments.Last();
 
         public override string ToString()
         {
--- a/Src/FluentAssertions/Common/StringExtensions.cs
+++ b/Src/FluentAssertions/Common/StringExtensions.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Linq;
+using System.Text.RegularExpressions;
 using FluentAssertions.Formatting;
 
 namespace FluentAssertions.Common
@@ -39,6 +40,22 @@ namespace FluentAssertions.Common
             string formattedString = Formatter.ToString(value.Substring(index, length));
 
             return $"{formattedString} (index {index})".EscapePlaceholders();
+        }
+
+        /// <summary>
+        /// Replaces all numeric indices from a path like "property[0].nested" and returns "property[].nested"
+        /// </summary>
+        public static string WithoutSpecificCollectionIndices(this string indexedPath)
+        {
+            return Regex.Replace(indexedPath, @"\[\d+\]", "[]");
+        }
+
+        /// <summary>
+        /// Determines whether a string contains a specific index like `[0]` instead of just `[]`.
+        /// </summary>
+        public static bool ContainsSpecificCollectionIndex(this string indexedPath)
+        {
+            return Regex.IsMatch(indexedPath, @"\[\d+\]");
         }
 
         /// <summary>
--- a/Src/FluentAssertions/Equivalency/INode.cs
+++ b/Src/FluentAssertions/Equivalency/INode.cs
@@ -16,7 +16,10 @@ namespace FluentAssertions.Equivalency
         /// <summary>
         /// Gets the name of this node.
         /// </summary>
-        string Name { get; }
+        /// <example>
+        /// "Property2"
+        /// </example>
+        string Name { get; set; }
 
         /// <summary>
         /// Gets the type of this node.
@@ -26,11 +29,17 @@ namespace FluentAssertions.Equivalency
         /// <summary>
         /// Gets the path from the root object UNTIL the current node, separated by dots or index/key brackets.
         /// </summary>
+        /// <example>
+        /// "Parent[0].Property2"
+        /// </example>
         string Path { get; }
 
         /// <summary>
         /// Gets the full path from the root object up to and including the name of the node.
         /// </summary>
+        /// <example>
+        /// "Parent[0]"
+        /// </example>
         string PathAndName { get; }
 
         /// <summary>
@@ -41,6 +50,9 @@ namespace FluentAssertions.Equivalency
         /// <summary>
         /// Gets the path including the description of the subject.
         /// </summary>
+        /// <example>
+        /// "property subject.Parent[0].Property2"
+        /// </example>
         string Description { get; }
 
         /// <summary>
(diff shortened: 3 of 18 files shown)


Src/FluentAssertions/Equivalency/INode.cs
```diff
@@ -16,7 +16,10 @@ public interface INode
         /// <summary>
         /// Gets the name of this node.
         /// </summary>
-        string Name { get; }
+        /// <example>
+        /// "Property2"
+        /// </example>
+        string Name { get; set; }
```
jnyrup, line 22: This is technically a breaking change.

This example will break
```cs
class MyNode : INode
{
    public string Name { get; }
}
```

Src/FluentAssertions/Common/MemberPath.cs
```diff
@@ -66,8 +74,33 @@ private bool IsChildOf(MemberPath candidate)
...
+        {
+            return path.WithoutSpecificCollectionIndex() == dottedPath.WithoutSpecificCollectionIndex();
+        }
+
+        public bool HasSameParentAs(MemberPath path)
+        {
+            return ParentSegments.SequenceEqual(path.ParentSegments);
+        }
+
+        private string[] ParentSegments => Segments.Take(Segments.Length - 1).ToArray();
```
jnyrup, line 90:
```suggestion
        public bool HasSameParentAs(MemberPath path)
        {
            return Segments.Length == path.Segments.Length
                && GetParentSegments().SequenceEqual(path.GetParentSegments());
        }

        private IEnumerable<string> GetParentSegments() => Segments.Take(Segments.Length - 1);
```

Src/FluentAssertions/Common/StringExtensions.cs
```diff
@@ -41,6 +42,14 @@ public static string IndexedSegmentAt(this string value, int index)
             return $"{formattedString} (index {index})".EscapePlaceholders();
         }
 
+        /// <summary>
+        /// Replaces the numeric index from a path like "property[0].nested" and returns "property[].nested"
+        /// </summary>
+        public static string WithoutSpecificCollectionIndex(this string indexedPath)
```
jnyrup, line 48: This method removes all indices, and not just a single index.
```suggestion
        /// <summary>
        /// Replaces the numeric indices from a path like "property[0].nested" and returns "property[].nested"
        /// </summary>
        public static string WithoutSpecificCollectionIndicesthis string indexedPath)
```

Src/FluentAssertions/Common/MemberPath.cs
```diff
@@ -66,8 +74,33 @@ private bool IsChildOf(MemberPath candidate)
...
 
+        /// <summary>
+        /// Returns the name of the member the current path points to without its parent path.
+        /// </summary>
+        public string MemberName => Segments.Last();
+
+        /// <summary>
+        /// Gets a value indicating whether the current path contains an indexer like `[1]` instead of `[]`.
+        /// </summary>
+        public bool ContainsSpecificCollectionIndex => dottedPath.WithoutSpecificCollectionIndex() != dottedPath;
```
jnyrup, line 102: Make this a method as it does a lot more than just returning a field

Src/FluentAssertions/Equivalency/Matching/MappedMemberMatchingRule.cs
```diff
@@ -0,0 +1,52 @@
...
+        private readonly string subjectMemberName;
+
+        public MappedMemberMatchingRule(string expectationMemberName, string subjectMemberName)
+        {
+            if (Regex.IsMatch(expectationMemberName, @"(\.|\[|\])"))
+            {
+                throw new ArgumentException("The expectation's member name cannot be a nested path");
+            }
+
+            if (Regex.IsMatch(subjectMemberName, $@"(\.|\[|\])"))
```
jnyrup, line 23:
```suggestion
            if (Regex.IsMatch(subjectMemberName, @"[\.\[\]]"))
```

Src/FluentAssertions/Common/StringExtensions.cs
```diff
@@ -41,6 +42,14 @@ public static string IndexedSegmentAt(this string value, int index)
             return $"{formattedString} (index {index})".EscapePlaceholders();
         }
 
+        /// <summary>
```
jnyrup, line 45: Consider adding
```cs
public static bool ContainsSpecificCollectionIndex(this string indexedPath)
    => Regex.IsMatch(indexedPath, @"\[\d+\]", "[]");
```
such that `MemberPath.ContainsSpecificCollectionIndex()` simply becomes
```cs
dottedPath.ContainsSpecificCollectionIndex()
```

Src/FluentAssertions/Equivalency/Matching/MappedPathMatchingRule.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        public IMember Match(IMember expectedMember, object subject, INode parent, IEquivalencyAssertionOptions options)
+        {
+            if (expectationPath.IsEquivalentTo(expectedMember.PathAndName))
+            {
+                var member = MemberFactory.Find(subject, subjectPath.MemberName, expectedMember.Type, parent);
+                if (member is null)
+                {
+                    throw new ArgumentException(
+                        $"Subject of type {subject?.GetType().Name} does not have member {subjectPath.MemberName}");
```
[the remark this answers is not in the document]
dennisdoomen: I thought about that scenario as well, but I discovered that `Match` is not even called in that case. Turns out we even documented this on `IMemberMatchingRule`

Src/FluentAssertions/Equivalency/INode.cs
dennisdoomen: Technically yes, but why would anybody add there own implementation of `INode`. You would have to replace a lot of logic to make that work. Everything is hard-coded to use `Field` or `Property`. And even if somebody would do that, they would override `Node`, which implements this change. So I think it's safe to introduce this change.

Src/FluentAssertions/Common/MemberPath.cs
dennisdoomen: It's `internal`, so I don't see much value in that.

Tests/FluentAssertions.Specs/Equivalency/BasicEquivalencySpecs.cs
```diff
@@ -3462,6 +3462,504 @@ public class OuterWithObject
...
+                });
+
+                // Act / Assert
+                subject.Should()
+                    .BeEquivalentTo(expectation, opt => opt
+                        .WithMapping<ExpectationWithProperty2, SubjectWithProperty1>("Property2", "Property1"));
+            }
+
+            [Fact]
+            public void Can_map_nested_fields_using_a_nested_type_and_field_names()
```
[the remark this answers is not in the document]
dennisdoomen: Good catch

Src/FluentAssertions/Common/MemberPath.cs
```diff
@@ -23,9 +23,18 @@ public MemberPath(IMember member, string parentPath)
         }
 
         public MemberPath(Type reflectedType, Type declaringType, string dottedPath)
+            : this(dottedPath)
         {
             this.reflectedType = reflectedType;
             this.declaringType = declaringType;
+        }
+
+        public MemberPath(string dottedPath)
```
jnyrup, line 32: I haven't dug into whether it's a realistic scenario, but if `IsSameAs(MemberPath candidate)` is called on an instance created from this constructor we could throw an NRE as `declaredType` might be null.

So as an extra precaution
```cs
declaringType?.IsAssignableFrom(candidate.reflectedType) == true
```

Src/FluentAssertions/Equivalency/MemberFactory.cs
```diff
@@ -19,5 +20,17 @@ public static IMember Create(MemberInfo memberInfo, INode parent)
 
             throw new NotSupportedException($"Don't know how to deal with a {memberInfo.MemberType}");
         }
+
+        internal static IMember Find(object target, string memberName, Type preferredMemberType, INode parent)
+        {
+            PropertyInfo property = target.GetType().FindProperty(memberName, preferredMemberType);
```
jnyrup, line 26: Note, I had to double check the implementation to verify that `FindProperty` includes inherited properties.
Have you double-checked that we can map between inherited properties?

Tests/FluentAssertions.Specs/Equivalency/BasicEquivalencySpecs.cs
```diff
@@ -3462,6 +3462,546 @@ public class OuterWithObject
...
+                var expectation = new ClassWithSomeFieldsAndProperties
+                {
+                    Field1 = "John",
+                };
+
+                // Act / Assert
+                subject.Should()
+                    .BeEquivalentTo(expectation, opt => opt
+                        .WithMapping("Field1", "Property1")
+                        .Including(e => e.Field1));
```
jnyrup, line 3664: Why is `.Including(e => e.Field1)` necessary?

Src/FluentAssertions/Common/MemberPath.cs
```diff
@@ -66,8 +75,34 @@ private bool IsChildOf(MemberPath candidate)
...
 
+        /// <summary>
+        /// Returns the name of the member the current path points to without its parent path.
+        /// </summary>
+        public string MemberName => Segments.Last();
+
+        /// <summary>
+        /// Gets a value indicating whether the current path contains an indexer like `[1]` instead of `[]`.
+        /// </summary>
+        public bool ContainsSpecificCollectionIndex => dottedPath.ContainsSpecificCollectionIndex();
```
jnyrup, line 104: Well... I read the internal code and find it useful as a guideline of the cost of invoking a member.
But as it's not public, it's not a hard requirement, only a wish.

Src/FluentAssertions/Common/MemberPath.cs
dennisdoomen: It's not a valid scenario, but I do agree the logic in `MemberPath` has become a bit fishy. I couldn't come up with a nicer solution yet.

Tests/FluentAssertions.Specs/Equivalency/BasicEquivalencySpecs.cs
dennisdoomen: Because I was too lazy to create another type with just one property and one field.

jnyrup: Pheew, I thought it was some requirement to make the field<->property mapping actually work.

Src/FluentAssertions/Common/MemberPath.cs
dennisdoomen: Because it bothers you so much, I've changed it into a method ;-)

Src/FluentAssertions/Equivalency/MemberFactory.cs
dennisdoomen: Another good catch. I've added an additional spec for that.

jnyrup (approved):

dennisdoomen:
Yeah!

PR #1743, opened 2021-11-17
Properly recognized non-class types as internal when using HaveAccessModifier assertion

### LINKED ISSUE #1741: Should().HaveAccessModifier() fails for internal interfaces
### Description

When trying to assert that an interface type has `internal` access, the following assertion message is thrown:

> Expected *expression* to be Internal, but it is InvalidForCSharp.

Assuming the following definition exists in the scope of the test:

```csharp
internal interface ITest {}
```

The following demonstrates the problem:

```csharp
typeof(ITest).Should().HaveAccessModifier(CSharpAccessModifier.Internal);
```

### Expected behavior:

The assertion should succeed.

### Actual behavior:

The following assertion message is thrown:

> Expected typeof(ITest) ITest to be Internal, but it is InvalidForCSharp.

### Versions

* Fluent Assertions version: 6.2.0
* Compiled against .NET 5.0
* Run against .NET 6.0

jonathonchase:
Removes `IsClass` check preventing enums and interfaces from being recognized as internal types when not nested.

Fixes #1741 

## IMPORTANT 

### DIFF
--- a/Tests/FluentAssertions.Specs/Types/TypeAssertionSpecs.HaveAccessModifier.cs
+++ b/Tests/FluentAssertions.Specs/Types/TypeAssertionSpecs.HaveAccessModifier.cs
@@ -59,6 +59,36 @@ namespace FluentAssertions.Specs.Types
         }
 
         [Fact]
+        public void When_asserting_an_internal_member_is_internal_it_succeeds()
+        {
+            // Arrange
+            Type type = typeof(IInternalInterface);
+
+            // Act
+            Action act = () =>
+                type.Should().HaveAccessModifier(CSharpAccessModifier.Internal);
+
+            // Assert
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_asserting_an_internal_member_is_protected_internal_it_throws()
+        {
+            // Arrange
+            Type type = typeof(IInternalInterface);
+
+            // Act
+            Action act = () =>
+                type.Should().HaveAccessModifier(
+                    CSharpAccessModifier.ProtectedInternal, "we want to test the failure {0}", "message");
+
+            // Assert
+            act.Should().Throw<XunitException>()
+                .WithMessage("Expected type IInternalInterface to be ProtectedInternal *failure message*, but it is Internal.");
+        }
+
+        [Fact]
         public void When_asserting_an_internal_type_is_protected_internal_it_throws()
         {
             // Arrange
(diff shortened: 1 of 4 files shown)


jonathonchase:
Closes #1741

jonathonchase:
This is my first time contributing, and I'm not sure if the check list is something I should be filling out, or if that's in place for a maintainer to fill.

Tests/FluentAssertions.Specs/Types/TypeAssertionSpecs.cs
```diff
@@ -136,6 +136,8 @@ public class ClassWithoutMembers { }
 
     public interface IPublicInterface { }
 
+    internal interface IInternalInterface { }
```
dennisdoomen, line 139: 🔧 I suggest to keep this test class with the test cases.

dennisdoomen (approved):

docs/_pages/releases.md
```diff
@@ -14,6 +14,8 @@ sidebar:
 
 ### Fixes
 
+* Resolve an issue preventing HaveAccessModifier from correctly recognizing internal interfaces and enums - [#1741](https://github.com/fluentassertions/fluentassertions/issues/1741)
```
dennisdoomen, line 17: ❌ The release note should point to the PR
🔧 The common convention is to use back-ticks around code identifiers such as `HaveAccessModifier`

dennisdoomen (changes requested):
Could you please give the PR itself also a functional description, for instance, by reusing part of the release notes? Because the title will appear in the auto-generated Github release note.

Tests/FluentAssertions.Specs/Types/TypeAssertionSpecs.cs
jnyrup: The type has to be non-nested to replicate the issue.

Tests/FluentAssertions.Specs/Types/TypeAssertionSpecs.HaveAccessModifier.cs
```diff
@@ -58,6 +58,36 @@ public void When_asserting_an_internal_type_is_internal_it_succeeds()
             act.Should().NotThrow();
         }
 
+        [Fact]
+        public void When_asserting_an_internal_member_is_internal_it_succeeds()
```
jnyrup, line 62: We should also test enums and structs.
You could modify
`When_asserting_an_internal_type_is_internal_it_succeeds` and `When_asserting_an_internal_type_is_protected_internal_it_throws`
from a `[Fact]` to a `[Theory]` covering `class`, `interface`, `enum`, `struct`.

dennisdoomen:
@jonathonchase are you planning to follow-up on @jnyrup's comments?

dennisdoomen:
Replaced by https://github.com/fluentassertions/fluentassertions/pull/1793

Issue #1735, opened 2021-11-08
Modify `HaveCount` report message order: start with items found and expected.


obiwanjacobi:
Please modify the order of the report for failing a `Should().HaveCount()` on a collection, by starting with the actual number of items found and expected before dumping the object's contents.

dennisdoomen:
Some of the other overloads of `HaveCount` don't include the entire collection, so maybe we should do the same here.

iliashkolyar:
@dennisdoomen This can be closed now.

PR #1760, opened 2021-12-26
Change HaveCount assertion message order to state number before dumpi…

iliashkolyar:
…ng object contents

Fixes Issue - https://github.com/fluentassertions/fluentassertions/issues/1735

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -1299,7 +1299,7 @@ namespace FluentAssertions.Collections
                 .Then
                 .Given(subject => subject.Count())
                 .ForCondition(actualCount => actualCount >= expected)
-                .FailWith("but found {0}: {1}.", actualCount => actualCount, actualCount => Subject)
+                .FailWith("but found {0}: {1}.", actualCount => actualCount, _ => Subject)
                 .Then
                 .ClearExpectation();
 
@@ -1331,7 +1331,7 @@ namespace FluentAssertions.Collections
                 .Then
                 .Given(subject => subject.Count())
                 .ForCondition(actualCount => actualCount > expected)
-                .FailWith("but found {0}: {1}.", actualCount => actualCount, actualCount => Subject)
+                .FailWith("but found {0}: {1}.", actualCount => actualCount, _ => Subject)
                 .Then
                 .ClearExpectation();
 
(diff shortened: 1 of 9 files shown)


Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -1299,7 +1299,7 @@ public AndConstraint<TAssertions> HaveCountGreaterThanOrEqualTo(int expected, st
                 .Then
                 .Given(subject => subject.Count())
                 .ForCondition(actualCount => actualCount >= expected)
-                .FailWith("but found {0}.", actualCount => actualCount)
+                .FailWith("but found {0}: {1}.", Subject.Count(), Subject)
```
dennisdoomen, line 1302: 🤔 This will cause another count of the subject. The original `FailWith` method provided that count through a Func providing the `actualCount` that was calculated on the `Given` line. Same for all the other overloads.

dennisdoomen (changes requested):
❌ Please also update the `releases.md` file as explained in the PR template.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
ilia-cy: Thanks for the feedback.
Updated release notes and modified the `FailWith` usage.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -1299,7 +1299,7 @@ public AndConstraint<TAssertions> HaveCountGreaterThanOrEqualTo(int expected, st
                 .Then
                 .Given(subject => subject.Count())
                 .ForCondition(actualCount => actualCount >= expected)
-                .FailWith("but found {0}.", actualCount => actualCount)
+                .FailWith("but found {0}: {1}.", actualCount => actualCount, actualCount => Subject)
```
dennisdoomen, line 1302: 🙃  It would probably use the `_` if I did not need the `actualCount`, but that's just me nitpicking

dennisdoomen (approved):

dennisdoomen (approved):

jnyrup (approved):
Thanks!
As a bonus info, I think this will please a colleague of mine who felt it disruptive to have the entire collection printed in the middle of a sentence.

Issue #1705, opened 2021-10-11
Calling Equals on Assertion object does not fail so that the test suggests correctness


akacoding:
Calling
```cs
1.Should().Equals(2); // simple example
```

does not assert nor fail.
It is clear that this is the natural behavior because `Equals` does not belong to FluentAssertions but to the base object class.

But some users ran already in this kind of problem, see [here on StackOverflow](https://stackoverflow.com/q/37846463/8051589) for example.

**Proposal:**
The `Equals` methods of all Assertion objects should throw or assert false with an appropiate message.
It would no longer be possible to write those wrong "assertion" statements with `Equals` because the test would fail.

Issue #1632, opened 2021-08-03
Non-generic overload for WithInnerExceptionExactly


lonix1:
## Background

I can do this:
````cs
action.Should().Throw<TargetInvocationException>()
  .WithInnerExceptionExactly<TException>()         // <--- known at compile time
  .Where(x => x.Message == message)
  .And.ParamName.Should().Be(paramName);
````

## Problem

The above assumes I know `TException` at compile time. But what if I don't?

I can rewrite it using [a trick](https://github.com/fluentassertions/fluentassertions/issues/1629#issuecomment-886243261) I learnt from @dennisdoomen, like this:
````cs
action.Should().Throw<TargetInvocationException>()
  .WithInnerException<Exception>().Which.Should().BeOfType(expectedType)
  //.Where(x => x.Message == message)
  //.And.ParamName.Should().Be(paramName);
````

But then I can't chain `.Where` etc.

## Feature Request

I read the [exception docs](https://fluentassertions.com/exceptions/) but couldn't find a workaround.

Please consider adding an overload that take an instance:
````cs
.WithInnerExceptionExactly(expectedType)
````

## Versions

* Which version of Fluent Assertions are you using? 5.10.3
* Which .NET runtime and version are you targeting? 5.0.301

dennisdoomen:
Sure. If anybody is willing to contribute this, we would welcome it.

PR #1769, opened 2022-01-07
Non-generic overload for WithInnerExceptionExactly

karenfarnes:
This fixes [#1705](https://github.com/fluentassertions/fluentassertions/issues/1632)

## IMPORTANT 


Src/FluentAssertions/Specialized/ExceptionAssertions.cs
```diff
@@ -191,6 +192,54 @@ public ExceptionAssertions(IEnumerable<TException> exceptions)
...
+
+            return expectedInnerExceptions;
+        }
+
+        private IEnumerable<Exception> GetInnerExceptionExactly(Type innerException, string because = null,
+            params object[] becauseArgs)
+        {
+            Guard.ThrowIfArgumentIsNull(innerException, nameof(innerException));
+
+            WithInnerException(innerException, because, becauseArgs);
```
dennisdoomen, line 229: 🔧 Instead of creating a seemingly recursive call (which it didn't after digging through the code), I would just call the `GetExpectedInnerExceptions` method (and reverse their order in the file for readability).

Src/FluentAssertions/Specialized/ExceptionAssertions.cs
```diff
@@ -191,6 +192,54 @@ public ExceptionAssertions(IEnumerable<TException> exceptions)
             return this;
         }
 
+        private IEnumerable<Exception> GetExpectedInnerExceptions(Type innerException, string because = null,
```
dennisdoomen, line 195: 🤔 I think we should rename this method to emphasize that it's primary function is to assert something. Same for the other `Get` method.

dennisdoomen (changes requested):
❌ Also update the `releases.md` file 
🔧 Also consider updating the `exceptions.md` file

Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
```diff
@@ -832,6 +832,21 @@ public async Task When_async_method_throws_the_expected_inner_exception_it_shoul
...
+            // Arrange
+            Func<Task> task = () => Throw.Async(new AggregateException(new InvalidOperationException()));
+
+            // Act
+            Func<Task> action = () => task
+                .Should().ThrowAsync<AggregateException>()
+                .WithInnerException(typeof(InvalidOperationException));
+
+            // Assert
+            await action.Should().NotThrowAsync();
```
jnyrup, line 847: I'm not seeing a failing test for `WithInnerException`.
Am I just missing it?

jnyrup (changes requested):

Src/FluentAssertions/Specialized/ExceptionAssertions.cs
```diff
@@ -191,6 +191,54 @@ public ExceptionAssertions(IEnumerable<TException> exceptions)
...
+                .FailWith("no exception was thrown.")
+                .Then
+                .ForCondition(Subject.Any(e => e.InnerException is not null))
+                .FailWith("the thrown exception has no inner exception.")
+                .Then
+                .ClearExpectation();
+
+            Exception[] expectedInnerExceptions = Subject
+                .Select(e => e.InnerException)
+                .Where(e => e?.GetType() == innerException)
```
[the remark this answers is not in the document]
dennisdoomen: Hmm, I missed that and thought this code merely moved.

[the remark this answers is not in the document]
karenfarnes: It is one of the few changes, so that it would work for both call types. Unfortunately I did not fully grasp the full extent of the OfType<> method. Will fix it!

Src/FluentAssertions/Specialized/ExceptionAssertions.cs
```diff
@@ -191,6 +191,54 @@ public ExceptionAssertions(IEnumerable<TException> exceptions)
...
+                .FailWith("no exception was thrown.")
+                .Then
+                .ForCondition(Subject.Any(e => e.InnerException is not null))
+                .FailWith("the thrown exception has no inner exception.")
+                .Then
+                .ClearExpectation();
+
+            Exception[] expectedInnerExceptions = Subject
+                .Select(e => e.InnerException)
+                .Where(e => e != null && (e.GetType().IsSubclassOf(innerException) || e.GetType() == innerException))
```
dennisdoomen, line 231: 🔧 You could combine the two conditions with `IsSameOrInherits`

dennisdoomen (approved):
You still need to rebase the PR on the latest `master`

Src/FluentAssertions/Specialized/ExceptionAssertions.cs
jnyrup: We need to use `IsAssignableFrom`, otherwise it won't work for interfaces.

dennisdoomen: That's what it uses internally

jnyrup: Right 👍 
I was mislead by `Inherits` since the terminology often is that base classes are inherited, while interfaces are implemented.

Src/FluentAssertions/Specialized/ExceptionAssertions.cs
```diff
@@ -148,19 +143,24 @@ public ExceptionAssertions(IEnumerable<TException> exceptions)
...
+        /// Asserts that the thrown exception contains an inner exception of the exact type <param name="innerException" /> (and not a derived exception type).
+        /// </summary>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public virtual ExceptionAssertions<Exception> WithInnerExceptionExactly(Type innerException, string because = null,
```
jnyrup, line 160: Remove `virtual` here as well.
As a rule-of-thumb, when we review PRs we only point out a single occurrence of something that should be changed.

jnyrup (changes requested):
As a (probably) last thing, you need to go through the check boxes if the PR description.
(hint, you're missing out on one of them, which the CI pipeline will also complain about.)

Src/FluentAssertions/Specialized/ExceptionAssertions.cs
karenfarnes: Yes, thank you for the Patience. It is my first time contributing to anything.

jnyrup:
Hmm... something went wrong with your last commit.
The `.verified.txt` should not have been deleted, but should be updated with the new functions you've added to the public API.

jnyrup (approved):

jnyrup:
@karenfarnes Congratulations with your first contribution 🎉

PR #1772, opened 2022-01-11
override Identifier for BufferedStream

amosonn:
## IMPORTANT 

* [x] ~~If the contribution adds a feature or fixes a bug, please update [**the release notes**](https://github.com/fluentassertions/fluentassertions/tree/master/docs), which are published on the [website](https://fluentassertions.com/releases).~~
* [x] ~~If the contribution changes the public API the changes needs to be included by running `AcceptApiChanges.ps1`/`AcceptApiChanges.sh`.~~
* [x] ~~If the contribution affects [the documentation](https://github.com/fluentassertions/fluentassertions/tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).~~


dennisdoomen (approved):

jnyrup (approved):

amosonn:
Yeah, I didn't think this would affect the API (since that class inherits the property anyway). I will run the AcceptApiChanges.sh and push again.

Src/FluentAssertions/Streams/BufferedStreamAssertions.cs
```diff
@@ -27,6 +27,8 @@ public BufferedStreamAssertions(BufferedStream stream)
         {
         }
 
+        protected override string Identifier => "bufferedstream";
```
jn-foreflight, line 30: I just had an extra look at how we do this elsewhere and as `Identifier` is included in a sentence it should be
```suggestion
        protected override string Identifier => "buffered stream";
```

jnyrup, line 30: I just had an extra look at how we do this elsewhere and as `Identifier` is included in a sentence it should be
```suggestion
        protected override string Identifier => "buffered stream";
```

amosonn: Changed this.

jnyrup:
Thanks for contributing this fix 👍

PR #1776, opened 2022-01-15
Fix determining caller identity for NET Native

jnyrup:
I managed to make a test reproducing the issue described in [#1149](https://github.com/fluentassertions/fluentassertions/issues/1149#issuecomment-925013141).
When the UWP application is build with `<UseDotNetNativeToolchain>` `StackFrame.GetMethod()` returns `null`, which led to a `NullReferenceException`.

I've excluded building this application from the solution and NUKE build.
I didn't get it to work 100% and the amount of quirks was disgusting.


Src/FluentAssertions/CallerIdentifier.cs
```diff
@@ -130,29 +130,29 @@ internal static bool OnlyOneFluentAssertionScopeOnCallStack()
...
         private static bool IsCustomAssertion(StackFrame frame)
         {
-            return frame.GetMethod().IsDecoratedWithOrInherit<CustomAssertionAttribute>();
+            return frame.GetMethod()?.IsDecoratedWithOrInherit<CustomAssertionAttribute>() == true;
         }
 
         private static bool IsDynamic(StackFrame frame)
         {
-            return frame.GetMethod().DeclaringType is null;
+            return frame.GetMethod() is { DeclaringType: null };
```
dennisdoomen, line 145: I like the use of this new language feature.

dennisdoomen (approved):

jnyrup:
@nirmal4g fyi, this fix will be part of the next release (v6.4.0)

jnyrup:
We don't have a private nuget feed, but you can download the newest nupkg built from master here:
https://github.com/fluentassertions/fluentassertions/actions/runs/1709514605

PR #1777, opened 2022-01-16
Adjust Identifier properties

jnyrup:
Inspired by #1772 I looked over the other `Identifier` properties to align them better.

### DIFF
--- a/Src/FluentAssertions/Types/MethodBaseAssertions.cs
+++ b/Src/FluentAssertions/Types/MethodBaseAssertions.cs
@@ -93,8 +93,6 @@ namespace FluentAssertions.Types
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
-        protected override string Identifier => "methodBase";
-
         internal static string GetParameterString(MethodBase methodBase)
         {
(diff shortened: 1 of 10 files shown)


Src/FluentAssertions/Types/MethodBaseAssertions.cs
```diff
@@ -93,8 +93,6 @@ public AndConstraint<TAssertions> NotHaveAccessModifier(CSharpAccessModifier acc
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
-        protected override string Identifier => "methodBase";
```
jnyrup, line 96: `MethodBaseAssertions` is `abstract` and its two derivatives `ConstructorInfoAssertions` and `MethodInfoAssertions` overrides `Identifier` with `"constructor"` and `"method"`, respectively.

dennisdoomen (approved):
Maybe we should also improve the documentation on the base-class property to emphasize it should be a user-friendly name.

jnyrup:
> Maybe we should also improve the documentation on the base-class property to emphasize it should be a user-friendly name.

Like this?
image

dennisdoomen:
> Like this?

Exactly

Issue #500, opened 2016-11-08
ShouldBeEquivalentTo: exclude a property of all of a list items


drauch:
Current state of the art:

> Maybe far-fetched, but you may even decide to exclude a member on a particular nested object by its index.

```cs
 orderDto.ShouldBeEquivalentTo(order, options => options.Excluding(o => o.Products[1].Status)); 
```
It would be nice to be able to exclude `Status` for all items in the list. One could use a special index instead for that, e.g., something along the lines of:

```cs
orderDto.ShouldBeEquivalentTo(order, options => options.Excluding(o => o.Products[Something.AllItems].Status));
```

This would be much better than the current workaround (using `PropertyPath`), as it is refactoring-safe.

Best regards,
D.R.

dennisdoomen:
It's a nice idea, but probably a lot of work to get done.

Jenan:
Hello,
I have tried to compare the lists but I can't find the property called `PropertyPath`.

I have used the version `4.19.4`.

How can I fix it? 
Thx.

dennisdoomen:
It's on the `ISubjectInfo` object that is passed as part of the overload of `Including` and `Excluding` that take `Func<ISubjectInfo, bool>`

Jenan:
I have tried this snippet for testing:
```
var clientDtos = new List<ClientDto>();
var clientDtosToCompare = new List<ClientDto>();

clientDtos.ShouldBeEquivalentTo(clientDtosToCompare, option => option.Excluding(x=> x.PropertyPath == "ClientName"));
```
But `PropertyPath` doesn't exist - can you please correct my code?

jnyrup:
`PropertyPath` was removed from `ISubjectInfo` in [v4](https://github.com/fluentassertions/fluentassertions/releases/tag/4.0.0).

Use `SelectedMemberPath` instead
```csharp
clientDtos.ShouldBeEquivalentTo(clientDtosToCompare, option => option.Excluding(x => x.SelectedMemberPath.EndsWith("ClientName")));
```

Jenan:
@jnyrup You are absolutely right, that works very well, thank you!

davisnw:
The suggested workaround of just checking "SelectedMemberPath" is too broad.  Sometimes I have the same property name on different classes, but only want to exclude it on one of the classes.  This is problematic when dealing with collections.

[...]

PR #1782, opened 2022-01-20
Add `For`/`Exclude` to allow exclusion of members inside a collection

### LINKED ISSUE #1771: Proposal to exclude properties on a nested list
### Description

I propose a new feature to exclude certain properties of an object in a nested list. (see: https://stackoverflow.com/questions/22142576/how-to-use-exclude-in-fluentassertions-for-property-in-collection).

Currently, it is only possible to exclude certain properties of an entity at a specified index or to use string-based property path matching. The first one does not offer the required functionality, the latter one is complicated, lacks IntelliSense, and is not straightforward.

### Example

Given the following objects, I want to exclude the `Text` property from the nested collection.

```C#
// Arrange
var subject = new
{
    Level = new
    {
        Collection = new[]
        {
            new { Number = 1, Text = "Actual" },
            new { Number = 2, Text = "Actual" }
        }
    }
};

var expected = new
{
    Level = new
    {
        Collection = new[]
        {
            new { Number = 1, Text = "Expected" },
            new { Number = 2, Text = "Expected" }
        }
    }
};
```

### Proposals
```c#
const int MATCH_ALL = 0;

// Option 1: By using a magic constant (Excluding would have to detect if MATCH_ALL was used)
Action option1 = () => subject.Should().BeEquivalentTo(expected,
    options => options.Excluding(x => x.Level.Collection[MATCH_ALL].Text));

// Option 2: By using the .Select function on IEnumerable
Action option2 = () => subject.Should().BeEquivalentTo(expected,
    options => options.Excluding(x => x.Level.Collection.Select(l => l.Text)));

// Option 3: By using the Excluding function again
Action option3 = () => subject.Should().BeEquivalentTo(expected,
    options => options.Excluding(x => x.Level.Collection.Excluding(l => l.Text)));
```

I am also interested in implementing this 🙂

whymatter:
Closes #1771 

## IMPORTANT 


whymatter:
@dennisdoomen I still have to do the release notes, however, you can already have a look at the `ThenExcluding` implementation. Unit tests are all green.

Src/FluentAssertions/Common/MemberPathSegmentEqualityComparer.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Text.RegularExpressions;
+
+namespace FluentAssertions.Common
+{
+    /// <summary>
+    /// Compares two segments of a <see cref="MemberPath"/>.
+    /// Sets the <see cref="AnyIndexQualifier"/> equal with any numeric index qualifier.
+    /// All other comparisons are default string equality.
+    /// </summary>
+    internal class MemberPathSegmentEqualityComparer : IEqualityComparer<string>
```
dennisdoomen, line 11: 🤔 When you check the latest `master` you'll see that we also started to allow `[]` as a valid separator for different purposes. I think it's worth combining that with this comparer.
🔧 I think it would be useful to have some examples in a `remarks` section. 
🔧 Maybe name the comparer to what it functionally does, e.g. something naming "indices"?

Src/FluentAssertions/Common/MemberPath.cs
```diff
@@ -55,18 +57,26 @@ private bool IsParentOf(MemberPath candidate)
...
         {
             string[] candidateSegments = candidate.Segments;
 
             return candidateSegments.Length < Segments.Length
-                   && candidateSegments.SequenceEqual(Segments.Take(candidateSegments.Length));
+                   && candidateSegments.SequenceEqual(Segments.Take(candidateSegments.Length),
+                       MemberPathSegmentEqualityComparer);
+        }
+
+        public MemberPath Extend(MemberPath nextPath, string separator = ".")
```
dennisdoomen, line 72: 🔧 Isn't `Combine` a better name? That's what .NET uses for combining file paths.

Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
```diff
@@ -38,6 +38,17 @@ public EquivalencyAssertionOptions<TExpectation> Excluding(Expression<Func<TExpe
             return this;
         }
 
+        /// <summary>
+        /// Excludes the specified (nested) member from the structural equality check.
+        /// Allows to navigate deeper by using the returned <see cref="EquivalencyAssertionOptionsBuilder{TExpectation, TNext}"/>.
+        /// </summary>
+        public EquivalencyAssertionOptionsBuilder<TExpectation, TNext> Excluding<TNext>(Expression<Func<TExpectation, IEnumerable<TNext>>> expression)
```
dennisdoomen, line 45: 🤔 Why do we need another overload?

Src/FluentAssertions/Equivalency/EquivalencyAssertionOptionsBuilder.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using FluentAssertions.Common;
+using FluentAssertions.Equivalency.Selection;
+
+namespace FluentAssertions.Equivalency
+{
+    public class EquivalencyAssertionOptionsBuilder<TExpectation, TCurrent> : EquivalencyAssertionOptions<TExpectation>
```
dennisdoomen, line 9: 🔧 I think something like `NestedExclusionOptionBuilder` would better convey the purpose of this class.

Src/FluentAssertions/Equivalency/Selection/ExcludeMemberByPathSelectionRule.cs
```diff
@@ -23,6 +23,12 @@ public ExcludeMemberByPathSelectionRule(MemberPath pathToExclude)
                 memberToExclude.IsSameAs(new MemberPath(member, parentPath)));
         }
 
+        public void ExtendPath(MemberPath nextPath)
+        {
+            memberToExclude = memberToExclude.Extend(nextPath, "[*]");
```
dennisdoomen, line 28: 🤔 To align with the new `WithMapping` options, I think we should use `[]` for that.

Src/FluentAssertions/Equivalency/Selection/SelectMemberByPathSelectionRule.cs
```diff
@@ -16,6 +16,11 @@ protected SelectMemberByPathSelectionRule(string selectedPath)
 
         public virtual bool IncludesMembers => false;
 
+        protected string SelectedPath
```
dennisdoomen, line 19: 🤔 We prefer to use a method instead of a write-only property

Tests/FluentAssertions.Specs/Equivalency/CollectionEquivalencySpecs.cs
```diff
@@ -715,6 +715,326 @@ public void When_a_deeply_nested_property_of_a_collection_with_an_invalid_value_
             act.Should().NotThrow();
         }
 
+        [Fact]
```
dennisdoomen, line 718: 🔧 We've [started to wrap](https://github.com/fluentassertions/fluentassertions/discussions/1551) related tests in a nested class. This also helps keep the test names shorter and functional. In this case, wrapping all tests dealing with `Excluding` and `ThenExcluding` would make total sense to me.
❌ Please avoid using the literal name of a method in test names. This makes the test name very difficult to read. Also applies to the other tests.

Tests/FluentAssertions.Specs/Equivalency/CollectionEquivalencySpecs.cs
```diff
@@ -715,6 +715,326 @@ public void When_a_deeply_nested_property_of_a_collection_with_an_invalid_value_
...
+        public void When_attribute_in_collection_is_excluded_with_then_excluding_it_should_not_throw()
+        {
+            // Arrange
+            var subject = new
+            {
+                Text = "Text",
```
dennisdoomen, line 724: 🔧 I would remove this property. It's not important for this test.

Tests/FluentAssertions.Specs/Equivalency/CollectionEquivalencySpecs.cs
```diff
@@ -715,6 +715,326 @@ public void When_a_deeply_nested_property_of_a_collection_with_an_invalid_value_
...
+                    }
+                }
+            };
+
+            Action act = () => subject.Should().BeEquivalentTo(expected,
+                options => options
+                    .Excluding(x => x.Level.Collection).ThenExcluding(x => x.Number)
+                );
+
+            act.Should().NotThrow();
```
dennisdoomen, line 769: 🔧 If a test is not supposed to throw, we've started to drop the extra `NotThrow` and `Action` logic.

Tests/FluentAssertions.Specs/Equivalency/CollectionEquivalencySpecs.cs
```diff
@@ -715,6 +715,326 @@ public void When_a_deeply_nested_property_of_a_collection_with_an_invalid_value_
...
+                        new
+                        {
+                            Number = 3,
+                            Text = "Actual"
+                        }
+                    }
+                }
+            };
+
+            Action act = () => subject.Should().BeEquivalentTo(expected,
```
dennisdoomen, line 764: 🙃 (nitpicking) Where's the `// Act/Assert`?

Tests/FluentAssertions.Specs/Equivalency/CollectionEquivalencySpecs.cs
```diff
@@ -715,6 +715,326 @@ public void When_a_deeply_nested_property_of_a_collection_with_an_invalid_value_
...
+                options => options
+                    .Excluding(x => x.Level.Collection).ThenExcluding(x => x.NextCollection)
+            );
+
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_attribute_in_collection_in_collection_is_excluded_with_then_excluding_it_should_not_throw()
```
dennisdoomen, line 851: 🙃 "Attribute"?

Tests/FluentAssertions.Specs/Equivalency/CollectionEquivalencySpecs.cs
```diff
@@ -715,6 +715,326 @@ public void When_a_deeply_nested_property_of_a_collection_with_an_invalid_value_
...
+            Action act = () => subject.Should().BeEquivalentTo(expected,
+                options => options
+                    .Excluding(x => x.Level.Collection).ThenExcluding(x => x.NextCollection).ThenExcluding(x => x.Text)
+            );
+
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_attribute_in_collection_is_excluded_with_then_excluding_followed_by_root_level_exclude_it_should_not_throw()
```
dennisdoomen, line 929: 🔧 Suggestion: "A_nested_exclusion_can_be_followed_by_a_root_level_exclusion"

dennisdoomen (commented):
I recommend waiting with the rebase until https://github.com/fluentassertions/fluentassertions/pull/1789 is merged because it requires moving the tests to a new file.

whymatter:
Hi, sorry I was offline for a few days, looking forward to fix the issues next week :)

Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
whymatter: I can not use the existing signature, because it does not have the `TNext` generic parameter. I can also not add the `TNext` to the existing signature, because I need to have `IEnumerable<TNext>` in the return parameter of the `Expression`.

Tests/FluentAssertions.Specs/Equivalency/CollectionEquivalencySpecs.cs
whymatter: You mean I should use "property"? If yes, changed it :)

Tests/FluentAssertions.Specs/Equivalency/CollectionEquivalencySpecs.cs
whymatter: Nice, also changed the following tests name to "A_nested_exclusion_can_be_**preceded**_by_a_root_level_exclusion"

Src/FluentAssertions/Common/MemberPath.cs
whymatter: ✅ Renamed to `Combine`

Src/FluentAssertions/Equivalency/EquivalencyAssertionOptionsBuilder.cs
whymatter: ✅ Renamed to `NestedExclusionOptionBuilder`

Src/FluentAssertions/Equivalency/Selection/SelectMemberByPathSelectionRule.cs
whymatter: ✅ New method `SetSelectedPath`

Tests/FluentAssertions.Specs/Equivalency/CollectionEquivalencySpecs.cs
whymatter: ✅ Extracted `ThenExcluding` tests to a nested class

Tests/FluentAssertions.Specs/Equivalency/CollectionEquivalencySpecs.cs
whymatter: ✅ Removed property

Tests/FluentAssertions.Specs/Equivalency/CollectionEquivalencySpecs.cs
whymatter: ✅ Removed the `NotThrow` and `Action`

Tests/FluentAssertions.Specs/Equivalency/CollectionEquivalencySpecs.cs
whymatter: ✅ Added `// Act/Assert`

Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
dennisdoomen: Clear!

Src/FluentAssertions/Common/MemberPathSegmentEqualityComparer.cs
whymatter: ❓Hm. If I use `[]` as valid `dottedPath` we can no longer use `SequenceEqual` to compare two `dottedPaths` because we are missing an element in the sequence. I. e. we get different path lengths for equivalent paths. I don't like that.

Example: 
```c#
"a[].b" // split: [a, b]
"a[1].b" // split: [a, 1, b]
```

Also with the introduction of `[]` the following is true and I belive this is a problem:
```c#
var path1 = new MemberPath("Parent.Property2.Test");
var path2 = new MemberPath("Parent.Property2");

path1.IsParentOrChildOf(path2).Should().BeTrue(); // correct

var path3 = new MemberPath("Parent[].Property2.Test");
var path4 = new MemberPath("Parent.Property2");

path3.IsParentOrChildOf(path4).Should().BeTrue(); // wrong: the paths are not parent / child
```

Src/FluentAssertions/Equivalency/Selection/ExcludeMemberByPathSelectionRule.cs
whymatter: Please see this comment https://github.com/fluentassertions/fluentassertions/pull/1782#discussion_r798841929

whymatter:
Updated the documentation and release notes, ran AcceptApiChanges.ps1 :)

Src/FluentAssertions/Common/MemberPathSegmentEqualityComparer.cs
dennisdoomen: Sorry for the late response, but needed to find some thing to think about the consequences. 

Is this a pure internal thing? In other words, this `*` will never leak into the consumer API? If so, then it's fine to keep it. If not, we need to change it.

Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
```diff
@@ -40,6 +39,17 @@ public EquivalencyAssertionOptions<TExpectation> Excluding(Expression<Func<TExpe
             return this;
         }
 
+        /// <summary>
+        /// Excludes the specified (nested) member from the structural equality check.
+        /// Allows to navigate deeper by using the returned <see cref="NestedExclusionOptionBuilder{TExpectation,TCurrent}"/>.
```
dennisdoomen, line 44: 🔧 I think it would be better to refer to the `ThenExcluding` member instead of the type. The former is what consumers will be using. The latter is an internal detail.

Src/FluentAssertions/Common/MemberPathSegmentEqualityComparer.cs
whymatter: Hm, it is visible in the output when a test failed.
```
With configuration:
- Exclude member Level.Collection[*]Number
```

A consumer could do something like this
```c#
subject.Should().BeEquivalentTo(expected,
    options =>
    {
        options = options
            .Excluding(x => x.Level.Collection)
            .ThenExcluding(x => x.Number);

        var selectionRules = ((IEquivalencyAssertionOptions)options).SelectionRules;
        selectionRules.Last().ToString(); // --> "x.Level.Collection[*]Number"

        return options;
    });
```

Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
whymatter: ✅ done

Src/FluentAssertions/Common/MemberPathSegmentEqualityComparer.cs
dennisdoomen: Then we need to change that, since it is inconsistent with the new `WithMapping` construct.

Src/FluentAssertions/Equivalency/Selection/ExcludeMemberByPathSelectionRule.cs
whymatter: Done 🙂

Src/FluentAssertions/Common/MemberPathSegmentEqualityComparer.cs
whymatter: Okay so I removed the `*` from the `dottedPath`. I add it back later when splitting the `dottedPath` into segments. The segments are private to the `MemberPath` and also not returned by any function.

A failed assertion returns something like this:
```
With configuration:
- Exclude member Collection[]Number
```

dennisdoomen (approved):

jnyrup (commented):
I tried to play a bit with this, but encountered something that seems a bit non-intuitive to me.

This passes as `Collection` is excluded from the equivalence
```cs
var subject = new { Member = "1", Collection = new[] { new { First = "A", Second = "B" } } };
var expected = new { Member = "1", Collection = new[] { new { First = "C", Second = "D" } } };

subject.Should().BeEquivalentTo(expected, opt => opt
    .Excluding(e => e.Collection));
```

Adding `.ThenExcluding(e => e.Second)` makes it fail as now only `Collection.Second` is excluded.
```cs
var subject = new { Member = "1", Collection = new[] { new { First = "A", Second = "B" } } };
var expected = new { Member = "1", Collection = new[] { new { First = "C", Second = "D" } } };

subject.Should().BeEquivalentTo(expected, opt => opt
    .Excluding(e => e.Collection).ThenExcluding(e => e.Second));
```

So the equivalency engine in fact _includes_ `Collection` _except_ its property `Second`.

I don't think the analogy of `Include` and `ThenInclude` from EF is completely translatable and invertible to `Exclude` and `ThenExclude`.
With `ThenInclude` you include something extra in addition to what's being included by `Include`.
If you have excluded something with `Exclude`, there's nothing left for `ThenExclude` to exclude.


I think e.g. combining `Including` with `ThenExcluding` comes closer to what's happening.
```cs
subject.Should().BeEquivalentTo(expected, opt => opt
    .Including(e => e.Collection).ThenExcluding(e => e.Second));
```

whymatter:
I see that `Including` followed by `ThenExcluding` is also an interesting case and I think we need it as well.
However, my intention was to help people using `Excluding` (like this guy #500).
If one has existing assertions that use `Excluding` and want to add attributes in nested collections and exclude them, it would be a pain to rewrite all test to use `Including`.

Src/FluentAssertions/Equivalency/NestedExclusionOptionBuilder.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        internal NestedExclusionOptionBuilder(EquivalencyAssertionOptions<TExpectation> equivalencyAssertionOptions,
+            ExcludeMemberByPathSelectionRule currentPathSelectionRule)
+            : base(equivalencyAssertionOptions)
+        {
+            this.currentPathSelectionRule = currentPathSelectionRule;
+        }
+
+        /// <summary>
+        /// Selects a property to use. This exists the <see cref="ThenExcluding"/> chain.
```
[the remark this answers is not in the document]
whymatter: Actually, it should be "exits". Changed it to "ends" 😄

Src/FluentAssertions/Common/MemberPathSegmentEqualityComparer.cs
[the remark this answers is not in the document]
whymatter: Tried it but `MemberPathSegmentEqualityComparer.Default` returns a default `string` comparer and not the `MemberPathSegmentEqualityComparer`. Am I missing something?

dennisdoomen:
> Adding .ThenExcluding(e => e.Second) makes it fail as now only Collection.Second is excluded.

> var subject = new { Member = "1", Collection = new[] { new { First = "A", Second = "B" } } };
> var expected = new { Member = "1", Collection = new[] { new { First = "C", Second = "D" } } };

> subject.Should().BeEquivalentTo(expected, opt => opt
>    .Excluding(e => e.Collection).ThenExcluding(e => e.Second));

I thought the entire purpose of this PR was to have a more type-safe way of doing this:

> subject.Should().BeEquivalentTo(expected, opt => opt
>    .Excluding(e => e.Path == "Collection[].Second"));

And the below statement will actually exclude the other property (`Member`) and then include only the `First` property.

> subject.Should().BeEquivalentTo(expected, opt => opt
    .Including(e => e.Collection).ThenExcluding(e => e.Second));

jnyrup:
> I thought the entire purpose of this PR was to have a more type-safe way of doing this:
>>    subject.Should().BeEquivalentTo(expected, opt => opt
    .Excluding(e => e.Path == "Collection[].Second"));

It certainly is (and I really like this feature), I'm just brain storming on the syntax

> And the below statement will actually exclude the other property (Member) and then include only the First property.
>>    subject.Should().BeEquivalentTo(expected, opt => opt
    .Including(e => e.Collection).ThenExcluding(e => e.Second));

Right, forgot how `Including` works.

--

Looking at the linked issues, one [proposes](https://github.com/fluentassertions/fluentassertions/issues/500#issuecomment-598032265) this syntax
```cs
actual.Should().BeEquivalentTo(expected, options => options
    .ForCollection(x => x.Items, o => o.Excluding(i => i.Id)));
```

It could also be adjusted to
```cs
actual.Should().BeEquivalentTo(expected, options => options
    .ForCollection(x => x.Items).Exclude(i => i.Id)));
```

What do you think about that syntax?

dennisdoomen:
> What do you think about that syntax?

I think the PR is fine as it is.

jnyrup:
@dennisdoomen Do you see my point about it being a bit odd that 
```cs
subject.Should().BeEquivalentTo(expected, opt => opt
    .Excluding(e => e.Collection).ThenExcluding(e => e.Second));
```
excludes _less_ than
```cs
subject.Should().BeEquivalentTo(expected, opt => opt
    .Excluding(e => e.Collection));
```

dennisdoomen:
> Do you see my point about it being a bit odd that

Funny. Now I suddenly do. I never read it the way you read it. But now that I get it, I can't unsee it ;-)

However, I haven't settled my mind on a better option just yet.

whymatter:
One of the proposals was the following:

```c#
subject.Should().BeEquivalentTo(expected, opt => opt
    .Excluding(e => e.Collection.Select(e => e.Second));
```

dennisdoomen:
subject.Should().BeEquivalentTo(expected, opt => opt
        .Excluding(e => e.Collection.Select(e => e.Second));

Maybe that's a better option. Or

 subject.Should().BeEquivalentTo(expected, opt => opt
    .Excluding(e => e.Collection,  c => c.Second));

whymatter:
@jnyrup @dennisdoomen So which option do you guys prefer?
I prefer the `.Excluding(e => e.Collection.Select(e => e.Second)`, however, I am fine with `.Excluding(e => e.Collection,  c => c.Second)` as well.

jnyrup:
I haven't been giving this PR much thinking the past days, other PRs have spend all (and more) of my time.

Wouldn't 
```cs
subject.Should().BeEquivalentTo(expected, opt => opt
    .Excluding(e => e.Collection.Select(e => e.Second));
```
conflict with the overload of `Excluding` used for current property exclusion?
E.g.
```cs
subject.Should().BeEquivalentTo(expected, opt => opt
    .Excluding(e => e.Collection);
```

Continuing from my previous comment, we could also leave out the "Collection" from "ForCollection".
```cs
subject.Should().BeEquivalentTo(expected, opt => opt
    .For(e => e.Collection).Exclude(e => e.Second));
```

Or use "From" 
(There's propably a slight semantical difference between "For" and "From" that I can't spot)
```cs
subject.Should().BeEquivalentTo(expected, opt => opt
    .From(e => e.Collection).Exclude(e => e.Second));
```



A nice-to-have: 
If we can find a syntax that also allows in a nicely fashion to exclude multiple properties from a collection property.

dennisdoomen:
> Wouldn't
> 
> ```cs
> subject.Should().BeEquivalentTo(expected, opt => opt
>     .Excluding(e => e.Collection.Select(e => e.Second));
> ```
> 
> conflict with the overload of `Excluding` used for current property exclusion? E.g.

We currently use an algorithm to build a property path from the expression. I guess it could be made smart enough to transform that example into `Collection[].Second`. 

> Continuing from my previous comment, we could also leave out the "Collection" from "ForCollection".
> 
> subject.Should().BeEquivalentTo(expected, opt => opt
>     .For(e => e.Collection).Exclude(e => e.Second));

I actually like this one.

jnyrup:
Seems we can agree on the `For`+`Exclude` combination. 

Adding two snippets of how I imagine usage of it. 

```cs 
subject.Should().BeEquivalentTo(expected, opt => opt
    .For(e => e.Collection)
        .For(e => e.NestedCollection)
            .Exclude(e => e.First));
```

```cs 
subject.Should().BeEquivalentTo(expected, opt => opt
    .For(e => e.Collection)
        .Exclude(e => e.First)
    .For(e => e.OtherCollection)
        .Exclude(e => e.Second));
```

whymatter:
@jnyrup sorry it took me some time to get back to the PR. Please have a look if my latest commit complies with your proposal.

Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
```diff
@@ -40,6 +39,17 @@ public EquivalencyAssertionOptions<TExpectation> Excluding(Expression<Func<TExpe
             return this;
         }
 
+        /// <summary>
+        /// Excludes the specified (nested) member from the structural equality check.
```
dennisdoomen, line 43: 🤔 I don't think this covers what `For` does, right? It's not that it excludes the specified property, but just identifies the collection that you want to exclude a property from, isn't it?

Src/FluentAssertions/Equivalency/Selection/ExcludeMemberByPathSelectionRule.cs
```diff
@@ -23,6 +23,12 @@ public ExcludeMemberByPathSelectionRule(MemberPath pathToExclude)
                 memberToExclude.IsSameAs(new MemberPath(member, parentPath)));
         }
 
+        public void CombinePath(MemberPath nextPath)
```
dennisdoomen, line 26: 🔧 Usually, a method like `Combine` implies it combines two or more arguments into something new. But I think it just _appends_ something to the existing rule.

Tests/FluentAssertions.Equivalency.Specs/CollectionSpecs.cs
```diff
@@ -556,6 +556,320 @@ public void When_a_deeply_nested_property_of_a_collection_with_an_invalid_value_
...
+                            }
+                        }
+                    }
+                };
+
+                // Act / Assert
+                subject.Should().BeEquivalentTo(expected,
+                    options => options
+                        .Excluding(x => x.Text)
+                        .For(x => x.Level.Collection).Exclude(x => x.Text));
```
dennisdoomen, line 869: 🤔 How about the scenario why you have two sets of `For` and `Exclude`? Is that possible?

dennisdoomen (changes requested):
🔧 You also need to increase the coverage

docs/_pages/objectgraphs.md
```diff
@@ -130,6 +130,23 @@ orderDto.Should().BeEquivalentTo(order, options =>
...
+           .Exclude(o => o.Status));
+```
+
+Using `For` you can navigate arbitrarily deep. Consider a `Product` has a collection of `Part`s and a `Part` has a name. Using `For` your can also exclude the `Name` of all `Part`s of all `Product`s.
+
+```csharp
+orderDto.Should().BeEquivalentTo(order, options =>
+    options.For(o => o.Products)
+           .For(o => o.Parts)
+           .Exclude(o => o.Status));
```
jnyrup, line 147:
```suggestion
           .Exclude(o => o.Name));
```

Src/FluentAssertions/Equivalency/NestedExclusionOptionBuilder.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using FluentAssertions.Common;
+using FluentAssertions.Equivalency.Selection;
+
+namespace FluentAssertions.Equivalency
+{
+    public class NestedExclusionOptionBuilder<TExpectation, TCurrent> : EquivalencyAssertionOptions<TExpectation>
```
jnyrup, line 9: I'm a bit unsure about this construction as it seems to allow applying methods from `EquivalencyAssertionOptions` on top of `NestedExclusionOptionBuilder` without taking effect.



```cs
// Arrange
var subject = new
{
    Text = "a",
    Inner = new[]
    {
        new { Number = 1 },
        new { Number = 2 },
        new { Number = 3 } // <-- should make the test fail
    }
};

var expected = new
{
    Text = "a",
    Inner = new[]
    {
        new { Number = 2 },
        new { Number = 1 }
    }
};

// Act / Assert
subject.Should().BeEquivalentTo(expected, opt => opt
    .WithStrictOrdering()
    .For(e => e.Inner)
        .WithoutStrictOrdering()
);
```

This also compiles
```cs
subject.Should().BeEquivalentTo(expected, opt => opt
    .For(e => e.Inner)
);
```
but should probably(?) just have been 

```cs
subject.Should().BeEquivalentTo(expected, opt => opt
    .Excluding(e => e.Inner)
);
```

Maybe `Restriction<T>`  which is used for the `Using<int>(ctx.Subject.Should().BeCloseTo(ctx.Expectation, 1)).WhenTypeIs<int>()` has some solutions?

jnyrup (commented):
No worries about taking your time, I really like reading the new tests 👍 

As I'm confident enough that the implementation does the expected in the expected cases, I also tried to turn the focus and see if I could make it do the unexpected.
History has shown that Murphy's Law applies to how the Fluent Assertions API will be used by developers.

Tests/FluentAssertions.Equivalency.Specs/CollectionSpecs.cs
whymatter: I added a test for that 👍

Src/FluentAssertions/Equivalency/Selection/ExcludeMemberByPathSelectionRule.cs
whymatter: I renamed it to `AppendPath`. I also renamed to `Combine` method (which I created) on the `MemberPath` to `Join`

Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
whymatter: Updated the description 👍

Src/FluentAssertions/Common/MemberPath.cs
```diff
@@ -64,15 +66,22 @@ private bool IsParentOf(MemberPath candidate)
...
         {
             string[] candidateSegments = candidate.Segments;
 
             return candidateSegments.Length < Segments.Length
-                   && candidateSegments.SequenceEqual(Segments.Take(candidateSegments.Length));
+                   && candidateSegments.SequenceEqual(Segments.Take(candidateSegments.Length),
+                       MemberPathSegmentEqualityComparer);
+        }
+
+        public MemberPath Combine(MemberPath nextPath, string separator = ".")
```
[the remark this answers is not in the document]
whymatter: I uses this method to join the 'already selected path' and the 'newly selected path'. I renamed the method to `Join` and removed the default `separator`

whymatter:
I am a bit puzzled about how to increase the coverage, I believe the problem is the `MemberPathSegmentEqualityComparer.GetHashCode` method. I a test like that okay?

```c#
var text = "Text";
var compare = new MemberPathSegmentEqualityComparer();

comparer.GetHashCode(text).Should().Be(text.GetHashCode())
```

Src/FluentAssertions/Equivalency/NestedExclusionOptionBuilder.cs
whymatter: Regarding the first issue where `.For(e => e.Inner)` actually excluded the whole `Inner` collection, this is fixed. I added a test `For_Does_Not_Exclude`. The issue was that I immediately called `AddSelectionRule` in the `For` method.

Is this fine or should the following not compile at all?
```c#
subject.Should().BeEquivalentTo(expected, opt => opt
    .For(e => e.Inner)
);
```

Src/FluentAssertions/Common/MemberPath.cs
```diff
@@ -64,15 +66,22 @@ private bool IsParentOf(MemberPath candidate)
...
         {
             string[] candidateSegments = candidate.Segments;
 
             return candidateSegments.Length < Segments.Length
-                   && candidateSegments.SequenceEqual(Segments.Take(candidateSegments.Length));
+                   && candidateSegments.SequenceEqual(Segments.Take(candidateSegments.Length),
+                       MemberPathSegmentEqualityComparer);
+        }
+
+        public MemberPath Join(MemberPath nextPath, string separator)
```
dennisdoomen, line 81: 🔧 Since you only use this member to join two paths using `[]`, I think we need to make it more explicit, e.g. something like `AsParentCollectionOf`

Src/FluentAssertions/Common/MemberPathSegmentEqualityComparer.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            if (y == AnyIndexQualifier)
+            {
+                return EqualsAnyIndexQualifier(x);
+            }
+
+            return x == y;
+        }
+
+        private static bool EqualsAnyIndexQualifier(string segment)
```
dennisdoomen, line 42: ⛏️ I think naming it `IsIndexQualifier` is a easier to understand.

Tests/FluentAssertions.Equivalency.Specs/CollectionSpecs.cs
```diff
@@ -558,6 +558,425 @@ public void When_a_deeply_nested_property_of_a_collection_with_an_invalid_value_
             act.Should().NotThrow();
         }
 
+        public class For
+        {
+            [Fact]
+            public void For_Does_Not_Exclude()
```
dennisdoomen, line 564: 🔧 This does not follow our naming convention.

dennisdoomen (changes requested):

dennisdoomen:
> I am a bit puzzled about how to increase the coverage, I believe the problem is the MemberPathSegmentEqualityComparer.GetHashCode method. I a test like that okay?

Don't bother. In fact, this `MemberPathSegmentEqualityComparer` is an implementation detail of something that is an implementation detail itself (`MemberPath`). We should never test them separately.

Tests/FluentAssertions.Equivalency.Specs/CollectionSpecs.cs
whymatter: Removed test since it does not compile anymore

Src/FluentAssertions/Common/MemberPathSegmentEqualityComparer.cs
whymatter: done 👍

Src/FluentAssertions/Common/MemberPath.cs
whymatter: done 👍

Src/FluentAssertions/Equivalency/NestedExclusionOptionBuilder.cs
whymatter: @dennisdoomen I made `NestedExclusionOptionBuilder` not inherit from `EquivalencyAssertionOptions`. This way a statement like this does not compile:

```c#
subject.Should().BeEquivalentTo(expected, opt => opt
    .For(e => e.Inner)
);
```

dennisdoomen:
> At least it should throw. Maybe you can set a flag that marks the For construction incomplete until you call the Including or Excluding option.

Did you address this?

whymatter:
> > At least it should throw. Maybe you can set a flag that marks the For construction incomplete until you call the Including or Excluding option.
> 
> Did you address this?

Yes, only using For does not compile. See here https://github.com/fluentassertions/fluentassertions/pull/1782#discussion_r847024590

whymatter:
Reran the `AcceptApiChanges.ps1`. Tests should be good now

jnyrup (approved):

jnyrup (approved):

Src/FluentAssertions/Common/MemberPathSegmentEqualityComparer.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </summary>
+        /// <param name="x">Left part of the comparison.</param>
+        /// <param name="y">Right part of the comparison.</param>
+        /// <returns>True if segments are equal, false if not.</returns>
+        public bool Equals(string x, string y)
+        {
+            if (x is null || y is null)
+            {
+                return x == y;
+            }
```
jnyrup, line 27: According to code coverage this branch is never entered.
I believe it is unreachable and can be deleted as `x` and `y` comes from `MemberPath.Segments` which will never contain `null` elements.

dennisdoomen: So if @whymatter removes this one, we can merge.

jnyrup: Yes 🤩

IT-VBFK: Please also be aware of the release notes 😉

whymatter: @dennisdoomen done 👍

jnyrup:
@whymatter Thanks for taking the time to contribute this feature, despite it took some time and iterations to get it merged.
I'm really pleased with the final result.

whymatter:
@jnyrup @dennisdoomen Thank you very much for accepting the proposal! I am really looking forward to using this feature! :)

PR #1786, opened 2022-01-22
Fix TryGetValue for dictionary like enumerables

### LINKED ISSUE #1783: Missing negation operator
The method `FluentAssertions.Common.DictionaryHelpers.TryGetValue` works incorrectly for non-dictionary classes, because there is a missing negation operator.

https://github.com/fluentassertions/fluentassertions/blob/ac04547a33db977d6edcfabcdf9fda8c055f98f2/Src/FluentAssertions/Common/DictionaryHelpers.cs#L63

jnyrup:
When I added `DictionaryHelpers` in #1298 I recall I did check that we had tests covering all the methods for `IEnumerable<KeyValuePair<TKey, TValue>>`.
Now I see that `DictionaryHelpers.TryGetValue` was only hit from `GenericDictionaryAssertions.ContainKey` which ignores the return value of `TryGetValue`.

This fixes #1783

### DIFF
--- a/docs/_pages/releases.md
+++ b/docs/_pages/releases.md
@@ -20,6 +20,7 @@ sidebar:
 * Variable name is not captured after await assertion - [#1770](https://github.com/fluentassertions/fluentassertions/pull/1770)
 * `OccurredEvent` ordering on monitored object is now done via thread-safe counter - [#1773](https://github.com/fluentassertions/fluentassertions/pull/1773)
 * Avoid a `NullReferenceException` when testing an application compiled with .NET Native - [#1776](https://github.com/fluentassertions/fluentassertions/pull/1776)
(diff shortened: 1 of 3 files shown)


docs/_pages/releases.md
```diff
@@ -20,6 +20,7 @@ sidebar:
 * Variable name is not captured after await assertion - [#1770](https://github.com/fluentassertions/fluentassertions/pull/1770)
 * `OccurredEvent` ordering on monitored object is now done via thread-safe counter - [#1773](https://github.com/fluentassertions/fluentassertions/pull/1773)
 * Avoid a `NullReferenceException` when testing an application compiled with .NET Native - [#1776](https://github.com/fluentassertions/fluentassertions/pull/1776)
+* Calling `ContainKey[s]` and `[Not]Contain(key, value)` on dictionary-like enumerables - [#1786](https://github.com/fluentassertions/fluentassertions/pull/1786)
```
dennisdoomen, line 23: Calling should do what?

dennisdoomen (approved):

PR #1787, opened 2022-01-22
Optimize build times

jnyrup:
Enabling `ProduceReferenceAssembly` produces a reference assembly for `FluentAssertions.dll` in the `ref/` folder which only contains the method signatures, but no implementations. 
Similarly to the relationship between an interface and an implementation.

Project references look at the reference assembly to see what methods are available.
When making changes in Core that does _not_ change the method signatures, the reference assembly is unchanged, so compiling the Specs only need to recompile the Core dll and copy it to Specs output.
It can skip compiling the Specs dll as no method signatures have changed.

My experiments show that making a change in Core and recompiling Specs with `ProduceReferenceAssembly` enabled changes the compile time from 25s to 7s.

Additionally only having analyzers enabled for the newest TFM brings the compile time of Core down from 7.1s to 2.5s.

All experiments were done using [MSBuild Binary and Structured Log Viewer](https://msbuildlog.com/)

### DIFF
--- a/Src/FluentAssertions/FluentAssertions.csproj
+++ b/Src/FluentAssertions/FluentAssertions.csproj
@@ -1,4 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk">
+  <!-- To reduce build times, we only enable analyzers for the newest TFM -->
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


dennisdoomen (approved):
It seems to help indeed. And it has become the default for .NET 5.0. Let's hope it doesn't have any side-effects such as stale assemblies.

PR #1789, opened 2022-01-23
Moved the equivalency specs to a separate test project and split them into separate classes

dennisdoomen:
To improve the developer experience while working on the many tests addressing `BeEquivalentTo`, I've moved them into a separate project. Also started to change regions into classes to make it easier to discover and navigate the tests.


Tests/FluentAssertions.Specs/FluentAssertions.Specs.csproj
```diff
@@ -9,7 +9,6 @@
     <IsPackable>false</IsPackable>
     <CodeAnalysisRuleSet>..\..\TestRules.ruleset</CodeAnalysisRuleSet>
     <!-- https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA0001.md -->
-    <DocumentationFile Condition="'$(TargetFramework)' == 'net5.0'">$(OutputPath)$(AssemblyName).xml</DocumentationFile>
```
jnyrup, line 12: Why is this removed?

jnyrup (commented):
I did a benchmark on this using MSBuild Log Viewer:

master:
Compiling .Specs
	no changes: 1.1s
	test changes: 17.5s
	core changes: 2.0s
	
PR
Compiling .Specs
	no changes: 1.0s
	test changes: 13s
	core changes: 1.8s
	
Compiling Equivalency.Specs
	no changes: 1.0s
	test changes: 2.9s
	core changes: 2.0s

jnyrup:
I really wish there were some sort of caching or incremental analysis we could benefit from instead of splitting our project into two to decrease build times.

Did you try moving/renaming in the files in a single commit?
That usually helps reducing the diffs when you both want to rename and change a file.

Tests/FluentAssertions.Specs/FluentAssertions.Specs.csproj
dennisdoomen: Because we don't need to analyze XML comments in test projects. I also disabled rule 5001 for that. I was thinking whether to put that in a separate PR, but couldn't be bothered. Should have known you would ask about that :-)

dennisdoomen:
> I really wish there were some sort of caching or incremental analysis we could benefit from instead of splitting our project into two to decrease build times.

I wish that too ;-)

> Did you try moving/renaming in the files in a single commit?
> That usually helps reducing the diffs when you both want to rename and change a file.

Why are you asking that?

dennisdoomen:
> To get a cleaner git history

You mean one commit that moves the files and another two split the files into smaller files? If you really think it's worth the time.

jnyrup:
> You mean one commit that moves the files and another two split the files into smaller files

Yes, that was my idea.
Right now you could have sneaked whatever in, since I didn't look at the source files due to the amount of changes.
Not a blocker for me, but this is what I preach to colleagues to ease PR review.

Tests/FluentAssertions.Specs/FluentAssertions.Specs.csproj
jnyrup: Roger, makes perfect sense 👍 

Then please remove the comment linking to `SA0001.md` from this and the new specs project.

dennisdoomen:
Took me almost an hour, and only the last commit makes everything compilable again.

jnyrup (approved):

PR #1794, opened 2022-01-29
Caller identification does not handle all arguments using `new`

### LINKED ISSUE #1695: CallerIdentification can't seem to handle `new[]{}`
I noticed that the subject in our assertion output often had "root[..]", which seemed odd so whipped up a quick test case to validate my concerns. The following test fails (output is included), but using `new List<int> { 1, 2, 3}` instead of `new[]` works just fine.

I was on master SHA: b5b5ceb445035e5c461eec03dfa1a0c72049db5a

```c#
[Fact]
public void When_the_caller_is_bob()
{
    // Arrange
    var foo = new Foo();

    // Act
    Action act = () => foo.GetFoo(new[] { 1, 2, 3 }.Sum() + "")
        .Should()
        .BeNull();

    // Assert
        .WithMessage("Expected foo.GetFoo(new[]{1,2,3}.Sum()+\"\") to be <null>*");
}
```

Output:
```
Expected exception message to match the equivalent of "Expected foo.GetFoo(new[]{1,2,3}.Sum()+"") to be <null>*", but "Expected object to be <null>, but found FluentAssertions.Specs.Execution.Foo
{
    Bar = "bar", 
    Field = "bar"
}." does not.
```

I can probably work on this in the future, but it might be a little as things are busy at the moment 👍.

dennisdoomen:
Certain constructs where `new` is used when constructing the argument of a method were not properly detected by the caller identification feature. 

Fixes #1695

### DIFF
--- a/Src/FluentAssertions/CallerIdentifier.cs
+++ b/Src/FluentAssertions/CallerIdentifier.cs
@@ -235,7 +235,7 @@ namespace FluentAssertions
 
         private static bool StartsWithNewKeyword(string candidate)
         {
-            return Regex.IsMatch(candidate, @"(^|s+)new(?:\s?\[|\s?\{|\s\w+)");
+            return Regex.IsMatch(candidate, @"(?:^|s+)new(?:\s?\[|\s?\{|\s\w+)");
         }
 
         private static bool IsStringLiteral(string candidate)
--- a/Src/FluentAssertions/CallerIdentifier.cs
+++ b/Src/FluentAssertions/CallerIdentifier.cs
@@ -173,7 +173,7 @@ namespace FluentAssertions
             {
                 Logger(statement);
                 if (!IsBooleanLiteral(statement) && !IsNumeric(statement) && !IsStringLiteral(statement) &&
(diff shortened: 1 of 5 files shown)


Src/FluentAssertions/CallerIdentifier.cs
```diff
@@ -233,9 +233,9 @@ private static string GetSourceCodeStatementFrom(StackFrame frame, StreamReader
             return sb.ToString();
         }
 
-        private static bool UsesNewKeyword(string candidate)
+        private static bool StartsWithNewKeyword(string candidate)
         {
-            return Regex.IsMatch(candidate, @"new(?:\s?\[|\s?\{|\s\w+)");
+            return Regex.IsMatch(candidate, @"(^|s+)new(?:\s?\[|\s?\{|\s\w+)");
```
jnyrup, line 238:
```suggestion
            return Regex.IsMatch(candidate, @"(?:^|s+)new(?:\s?\[|\s?\{|\s\w+)");
```

jnyrup (approved):

PR #1806, opened 2022-02-12
Attempt to fix System.Http.Net compatibility in .NET 4.7.2

### LINKED ISSUE #1805: v6.4.0 breaks build under mono when targeting net48
### Description
v6.4.0 and newer now depend on the System.Net.Http nuget package. This package is problematic under mono because it includes compiled versions of the System.Net.Http dll for .Net Framework that differ from the mono versions and will ultimately cause projects using mono to no longer build with FluentAssertions referenced.

This is a known problem with the System.Net.Http package, e.g. discussion at https://github.com/dotnet/runtime/issues/21777
The package is effectively abandoned and not recommended for use in new projects, so if possible can it be removed as a dependency in FluentAssertions and alternatives be used if necessary?

Blank project targeting net48 with fluent assertions v6.4.0 or newer referenced & attempt to compile with mono.

### Expected behavior:
Builds

### Actual behavior:
Build fails with error:
```
CSC : error CS1705: Assembly 'FluentAssertions' with identity 'FluentAssertions, Version=6.5.1.0, Culture=neutral, PublicKeyToken=33f2691a05b67b6a' uses 'System.Net.Http, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' which has a higher version than referenced assembly 'System.Net.Http' with identity 'System.Net.Http, Version=4.1.1.3, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' [/__w/xRetry/xRetry/test/UnitTests/UnitTests.csproj]
    1 Warning(s)
    1 Error(s)
```

### Versions

* FluentAssertions 6.4.0, 6.5.0 & 6.5.1
* net468 under mono

### Additional Information
Example of a PR with the CI build failing with the upgrade from FluentAssertions 6.3.0 => 6.4.0: https://github.com/JoshKeegan/xRetry/pull/114

dennisdoomen:
Fixes #1805

### DIFF
--- a/Src/FluentAssertions/FluentAssertions.csproj
+++ b/Src/FluentAssertions/FluentAssertions.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <!-- To reduce build times, we only enable analyzers for the newest TFM -->
   <PropertyGroup>
-    <TargetFrameworks>net47;netcoreapp2.1;netcoreapp3.0;netstandard2.0;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>net47;net472;netcoreapp2.1;netcoreapp3.0;netstandard2.0;netstandard2.1</TargetFrameworks>
     <SignAssembly>True</SignAssembly>
     <AssemblyOriginatorKeyFile>FluentAssertions.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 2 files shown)


dennisdoomen:
@jnyrup I don't feel very fond of duplicating all the tests for .NET 4.7 to include .NET 4.7.2 as well. What do you think?

PR #1807, opened 2022-02-15
Add the ability to exclude non-browsabe members from equivalency tests

logiclrd:
This PR:

* Adds a property `ExcludeNonBrowsable` to `IEquivalencyAssertionOptions`, and implements it in `SelfReferenceEquivalencyAssertionOptions`, `CollectionMemberAssertionOptionsDecorator` and UsersOfGetClosedGenericInterfaces`.
* Adds a property `IsBrowsable` to ``IMember` and implements it in `Field` and `Property`.
* Uses the new properties in `AssertMemberEquality` in `StructuralEqualityEquivalencyStep` to allow for non-browsable members to be skipped when checking equivalence.
* Adds automated tests of the new functionality to `SelectionRulesSpec`.
* Accepted API changes into the approved API.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Equivalency/Field.cs
+++ b/Src/FluentAssertions/Equivalency/Field.cs
@@ -1,4 +1,5 @@
 using System;
+using System.ComponentModel;
 using System.Reflection;
 using FluentAssertions.Common;
 
@@ -26,6 +27,12 @@ namespace FluentAssertions.Equivalency
             Name = fieldInfo.Name;
             Type = fieldInfo.FieldType;
             RootIsCollection = parent.RootIsCollection;
+
+            bool isNonBrowsable =
+                (fieldInfo.GetCustomAttribute<EditorBrowsableAttribute>() is EditorBrowsableAttribute attribute) &&
+                (attribute.State == EditorBrowsableState.Never);
+
+            IsBrowsable = !isNonBrowsable;
         }
 
         public Type ReflectedType { get; }
@@ -42,5 +49,7 @@ namespace FluentAssertions.Equivalency
         public CSharpAccessModifier GetterAccessibility => fieldInfo.GetCSharpAccessModifier();
 
         public CSharpAccessModifier SetterAccessibility => fieldInfo.GetCSharpAccessModifier();
+
+        public bool IsBrowsable { get; private set; }
     }
 }
--- a/Src/FluentAssertions/Equivalency/IEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/IEquivalencyAssertionOptions.cs
@@ -1,5 +1,7 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
+
 using FluentAssertions.Equivalency.Tracing;
 
 namespace FluentAssertions.Equivalency
@@ -70,6 +72,12 @@ namespace FluentAssertions.Equivalency
         /// Gets a value indicating whether and which fields should be considered.
         /// </summary>
         MemberVisibility IncludedFields { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether members marked with [EditorBrowsable]
+        /// and an EditorBrowsableState of Never should be excluded.
+        /// </summary>
+        bool ExcludeNonBrowsable { get; }
 
         /// <summary>
         /// Gets a value indicating whether records should be compared by value instead of their members
--- a/Src/FluentAssertions/Equivalency/Property.cs
+++ b/Src/FluentAssertions/Equivalency/Property.cs
@@ -1,4 +1,5 @@
 using System;
+using System.ComponentModel;
 using System.Reflection;
 using FluentAssertions.Common;
 
@@ -27,6 +28,12 @@ namespace FluentAssertions.Equivalency
             Path = parent.PathAndName;
             GetSubjectId = parent.GetSubjectId;
             RootIsCollection = parent.RootIsCollection;
+
+            bool isNonBrowsable =
+                (propertyInfo.GetCustomAttribute<EditorBrowsableAttribute>() is EditorBrowsableAttribute attribute) &&
+                (attribute.State == EditorBrowsableState.Never);
+
+            IsBrowsable = !isNonBrowsable;
         }
 
         public object GetValue(object obj)
@@ -43,5 +50,7 @@ namespace FluentAssertions.Equivalency
         public CSharpAccessModifier GetterAccessibility => propertyInfo.GetGetMethod(nonPublic: true).GetCSharpAccessModifier();
 
         public CSharpAccessModifier SetterAccessibility => propertyInfo.GetSetMethod(nonPublic: true).GetCSharpAccessModifier();
+
+        public bool IsBrowsable { get; private set; }
     }
 }
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
@@ -56,6 +56,7 @@ namespace FluentAssertions.Equivalency
 
         private MemberVisibility includedProperties;
         private MemberVisibility includedFields;
+        private bool excludeNonBrowsable;
 
         private bool compareRecordsByValue;
 
@@ -80,6 +81,7 @@ namespace FluentAssertions.Equivalency
             useRuntimeTyping = defaults.UseRuntimeTyping;
             includedProperties = defaults.IncludedProperties;
             includedFields = defaults.IncludedFields;
+            excludeNonBrowsable = defaults.ExcludeNonBrowsable;
             compareRecordsByValue = defaults.CompareRecordsByValue;
 
             ConversionSelector = defaults.ConversionSelector.Clone();
@@ -161,6 +163,8 @@ namespace FluentAssertions.Equivalency
         MemberVisibility IEquivalencyAssertionOptions.IncludedProperties => includedProperties;
 
         MemberVisibility IEquivalencyAssertionOptions.IncludedFields => includedFields;
+
+        bool IEquivalencyAssertionOptions.ExcludeNonBrowsable => excludeNonBrowsable;
 
         public bool CompareRecordsByValue => compareRecordsByValue;
 
@@ -309,6 +313,26 @@ namespace FluentAssertions.Equivalency
         public TSelf ExcludingProperties()
         {
             includedProperties = MemberVisibility.None;
+            return (TSelf)this;
+        }
+
+        /// <summary>
+        /// Instructs the comparison to include non-browsable members (members with an EditorBrowsableState of Never).
+        /// </summary>
+        /// <returns></returns>
+        public TSelf IncludingNonBrowsableMembers()
+        {
+            excludeNonBrowsable = false;
+            return (TSelf)this;
+        }
+
+        /// <summary>
+        /// Instructs the comparison to exclude non-browsable members (members with an EditorBrowsableState of Never).
+        /// </summary>
+        /// <returns></returns>
+        public TSelf ExcludingNonBrowsableMembers()
+        {
+            excludeNonBrowsable = true;
             return (TSelf)this;
         }
 
@@ -724,6 +748,15 @@ namespace FluentAssertions.Equivalency
             else
             {
                 builder.AppendLine("- Compare records by their members");
+            }
+
+            if (excludeNonBrowsable)
+            {
+                builder.AppendLine("- Exclude non-browsable members");
+            }
+            else
+            {
+                builder.AppendLine("- Include non-browsable members");
             }
 
             foreach (Type valueType in valueTypes)
--- a/Src/FluentAssertions/Equivalency/Steps/StructuralEqualityEquivalencyStep.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/StructuralEqualityEquivalencyStep.cs
@@ -55,21 +55,24 @@ namespace FluentAssertions.Equivalency.Steps
             IMember matchingMember = FindMatchFor(selectedMember, context.CurrentNode, comparands.Subject, options);
             if (matchingMember is not null)
             {
-                var nestedComparands = new Comparands
+                if (matchingMember.IsBrowsable || !options.ExcludeNonBrowsable)
                 {
-                    Subject = matchingMember.GetValue(comparands.Subject),
-                    Expectation = selectedMember.GetValue(comparands.Expectation),
-                    CompileTimeType = selectedMember.Type
-                };
+                    var nestedComparands = new Comparands
+                    {
+                        Subject = matchingMember.GetValue(comparands.Subject),
+                        Expectation = selectedMember.GetValue(comparands.Expectation),
+                        CompileTimeType = selectedMember.Type
+                    };
 
-                if (selectedMember.Name != matchingMember.Name)
-                {
-                    // In case the matching process selected a different member on the subject,
-                    // adjust the current member so that assertion failures report the proper name.
-                    selectedMember.Name = matchingMember.Name;
+                    if (selectedMember.Name != matchingMember.Name)
+                    {
+                        // In case the matching process selected a different member on the subject,
+                        // adjust the current member so that assertion failures report the proper name.
+                        selectedMember.Name = matchingMember.Name;
+                    }
+
+                    parent.RecursivelyAssertEquality(nestedComparands, context.AsNestedMember(selectedMember));
                 }
-
-                parent.RecursivelyAssertEquality(nestedComparands, context.AsNestedMember(selectedMember));
             }
         }
 
--- a/Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
+++ b/Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
@@ -803,6 +803,7 @@ namespace FluentAssertions.Equivalency
         public System.Type DeclaringType { get; set; }
         public override string Description { get; }
         public FluentAssertions.Common.CSharpAccessModifier GetterAccessibility { get; }
+        public bool IsBrowsable { get; }
         public System.Type ReflectedType { get; }
         public FluentAssertions.Common.CSharpAccessModifier SetterAccessibility { get; }
         public object GetValue(object obj) { }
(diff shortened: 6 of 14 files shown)


Src/FluentAssertions/Equivalency/Field.cs
```diff
@@ -26,6 +27,12 @@ public Field(Type reflectedType, FieldInfo fieldInfo, INode parent)
             Name = fieldInfo.Name;
             Type = fieldInfo.FieldType;
             RootIsCollection = parent.RootIsCollection;
+
+            bool isNonBrowsable =
```
dennisdoomen, line 31: 🔧 I'm not 100% sure it is really needed, but I would use a lazy-evaluated solution to not add more reflection if it isn't really needed. Same for the `Property` member.

Src/FluentAssertions/Equivalency/IEquivalencyAssertionOptions.cs
```diff
@@ -71,6 +73,12 @@ public interface IEquivalencyAssertionOptions
         /// </summary>
         MemberVisibility IncludedFields { get; }
 
+        /// <summary>
+        /// Gets a value indicating whether members marked with [EditorBrowsable]
```
dennisdoomen, line 77: 🔧 Would be nice to use `<see` constructs when referring to types and members.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1489,5 +1490,89 @@ public void Including_an_interface_property_through_inheritance_should_work()
                 .Including(a => a.Value2)
                 .RespectingRuntimeTypes());
         }
+
+        [Theory]
+        [MemberData(nameof(TestMembers))]
```
dennisdoomen, line 1495: 🤔 I found it very hard to understand what this is doing with the parameter. Because of that, I strongly believe it is much better to drop the parameterized tests and have individual tests for the relevant test cases.

dennisdoomen (changes requested):

Src/FluentAssertions/Equivalency/IEquivalencyAssertionOptions.cs
logiclrd: Easy enough to do. :-)

Src/FluentAssertions/Equivalency/Field.cs
logiclrd: Easy enough to do. :-)

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
```diff
@@ -312,6 +316,26 @@ public TSelf ExcludingProperties()
             return (TSelf)this;
         }
 
+        /// <summary>
+        /// Instructs the comparison to include non-browsable members (members with an EditorBrowsableState of Never).
+        /// </summary>
+        /// <returns></returns>
+        public TSelf IncludingNonBrowsableMembers()
```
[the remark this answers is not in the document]
logiclrd: I was looking at the existing `IncludedFields` and `IncludedProperties` members -- I don't see how those can be applied to just certain types. What am I missing??

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: Oh? This is just how Xunit works. Maybe this can be improved with comments and/or better naming??

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
```diff
@@ -858,6 +860,7 @@ namespace FluentAssertions.Equivalency
     {
         System.Type DeclaringType { get; }
         FluentAssertions.Common.CSharpAccessModifier GetterAccessibility { get; }
+        bool IsBrowsable { get; }
```
jnyrup, line 863: Just for the record, as adding a member to a public interface is a source breaking change.
In https://github.com/fluentassertions/fluentassertions/pull/1742#discussion_r790274160 we agreed changes to `INode` are justified as we believe(/hope) that they are not implemented by anyone else.
I'm fine with applying the same argument to changes to `IMember`.

Src/FluentAssertions/Equivalency/Steps/StructuralEqualityEquivalencyStep.cs
```diff
@@ -55,21 +55,24 @@ public class StructuralEqualityEquivalencyStep : IEquivalencyStep
             IMember matchingMember = FindMatchFor(selectedMember, context.CurrentNode, comparands.Subject, options);
             if (matchingMember is not null)
             {
-                var nestedComparands = new Comparands
+                if (matchingMember.IsBrowsable || !options.ExcludeNonBrowsable)
```
jnyrup, line 58: If we swapped the left and right hand sides of `||` and made `IsBrowsable` lazy, we would only evaluate the `GetCustomAttribute<EditorBrowsableAttribute>()` when strictly necessary.
Could be nice with a small benchmark to see what the actual overhead of determining `IsBrowsable` is.

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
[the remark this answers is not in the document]
dennisdoomen: Like `ComparingByMembers<T>` and `ComparingByMembers(Type type)`

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
dennisdoomen: I strongly disagree. Parameterized tests are only useful if you have a simple list of values that you use in combination with a `Theory` attribute. For anything else, they obscure the cause and effect of the test. In this case, having 6 different tests would be much clearer.

logiclrd: Oh dear :-P Which way should I take it?

Src/FluentAssertions/Equivalency/Field.cs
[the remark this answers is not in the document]
logiclrd: I've never seen this syntax before. Interesting!

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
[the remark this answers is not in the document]
logiclrd: Hmm, that feels like it might be overcomplicating things a bit. Is it likely that someone will have an object graph where they need to compare non-browsable properties in one part of the graph and not in another part of the same graph??

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
jnyrup: I find the difference between having primitives in attributes and _my example_ to be insignificant. 
If the test case generator is named something with "browsable" I even think it has benefits as the test names don't have to be cluttered with implementation details such as "always" and "advanced". 
So I think dynamic test cases have their moments, but certainly many pitfalls that harm "readability". 

Dennis and don't agree on everything, but (surprisingly?) on quite many things. 
In the end Dennis has the final call.

Src/FluentAssertions/Equivalency/Steps/StructuralEqualityEquivalencyStep.cs
logiclrd: Looks like it is considerably more expensive if the attribute is present, but still only a minor cost at about 1 μs (on my development laptop) per lookup. If the attribute is not present, then the same code runs in about a fifth the time.

```
|                   Method | IsBrowsable |       Mean |    Error |   StdDev |  Gen 0 | Allocated |
|------------------------- |------------ |-----------:|---------:|---------:|-------:|----------:|
| CheckIfMemberIsBrowsable |       False | 1,121.7 ns | 29.20 ns | 84.71 ns | 0.0305 |     192 B |
| CheckIfMemberIsBrowsable |        True |   241.7 ns |  4.82 ns |  9.96 ns |      - |         - |
```

Will commit the benchmark and swap the order in the `if` :-)

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
dennisdoomen: > What about having a test method that isn't marked [Test] and 6 "head" methods that call the test implementation with different parameters?

I think it would improve the situation a bit. But in the end, I would just spell out the entire test case. Trying to achieve DRY in unit testing usually backfires.

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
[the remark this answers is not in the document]
dennisdoomen: > Is it likely that someone will have an object graph where they need to compare non-browsable properties in one part of the graph and not in another part of the same graph??

That's exactly what I was wondering. I haven't used that attribute myself a lot, so I can't judge that. Do you?

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: In this particular case, though, these really are variations on the _same test_. That's why I personally went for the parameterized approach in the first place. If the nature of the test ever changes, it's guaranteed to need the same edit to every repeated copy of the test. I agree that DRY is _in general_ not a good principle in unit tests, but there are cases like this where I think it makes sense.

jnyrup: Interesting 🙃 I prefer the parameterized one to this, as I don't have to jump between method bodies.

Conditionals in tests is a (99.9%) no-go for me. 
For parameterized tests I would create separate test case generators to avoid that.

Seems Dennis and I can agree on spelling out the entire test cases.

logiclrd: Okay, conditionals are gone. :-)

dennisdoomen: Sorry, was trying to understand each test, but still had to dig through the private methods to understand what each test was doing. I really want them to be spelled out and have functional names.

logiclrd: I have pushed a commit that gives the methods in question much more explicit names. Is that what you had in mind? :-)

dennisdoomen: Unfortunately not. In fact, it makes it worse since you refer to the literal names of the APIs, which is an [anti-pattern for me](https://www.continuousimprover.com/2021/10/laws-test-driven-development.html). What I meant is to drop the parameterized tests completely.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1493,86 +1493,131 @@ public void Including_an_interface_property_through_inheritance_should_work()
 
         [Fact]
         public void Including_non_browsable_members_when_testing_NotEquivalent_should_work_when_browsable_field_differs()
```
dennisdoomen, line 1495: Almost ready...

We need to avoid to use the literal name of APIs like `NotBeEquivalentTo` and `BeEquivalentTo`. In this case, I would focus the scenarios using `BeEquivalentTo` because that's what `NotBeEquivalentTo` internally uses.  E.g. `Browsable_fields_can_be_included_in_the_assertion`

Also, I would explicitly show the value of `BrowsableField` of the subject.

Also applies on the other scenarios.

dennisdoomen (changes requested):

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: What do you mean by

> explicitly show the value of `BrowsableField` of the subject

?? The test should write it to test output??

dennisdoomen: [Fact]
        public void When_browsable_field_differs_including_non_browsable_members_should_not_affect_result()
        {
            // Arrange
            var subject = new ClassWithNonBrowsableMembers() { BrowsableField = 2 };
            var expectation = new ClassWithNonBrowsableMembers() { BrowsableField = 1 };

            // Act
            Action action =
                () => subject.Should().BeEquivalentTo(expectation, config => config.IncludingNonBrowsableMembers());

            // Assert
        }

logiclrd: Ahh, okay, don't use the implicit value of 0. Got it :-)

logiclrd: I've pushed a commit to this effect. It explicitly assigns to the member where a difference is benig set up for both the subject and the expectation now.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1489,5 +1490,262 @@ public void Including_an_interface_property_through_inheritance_should_work()
...
+            // Act
+            Action action =
+                () => subject.Should().BeEquivalentTo(expectation, config => config.IncludingNonBrowsableMembers());
+
+            // Assert
+            action.Should().Throw<XunitException>();
+        }
+
+        [Fact]
+        public void When_browsable_field_differs_excluding_non_browsable_members_should_not_affect_result()
```
jnyrup, line 1615: Many of the `_excluding_` tests use `IncludingNonBrowsableMembers`.

logiclrd: Whoops :-P

Src/FluentAssertions/Equivalency/Property.cs
```diff
@@ -27,6 +29,9 @@ public Property(Type reflectedType, PropertyInfo propertyInfo, INode parent)
             Path = parent.PathAndName;
             GetSubjectId = parent.GetSubjectId;
             RootIsCollection = parent.RootIsCollection;
+
+            isBrowsable = new Lazy<bool>(
```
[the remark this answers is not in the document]
logiclrd: That's fair :-) My gut is to write out the accessor more explicitly rather than using the null-coalescing operator, for readability.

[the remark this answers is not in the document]
logiclrd: Hmm, true, there are a few places, but it feels like it doesn't match the emphasis on readability elsewhere.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
dennisdoomen: > Ahh, okay, don't use the implicit value of 0. Got it :-)

Exactly, because you don't know that if you look at the test. In other words, if it's important to understand the test case, it's very important to show it ;-)

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
```diff
@@ -823,6 +824,7 @@ namespace FluentAssertions.Equivalency
         FluentAssertions.Equivalency.ConversionSelector ConversionSelector { get; }
         FluentAssertions.Equivalency.CyclicReferenceHandling CyclicReferenceHandling { get; }
         FluentAssertions.Equivalency.EnumEquivalencyHandling EnumEquivalencyHandling { get; }
+        bool ExcludeNonBrowsable { get; }
```
jnyrup, line 827: I'm a little more worried about this one.
So I wanted to search the web a bit, to see if I can find someone who implements this interface

E.g. it _almost_ conflicts with https://github.com/fluentassertions/fluentassertions.json/pull/59/files#diff-cbc228d7e59fc36a3c426829ebf2bae9f6039116b210107e5506e15b96707aef

logiclrd: The only implementors I can find on GitHub are people who have imported the FluentAssertions code whole-hog into their own codebase for whatever reason. They, of course, won't be seeing this change until/unless they upgrade their imported copy, which'll bring in the new implementations at the same time.

jnyrup: I cloned the repos listed below and `grep`'ed for `IEquivalencyAssertionOptions`.
As none of those implemented the interface I'm okay with this.

```
atc-test/
atc/
FacilityCSharp/
FluentAssertions.ArgumentMatchers.Moq/
fluentassertions.aspnetcore.mvc/
FluentAssertions.AssertMultiple/
fluentassertions.autofac/
fluentassertions.bUnit/
FluentAssertions.Extensions/
FluentAssertions.Http/
fluentassertions.ioc.ninject/
fluentassertions.json/
FluentAssertions.LanguageExt/
fluentassertions.mvc/
FluentAssertions.NodaTime/
fluentassertions.reactive/
FluentAssertions.Web/
FluentResults/
Flushot/
functional-maybe-extensions/
Functional.Utilities/
Nut.Results/
playwright-dotnet-contrib/
RecShark/
RestVerifier/
SerilogSinksInMemory/
WireMock.Net/
xassert/
```

logiclrd:
..but a merge conflict on `releases.md` provides an excuse for a new commit :-)

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
dennisdoomen: Gee. I think you need to find a hobby 😅

dennisdoomen:
Thanks for the great work.

PR #1809, opened 2022-02-16
Fix typo in comment for ComparingRecordsByMembers

kijanawoodard:
This appeared to be copy/pasted from `ComparingRecordsByValue`

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
@@ -576,7 +576,7 @@ namespace FluentAssertions.Equivalency
         }
 
         /// <summary>
-        /// Ensures records by default are compared by its members even though it may override
+        /// Ensures records by default are compared by their members even though they override 
         /// the <see cref="object.Equals(object)" /> method.
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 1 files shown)


dennisdoomen (approved):

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
```diff
@@ -576,7 +576,8 @@ public TSelf ComparingRecordsByValue()
         }
 
         /// <summary>
-        /// Ensures records by default are compared by value instead of their members.
+        /// Ensures records by default are compared by its members even though it may override
```
jnyrup, line 579: Doesn't a record _always_ override `object.Equals`?

My grammar tells me that `its` and `it` should `their` and `they` as `records` is plural.
```
Ensures records by default are compared by their members 
even though they override the <see cref="object.Equals(object)" /> method.
```

kijanawoodard: maybe. that's _me_ doing copy/paste from the comments below and trying to "align them". I didn't think about it much past that point. 

but was my initial suspicion correct that the existing comment is incorrect?

dennisdoomen: You can always count on @jnyrup for a keen eye for detail ;-)

kijanawoodard: my next question is figuring out how to edit the file. 
(I made the PR right in github web) :-D

jnyrup (approved):

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
kijanawoodard: found it!
updated.

PR #1814, opened 2022-02-17
Fill gaps in System.Data unit testing

logiclrd:
This PR backfills missing automated tests for System.Data.

## IMPORTANT 

### DIFF
--- a/Tests/FluentAssertions.Equivalency.Specs/DataRowSpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/DataRowSpecs.cs
@@ -343,7 +343,7 @@ namespace FluentAssertions.Equivalency.Specs
 
             var dataRow = dataSet.TypedDataTable1[0];
 
-            string expectedColumnName = dataSet.TypedDataTable1.Columns.OfType<DataColumn>().Last().ColumnName;
+            string expectedColumnName = dataSet.TypedDataTable1.Columns.Cast<DataColumn>().Last().ColumnName;
 
             // Act & Assert
             dataRow.Should().HaveColumn(expectedColumnName);
@@ -373,7 +373,7 @@ namespace FluentAssertions.Equivalency.Specs
 
             var dataRow = dataSet.TypedDataTable1[0];
 
-            var subsetOfColumnNames = dataRow.Table.Columns.OfType<DataColumn>()
+            var subsetOfColumnNames = dataRow.Table.Columns.Cast<DataColumn>()
                 .Take(dataRow.Table.Columns.Count - 2)
                 .Select(column => column.ColumnName);
 
@@ -389,7 +389,7 @@ namespace FluentAssertions.Equivalency.Specs
 
             var dataRow = dataSet.TypedDataTable1[0];
 
-            var subsetOfColumnNamesWithUnicorn = dataRow.Table.Columns.OfType<DataColumn>()
+            var subsetOfColumnNamesWithUnicorn = dataRow.Table.Columns.Cast<DataColumn>()
                 .Take(dataRow.Table.Columns.Count - 2)
                 .Select(column => column.ColumnName)
                 .Concat(new[] { "Unicorn" });
--- a/Tests/FluentAssertions.Equivalency.Specs/DataRowSpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/DataRowSpecs.cs
@@ -5,328 +5,328 @@ using Xunit.Sdk;
 
 namespace FluentAssertions.Equivalency.Specs
 {
-        public class DataRowSpecs : DataSpecs
+    public class DataRowSpecs : DataSpecs
+    {
+        [Fact]
+        public void When_DataRows_are_identical_it_should_succeed()
         {
-            [Fact]
-            public void When_DataRows_are_identical_it_should_succeed()
-            {
-                // Arrange
-                var dataSet1 = CreateDummyDataSet<TypedDataSetSubclass>();
-                var dataSet2 = new TypedDataSetSubclass(dataSet1);
-
-                var dataTable1 = dataSet1.TypedDataTable1;
-                var dataTable2 = dataSet2.TypedDataTable1;
-
-                // Act & Assert
-                dataTable1[0].Should().BeEquivalentTo(dataTable2[0]);
-            }
-
-            [Fact]
-            public void When_DataRows_are_both_null_it_should_succeed()
-            {
-                // Act & Assert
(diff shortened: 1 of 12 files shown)


Tests/FluentAssertions.Equivalency.Specs/DataRowSpecs.cs
```diff
@@ -1,332 +1,421 @@
...
 
-                var dataSet2 = new TypedDataSetSubclass(dataSet1);
+            // Act & Assert
+            dataRow1.Should().BeEquivalentTo(dataRow2, config => config.ExcludingColumn(dataSet2.TypedDataTable1.DateTimeColumn));
+        }
 
-                var dataRow1 = dataSet1.TypedDataTable1[0];
-                var dataRow2 = dataSet2.TypedDataTable1[0];
+        [Fact]
+        public void When_data_has_column_asserting_HaveColumn_should_succeed()
```
dennisdoomen, line 336: ❌ Keep the names functional and concise without using the literal names of code elements. Same for all the other tests.

Tests/FluentAssertions.Equivalency.Specs/DataRowSpecs.cs
```diff
@@ -1,332 +1,421 @@
...
+            // Arrange
+            var dataSet = CreateDummyDataSet<TypedDataSetSubclass>();
 
-                dataRow1.DateTime = DateTime.UtcNow;
-                dataRow2.DateTime = dataRow1.DateTime;
+            var dataRow = dataSet.TypedDataTable1[0];
 
-                dataSet2.AcceptChanges();
+            // Act & Assert
+            dataRow.Should().HaveColumn("ForeignRowID");
```
dennisdoomen, line 344: 🤔 How do I know from the test that it doesn't have that column?

Tests/FluentAssertions.Equivalency.Specs/DataRowSpecs.cs
```diff
@@ -1,332 +1,421 @@
...
-            {
-                // Arrange
-                var dataSet1 = CreateDummyDataSet<TypedDataSetSubclass>();
+            // Act
+            Action action =
+                () => dataRow.Should().HaveColumn("Unicorn");
 
-                var dataSet2 = new TypedDataSetSubclass(dataSet1);
+            // Assert
+            action.Should().Throw<XunitException>();
```
dennisdoomen, line 360: 🤔 How do I know the exception was raised for the right reasons?
🤔 How do I know if that column exists or not from this test?

Tests/FluentAssertions.Equivalency.Specs/DataRowSpecs.cs
```diff
@@ -1,332 +1,421 @@
...
-                dataSet2.AcceptChanges();
+            var columnNames = dataRow.Table.Columns.OfType<DataColumn>()
+                .Take(3)
+                .Select(column => column.ColumnName);
 
-                // Act & Assert
-                dataRow1.Should().BeEquivalentTo(dataRow2, config => config.Excluding(dataRow => dataRow.RowState));
-            }
+            // Act & Assert
+            dataRow.Should().HaveColumns(columnNames);
```
dennisdoomen, line 376: 🤔 I don't get this assertion. Why are you _taking_ only 3 and then expect this to match to the total number of rows?

Tests/FluentAssertions.Equivalency.Specs/DataSetSpecs.cs
```diff
@@ -610,5 +612,108 @@ public void When_Tables_contain_different_data_it_should_fail()
             // Assert
             action.Should().Throw<XunitException>();
         }
+
+        [Fact]
+        public void When_asserting_table_count_if_it_matches_it_should_succeed()
```
dennisdoomen, line 617: 🙃 That's a weird sentence.

dennisdoomen (changes requested):
Thanks for taking the time for this. I stopped repeating myself, so please consider my comments to be applicable on every new test in the PR.

Tests/FluentAssertions.Equivalency.Specs/DataSetSpecs.cs
dennisdoomen: If C# names supported commas, then no.

Tests/FluentAssertions.Equivalency.Specs/DataRowSpecs.cs
dennisdoomen: But why 3 specifically? I can't see that from the test. If it's just a subset, I would name `columnNames` something like `subsetOfColumnNames` and take the number of columns in that table minus 1.

dennisdoomen:
@eNeRGy164 why does a PR like this not report code coverage?

dennisdoomen (approved):

eNeRGy164:
> @eNeRGy164 why does a PR like this not report code coverage?

Because develop is not (recently) pulled into this branch.

Tests/FluentAssertions.Equivalency.Specs/DataRowSpecs.cs
```diff
@@ -1,332 +1,426 @@
...
+            // Arrange
+            var dataSet = CreateDummyDataSet<TypedDataSetSubclass>();
 
-                var dataRow1 = dataSet1.TypedDataTable1[0];
-                var dataRow2 = dataSet2.TypedDataTable1[0];
+            var dataRow = dataSet.TypedDataTable1[0];
 
-                dataRow1.DateTime = DateTime.UtcNow;
-                dataRow2.DateTime = dataRow1.DateTime;
+            string expectedColumnName = dataSet.TypedDataTable1.Columns.OfType<DataColumn>().Last().ColumnName;
```
jnyrup, line 346: Could this be `.Cast<DataColumn>()` instead?
That avoids unexpectedly leaving any `nulls` out.

I see that `DataRowEquivalencyStep` and `DataSetEquivalencyStep` uses `OfType` as wel..

jnyrup (commented):
Thanks for adding the extra tests.
Please rebase on `develop`, so the PR will trigger the newly added code coverage.

Tests/FluentAssertions.Equivalency.Specs/DataRowSpecs.cs
logiclrd: I don't think it's possible to put a `null` into any of the System.Data collections. `OfType` and `Cast` should be functionally equivalent in this context. I can change them to `Cast`.

logiclrd:
Rebased. :-)

jnyrup (approved):

PR #1820, opened 2022-02-19
More Code Coverage

jnyrup:
* In #1789 we forgot to add the new `FluentAssertions.Equivalency.Specs.csproj` to the Test task.
* Adds some code coverage from the test frameworks that targets net5.0. I couldn't find an easy way to include code coverage for xunit2 and nspec
* Exclude `StringSyntaxAttribute` from code coverage


dennisdoomen (approved):
Awesome. I was already kind of disappointed that our code coverage was so low..

PR #1822, opened 2022-02-20
Better support for NaN in numeric assertions on floats and doubles

### LINKED ISSUE #1690: NumericAssertions does not handle NaN correctly
### Description

`NumericAssertions` uses `CompareTo` to decide if a subject is e.g. smaller than the expected.
That is technically mixing different concepts, as `CompareTo` only decides the relative ordering between two items, i.e. how they should be sorted.
What `NumericAssertions` should be doing when "comparing" is arithmetic comparisons that corresponds to using operators like `<`, `==` and so on.

```csharp
double.NaN.Should().BeLessThan(0);
```

### Expected behavior:
Should fail, as comparing `NaN` to anything using `<`, `<=`, `==`, `>=` or `>` returns false.

### Actual behavior:
The test passes.

### Versions

  * 6.1.0 (master)

dennisdoomen:
Assertions where floats and doubles were compared to NaN did not always work as expected, especially when "larger than",  "smaller than", ranges and approximation were involved. Also, using NaN as an argument is also no longer supported (and not desired).

Fixes #1690

**Note** Review without whitespace

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Numeric/NumericAssertions.cs
+++ b/Src/FluentAssertions/Numeric/NumericAssertions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
@@ -165,7 +165,7 @@ namespace FluentAssertions.Numeric
         public AndConstraint<TAssertions> BeNegative(string because = "", params object[] becauseArgs)
         {
             Execute.Assertion
-                .ForCondition(Subject.HasValue && Subject.Value.CompareTo(default(T)) < 0)
+                .ForCondition(Subject.HasValue && !IsNaN(Subject.Value) && Subject.Value.CompareTo(default(T)) < 0)
                 .BecauseOf(because, becauseArgs)
                 .FailWith("Expected {context:value} to be negative{reason}, but found {0}.", Subject);
 
@@ -185,8 +185,13 @@ namespace FluentAssertions.Numeric
         /// </param>
         public AndConstraint<TAssertions> BeLessThan(T expected, string because = "", params object[] becauseArgs)
         {
+            if (IsNaN(expected))
+            {
+                throw new ArgumentException("A value can never be less than NaN", nameof(expected));
+            }
+           
             Execute.Assertion
-                .ForCondition(Subject.HasValue && Subject.Value.CompareTo(expected) < 0)
+                .ForCondition(Subject.HasValue && !IsNaN(Subject.Value) && Subject.Value.CompareTo(expected) < 0)
                 .BecauseOf(because, becauseArgs)
                 .FailWith("Expected {context:value} to be less than {0}{reason}, but found {1}.", expected, Subject);
 
@@ -207,8 +212,13 @@ namespace FluentAssertions.Numeric
         public AndConstraint<TAssertions> BeLessThanOrEqualTo(T expected, string because = "",
             params object[] becauseArgs)
         {
+            if (IsNaN(expected))
+            {
+                throw new ArgumentException("A value can never be less than or equal to NaN", nameof(expected));
+            }
+            
             Execute.Assertion
-                .ForCondition(Subject.HasValue && Subject.Value.CompareTo(expected) <= 0)
+                .ForCondition(Subject.HasValue && !IsNaN(Subject.Value) && Subject.Value.CompareTo(expected) <= 0)
                 .BecauseOf(because, becauseArgs)
                 .FailWith("Expected {context:value} to be less than or equal to {0}{reason}, but found {1}.", expected, Subject);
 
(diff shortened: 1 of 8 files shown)


Src/FluentAssertions/Numeric/NumericAssertions.cs
```diff
@@ -208,7 +213,7 @@ public AndConstraint<TAssertions> BeLessThan(T expected, string because = "", pa
             params object[] becauseArgs)
         {
```
jnyrup, line 214: Add `IsNaN(expected)` check

Src/FluentAssertions/Numeric/NumericAssertions.cs
```diff
@@ -449,5 +474,7 @@ public AndConstraint<TAssertions> NotBeOfType(Type unexpectedType, string becaus
         /// <inheritdoc/>
         public override bool Equals(object obj) =>
             throw new NotSupportedException("Calling Equals on Assertion classes is not supported.");
+
+        internal virtual bool IsNaN(T value) => false;
```
jnyrup, line 478: If you bother this could be `private protected` for minimum visibility.

Src/FluentAssertions/NumericAssertionsExtensions.cs
```diff
@@ -1181,6 +1196,11 @@ public static class NumericAssertionsExtensions
             double unexpectedValue, double precision, string because = "",
             params object[] becauseArgs)
         {
+            if (double.IsNaN(unexpectedValue))
```
jnyrup, line 1199: This is the only of the eight extension methods on `NullableNumericAssertions` where the `IsNaN` check is added.

dennisdoomen: Unless, I'm missing the point, all of them end up calling the same methods, which are protected by `IsNan`.

jnyrup: Roger.
Then this one is not necessary, as `unexpectedValue` is checked inside the call to `NotBeApproximately` on line 1212.

What triggered me was that I saw 5, i.e. an odd number of added calls to `IsNaN`, either float or double had too many/few calls to `IsNaN`.

Tests/FluentAssertions.Specs/Numeric/NumericAssertionSpecs.cs
```diff
@@ -666,6 +817,54 @@ public void When_a_nullable_numeric_null_value_is_not_greater_than_or_equal_to_i
                     .Should().Throw<XunitException>()
                     .WithMessage("*null*");
             }
+
+            [Fact]
+            public void NaN_is_never_greater_or_equal_to_another_float()
```
jnyrup, line 822: greather_ _than_ _or_equal_to

Tests/FluentAssertions.Specs/Numeric/NumericAssertionSpecs.cs
```diff
@@ -817,6 +1064,54 @@ public void When_a_nullable_numeric_null_value_is_not_less_than_or_equal_to_it_s
                     .Should().Throw<XunitException>()
                     .WithMessage("*null*");
             }
+
+            [Fact]
+            public void NaN_is_never_less_than_equal_to_another_float()
```
jnyrup, line 1069: less_than_ _or_ _equal_to

Tests/FluentAssertions.Specs/Numeric/NumericAssertionSpecs.cs
```diff
@@ -864,6 +1159,72 @@ public void When_a_nullable_numeric_null_value_is_not_in_range_it_should_throw()
                     .Should().Throw<XunitException>()
                     .WithMessage("*null*");
             }
+
+            [Fact]
+            public void A_NaN_is_never_in_range_of_two_floats()
```
jnyrup, line 1164:
```suggestion
            public void NaN_is_never_in_range_of_two_floats()
```

Tests/FluentAssertions.Specs/Numeric/NumericAssertionSpecs.cs
```diff
@@ -864,6 +1159,72 @@ public void When_a_nullable_numeric_null_value_is_not_in_range_it_should_throw()
...
+
+                // Assert
+                act
+                    .Should().Throw<XunitException>()
+                    .WithMessage(
+                        "Expected value to be between*4* and*5*, but found*NaN*");
+            }
+
+            [Fact]
+            public void A_float_can_never_be_in_a_range_containing_NaN()
```
jnyrup, line 1180: Use `InlineData` like for double

Tests/FluentAssertions.Specs/Numeric/NumericAssertionSpecs.cs
```diff
@@ -38,7 +38,7 @@ public void When_a_negative_value_is_positive_it_should_throw()
             public void When_a_zero_value_is_positive_it_should_throw()
             {
                 // Arrange
-                int value = 0;
+                double value = double.NaN;
```
jnyrup, line 41: Was this change intentional?
The method name says "zero value", but body uses NaN

jnyrup (approved):

Tests/FluentAssertions.Specs/Numeric/NumericAssertionSpecs.cs
dennisdoomen: Good catch. Must have been experimenting with something.

PR #1827, opened 2022-02-25
Add the ability to exclude non-browsable members from equivalency tests

logiclrd:
In #1807, changes were made to allow non-browsable members to be excluded from equivalency tests. But, these changes ended up being reverted because they affected processing of both subject & expectation, but include/exclude functions are supposed to apply only to the expectation. In addition, the original changes only worked properly when the subject & expectation were of the same data type. In my actual usage, the subject and expectation are different data types and _only one of them_ marks the member as non-browsable. This wasn't factored in in the original implementation, for which the unit tests exclusively test the case where the subject and expectation are the same data type. The changes in #1807 got reverted since they were in an incomplete/uncertain state and were blocking a release.

This PR makes another attempt at adding this feature. In addition to the tests added in #1807, the PR adds unit tests that comprehensively cover the behaviour when the subject and expectation are different types and only one of them marks a member as non-browsable.

This PR then also provides an implementation that allows excluding non-browsable members from the expectation using an `Excluding` method, and from the subject using a separate `Ignore` method.

### DIFF
--- a/Src/FluentAssertions/Equivalency/Matching/MustMatchByNameRule.cs
+++ b/Src/FluentAssertions/Equivalency/Matching/MustMatchByNameRule.cs
@@ -36,6 +36,13 @@ namespace FluentAssertions.Equivalency.Matching
                     $"Expectation has {expectedMember.Description} that the other object does not have.");
             }
 
+            if (config.IgnoreNonBrowsableOnSubject && !subjectMember.IsBrowsable)
+            {
+                Execute.Assertion.FailWith(
+                    $"Expectation has {expectedMember.Description} that is non-browsable in the other object, and non-browsable " +
+                    $"members on the subject are ignored with the current configuration");
+            }
+
             return subjectMember;
         }
 
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
@@ -326,19 +327,9 @@ namespace FluentAssertions.Equivalency
         }
 
         /// <summary>
-        /// Instructs the comparison to include non-browsable members (members with an EditorBrowsableState of Never).
-        /// </summary>
-        /// <returns></returns>
-        public TSelf IncludingNonBrowsableMembers()
-        {
-            excludeNonBrowsableOnExpectation = false;
-            return (TSelf)this;
-        }
-
-        /// <summary>
-        /// Instructs the comparison to exclude non-browsable members (members with an EditorBrowsableState of Never).
-        /// They can be marked non-browsable in the subject or the expectation. It is not required that they be marked
-        /// non-browsable in both the subject and the expectation.
+        /// Instructs the comparison to exclude non-browsable members in the expectation (members set to
+        /// <see cref="EditorBrowsableState.Never"/>). It is not required that they be marked non-browsable in the subject. Use
+        /// <see cref="IgnoringNonBrowsableMembersOnSubject"/> to ignore non-browsable members in the subject.
         /// </summary>
         /// <returns></returns>
         public TSelf ExcludingNonBrowsableMembers()
@@ -347,6 +338,11 @@ namespace FluentAssertions.Equivalency
             return (TSelf)this;
         }
 
+        /// <summary>
+        /// Instructs the comparison to treat non-browsable members in the subject as though they do not exist. If you need to
+        /// ignore non-browsable members in the expectation, use <see cref="ExcludingNonBrowsableMembers"/>.
+        /// </summary>
+        /// <returns></returns>
         public TSelf IgnoringNonBrowsableMembersOnSubject()
         {
             ignoreNonBrowsableOnSubject = true;
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
@@ -56,6 +56,7 @@ namespace FluentAssertions.Equivalency
 
         private MemberVisibility includedProperties;
         private MemberVisibility includedFields;
+        private bool excludeNonBrowsable;
 
         private bool compareRecordsByValue;
 
@@ -80,6 +81,7 @@ namespace FluentAssertions.Equivalency
             useRuntimeTyping = defaults.UseRuntimeTyping;
             includedProperties = defaults.IncludedProperties;
             includedFields = defaults.IncludedFields;
+            excludeNonBrowsable = defaults.ExcludeNonBrowsable;
             compareRecordsByValue = defaults.CompareRecordsByValue;
 
             ConversionSelector = defaults.ConversionSelector.Clone();
@@ -161,6 +163,8 @@ namespace FluentAssertions.Equivalency
         MemberVisibility IEquivalencyAssertionOptions.IncludedProperties => includedProperties;
 
         MemberVisibility IEquivalencyAssertionOptions.IncludedFields => includedFields;
+
+        bool IEquivalencyAssertionOptions.ExcludeNonBrowsable => excludeNonBrowsable;
 
         public bool CompareRecordsByValue => compareRecordsByValue;
 
@@ -309,6 +313,26 @@ namespace FluentAssertions.Equivalency
         public TSelf ExcludingProperties()
         {
             includedProperties = MemberVisibility.None;
+            return (TSelf)this;
+        }
+
+        /// <summary>
+        /// Instructs the comparison to include non-browsable members (members with an EditorBrowsableState of Never).
+        /// </summary>
+        /// <returns></returns>
+        public TSelf IncludingNonBrowsableMembers()
+        {
+            excludeNonBrowsable = false;
+            return (TSelf)this;
+        }
+
+        /// <summary>
+        /// Instructs the comparison to exclude non-browsable members (members with an EditorBrowsableState of Never).
+        /// </summary>
+        /// <returns></returns>
+        public TSelf ExcludingNonBrowsableMembers()
+        {
+            excludeNonBrowsable = true;
             return (TSelf)this;
         }
 
@@ -725,6 +749,15 @@ namespace FluentAssertions.Equivalency
             else
             {
                 builder.AppendLine("- Compare records by their members");
+            }
+
+            if (excludeNonBrowsable)
+            {
+                builder.AppendLine("- Exclude non-browsable members");
+            }
+            else
+            {
+                builder.AppendLine("- Include non-browsable members");
             }
 
             foreach (Type valueType in valueTypes)
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
@@ -56,7 +56,8 @@ namespace FluentAssertions.Equivalency
 
         private MemberVisibility includedProperties;
         private MemberVisibility includedFields;
-        private bool excludeNonBrowsable;
+        private bool ignoreNonBrowsableOnSubject;
+        private bool excludeNonBrowsableOnExpectation;
 
         private bool compareRecordsByValue;
 
@@ -81,7 +82,8 @@ namespace FluentAssertions.Equivalency
             useRuntimeTyping = defaults.UseRuntimeTyping;
             includedProperties = defaults.IncludedProperties;
             includedFields = defaults.IncludedFields;
-            excludeNonBrowsable = defaults.ExcludeNonBrowsable;
+            ignoreNonBrowsableOnSubject = defaults.IgnoreNonBrowsableOnSubject;
+            excludeNonBrowsableOnExpectation = defaults.ExcludeNonBrowsableOnExpectation;
             compareRecordsByValue = defaults.CompareRecordsByValue;
 
             ConversionSelector = defaults.ConversionSelector.Clone();
@@ -115,6 +117,11 @@ namespace FluentAssertions.Equivalency
                 if (includedFields.HasFlag(MemberVisibility.Public) && !hasConflictingRules)
                 {
                     yield return new AllFieldsSelectionRule();
+                }
+
+                if (excludeNonBrowsableOnExpectation)
+                {
+                    yield return new ExcludeNonBrowsableMembersRule();
                 }
 
                 foreach (IMemberSelectionRule rule in selectionRules)
@@ -164,7 +171,9 @@ namespace FluentAssertions.Equivalency
 
         MemberVisibility IEquivalencyAssertionOptions.IncludedFields => includedFields;
 
-        bool IEquivalencyAssertionOptions.ExcludeNonBrowsable => excludeNonBrowsable;
+        bool IEquivalencyAssertionOptions.IgnoreNonBrowsableOnSubject => ignoreNonBrowsableOnSubject;
+
+        bool IEquivalencyAssertionOptions.ExcludeNonBrowsableOnExpectation => excludeNonBrowsableOnExpectation;
 
         public bool CompareRecordsByValue => compareRecordsByValue;
 
@@ -322,17 +331,25 @@ namespace FluentAssertions.Equivalency
         /// <returns></returns>
         public TSelf IncludingNonBrowsableMembers()
         {
-            excludeNonBrowsable = false;
+            excludeNonBrowsableOnExpectation = false;
             return (TSelf)this;
         }
 
         /// <summary>
         /// Instructs the comparison to exclude non-browsable members (members with an EditorBrowsableState of Never).
+        /// They can be marked non-browsable in the subject or the expectation. It is not required that they be marked
+        /// non-browsable in both the subject and the expectation.
         /// </summary>
         /// <returns></returns>
         public TSelf ExcludingNonBrowsableMembers()
         {
-            excludeNonBrowsable = true;
+            excludeNonBrowsableOnExpectation = true;
+            return (TSelf)this;
+        }
+
+        public TSelf IgnoringNonBrowsableMembersOnSubject()
+        {
+            ignoreNonBrowsableOnSubject = true;
             return (TSelf)this;
         }
 
@@ -722,6 +739,11 @@ namespace FluentAssertions.Equivalency
                 .Append(useRuntimeTyping ? "runtime" : "declared")
                 .AppendLine(" types and members");
 
+            if (ignoreNonBrowsableOnSubject)
+            {
+                builder.AppendLine("- Do not consider members marked non-browsable on the subject");
+            }
+
             if (isRecursive)
             {
                 if (allowInfiniteRecursion)
@@ -751,15 +773,6 @@ namespace FluentAssertions.Equivalency
                 builder.AppendLine("- Compare records by their members");
             }
 
-            if (excludeNonBrowsable)
-            {
-                builder.AppendLine("- Exclude non-browsable members");
-            }
-            else
-            {
-                builder.AppendLine("- Include non-browsable members");
-            }
-
             foreach (Type valueType in valueTypes)
             {
                 builder.AppendLine($"- Compare {valueType} by value");
@@ -768,6 +781,15 @@ namespace FluentAssertions.Equivalency
             foreach (Type type in referenceTypes)
             {
                 builder.AppendLine($"- Compare {type} by its members");
+            }
+
+            if (excludeNonBrowsableOnExpectation)
+            {
+                builder.AppendLine("- Exclude non-browsable members");
+            }
+            else
+            {
+                builder.AppendLine("- Include non-browsable members");
             }
 
             foreach (IMemberSelectionRule rule in selectionRules)
--- a/Src/FluentAssertions/Equivalency/Steps/StructuralEqualityEquivalencyStep.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/StructuralEqualityEquivalencyStep.cs
@@ -55,21 +55,24 @@ namespace FluentAssertions.Equivalency.Steps
             IMember matchingMember = FindMatchFor(selectedMember, context.CurrentNode, comparands.Subject, options);
             if (matchingMember is not null)
             {
-                var nestedComparands = new Comparands
+                if (!options.ExcludeNonBrowsable || matchingMember.IsBrowsable)
                 {
-                    Subject = matchingMember.GetValue(comparands.Subject),
-                    Expectation = selectedMember.GetValue(comparands.Expectation),
-                    CompileTimeType = selectedMember.Type
-                };
+                    var nestedComparands = new Comparands
+                    {
+                        Subject = matchingMember.GetValue(comparands.Subject),
+                        Expectation = selectedMember.GetValue(comparands.Expectation),
+                        CompileTimeType = selectedMember.Type
+                    };
 
-                if (selectedMember.Name != matchingMember.Name)
-                {
-                    // In case the matching process selected a different member on the subject,
-                    // adjust the current member so that assertion failures report the proper name.
-                    selectedMember.Name = matchingMember.Name;
+                    if (selectedMember.Name != matchingMember.Name)
+                    {
+                        // In case the matching process selected a different member on the subject,
+                        // adjust the current member so that assertion failures report the proper name.
+                        selectedMember.Name = matchingMember.Name;
+                    }
+
+                    parent.RecursivelyAssertEquality(nestedComparands, context.AsNestedMember(selectedMember));
                 }
-
-                parent.RecursivelyAssertEquality(nestedComparands, context.AsNestedMember(selectedMember));
             }
         }
 
--- a/Src/FluentAssertions/Equivalency/Steps/StructuralEqualityEquivalencyStep.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/StructuralEqualityEquivalencyStep.cs
@@ -55,24 +55,21 @@ namespace FluentAssertions.Equivalency.Steps
             IMember matchingMember = FindMatchFor(selectedMember, context.CurrentNode, comparands.Subject, options);
             if (matchingMember is not null)
             {
-                if (!options.ExcludeNonBrowsable || matchingMember.IsBrowsable)
+                var nestedComparands = new Comparands
                 {
-                    var nestedComparands = new Comparands
-                    {
-                        Subject = matchingMember.GetValue(comparands.Subject),
-                        Expectation = selectedMember.GetValue(comparands.Expectation),
-                        CompileTimeType = selectedMember.Type
-                    };
+                    Subject = matchingMember.GetValue(comparands.Subject),
+                    Expectation = selectedMember.GetValue(comparands.Expectation),
+                    CompileTimeType = selectedMember.Type
+                };
 
-                    if (selectedMember.Name != matchingMember.Name)
-                    {
-                        // In case the matching process selected a different member on the subject,
-                        // adjust the current member so that assertion failures report the proper name.
-                        selectedMember.Name = matchingMember.Name;
-                    }
-
-                    parent.RecursivelyAssertEquality(nestedComparands, context.AsNestedMember(selectedMember));
+                if (selectedMember.Name != matchingMember.Name)
+                {
+                    // In case the matching process selected a different member on the subject,
+                    // adjust the current member so that assertion failures report the proper name.
+                    selectedMember.Name = matchingMember.Name;
                 }
+
+                parent.RecursivelyAssertEquality(nestedComparands, context.AsNestedMember(selectedMember));
             }
         }
 
@@ -84,6 +81,11 @@ namespace FluentAssertions.Equivalency.Steps
                 let match = rule.Match(selectedMember, subject, currentNode, config)
                 where match is not null
                 select match;
+
+            if (config.IgnoreNonBrowsableOnSubject)
+            {
+                query = query.Where(member => member.IsBrowsable);
+            }
 
             return query.FirstOrDefault();
         }
--- a/Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
@@ -1632,6 +1632,23 @@ namespace FluentAssertions.Equivalency.Specs
         }
 
         [Fact]
+        public void When_non_browsable_property_on_subject_is_ignored_but_is_present_on_expectation_it_should_fail()
+        {
+            // Arrange
+            var subject = new ClassWhereMemberThatCouldBeNonBrowsableIsNonBrowsable() { PropertyThatMightBeNonBrowsable = 0 };
+            var expectation = new ClassWhereMemberThatCouldBeNonBrowsableIsBrowsable() { PropertyThatMightBeNonBrowsable = 1 };
+
+            // Act
+            Action action =
+                () => subject.Should().BeEquivalentTo(expectation, config => config.IgnoringNonBrowsableMembersOnSubject());
+
+            // Assert
+            action.Should().Throw<XunitException>().WithMessage(
+                $"Expectation has * subject.*ThatMightBeNonBrowsable that is non-browsable in the other object, and non-browsable " +
+                $"members on the subject are ignored with the current configuration*");
+        }
+
+        [Fact]
         public void When_property_is_non_browsable_only_in_expectation_excluding_non_browsable_members_should_make_it_succeed()
         {
             // Arrange
--- a/Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
@@ -1723,29 +1723,318 @@ namespace FluentAssertions.Equivalency.Specs
             subject.Should().BeEquivalentTo(expectation, config => config.ExcludingNonBrowsableMembers());
         }
 
+        [Fact]
+        public void When_property_is_non_browsable_only_in_subject_excluding_non_browsable_members_should_not_make_it_succeed()
+        {
+            // Arrange
+            var subject = new ClassWhereMemberThatCouldBeNonBrowsableIsNonBrowsable() { PropertyThatMightBeNonBrowsable = 0 };
+            var expectation = new ClassWhereMemberThatCouldBeNonBrowsableIsBrowsable() { PropertyThatMightBeNonBrowsable = 1 };
+
+            // Act
+            Action action =
+                () => subject.Should().BeEquivalentTo(expectation, config => config.ExcludingNonBrowsableMembers());
+
+            // Assert
+            action.Should().Throw<XunitException>().WithMessage("Expected property subject.PropertyThatMightBeNonBrowsable to be 1, but found 0.*");
+        }
+
+        [Fact]
+        public void When_property_is_non_browsable_only_in_subject_ignoring_non_browsable_members_on_subject_should_make_it_succeed()
+        {
+            // Arrange
+            var subject = new ClassWhereMemberThatCouldBeNonBrowsableIsNonBrowsable() { PropertyThatMightBeNonBrowsable = 0 };
+            var expectation = new ClassWhereMemberThatCouldBeNonBrowsableIsBrowsable() { PropertyThatMightBeNonBrowsable = 1 };
+
+            // Act & Assert
+            subject.Should().BeEquivalentTo(expectation, config => config.IgnoringNonBrowsableMembersOnSubject());
+        }
+
+        [Fact]
+        public void When_property_is_non_browsable_only_in_expectation_excluding_non_browsable_members_should_make_it_succeed()
+        {
+            // Arrange
+            var subject = new ClassWhereMemberThatCouldBeNonBrowsableIsBrowsable() { PropertyThatMightBeNonBrowsable = 0 };
+            var expectation = new ClassWhereMemberThatCouldBeNonBrowsableIsNonBrowsable() { PropertyThatMightBeNonBrowsable = 1 };
+
+            // Act & Assert
+            subject.Should().BeEquivalentTo(expectation, config => config.ExcludingNonBrowsableMembers());
+        }
+
+        [Fact]
+        public void When_field_is_non_browsable_only_in_subject_excluding_non_browsable_members_should_not_make_it_succeed()
+        {
+            // Arrange
+            var subject = new ClassWhereMemberThatCouldBeNonBrowsableIsNonBrowsable() { FieldThatMightBeNonBrowsable = 0 };
+            var expectation = new ClassWhereMemberThatCouldBeNonBrowsableIsBrowsable() { FieldThatMightBeNonBrowsable = 1 };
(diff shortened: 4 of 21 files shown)


dennisdoomen (approved):

jnyrup:
Thanks for dogfooding the feature 👍 

My first concern was that we normally just "let the the expectation drive the comparison", e.g. when using an anonymous object with a subset of members.

But it seems we have approached this issue from the wrong angle.

This example 

```cs
var subject = new { Browsable = 1 };
var expected = new NonBrowable { Browsable = 1, NonBrowsable = 2 };
subject.Should().BeEquivalentTo(expected, opt => opt.ExcludingNonBrowsableMembers());

private class NonBrowable
{
    public int Browsable { get; set; }

    [EditorBrowsable(EditorBrowsableState.Never)]
    public int NonBrowsable { get; set; }
}
```

fails with
```
Expectation has property subject.NonBrowsable that the other object does not have.

With configuration:
- Compare enums by value
- Exclude non-browsable members
```

In short, it fails on the missing `NonBrowsable` on `subject` before it sees that it can be ignored.

The example above works if we instead implement this feature using an `IMemberSelectionRule`.
See https://github.com/jnyrup/fluentassertions/commit/476069d781afa127ff81d96e0e8dc2953f9a7b03

logiclrd:
That's a clever approach :-) I hadn't thought of this possibility at all, but it makes a lot of sense.

I'll try this alternate implementation against our test suites, and I anticipate cherry-picking your commit into this branch :-)

logiclrd:
Hmm, as written it doesn't actually do everything that's needed, because it only excludes members if they're marked non-browsable in the `expectation`. If only the `subject` marks them non-browsable, then this isn't noticed because the filter only applies to members of the `expectation`.

With this implementation, tests `When_field_is_non_browsable_only_in_subject_excluding_non_browsable_members_should_make_it_succeed` and `When_property_is_non_browsable_only_in_subject_excluding_non_browsable_members_should_make_it_succeed` fail.

Is there a way to resolve this? Or is it fundamentally impossible to have a selection rule guide itself based on aspects of the subject?

logiclrd:
I've done a more direct fix. We can backtrack if there is a more elegant solution.

Src/FluentAssertions/Equivalency/Matching/MustMatchByNameRule.cs
```diff
@@ -32,8 +32,13 @@ public IMember Match(IMember expectedMember, object subject, INode parent, IEqui
 
             if (subjectMember is null)
             {
-                Execute.Assertion.FailWith(
-                    $"Expectation has {expectedMember.Description} that the other object does not have.");
+                bool ignoreMemberBecauseItIsNotBrowsable = config.ExcludeNonBrowsable && !expectedMember.IsBrowsable;
+
+                if (!ignoreMemberBecauseItIsNotBrowsable)
```
jnyrup, line 37: `IMemberMatchingRule`s are currently implemented as additive functions, i.e. if a member on `subject` matching `expectedMember` cannot be found, it will try the next rule.

See `StructuralEqualityEquivalencyStep.FindMatchFor`
```cs
private static IMember FindMatchFor(IMember selectedMember, INode currentNode, object subject,
    IEquivalencyAssertionOptions config)
{
    IEnumerable<IMember> query =
        from rule in config.MatchingRules
        let match = rule.Match(selectedMember, subject, currentNode, config)
        where match is not null
        select match;

    return query.FirstOrDefault();
}
```

What we want (it seems) here is to discard the `match` if either `selectedMember` or `match` is non-browsable.

So the check can not only be inside `MustMatchBeNameRule` as another `IMemberMatchingRule`, e.g. `TryMatchByNameRule` or even a user-provided `IMemberMatchingRule` could find a `match`.

logiclrd: It seems that this breaks the nice abstraction & isolation of the design of the match rules. Now, before including a member from the subject, it needs to check for a matching rule in the expectation and vice versa. This is feeling like more code, more convoluted code, and less well-structured code. Are you sure it's preferable??

jnyrup: Currently I don't have any idea how to solve the problem at all.
I just point out what does _not_ work.
Don't take it as critique, I'm just pressure testing the code to avoid releasing a half-baked feature.

dennisdoomen: I would not expect the `ExcludingNonBrowsableMembers` to have any affect on the subject. If the expectation contains a property which is non-browsable, and you use the `ExcludingNonBrowsableMembers` option, it should just ignore that property and don't bother with it. Implementing that as a `IMemberSelectionRule` won't work either, since any of them can readd that property. I think you can solve this completely in the `StructuralEqualityEquivalencyStep` by first removing it from the final collection that `GetMembersFromExpectation` produces and then removing the subject property in `AssertMemberEquality` as you do now.

logiclrd: Is that a preferable way to implement the feature? It seems neither offers perfect isolation, but the current implementation excludes both subject and expectation fields in the same `if` condition, and has one extra `if` to avoid complaining about missing members if they're excluded. If I'm understanding correctly, what you propose would use one mechanism for the expectation by post-processing the results of the selection rules, and then another, different mechanism for the subject. It should have the same effect but it feels more convoluted to me.

dennisdoomen: In my humble opinion, it puts the logic where it belongs since these are two separate steps in the algorithm
1. Determine the members of the expectation to include
2. Find the matching member on the subject

logiclrd: Hmm, okay. :-) I'll give it a go, see if I can't make it look like you're imagining.

logiclrd: Okay that was a smaller change than I was expecting :-)

logiclrd: Even better, I moved the subject check from `AssertMemberEquality` to `FindMatchFor` and it's now very symmetrical :-)

dennisdoomen: But then you don't need these changes in `MustMatchByNameRule`, right?

logiclrd: Indeed!

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1723,29 +1723,231 @@ public void When_non_browsable_property_differs_excluding_non_browsable_members_
             subject.Should().BeEquivalentTo(expectation, config => config.ExcludingNonBrowsableMembers());
         }
 
+        [Fact]
+        public void When_property_is_non_browsable_only_in_subject_excluding_non_browsable_members_should_make_it_succeed()
+        {
+            // Arrange
+            var subject = new ClassWithNonBrowsableMembers() { NonBrowsableProperty = 0 };
+            var expectation = new ClassWithMatchingMemberNamesButEverythingIsBrowsable() { NonBrowsableProperty = 1 };
```
dennisdoomen, line 1731: 🤔 This confuses me. The property is called `NonBrowsableProperty`, but it is still browsable.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1723,29 +1723,231 @@ public void When_non_browsable_property_differs_excluding_non_browsable_members_
             subject.Should().BeEquivalentTo(expectation, config => config.ExcludingNonBrowsableMembers());
         }
 
+        [Fact]
+        public void When_property_is_non_browsable_only_in_subject_excluding_non_browsable_members_should_make_it_succeed()
```
dennisdoomen, line 1727: 🤔 Shouldn't this test fail? The `expectation` has a property that is browsable, but the equally named property on the `subject` is not-browsable and thus missing.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: I thought `Class With Matching Member Names But Everything Is Browsable` would explain it, but I can make make data types specific to that test.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: No, `Exclude Non Browsable Members` doesn't mean "pretend that anything marked non-browsable doesn't exist", it means "if you see a member marked non-browsable, exclude it from the entire comparison process".

In my immediate use case, I have a legacy data type in a NuGet package that I can't update, and I have a new & better data type with mostly matching members in the codebase under test. The new & better data type includes a couple of members that the old type contains, but they aren't functional. They throw NIE and they're marked non-browsable. I can't go back in time and mark the corresponding properties in the legacy library as non-browsable.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
dennisdoomen: It's not a critical issue, but I'm pretty sure it confuses us again if we look at it in 6 months.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
dennisdoomen: This is not very intuitive to me. Up to now, methods like `Including` and `Excluding` only affect the expectation, and never the subject. What do you think @jnyrup ?

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: I reworked those tests so that the member name is now `PropertyThatMightBeNonBrowsable` and the type name is very clear about whether it does or does not actually mark it non-browsable.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
jnyrup: I'm pretty occupied these days, but should have time during the weekend to look more into this.

jnyrup: I agree with with @dennisdoomen that only the members to select on the expectation should be affected.

Be aware that `BeEquivalentTo` is not meant to be a symmetrical assertion.

```cs
var subject = new { Important = 1, Unimportant = 2 };
var expected = new { Important = 1 };

subject.Should().BeEquivalentTo(expected);
```

Src/FluentAssertions/Equivalency/Steps/StructuralEqualityEquivalencyStep.cs
```diff
@@ -85,6 +82,13 @@ public class StructuralEqualityEquivalencyStep : IEquivalencyStep
                 where match is not null
                 select match;
 
+            // This excludes non-browsable members from the subject. If a member is marked non-browsable
+            // in the expectation, it gets excluded in GetMembersFromExpectation.
+            if (config.ExcludeNonBrowsable)
```
jnyrup, line 87: > Implementing that as a IMemberSelectionRule won't work either, since any of them can readd that property.

Isn't it fine for a subsequent rule to readd a non-browsable member?

```cs
private class MyClass
{
    [EditorBrowsable(EditorBrowsableState.Never)]
    public int NonBrowsable { get; set; }

    public int Dummy { get; set; }
}

var subject = new { Dummy = 1, NonBrowsable = 1 };
var expected = new MyClass { Dummy = 1, NonBrowsable = 2 };

subject.Should().NotBeEquivalentTo(expected, opt => opt
    .ExcludingNonBrowsableMembers()
    .Including(e => e.NonBrowsable));
```

logiclrd:
> I agree with with @dennisdoomen that only the members to select on the expectation should be affected.
> 
> Be aware that BeEquivalentTo is not meant to be a symmetrical assertion.

Okay, this makes sense, _but_, how, then, do I achieve my required use case, which is specifically that non-browsable members of the _subject_ should be skipped?

Src/FluentAssertions/Equivalency/Steps/StructuralEqualityEquivalencyStep.cs
logiclrd: If "include" and "exclude" are meant to be cumulative edits to the set and specifically refer to expectation members, then perhaps `ExcludingNonBrowsableMembers` isn't the right name for this functionality -- it shouldn't be lumped in with those.

What about renaming it to something like `SkipAllNonBrowsableMembers` so that there's no confusion about it being associated with "include" and "exclude" functions??

dennisdoomen:
> Okay, this makes sense, _but_, how, then, do I achieve my required use case, which is specifically that non-browsable members of the _subject_ should be skipped?

If the expectation doesn't contain that property or it is excluded, it should not even look for a property on the subject.

Src/FluentAssertions/Equivalency/Steps/StructuralEqualityEquivalencyStep.cs
dennisdoomen: Why do you think `ExcludingNonBrowsableMembers` isn't the right name?

logiclrd:
The expectation _does_ contain the property. It is marked non-browsable on the subject only.

Src/FluentAssertions/Equivalency/Steps/StructuralEqualityEquivalencyStep.cs
logiclrd: Because it currently operates as a blanket flag. It doesn't update the set cumulatively. If it's called, then at the end, any member that is marked non-browsable is excluded, even if you call `.Including` on it explicitly as shows in @jnyrup's example.

dennisdoomen:
> The expectation _does_ contain the property. It is marked non-browsable on the subject only.

Ah, I didn't get that scenario. Maybe by having an additional option called `TreatingNonBrowsableMembersAsMissing` so that the default `ThrowingOnMissingMembers` will do what you had in mind.

logiclrd:
Mm, so then ... `TreatingNonBrowsableMembersAsMissing` would make the logic pretend that the member straight up didn't exist on the subject, and if you wanted that situation to be a soft "ignore" then you simply don't call `ThrowingOnMissingMembers`? That could work, though, it does make it impossible to both ignore non-browsable members _and_ throw a hard error if members are _actually_ missing...

dennisdoomen:
> Mm, so then ... `TreatingNonBrowsableMembersAsMissing` would make the logic pretend that the member straight up didn't exist on the subject, and if you wanted that situation to be a soft "ignore" then you simply don't call `ThrowingOnMissingMembers`?

No, `ThrowingOnMissingMembers` is the default. But you can override that using `ExcludingMissingMembers`. 

> That could work, though, it does make it impossible to both ignore non-browsable members _and_ throw a hard error if members are _actually_ missing...

Sorry, I don't think I get that scenario. If you exclude a member from the expectation, than it's completely ignored.

logiclrd:
How about we keep the functionality the way it's currently implemented, but rename it from `.ExcludingNonBrowsableMembers`, which by convention _should_ only look at the expectation, to `.DisregardingAllNonBrowsableMembers`, which makes it clear that it isn't a part of the `Including`/`Excluding` world that only implicates the expectation?

logiclrd:
@jnyrup @dennisdoomen How about my suggestion of simply changing the name so that this feature isn't implicitly expected to follow the same rules as other `Including`/`Excluding` methods?

dennisdoomen:
It's kind of a niche feature (which nobody asked for in over 10 years), so either we do it correctly (using two options like `TreatingNonBrowsableMembersAsMissing` and `ExcludingNonBrowsableMembers`) or we don't. FYI, we've reverted the previous PR to keep `develop` consistent until we've agreed on a proper solution (which I just did).

logiclrd:
That's fair. But, with `TreatingNonBrowsableMembersAsMissing` and `ExcludingNonBrowsableMembers` (on the expectation only), how do you exclude members from consideration that are non-browsable on the subject only??

dennisdoomen:
`TreatingNonBrowsableMembersAsMissing ` would only affect the subject.

logiclrd:
Ah, okay, and the default logic doesn't care if the expectation has extra members? So then, with what you describe, those two methods are basically doing exactly the same thing, just one on the subject and one on the expectation.

The wording `TreatingNonBrowsableMembersAsMissing` doesn't feel "clear" to me -- at first glance, it seems to suggest that non-browsable members will be an _error_ condition, and it isn't obvious that it's referring to the subject. What about `IgnoringSubjectNonBrowsableMembers`??

dennisdoomen:
> Ah, okay, and the default logic doesn't care if the expectation has extra members?

It does. The expectation defines the members that are expected. 

> What about IgnoringSubjectNonBrowsableMembers??

I chose `TreatingNonBrowsableMembersAsMissing` because it felt like a good alignment to `ThrowingOnMissingMembers`

But maybe `IgnoringSubjectsNonBrowsableMembers` (with the extra `s`) or `IgnoringNonBrowsableMembersOnSubject` would be better. I'm fine with both.

logiclrd:
In previous discussion, the `Including` and `Excluding` methods were described as sort of "editors" that are used to build up the desired state in a stepwise manner. For instance, you ought to be able to do something like:
```
.ExcludingProperties()
.Including(e => e.SpecificProperty)
```
But, looking at the implementation, it doesn't seem to work that way. The various `Excluding` and `Including` methods aren't actually editing a list and building it up as they go, they're just setting flags that get processed later on down the line. Only the `.Excluding` and `.Including` members defined on the `EquivalencyAssertionOptions` subclass work by building up a sequence of changes to the state.

Furthermore, it appears that if you explicitly include any member, then the default inclusion of public properties & fields is disabled. So, it isn't possible to do something like:
```
.ExcludingNonBrowsableMembers()
.Including(e => e.ExceptForThisSpecificNonBrowsableMember)
```
The current logic sees the `.Including` and disables the default inclusion of members, and this can't be overridden even by explicitly trying to include them:
```
.IncludingProperties()
.ExcludingNonBrowsableMembers()
.Including(e => e.ExceptForThisSpecificNonBrowsableProperty)
```
...because the `.IncludingProperties` isn't actually a selection rule in the sequence, it's just a flag.

I think if we want this to work consistently and in an easy-to-understand way, then this needs to be reworked so that `AllPropertiesSelectionRule` and `AllFieldsSelectionRule` can be explicitly specified in the sequence of selections.

I'm inclined to instead make an `IMemberSelectionRule` that excludes non-browsable members, and return it explicitly prior to any custom selection rules. This is a smaller change to the code. But, it does mean that the logic continues to have special cases rather than being truly a generic sequence of edits.

logiclrd:
I have rebased the changes here, collapsing the commits that were part of #1807 into a single commit, and the edits made in this PR into a single commit. The character of the changes changed sufficiently that I didn't see any value in preserving the edit history that went down a different path. With this rebase, this PR can be merged and will reintroduce the changes reverted in #1842 as a base onto which the new approach separating the subject from the expectation is applied.

dennisdoomen:
> I'm inclined to instead make an `IMemberSelectionRule` that excludes non-browsable members, and return it explicitly prior to any custom selection rules. This is a smaller change to the code. But, it does mean that the logic continues to have special cases rather than being truly a generic sequence of edits

I guess in the current state of the code base, there's no other way to make this work.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1489,5 +1490,551 @@ public void Including_an_interface_property_through_inheritance_should_work()
                 .Including(a => a.Value2)
                 .RespectingRuntimeTypes());
         }
+
+        [Fact]
+        public void When_browsable_field_differs_including_non_browsable_members_should_not_affect_result()
```
dennisdoomen, line 1495: 🤔 Why would including non browsable members effect the test at all? It will throw regardless and your assertion is too generic to see the difference.
⛏️ Since you can't use commas in test names, add `then` to make it readable, e.g.  `When_browsable_field_differs_then_including_non_browsable_members_should_not_affect_result`

Same comment on all other similar tests

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1489,5 +1490,551 @@ public void Including_an_interface_property_through_inheritance_should_work()
...
+            // Act
+            Action action =
+                () => subject.Should().BeEquivalentTo(expectation, config => config.IncludingNonBrowsableMembers());
+
+            // Assert
+            action.Should().Throw<XunitException>();
+        }
+
+        [Fact]
+        public void When_non_browsable_field_differs_including_non_browsable_members_should_not_affect_result()
```
dennisdoomen, line 1585: 🤔 This _should_ affect the result, shouldn't it?

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1489,5 +1490,551 @@ public void Including_an_interface_property_through_inheritance_should_work()
...
+            // Act
+            Action action =
+                () => subject.Should().BeEquivalentTo(expectation, config => config.IncludingNonBrowsableMembers());
+
+            // Assert
+            action.Should().Throw<XunitException>();
+        }
+
+        [Fact]
+        public void When_browsable_field_differs_excluding_non_browsable_members_should_not_affect_result()
```
dennisdoomen, line 1615: 🤔 Same comment as for `When_browsable_field_differs_including_non_browsable_members_should_not_affect_result`. There are no non-browsable fields which value differs, so `ExcludingNonBrowsableMembers` would never affect the test.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1489,5 +1490,551 @@ public void Including_an_interface_property_through_inheritance_should_work()
...
+            // Act
+            Action action =
+                () => subject.Should().BeEquivalentTo(expectation, config => config.ExcludingNonBrowsableMembers());
+
+            // Assert
+            action.Should().Throw<XunitException>();
+        }
+
+        [Fact]
+        public void When_explicilty_browsable_property_differs_excluding_non_browsable_members_should_not_affect_result()
```
dennisdoomen, line 1690: ⛏️ Typo in "explicilty"

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
```diff
@@ -312,6 +325,34 @@ public TSelf ExcludingProperties()
...
+        /// non-browsable in both the subject and the expectation.
+        /// </summary>
+        /// <returns></returns>
+        public TSelf ExcludingNonBrowsableMembers()
+        {
+            excludeNonBrowsableOnExpectation = true;
+            return (TSelf)this;
+        }
+
+        public TSelf IgnoringNonBrowsableMembersOnSubject()
```
dennisdoomen, line 350: 🔧 Please add the relevant XML comments and mention it in the docs.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1489,5 +1490,551 @@ public void Including_an_interface_property_through_inheritance_should_work()
...
+
+            // Act & Assert
+            subject.Should().BeEquivalentTo(expectation, config => config.ExcludingNonBrowsableMembers());
+        }
+
+        [Fact]
+        public void When_property_is_non_browsable_only_in_subject_excluding_non_browsable_members_should_not_make_it_succeed()
+        {
+            // Arrange
+            var subject = new ClassWhereMemberThatCouldBeNonBrowsableIsNonBrowsable() { PropertyThatMightBeNonBrowsable = 0 };
```
dennisdoomen, line 1730: 🤔 Why is this property named with "might" even though it is marked with `EditorBrowsableState.Never`?

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1489,5 +1490,551 @@ public void Including_an_interface_property_through_inheritance_should_work()
...
+
+        [Fact]
+        public void When_property_is_non_browsable_only_in_subject_ignoring_non_browsable_members_on_subject_should_make_it_succeed()
+        {
+            // Arrange
+            var subject = new ClassWhereMemberThatCouldBeNonBrowsableIsNonBrowsable() { PropertyThatMightBeNonBrowsable = 0 };
+            var expectation = new ClassWhereMemberThatCouldBeNonBrowsableIsBrowsable() { PropertyThatMightBeNonBrowsable = 1 };
+
+            // Act & Assert
+            subject.Should().BeEquivalentTo(expectation, config => config.IgnoringNonBrowsableMembersOnSubject());
```
dennisdoomen, line 1749: 🤔 This should throw right? Because the expectation has a property `PropertyThatMightBeNonBrowsable` that it will try to find on the subject.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1489,5 +1490,551 @@ public void Including_an_interface_property_through_inheritance_should_work()
...
+            var expectation = new ClassWhereMemberThatCouldBeNonBrowsableIsNonBrowsable() { FieldThatMightBeNonBrowsable = 1 };
+
+            // Act & Assert
+            subject.Should().BeEquivalentTo(expectation, config => config.ExcludingNonBrowsableMembers());
+        }
+
+        public class NonBrowsableOnOneButMissingFromTheOther
+        {
+            [Fact]
+            public void When_property_is_missing_from_subject_excluding_non_browsable_members_should_make_it_succeed()
```
dennisdoomen, line 1803: 🤔 How do these tests differ from the ones above?

docs/_pages/releases.md
```diff
@@ -13,6 +13,7 @@ sidebar:
 * Annotated `[Not]MatchRegex(string)` with `[StringSyntax("Regex")]` which IDEs can use to colorize the regular expression argument - [#1816](https://github.com/fluentassertions/fluentassertions/pull/1816)
 * Added support for .NET6 `DateOnly` struct - [#1844](https://github.com/fluentassertions/fluentassertions/pull/1844)
 * Added support for .NET6 `TimeOnly` struct - [#1848](https://github.com/fluentassertions/fluentassertions/pull/1848)
+* Added the ability to exclude fields & properties marked as non-browsable in the code editor from structural equality equivalency comparisons - [#1807](https://github.com/fluentassertions/fluentassertions/pull/1807) & [#1812](https://github.com/fluentassertions/fluentassertions/pull/1812)
```
dennisdoomen, line 16: ⛏️ "equality equivalency" is superfluous

dennisdoomen (changes requested):
We're almost there ;-)

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
```diff
@@ -312,6 +325,34 @@ public TSelf ExcludingProperties()
...
+        /// <returns></returns>
+        public TSelf IncludingNonBrowsableMembers()
+        {
+            excludeNonBrowsableOnExpectation = false;
+            return (TSelf)this;
+        }
+
+        /// <summary>
+        /// Instructs the comparison to exclude non-browsable members (members with an EditorBrowsableState of Never).
+        /// They can be marked non-browsable in the subject or the expectation. It is not required that they be marked
```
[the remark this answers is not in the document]
logiclrd: Indeed, out-of-date documentation. Oops :-)

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: Yeah I think I'll just remove the `IncludingNonBrowsableMembers` method. All it does is undoes the flag that `ExcludingNonBrowsableMembers` sets. In retrospect, I don't really see a use case for this.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: Unnecessary test is gone now anyway. :-)

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: The property has to have the exact same name in two different classes, and in one of them it is marked with `EditorBrowsableState.Never` and the other one it isn't.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: I thought members on the expectation that were missing on the subject were simply ignored.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: The ones above test the case where both the subject and the expectation have the property or field, but it's marked non-browsable in one and then the corresponding exclude/ignore functionality is activated.

The tests in this nested class test the case where only one of the subject and the expectation _actually has the member at all_, and they test that if it's marked non-browsable and the corresponding exclude/ignore functionality is activated, then the fact that it's missing from the other type should be ignored.

docs/_pages/releases.md
logiclrd: Well, the type in the code is called `StructuralEqualityEquivalencyStep`. I guess I can omit the word "equivalency", but that's why it's there. :-)

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: Or is it the other way around?

logiclrd: Hmm, my use case requires a function like this which does _not_ throw. I don't want to use `.ExcludingMissingMembers` (which, I just noticed, is actually affecting how the _subject_ is processed despite having the word `Excluding` in its name 🤔), because then members that _aren't_ non-browsable that are missing won't cause test failures.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
dennisdoomen: Ah yes, unless you use `WithMapping` of course ;-)

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
dennisdoomen: Damn. I can't see the forest from the trees with all those cases ;-)

docs/_pages/releases.md
dennisdoomen: Hmm, that's a stupid name indeed. But it's `public`, so we can't rename it.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: I'm looking at the code in `StructuralEqualityEquivalencyStep` and I don't see why it would throw:
```
                IMember[] selectedMembers = GetMembersFromExpectation(context.CurrentNode, comparands, context.Options).ToArray();
                if (context.CurrentNode.IsRoot && !selectedMembers.Any())
                {
                    throw new InvalidOperationException(
                        "No members were found for comparison. " +
                        "Please specify some members to include in the comparison or choose a more meaningful assertion.");
                }

                foreach (IMember selectedMember in selectedMembers)
                {
                    AssertMemberEquality(comparands, context, nestedValidator, selectedMember, context.Options);
                }
```
So, it'll find all the members on the expectation, and then for each one it calls `AssertMemberEquality`:
```
        private static void AssertMemberEquality(Comparands comparands, IEquivalencyValidationContext context,
            IEquivalencyValidator parent, IMember selectedMember, IEquivalencyAssertionOptions options)
        {
            IMember matchingMember = FindMatchFor(selectedMember, context.CurrentNode, comparands.Subject, options);
            if (matchingMember is not null)
            {
                ...
            }
        }
```
If `FindMatchFor` returns `null` then it simply does nothing. Here's `FindMatchFor` before this PR:
```
        private static IMember FindMatchFor(IMember selectedMember, INode currentNode, object subject,
            IEquivalencyAssertionOptions config)
        {
            IEnumerable<IMember> query =
                from rule in config.MatchingRules
                let match = rule.Match(selectedMember, subject, currentNode, config)
                where match is not null
                select match;

            return query.FirstOrDefault();
        }
```
So, if the subject is missing a member found on the expectation, `FindMatchFor` returns `null` and `AssertMemberEquality` simply does nothing.

This PR extends `FindMatchFor` so that it won't find non-browsable members if `IgnoreNonBrowsableOnSubject` is set in `config`.

Am I missing something??

dennisdoomen: Sorry, I'm lost now. It's probably all the double negatives in that sentence 🤔.

With respect to `ExcludingMissingMembers`, yes, it should have been called something like `IgnoringMissingMembers`. I think that in previous major versions the comparison was driven through the subject instead of the expectation.

dennisdoomen: > I'm looking at the code in StructuralEqualityEquivalencyStep and I don't see why it would throw

Because `MustMatchByNameRule` will throw.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: Should I be doing that in these tests??

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: Hmm, but in practice, it doesn't throw. The tests as written pass. 🤔

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
dennisdoomen: Up to you. It may help understand the test better.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: Ohh, it's because of where it's excluding the non-browsable members. It does it _after_ the matching rule runs. So, `MustMatchByNameRule` doesn't throw because it _does_ find a matching member, but then the _call site_ throws it out.

logiclrd: I've devised a fix for this. :-) It'll be in the next commit.

logiclrd: And the test still works, it just has to include `.ExcludingMissingMembers()` in the config as well.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
logiclrd: I think it would make the tests more complex, and also mean that they're testing things beyond their scope.

dennisdoomen: Debatable since it would have prevented me from asking the original question.

logiclrd: But then the test is, in a very real sense, testing multiple things simultaneously. If `WithMapping` breaks, then these tests break too as a side-effect.

dennisdoomen: That's a non-issue. We do that all the time when we use `Should().Throw()`

dennisdoomen (approved):

Src/FluentAssertions/Equivalency/Matching/MustMatchByNameRule.cs
```diff
@@ -36,6 +36,13 @@ public IMember Match(IMember expectedMember, object subject, INode parent, IEqui
                     $"Expectation has {expectedMember.Description} that the other object does not have.");
             }
 
+            if (config.IgnoreNonBrowsableOnSubject && !subjectMember.IsBrowsable)
```
jnyrup, line 39: Some observations:

I'm not sure what case we are testing here (I've been a bit out of the loop) and no tests fails if I remove it.

As this is defined in `MustMatchByNameRule` it won't be executed when combing with `ExcludingMissingMembers`.

This passes
```cs
var subject = new NonBrowable { Browsable = 1, NonBrowsable = 2 };
var expected = new NonBrowable { Browsable = 1, NonBrowsable = 3 };

subject.Should().BeEquivalentTo(expected, opt => opt
    .IgnoringNonBrowsableMembersOnSubject());
```

This fails
```cs
var subject = new NonBrowable { Browsable = 1, NonBrowsable = 2 };
var expected = new NonBrowable { Browsable = 1, NonBrowsable = 3 };

subject.Should().BeEquivalentTo(expected, opt => opt
    .ExcludingMissingMembers()
    .IgnoringNonBrowsableMembersOnSubject());
```

logiclrd: Mmm, indeed, this isn't covered by a test. This was added in response to feedback from @dennisdoomen. You can see it in action by removing the call to `.ExcludingMissingMembers` from the test `When_property_is_non_browsable_only_in_subject_ignoring_non_browsable_members_on_subject_should_make_it_succeed`.

logiclrd: I've added a test that covers this case in f730ff4. It fails if the lines in question are removed from `MustMatchByNameRule.cs`.

logiclrd: The short explanation is that `.IgnoringNonBrowsableMembersOnSubject()` is supposed to pretend that the members _don't exist on the subject at all_, which means that if they _are_ present on the expectation, then without `.ExcludingMissingMembers()`, it should fail the equivalency test -- but without these lines, that failure doesn't happen because `MustMatchByNameRule` has already finished executing before `IgnoreNonBrowsableOnSubject` is processed by `FindMatchFor` in `StructuralEqualityEquivalencyStep`, and the member isn't _actually_ physically missing.

logiclrd:
(rebased to resolve `release.md` conflict)

jnyrup (approved):

dennisdoomen:
Wow. That was a big and long-running PR.

PR #1839, opened 2022-03-07
Improve code coverage and fix some test mutations

eNeRGy164:
Improve code coverage and kill mutations

* Split some test files up
* Renamed some test
* Added some tests
* Improves messages

## IMPORTANT 


Tests/FluentAssertions.Specs/Exceptions/ExceptionAssertionSpecs.cs
```diff
@@ -896,6 +896,20 @@ public void When_invoking_a_function_on_a_null_subject_it_should_throw()
                 .WithParameterName("subject");
         }
 
+        [Fact]
+        public void When_invoking_a_function_with_null_it_should_throw()
```
dennisdoomen, line 900: Would be cool if you could start using a shorter naming convention, like `Invoking_with_null_is_not_allowed`

eNeRGy164:
@dennisdoomen As `Invoking` is actually part of `AssertionExtensions`, should I move these specs? They are not really part of `Exceptions` specs

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -887,8 +886,8 @@ public void When_asserting_a_document_has_a_null_root_element_it_should_fail()
             Action act = () => theDocument.Should().HaveRoot(null);
 
             // Assert
-            act.Should().Throw<ArgumentNullException>().WithMessage(
-                "Cannot assert the document has a root element if the element name is <null>*");
+            act.Should().ThrowExactly<ArgumentNullException>().WithMessage(
```
dennisdoomen, line 889: ❓ Why was that necessary? Do we care?

Tests/FluentAssertions.Specs/Execution/MessageFormatingSpecs.cs
```diff
@@ -99,52 +103,15 @@ public void When_an_assertion_fails_on_ContainSingle_succeeding_message_should_b
...
             };
 
-            // Assert
-            act.Should().Throw<XunitException>()
-                .WithMessage("Expected*to contain a single item, but the collection is empty*" +
-                "Expected*to contain a single item, but the collection is empty*");
-        }
-
-        [Fact]
-        public void When_an_assertion_fails_on_BeOfType_succeeding_message_should_be_included()
```
dennisdoomen: ❓ Why were they removed?

Tests/FluentAssertions.Specs/Execution/ContextDataSpecs.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+using FluentAssertions.Execution;
+using Xunit;
+using Xunit.Sdk;
+
+namespace FluentAssertions.Specs.Execution
+{
+    public class ContextDataSpecs
```
dennisdoomen, line 8: 🤔 Aren't they still `AssertionScope` tests and thus should be named accordingly?

dennisdoomen (approved):
Nice work!

Tests/FluentAssertions.Specs/Execution/MessageFormatingSpecs.cs
eNeRGy164: I couldn't think of a reason why this should be a test related to formatting. It is covered by type tests related to `BeOfType`.

Tests/FluentAssertions.Specs/Execution/ContextDataSpecs.cs
eNeRGy164: Yeah, I was not sure on this, I had it as `AssertionScope.ContextDataSpecs.cs` and `AssertionScopeContextDataSpecs.cs` at some point. But as `AssertionScopeSpecs` is really about live cycle, and I felt this was really about a specific aspect not related to that.
Any suggestion what you would like?

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
eNeRGy164: In general, I try to be as specific as possible when testing for exceptions, so I try to always use `ThrowExactly`.
So, no `AgrumentException` for instance, which would also be fine with `Throw`. But in the end there is currently no difference as long as .NET does not introduce an inherited Exception that inherits from `ArgumentNullException`.

I can revert this change, my point was to have clearer messages, and this was collateral.

jnyrup (commented):
I haven't reviewed 21d2bda0a20835a9b0c6ea349b3946f314a33a50 and 029021d6cab380d2176b138d46328e876ee52f90 rigorously, but I'll trust you that they are purely moving methods around.

Tests/FluentAssertions.Specs/Execution/MessageFormatingSpecs.cs
```diff
@@ -333,11 +333,14 @@ public void Message_should_have_named_placeholder_be_replaced_by_nonreportable_v
             var scope = new AssertionScope();
```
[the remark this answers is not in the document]
eNeRGy164: When I wrote the original test, I didn't know that nonreportable values were available as replacement and could be tested that way, and used `Get<string>` at the time for added code coverage.

However, this resulted in a mutation surviving as the rendering (`requiresFormatting: false` -> `true`) did not have an effect on any unit test;

The same for the `AddReportable` with a defered value. By only testing if the value is in there at the beginning, the mutations for `reportable: true` -> `false` and  `requiresFormatting: false` -> `true` survived.
By making the rendering explicit, I cover these mutations.

To still cover `Get<T>` I made separate tests.

Tests/FluentAssertions.Specs/Execution/MessageFormatingSpecs.cs
eNeRGy164: I can move them to the respective sections? I don't see why they were in the `Message Formatting` section of `AssertionScope`. But maybe it's the test name that is confusing me.

Tests/FluentAssertions.Specs/Execution/MessageFormatingSpecs.cs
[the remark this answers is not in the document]
jnyrup: Roger.
I see that the original `When_adding_non_reportable_value_it_should_be_retrievable_from_context` is now covered by `ContextDataSpecs`.

Tests/FluentAssertions.Specs/Execution/MessageFormatingSpecs.cs
jnyrup: They were probably put here because I did that for `ContainKey` and `ContainSingle` in b4d67280bd2f48fd9aa15125f1157e55f65d7955.

But you're right, they all four belong in their respective Specs files and not in formatting.

eNeRGy164:
> I haven't reviewed [21d2bda](https://github.com/fluentassertions/fluentassertions/commit/21d2bda0a20835a9b0c6ea349b3946f314a33a50) and [029021d](https://github.com/fluentassertions/fluentassertions/commit/029021d6cab380d2176b138d46328e876ee52f90) rigorously, but I'll trust you that they are purely moving methods around.

Yes, intentionally to make reviewing the other changes easier.

Tests/FluentAssertions.Specs/Execution/MessageFormatingSpecs.cs
eNeRGy164: I'll move the original tests around.

Tests/FluentAssertions.Specs/Execution/ContextDataSpecs.cs
dennisdoomen: I prefer `AssertionScope.ContextDataSpecs.cs`.

eNeRGy164: I'll take it the same applies for `ChainingApiSpecs.cs` and `MessageFormattingSpecs.cs`?

eNeRGy164:
Updated after processing remarks

Tests/FluentAssertions.Specs/Types/TypeAssertionSpecs.BeAssignableTo.cs
```diff
@@ -156,6 +157,25 @@ public void When_asserting_a_type_to_be_assignable_to_null_it_should_throw()
...
 
+        [Fact]
+        public void When_an_assertion_fails_on_BeAssignableTo_succeeding_message_should_be_included()
+        {
+            // Act
+            Action act = () =>
+            {
+                using var _ = new AssertionScope();
+                var item = string.Empty;
+                item.Should().BeAssignableTo<int>();
```
jnyrup, line 168: This is the `ReferenceTypeAssertions.BeAssignableTo<T>` and not `TypeAssertions.BeAssignableTo<T>`.

eNeRGy164: oops

jnyrup (approved):

Issue #1754, opened 2021-12-02
Add assertions for net6.0 DateOnly/TimeOnly


jnyrup:
### Description
.NET 6 introduced two new interesting time structures, [`DateOnly`](https://docs.microsoft.com/en-us/dotnet/api/system.dateonly) and [`TimeOnly`](https://docs.microsoft.com/en-us/dotnet/api/system.timeonly), see [this PR](https://github.com/dotnet/runtime/pull/50980).

Currently one can workaround the missing APIs in Fluent Assertions by converting `DateOnly` to `DateTime` and `TimeOnly` to `TimeSpan`.
```cs
new DateOnly().ToDateTime(TimeOnly.MinValue).Should();
new TimeOnly().ToTimeSpan().Should();
```

It would be nice if we had native APIs for asserting on `DateOnly` and `TimeOnly`.
I believe most assertions can simply be copied from the `[Nullable]DateTimeAssertions` and `[Nullable]SimpleTimeSpanAssertions`.

julealgon:
@MullerWasHere can you confirm that you are working on this issue?

MullerWasHere:
@julealgon I was working on it previously, haven't progressed much and I'm lacking time. If you want you can open a PR for it.

iliashkolyar:
@jnyrup
Wondering regarding this one.
Shouldn't there be some prerequisite PR that adds `net6.0` support to all relevant projects and builds before being able to add support to these new time structures?

jnyrup:
My initial thought was that adding .NET 6 could be done as part of adding these assertions, as there otherwise is no/little value in adding the extra target framework.

iliashkolyar:
Well, I tried to do so but it wasn't that straightforward.
There are a several compilation flags that check the `TargetFramework` and Include/Remove dependencies accordingly (like `Common/NullConfigurationStore.cs`). These are spread out across multiple project, and the code analyzers are also configured via these flags.
There are several obsolete methods that fail compilation after the addition of the `net6.0` framework such as `BinaryFormatter.Serialize(Stream, object)` and `BinaryFormatter.Deserialize(Stream)` (https://aka.ms/binaryformatter).

I think that these changes are not in scope for this specific feature implementation and should probably be performed by someone with better familiarity with the project.
What do you think?

jnyrup:
Oh, thought it would be more straight forward. 
I'll take a stab at adding .NET 6.

jnyrup:
[...]

PR #1844, opened 2022-03-11
Adding support for .NET6 DateOnly struct

iliashkolyar:
## IMPORTANT 

Fixes part of issue https://github.com/fluentassertions/fluentassertions/issues/1754


Tests/FluentAssertions.Specs/AssertionExtensionsSpecs.cs
```diff
@@ -32,6 +32,8 @@ private static bool OverridesEquals(Type t)
             return equals is not null;
         }
 
+#if !NET6_0_OR_GREATER
```
iliashkolyar, line 35: I 'commented out' this test since I could not understand how to add my new `FluentAssertions.Primitives.DateTimeOffsetAssertions[TAssertions]` to the `fakeOverloads` that this test expects.

Would be happy if someone points me to the right direction here.

The test failure:
```
Expected fakeOverloads to be a collection with 14 item(s), but {FluentAssertions.Primitives.ReferenceTypeAssertions`2[TSubject,TAssertions], FluentAssertions.Primitives.BooleanAssertions`1[TAssertions], FluentAssertions.Primitives.DateTimeAssertions`1[TAssertions], FluentAssertions.Primitives.DateTimeOffsetAssertions`1[TAssertions], FluentAssertions.Specialized.ExecutionTimeAssertions, FluentAssertions.Primitives.GuidAssertions`1[TAssertions], FluentAssertions.Types.MethodInfoSelectorAssertions, FluentAssertions.Numeric.NumericAssertions`2[TSubject,TAssertions], FluentAssertions.Types.PropertyInfoSelectorAssertions, FluentAssertions.Primitives.SimpleTimeSpanAssertions`1[TAssertions], FluentAssertions.Specialized.TaskCompletionSourceAssertions`1[TSubject], FluentAssertions.Types.TypeSelectorAssertions, FluentAssertions.Primitives.EnumAssertions`2[TEnum,TAssertions]}"
00:44:48 [DBG] "contains 1 item(s) less than"
00:44:48 [DBG] "{FluentAssertions.Primitives.ReferenceTypeAssertions`2[TSubject,TAssertions], FluentAssertions.Specialized.TaskCompletionSourceAssertions`1[T], FluentAssertions.Specialized.ExecutionTimeAssertions, FluentAssertions.Primitives.BooleanAssertions`1[TAssertions], FluentAssertions.Primitives.GuidAssertions`1[TAssertions], FluentAssertions.Primitives.DateTimeAssertions`1[TAssertions], FluentAssertions.Primitives.DateTimeOffsetAssertions`1[TAssertions], FluentAssertions.Primitives.DateOnlyAssertions`1[TAssertions], FluentAssertions.Numeric.NumericAssertions`2[T,TAssertions], FluentAssertions.Primitives.SimpleTimeSpanAssertions`1[TAssertions], FluentAssertions.Types.TypeSelectorAssertions, FluentAssertions.Types.MethodInfoSelectorAssertions, FluentAssertions.Types.PropertyInfoSelectorAssertions, FluentAssertions.Primitives.EnumAssertions`2[TEnum,TAssertions]}.
```

Src/FluentAssertions/Formatting/DateOnlyValueFormatter.cs
```diff
@@ -0,0 +1,33 @@
...
+            return value is DateOnly;
+        }
+
+        public void Format(object value, FormattedObjectGraph formattedGraph, FormattingContext context, FormatChild formatChild)
+        {
+            var dateOnly = (DateOnly)value;
+
+            formattedGraph.AddFragment("<");
+            formattedGraph.AddFragment(dateOnly.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture));
+            formattedGraph.AddFragment(">");
```
jnyrup, line 28:
```suggestion
            formattedGraph.AddFragment(dateOnly.ToString("<yyyy-MM-dd>", CultureInfo.InvariantCulture));
```

Src/FluentAssertions/Extensions/FluentDateOnlyExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+    [DebuggerNonUserCode]
+    public static class FluentDateOnlyExtensions
+    {
+        /// <summary>
+        /// Returns a new <see cref="DateOnly"/> value that is the current <see cref="TimeSpan"/> before the
+        /// specified <paramref name="sourceDateOnly"/>.
+        /// </summary>
+        public static DateOnly Before(this TimeSpan timeDifference, DateOnly sourceDateOnly)
+        {
+            var dayDifference = (int)timeDifference.TotalDays;
```
jnyrup, line 20: I'm not sure about whether we should add these extension methods.

One _could_ write this piece of code, which would fail.
```cs
2.Hours().Before(dateOnly).Should().NotBe(dateOnly);
```

ilia-cy: I get your point.
To me it seems that trying to modify `DateOnly` with hours, minutes, etc will not change the original date (as it doesn't refer to time frames), but I see how it can be confusing.
I'll remove it.

iliashkolyar: I get your point.
To me it seems that trying to modify DateOnly with hours, minutes, etc will not change the original date (as it doesn't refer to time frames), but I see how it can be confusing.
I'll remove it.

Src/FluentAssertions/Primitives/DateOnlyAssertions.cs
```diff
@@ -0,0 +1,527 @@
...
+        /// </summary>
+        /// <param name="expected">The expected value</param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> Be(DateOnly expected, string because = "", params object[] becauseArgs)
```
dennisdoomen, line 61: 🤔 Would it make sense to reuse the `DateTimeAssertions` with additional overloads and change the `Should(this DateOnly`) to reuse `DateTimeAssertions` by casting the `DateOnly` to `DateTime`?

Tests/FluentAssertions.Specs/Primitives/DateOnlyAssertionSpecs.cs
```diff
@@ -0,0 +1,863 @@
...
+{
+    public class DateOnlyAssertionSpecs
+    {
+        #region (Not) Be
+
+        [Fact]
+        public void Should_succeed_when_asserting_dateonly_value_is_equal_to_the_same_value()
+        {
+            // Arrange
+            DateOnly dateOnly = new(2016, 06, 04);
```
dennisdoomen, line 17: 🤔 Looking at this, I'm wondering if we should change the month-based extensions to return `DateOnly` for .NET 6. Because normally you can write that date as `4.April(2016)`.

Tests/FluentAssertions.Specs/Primitives/DateOnlyAssertionSpecs.cs
```diff
@@ -0,0 +1,863 @@
...
+using Xunit;
+using Xunit.Sdk;
+
+#if NET6_0_OR_GREATER
+
+namespace FluentAssertions.Specs.Primitives
+{
+    public class DateOnlyAssertionSpecs
+    {
+        #region (Not) Be
```
dennisdoomen, line 11: 🔧 We should be using the nested class construct as we've been doing recently. See also https://github.com/fluentassertions/fluentassertions/issues/1340

Tests/FluentAssertions.Specs/Primitives/DateOnlyAssertionSpecs.cs
```diff
@@ -0,0 +1,863 @@
...
+        {
+            // Arrange
+            DateOnly dateOnly = new(2016, 06, 04);
+            DateOnly sameDateOnly = new(2016, 06, 04);
+
+            // Act
+            Action act = () => dateOnly.Should().Be(sameDateOnly);
+
+            // Assert
+            act.Should().NotThrow();
```
dennisdoomen, line 24: 🔧 We prefer to not use an action and `NotThrow` if an exception is not excepted. You just call `dateOnly.Should().Be(sameDateOnly)` with a comment statement `// Act/Assert`

dennisdoomen (changes requested):

Tests/FluentAssertions.Specs/Primitives/DateOnlyAssertionSpecs.cs
eNeRGy164: Would be interesting. And if someone wants a datetime as return type, they would have to add time explicitly? 
Something like `4.April(2016).AsDateTime()` or `4.April(2016).WithTime()`?

Src/FluentAssertions/Primitives/DateOnlyAssertions.cs
iliashkolyar: This will lead to failed assertions printing the "time part" of the `DateTime` which is always irrelevant for `DateOnly` structs.

By "additional overloads" do you mean telling the different methods whether to print "date and time" or only "date"?

Tests/FluentAssertions.Specs/Primitives/DateOnlyAssertionSpecs.cs
```diff
@@ -0,0 +1,863 @@
...
+
+            // Act
+            Action act = () => dateOnly.Should().NotBe(otherDateOnly);
+
+            // Assert
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_dateonly_value_is_not_equal_to_a_different_nullable_value_notbe_should_succeed()
```
jnyrup, line 99: The method body is identical to `Should_succeed_when_asserting_dateonly_value_is_not_equal_to_a_different_value` above.

Tests/FluentAssertions.Specs/Primitives/DateOnlyAssertionSpecs.cs
jnyrup: I like the idea.
If you wanted a `DateTime` instead of `DateOnly` you could write `4.April(2016).At(00, 00)`.

A problem here is that if we change the return type of `April(int year)` from `DateTime` to `DateOnly` this example breaks if upgrade FA and then later upgrade to net6.0
```cs
DateTime date = 4.April(2016);
```
with `CS0029	Cannot implicitly convert type 'System.DateOnly' to 'System.DateTime'`

Src/FluentAssertions/Primitives/DateOnlyAssertions.cs
jnyrup: Other reasons why not to reuse `DateTimeAssertions` (directly):
 
 `dateOnly.Should().Subject`  would return a `DateTime` instead of a `DateOnly`.
 
 Assertions on `DateTimeAssertions` that are meaningless to `DateOnly`

* `HaveSecond`/`HaveMinute`/`HaveHour`
* `BeIn(DateTimeKind)`
* `BeSameDateAs` (this is only irrelevant as it is identical to `Be`)
* `BeCloseTo(TimeSpan)`

dennisdoomen: > This will lead to failed assertions printing the "time part" of the DateTime which is always irrelevant for DateOnly structs.
> Other reasons why not to reuse DateTimeAssertions (directly):

Fair enough

Tests/FluentAssertions.Specs/Primitives/DateOnlyAssertionSpecs.cs
dennisdoomen: True. I thought that you just guessed with that error message, but it's the real deal ;-)

Kind of disappointed. It would make total sense to have implicit conversion from `DateOnly` to `DateTime`.

dennisdoomen (approved):
Please make sure that you cover all lines with tests.

Tests/FluentAssertions.Specs/Primitives/DateOnlyAssertionSpecs.cs
jnyrup: There's an issue about it https://github.com/dotnet/runtime/issues/58734

Src/FluentAssertions/Primitives/DateOnlyAssertions.cs
jnyrup: But thanks for bringing up the point to see if it was a feasible path.

jnyrup:
> Please make sure that you cover all lines with tests.

I think the uncovered line is the fake `Should()` overload to guard.
I'll probably try picking up the snippet posted on Slack to write a single test exercising this these overloads.

iliashkolyar:
> > Please make sure that you cover all lines with tests.
> 
> I think the uncovered line is the fake `Should()` overload to guard. I'll probably try picking up the snippet posted on Slack to write a single test exercising this these overloads.

@dennisdoomen, so is there anything else needed?

jnyrup (approved):

docs/_pages/releases.md
```diff
@@ -11,6 +11,7 @@ sidebar:
 
 ### What's New
 * Annotated `[Not]MatchRegex(string)` with `[StringSyntax("Regex")]` which IDEs can use to colorize the regular expression argument - [#1816](https://github.com/fluentassertions/fluentassertions/pull/1816)
+* Added support for .NET6 `DateOnly` struct
```
jnyrup, line 14: Please add a link to this PR like how the line above does.

Src/FluentAssertions/Primitives/NullableDateOnlyAssertions.cs
```diff
@@ -0,0 +1,105 @@
...
+
+namespace FluentAssertions.Primitives
+{
+    /// <summary>
+    /// Contains a number of methods to assert that a nullable <see cref="DateOnly"/> or
+    /// </summary>
+    [DebuggerNonUserCode]
+    public class NullableDateOnlyAssertions : NullableDateOnlyAssertions<NullableDateOnlyAssertions>
+    {
+        public NullableDateOnlyAssertions(DateOnly? expected)
```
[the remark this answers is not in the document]
jnyrup: @iliashkolyar Let's just stay consistent with `DateOnlyAssertions` and rename `expected` to `value` in the constructor of both `NullableDateOnlyAssertions` classes.

PR #1847, opened 2022-03-13
Improve code coverage of fake should overloads

jnyrup:
This PR adds tests covering the fake `Should` overloads.

I first aimed for test that would find all the relevant types itself, but even before reaching that goal the test became quite unreadable.
So went with a `Theory`, which should be manageable for contributors, as `Should_methods_have_a_matching_overload_to_guard_against_chaining_and_constraints` will complain about missing `Should` overloads.

Thanks to @eNeRGy164 for the inspiration.

### DIFF
--- a/Build/Build.cs
+++ b/Build/Build.cs
@@ -50,12 +50,15 @@ class Build : NukeBuild
 
     AbsolutePath ArtifactsDirectory => RootDirectory / "Artifacts";
 
+    AbsolutePath TestResultsDirectory => RootDirectory / "TestResults";
+
     string SemVer;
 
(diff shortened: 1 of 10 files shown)


eNeRGy164:
Totally not important, but would an expression body/lambda style method fix this line counted as not covered?
image

jnyrup:
I should be able to just to annotate `InvalidShouldCall` with `[DoesNotReturn]`
https://github.com/coverlet-coverage/coverlet/pull/904

eNeRGy164:
Nice!

dennisdoomen (approved):
Nice

Issue #1823, opened 2022-02-21
Improve code coverage


jnyrup:
### Description

We now have automatic code coverage as part of the CI pipeline, posting the result to https://coveralls.io/github/fluentassertions/fluentassertions

At the time of writing our line coverage is 95.75% and branch coverage is 92.45%.
We do not seek 100% coverage at all costs as some parts might not/hardly be testable, but there are _clearly_ spots were it's easy to write the missing tests.

Ideally all code should be testable through the public API and shouldn't require any changes to the implementation.
If you think the core project needs changes, please discuss those changes here first.

chvollm:
I looked through a lot of code. From what I see most uncovered is either boilerplate (GetEnumerator, etc.) or "ThisShouldn'tHappen".

dennisdoomen:
You should be able to see what's missing from https://coveralls.io/github/fluentassertions/fluentassertions?branch=develop. You'll need to log on yourself. You can also run the build script locally or use an IDE like Rider to get coverage results.

chvollm:
It might also be possible to improve coverage by removing unnecessary code.

chvollm:
> You should be able to see what's missing from https://coveralls.io/github/fluentassertions/fluentassertions?branch=develop. You'll need to log on yourself. You can also run the build script locally or use an IDE like Rider to get coverage results.

I do use my Visual Studio for this. It works great, I just wanted to stress my findings. But I will dig deeper into the areas where it's possible.

ITaluone:
How can this happen: 
https://github.com/fluentassertions/fluentassertions/blob/d53e8d996fcfb9141a89ebf8230115aebe5263fd/Src/FluentAssertions/Equivalency/Steps/DataSetEquivalencyStep.cs#L31-L34

If not should I remove this (also related to `DataColumEquivalencyStep` and `DataRowEquivalencyStep`)

jnyrup:
Here's an example of how to hit that line

```cs
// Arrange
var subject = new
{
    DataSet = "foobar"
};

var expected = new
{
    DataSet = new DataSet()
};

// Act
Action act = () => subject.Should().BeEquivalentTo(expected);

act.Should().Throw<XunitException>()
    .WithMessage("Expected*System.Data.DataSet*found System.String*");
```

ITaluone:
I have bend over to reach this line of code:
[...]

PR #915, opened 2018-09-16
Report mismatch segment for strings of unequal lengths

krajek:
Proposal to fix #907.

I used the same mechanism to display differences as in part of the code that handles strings with equal lengths, with few adjustments. 

BTW Possible follow-ups: 

- "lazy" version of FailWith, that delays the calculation of message and its params
- display differences improvement reporting line within the text when the first difference occurs, maybe only for multiline subjects

PR #1849, opened 2022-03-14
Removes unnecessary code

chvollm:
## IMPORTANT 

#1823

### DIFF
--- a/Src/FluentAssertions/Primitives/StringEqualityValidator.cs
+++ b/Src/FluentAssertions/Primitives/StringEqualityValidator.cs
@@ -44,26 +44,14 @@ namespace FluentAssertions.Primitives
         {
             int indexOfMismatch = Subject.IndexOfFirstMismatch(Expected, comparisonMode);
 
-            // If there is no difference it means that either
-            // * subject starts with expected or
-            // * expected starts with subject
+            // If there is no difference it means that expected starts with subject and subject is shorter than expected
             if (indexOfMismatch == -1)
             {
-                // If subject is shorter we are sure that expected starts with subject
-                if (Subject.Length < Expected.Length)
-                {
-                    // Subject is shorter so we point at its last character.
-                    // We would like to point at next character as it is the real
-                    // index of first mismatch, but we need to point at character existing in
-                    // subject, so the next best thing is the last subject character.
-                    indexOfMismatch = Math.Max(0, Subject.Length - 1);
-                }
-                else
-                {
-                    // If subject is longer we are sure that subject starts with expected
-                    // and we point at first character after expected.
-                    indexOfMismatch = Expected.Length;
-                }
+                // Subject is shorter so we point at its last character.
+                // We would like to point at next character as it is the real
+                // index of first mismatch, but we need to point at character existing in
+                // subject, so the next best thing is the last subject character.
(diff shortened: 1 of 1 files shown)


Src/FluentAssertions/Primitives/StringEqualityValidator.cs
```diff
@@ -44,26 +44,14 @@ private string GetMismatchSegmentForStringsOfDifferentLengths()
...
-                {
-                    // Subject is shorter so we point at its last character.
-                    // We would like to point at next character as it is the real
-                    // index of first mismatch, but we need to point at character existing in
-                    // subject, so the next best thing is the last subject character.
-                    indexOfMismatch = Math.Max(0, Subject.Length - 1);
-                }
-                else
-                {
-                    // If subject is longer we are sure that subject starts with expected
```
dennisdoomen, line 63: What makes you think this scenario is not necessary anymore?

chvollm: It was red in my VS coverage. I then debugged all corresponding tests, e.g. `When_string_start_is_compared_with_string_that_is_longer_it_should_throw()` , they all don't go into the else branch. I even added a new test to be sure. 
Looking at `IndexOfFirstMismatch()`, it compares Subject with Expected by iterating over the characters of Subject. So the only scenario for -1 can be if Subject is shorter than Expected (or equal) and up until the end of Subject the characters are the same, then it stops at the end of Subject and doesn't compare the rest. Otherwise, if Subject is longer than Expected, we get the position of the first differing character.

jnyrup: I haven't dug into the logic, but have you tried wrapping your tests in an assertion scope to if the logic handles continuing after the first failure?

The commit/PR adding the logic maybe also has some useful insights.

dennisdoomen: It seems that @chvollm is right. That branch isn't covered by any tests and I can't seem to find an edge case where it fails.

dennisdoomen (approved):

Src/FluentAssertions/Primitives/StringEqualityValidator.cs
chvollm: Thanks. I spent quite some time on it but was still a bit unsure, so good to see that you can confirm my observation.

jnyrup (approved):
This code was added in #915.
https://github.com/fluentassertions/fluentassertions/pull/915/files#diff-ca701e8c572b2fa147076087dff48e19fc01b574423edabca1cc8c44e9af96a6R58-R75 adds three possible paths, but only two tests was added.

I tried running this, which should exercise the possible combinations of two strings not being equal.
None of the cases hit 
```cs
[Theory]
[InlineData(null, "")]
[InlineData(null, "a")]
[InlineData("", null)]
[InlineData("", "a")]
[InlineData("a", null)]
[InlineData("a", "")]
[InlineData("a", "aa")]
[InlineData("aa", "a")]
public void MyTestMethod(string a, string b)
{
    // Act
    var ass = new AssertionScope();
    a.Should().Be(b);
}
```

I'm confident with this change 👍

PR #1852, opened 2022-03-16
Fix A_NaN_is_never_in_range_of_two_doubles

jnyrup:
This wasn't caught be code coverage as `BeInRange` is generic and the paths were already covered by `float`.


dennisdoomen (approved):
Sloppy-paste on my side ;-)

PR #1853, opened 2022-03-16
Cleanups in Specs

jnyrup:
Some cleanups in Specs. 
A little dogfooding and getting rid of a distraction method. 

I haven't heard it used before, so I'll pretend I coined that term for "helper methods" that are unhelpful and simply a distraction.

### DIFF
--- a/Tests/FluentAssertions.Specs/Exceptions/InnerExceptionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Exceptions/InnerExceptionSpecs.cs
@@ -50,7 +50,7 @@ namespace FluentAssertions.Specs.Exceptions
         public void WithInnerExceptionExactly_no_parameters_when_subject_throws_subclass_of_expected_inner_exception_it_should_throw_with_clear_description()
         {
             // Arrange
-            var innerException = new ArgumentNullException();
+            var innerException = new ArgumentNullException("InnerExceptionMessage", (Exception)null);
 
             Action act = () => throw new BadImageFormatException("", innerException);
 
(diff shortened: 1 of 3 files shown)


Tests/FluentAssertions.Specs/Exceptions/MiscellaneousExceptionSpecs.cs
```diff
@@ -62,52 +62,33 @@ private static IEnumerable<char> MethodThatUsesYield(string bar)
...
-            {
-                // Assert
-                exc.Message.Should().StartWith(
-                    "Expected exception where (e.Message == \"Error\"), but the condition was not met");
-            }
-            catch (Exception exc)
-            {
-                exc.Message.Should().StartWith(
-                    "Expected exception where (e.Message == \"Error\"), but the condition was not met");
-            }
```
jnyrup, line 110: I have no idea why this was implemented with two `catch`es...

Tests/FluentAssertions.Specs/Exceptions/InnerExceptionSpecs.cs
```diff
@@ -165,77 +139,53 @@ public void WithInnerExceptionExactly_when_subject_throws_expected_inner_excepti
                     .WithInnerExceptionExactly<ArgumentNullException>("because {0} should do just that", "the action");
         }
 
-        private static string BuildExpectedMessageForWithInnerExceptionExactly(string because, string innerExceptionMessage)
```
jnyrup, line 168: One ~helper~ distractor method less 🔥

dennisdoomen:
The reason why I used the old-style `try...catch` construct is to avoid false negatives because a bug in `Should().Throw()` may get hidden by itself.

jnyrup:
> The reason why I used the old-style `try...catch` construct is to avoid false negatives because a bug in `Should().Throw()` may get hidden by itself.

Ahh of course.

I'll revert the ones that had `throw new XunitException("This point should not be reached.")`

dennisdoomen:
I guess it is a bit hard which require explicit `try...catch` and which ones are fine to be dogfooding using `Should().Throw()`.

dennisdoomen (approved):

dennisdoomen:
You are having second thoughts @jnyrup? Or did you lost track of your own PR? 😜

jnyrup:
Partly on both. Been more busy on others PRs.

PR #1859, opened 2022-03-19
Add difference to numeric assertion failure messages

iliashkolyar:
…ctual and expected

## IMPORTANT 

Fixes https://github.com/fluentassertions/fluentassertions/issues/1412.


Src/FluentAssertions/Numeric/NumericAssertions.cs
```diff
@@ -60,7 +63,7 @@ public AndConstraint<TAssertions> Be(T expected, string because = "", params obj
             Execute.Assertion
                 .ForCondition(Subject.HasValue && Subject.Value.CompareTo(expected) == 0)
                 .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context:value} to be {0}{reason}, but found {1}.", expected, Subject);
+                .FailWith("Expected {context:value} to be {0}{reason}, but found {1}" + GenerateDifferenceMessage(expected), expected, Subject);
```
iliashkolyar, line 66: Would be happy to find an alternative to this string concatenation.

I tried using `.FailWith("Expected {context:value} to be {0}{reason}, but found {1}{2}.", expected, Subject, GenerateDifferenceMessage(expected)`
but the `MessageBuilder.FormatArgumentPlaceholders` added the string along with its double quotes:
`"Expected value to be 2 because we want to test the failure message, but found 1" (which differs by -1)."."` 

Any suggestions on how to make it cleaner?

iliashkolyar: I could possibly create a `NumericComparisonResult<T>` class that will hold the `CalculateDifference` result and add that as an argument to `FailWith`.
Then create a dedicated `NumericComparisonResultFormatter` which will hold the logic that currently resides in `GenerateDifferenceMessage`.

Does that make sense?

dennisdoomen (changes requested):
Nice work. I think we should try to fine-tune the output a little bit more.

And please check the code coverage. We should at least remain at the same level.

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.AllSatisfy.cs
```diff
@@ -108,11 +108,11 @@ public void Any_items_not_satisfying_inspector_should_throw()
 *Expected customer.Age to be less than 21, but found 21
 *Expected customer.Items to contain only items satisfying the inspector:
 *At index 0:
-*Expected item to be 3, but found 1
+*Expected item to be 3, but found 1 (which differs by -2)
 *At index 1:
-*Expected item to be 3, but found 2
+*Expected item to be 3, but found 2 (which differs by -1)
```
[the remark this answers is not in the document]
iliashkolyar: For integers you may be right, but what about decimal values?
In those cases I think it does add value to see the difference when it's less than 10 (or even 1 for that matter).
So the real question is whether we run specific "filtering logic" per numeric type?

Tests/FluentAssertions.Specs/Numeric/NumericAssertionSpecs.cs
```diff
@@ -491,7 +491,7 @@ public void When_asserting_that_a_double_value_is_equal_to_a_different_value_it_
                 act
                     .Should().Throw<XunitException>()
                     .WithMessage(
-                        "Expected value to be 3.4 because we want to test the error message, but found 3.5.");
+                        "Expected value to be 3.4 because we want to test the error message, but found 3.5 (which differs by 0.10000000000000009).");
```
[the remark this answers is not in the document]
iliashkolyar: The `difference` is not consistent in different runtime environments (for example in .NET4.7 the difference is 0.1 and in .NET6 the difference is the one you see above).
Something similar to what exists in the [FormatterSpecs file](https://github.com/fluentassertions/fluentassertions/blob/develop/Tests/FluentAssertions.Specs/Formatting/FormatterSpecs.cs#L558).
I changed the test to use `0.1*`, but this is just a workaround.

Regarding your suggestion, how do you see it happening?
What if the `subject` has X numbers in the fraction and the `expected` has Y digits?
Feels to me that this is not properly defined and could lead to misunderstandings (without even addressing the runtime issue mentioned above).

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.AllSatisfy.cs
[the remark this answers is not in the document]
dennisdoomen: > For integers you may be right, but what about decimal values?

There the details are important. 

> So the real question is whether we run specific "filtering logic" per numeric type?

I think we should. And it'll be relatively easy to do that since you have separate assertion classes per numeric type now.

Tests/FluentAssertions.Specs/Numeric/NumericAssertionSpecs.cs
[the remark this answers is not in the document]
dennisdoomen: Maybe by rendering the run-time value and determining the number of fractional digits and taking the highest number? It's just a thought.

dennisdoomen:
One more thought. Apparently there's a [subtle difference](https://twitter.com/MrPeterLMorris/status/1505495747596361729) between _differs_ and _difference_.

Src/FluentAssertions/Numeric/NumericAssertions.cs
```diff
@@ -481,5 +481,53 @@ public AndConstraint<TAssertions> NotBeOfType(Type unexpectedType, string becaus
...
+                return noDifferenceMessage;
+            }
+
+            var difference = CalculateDifferenceForFailureMessage(expected.Value);
+            if (difference is null)
+            {
+                return noDifferenceMessage;
+            }
+
+            if (difference.Value.CompareTo(GetMinimalDifferenceThresholdForFailureMessage()) < 0 ||
```
dennisdoomen, line 524: 🤔 I think you've misunderstood me. What I meant is that if the subject and expectation are both under 10, you don't need FA to show you the difference. 
🤔 I wonder why you decided to introduce so much complexity if the `CalculateDifferenceForFailureMessage` can determine itself whether it's worth to return a difference for small numbers or return `null`.

iliashkolyar: I wanted a single location to decide whether or not to generate the "difference message", which logically should be in the `GenerateDifferenceMessage` method.
Unfortunately, I could not simply run the logic on the result that is returned from `CalculateDifferenceForFailureMessage`, as the result is a `T` on which I cannot perform `<` and `>` operations.
Therefore I went with the approach of comparing the `T` difference with concrete values returned from the Minimal/Maximal methods.

Anyway, your suggestion is a lot simpler.
I'll add logic to specific `CalculateDifferenceForFailureMessage` methods to return `null` if we `subject` and `expectation` are under 10 (int, uint, short, ushort, long, ulong).
And for the rest (decimal, float, double, byte, sbyte) we will always return the message.
WDYT?

dennisdoomen: Yeah, sounds like a good idea.

dennisdoomen: Yes, I agree. But I can imagine that when you have large numbers like 12345784, a difference of 3 might be less obvious to the eye.

docs/_pages/releases.md
```diff
@@ -13,6 +13,7 @@ sidebar:
 * Annotated `[Not]MatchRegex(string)` with `[StringSyntax("Regex")]` which IDEs can use to colorize the regular expression argument - [#1816](https://github.com/fluentassertions/fluentassertions/pull/1816)
 * Added support for .NET6 `DateOnly` struct - [#1844](https://github.com/fluentassertions/fluentassertions/pull/1844)
 * Added support for .NET6 `TimeOnly` struct - [#1848](https://github.com/fluentassertions/fluentassertions/pull/1848)
+* Added comparison difference to numeric assertion failure messages - [#1859](https://github.com/fluentassertions/fluentassertions/pull/1859)
```
dennisdoomen, line 16: ⛏️ I'm nitpicking, but I think you don't need the work _comparison_.

dennisdoomen (approved):

Src/FluentAssertions/Numeric/IntAssertions.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Diagnostics;
+
+namespace FluentAssertions.Numeric
+{
+    /// <summary>
+    /// Contains a number of methods to assert that a <see cref="int"/> is in the expected state.
+    /// </summary>
+    [DebuggerNonUserCode]
+    public class IntAssertions : NumericAssertions<int>
```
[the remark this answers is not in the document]
jnyrup: Should we be consistent in this?
Short -> Int16
Long -> Int64
Float -> Single

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types

Src/FluentAssertions/Numeric/NumericAssertions.cs
```diff
@@ -481,5 +481,27 @@ public AndConstraint<TAssertions> NotBeOfType(Type unexpectedType, string becaus
             throw new NotSupportedException("Calling Equals on Assertion classes is not supported.");
 
         private protected virtual bool IsNaN(T value) => false;
+
+        /// <summary>
+        /// A method to generate additional information upon comparison failures.
+        /// </summary>
+        /// <param name="expected">The value to compare the current numeric value with.</param>
+        /// <returns>
+        /// Returns the difference between the integral number value and the <paramref name="expected" /> value.
```
jnyrup, line 490: This method is not only for integral values, floats, doubles and decimals are e.g. also covered.

Src/FluentAssertions/Numeric/SByteAssertions.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Diagnostics;
+
+namespace FluentAssertions.Numeric
+{
+    /// <summary>
+    /// Contains a number of methods to assert that a <see cref="sbyte"/> is in the expected state.
```
jnyrup, line 6: With the assumption that "sbyte" is pronounced "esbyte"
```suggestion
    /// Contains a number of methods to assert that an <see cref="sbyte"/> is in the expected state.
```

Tests/FluentAssertions.Specs/Numeric/NumericDifferenceAssertionsSpecs.cs
```diff
@@ -0,0 +1,703 @@
...
+            public void When_an_int_difference_between_large_numbers_it_should_add_difference_message(int value, int expected)
+            {
+                // Act
+                Action act = () =>
+                    value.Should().Be(expected, "because we want to test the failure {0}", "message");
+
+                // Assert
+                act
+                    .Should().Throw<XunitException>()
+                    .WithMessage($"Expected value to be {expected} because we want to test the failure message, but found {value} (difference of {value - expected}).");
```
jnyrup, line 40: 🤔 Thinking, should we add the expected difference to the `InlineData` instead of calculating it for the assertion?

Tests/FluentAssertions.Specs/Numeric/NumericDifferenceAssertionsSpecs.cs
```diff
@@ -0,0 +1,703 @@
...
+namespace FluentAssertions.Specs.Numeric
+{
+    public class NumericDifferenceAssertionsSpecs
+    {
+        public class Be
+        {
+            [Theory]
+            [InlineData(8, 5)]
+            [InlineData(1, 9)]
+            public void When_an_int_difference_between_small_numbers_it_should_not_add_difference_message(int value, int expected)
```
jnyrup, line 14: We're trying to move away from the "When_abc_it_should_def" pattern heavily used in these tests as we realized it sometimes led to slightly awkward test names.
The new "pattern" (or lack of) is still about focusing of the behavior and not the implementation details, but relaxing/removing the requirement about having to fit into a rigid naming scheme.
See https://enterprisecraftsmanship.com/posts/you-naming-tests-wrong/ for more details.
 
```suggestion
            public void The_difference_between_small_ints_is_not_included_in_the_message(int value, int expected)
```

Tests/FluentAssertions.Specs/Numeric/NumericDifferenceAssertionsSpecs.cs
```diff
@@ -0,0 +1,703 @@
...
+                const double expected = 1;
+
+                // Act
+                Action act = () =>
+                    value.Should().Be(expected, "because we want to test the failure {0}", "message");
+
+                // Assert
+                act
+                    .Should().Throw<XunitException>()
+                    .WithMessage($"Expected value to be 1.0 because we want to test the failure message, but found {value} (difference of {value - expected}).");
```
jnyrup, line 330: Formatting floating points is culture sensitive due to different decimal point separators.
So this would fail with e.g. Danish culture as `Be` would use the english `.` as decimal separator, but `{value}` would be formatting as `1,5` in Danish.

The solution is to remove the string interpolations and embed the expected values directly.
E.g. 
```cs
WithMessage("Expected value to be 1.0 because we want to test the failure message, but found 1.5 (difference of 0.5).")
```

jnyrup (changes requested):

Src/FluentAssertions/Numeric/NullableIntAssertions.cs
```diff
@@ -0,0 +1,27 @@
...
+    public class NullableIntAssertions : NullableNumericAssertions<int>
+    {
+        public NullableIntAssertions(int? value)
+            : base(value)
+        {
+        }
+
+        private protected override int? CalculateDifferenceForFailureMessage(int expected)
+        {
+            if (Subject!.Value > 0 && Subject!.Value < 10 && expected > 0 && expected < 10)
```
jnyrup, line 18: Could also be simplified with pattern matching as you did with most others.

Src/FluentAssertions/Numeric/NullableShortAssertions.cs
```diff
@@ -0,0 +1,27 @@
...
+        }
+
+        private protected override short? CalculateDifferenceForFailureMessage(short expected)
+        {
+            if (Subject < 10 && expected < 10)
+            {
+                return null;
+            }
+
+            var difference = (short?)(Subject - expected);
```
jnyrup, line 23: Isn't this cast unnecessary?

Src/FluentAssertions/Numeric/NumericAssertions.cs
```diff
@@ -193,8 +193,7 @@ public AndConstraint<TAssertions> BeLessThan(T expected, string because = "", pa
             Execute.Assertion
                 .ForCondition(Subject.HasValue && !IsNaN(Subject.Value) && Subject.Value.CompareTo(expected) < 0)
                 .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context:value} to be less than {0}{reason}, but found {1}.", expected, Subject);
-
+                .FailWith("Expected {context:value} to be less than {0}{reason}, but found {1}" + GenerateDifferenceMessage(expected), expected, Subject);
```
jnyrup, line 196: Nit, restore the empty line

Src/FluentAssertions/Numeric/NullableShortAssertions.cs
iliashkolyar: image

Src/FluentAssertions/Numeric/IntAssertions.cs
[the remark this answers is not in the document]
jnyrup: @iliashkolyar in case this got buried in the other comments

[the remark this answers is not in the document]
iliashkolyar: @jnyrup done.

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
```diff
@@ -1665,6 +1665,10 @@ namespace FluentAssertions.Formatting
 }
 namespace FluentAssertions.Numeric
 {
+    public class ByteAssertions : FluentAssertions.Numeric.NumericAssertions<byte>
```
jnyrup, line 1668: Since we still return `NumericAssertions<byte>` in `AssertionExtensions.cs` can we make these classes `internal`?

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
```diff
@@ -1689,6 +1693,50 @@ namespace FluentAssertions.Numeric
...
+    }
+    public class NullableByteAssertions : FluentAssertions.Numeric.NullableNumericAssertions<byte>
+    {
+        public NullableByteAssertions(byte? value) { }
+    }
+    public class NullableDecimalAssertions : FluentAssertions.Numeric.NullableNumericAssertions<decimal>
+    {
+        public NullableDecimalAssertions(decimal? value) { }
+    }
+    public class NullableDoubleAssertions : FluentAssertions.Numeric.NullableNumericAssertions<double>
```
jnyrup, line 1724: FYI, that you had to add `NullableDoubleAssertions` led me to discover that we should have had it.
So I've opened #1867 to add it.

Src/FluentAssertions/Numeric/NumericAssertions.cs
```diff
@@ -481,5 +481,27 @@ public AndConstraint<TAssertions> NotBeOfType(Type unexpectedType, string becaus
...
+        private string GenerateDifferenceMessage(T? expected)
+        {
+            const string noDifferenceMessage = ".";
+            if (!Subject.HasValue || !expected.HasValue)
+            {
+                return noDifferenceMessage;
+            }
+
+            var difference = CalculateDifferenceForFailureMessage(expected.Value);
+            return difference is null ? noDifferenceMessage : $" (difference of {difference}).";
```
jnyrup, line 504: `{difference}` is formatted using `CurrentCulture`.

If you add `using static System.FormattableString;` to the usings, you can write

```cs
return difference is null ? noDifferenceMessage : Invariant($" (difference of {difference}).");
```

jnyrup:
I didn't get time to look deeper into this, but we probably need to handle overflow in the difference calculation.

```cs
int.MinValue.Should().Be(int.MaxValue);
```

```
Expected int.MinValue to be 2147483647, but found -2147483648 (difference of 1).
```

iliashkolyar:
> I didn't get time to look deeper into this, but we probably need to handle overflow in the difference calculation.
> 
> ```cs
> int.MinValue.Should().Be(int.MaxValue);
> ```
> 
> ```
> Expected int.MinValue to be 2147483647, but found -2147483648 (difference of 1).
> ```

This seems like the best approach - https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/checked

Would you like me to use it?

Src/FluentAssertions/Numeric/NumericAssertions.cs
```diff
@@ -6,17 +6,18 @@
...
+using FluentAssertions.Formatting;
 
 namespace FluentAssertions.Numeric
 {
     /// <summary>
     /// Contains a number of methods to assert that an <see cref="IComparable{T}"/> is in the expected state.
     /// </summary>
     [DebuggerNonUserCode]
-    public class NumericAssertions<T> : NumericAssertions<T, NumericAssertions<T>>
+    public abstract class NumericAssertions<T> : NumericAssertions<T, NumericAssertions<T>>
```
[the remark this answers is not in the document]
dennisdoomen: In #1822, I introduced a `private protected virtual` method. That means people can override `NumericAssertions` but can't handle `NaN` properly, like the `Half` you just mentioned.

jnyrup:
For all integral types I think (only fiddled for a couple of minutes) we could upcast the values to `decimal`.

```cs
private protected override string CalculateDifferenceForFailureMessage(byte expected)
{
    decimal difference = ((decimal)Subject) - expected;
    return difference != 0 ? difference.Tostring(CultureInfo.InvariantCulture) : null;
}
```

Not all `float`s and `double`s are representable by `decimal` 
```cs
private protected override string CalculateDifferenceForFailureMessage(float expected)
{
    try
    {
        decimal difference = ((decimal)Subject) - expected;
        return difference != 0 ? difference.Tostring(CultureInfo.InvariantCulture) : null;
    }
    catch (OverflowException)
    {
        return null;
    }
}
```

This should allow us to print differences that are not presentable with the type itself.
E.g. `byte.MinValue - byte.MaxValue` is -255, but -255 is not representable as a `byte`.

jnyrup:
> Would it be possible to handle these types of tweaks in a future PR?
This PR already holds a lot of content and is becoming a bit hard to maintain (conflicts-wise).

Sure and thanks for pointing out that my comments begin to scope creep 👍 
If you remove the try/catch+checked from this PR, then handling overflow and related tests can be postponed to another PR.

To get this PR ready:
* As the PR title is used when we generate release notes for GitHub releases, it should be a complete sentence, probably quite close to the line you added to `releases.md`
* The code coverage should not decrease.

iliashkolyar:
> > Would it be possible to handle these types of tweaks in a future PR?
> > This PR already holds a lot of content and is becoming a bit hard to maintain (conflicts-wise).
> 
> Sure and thanks for pointing out that my comments begin to scope creep 👍 If you remove the try/catch+checked from this PR, then handling overflow and related tests can be postponed to another PR.
> 
> To get this PR ready:
> 
> * As the PR title is used when we generate release notes for GitHub releases, it should be a complete sentence, probably quite close to the line you added to `releases.md`
> * The code coverage should not decrease.

1. I've updated the PR title. Tell me if that is OK.
2. I'm trying to see how my PR decreased the code coverage. 
Looking [at the report](https://coveralls.io/builds/47778935) I see files that were not modified in this PR, so i'm probably missing something.
Could you please point me to the uncovered code that is related to my PR?
3. I was hoping that we could merge this PR in its current state and deal with the non-integral edge cases later on.
The current state includes handling Overflows and IMO merging it like this will provide better functionality then removing the Overflow handling all-together.
I believe that handling Overflows (without the edge cases in presentable differences) is better than not handling them at all.
WDYT?

jnyrup:
1: The new title is fine.

2: You'll have to login to Coveralls with your GitHub account to see the lines.
Here's a screenshot
image

3: I forgot that [u]short and [s]byte are automatically upcasted to int when performing a subtraction, so they are fine.
Is it intentional that you didn't add `checked` to unsigned integrals?

jnyrup (approved):

jnyrup:
Not sure what goes on with Code Coverage. 
It says that coverage decreased, but doesn't list any files with decreased coverage 😕 
cc: @eNeRGy164

dennisdoomen:
> It says that coverage decreased, but doesn't list any files with decreased coverage

It does. The `DoubleAssertions.CalculateDifferenceForFailureMessage` isn't covered. And `NullableDoubleAssertions` isn't covered in .NET 6.0

iliashkolyar:
> > It says that coverage decreased, but doesn't list any files with decreased coverage
> 
> It does. The `DoubleAssertions.CalculateDifferenceForFailureMessage` isn't covered. And `NullableDoubleAssertions` isn't covered in .NET 6.0

@dennisdoomen, I'm not sure where it is stated or how it is calculated, but I just verified again.
1. [The following test](https://github.com/iliashkolyar/fluentassertions/blob/display-diff-numeric-assertions/Tests/FluentAssertions.Specs/Numeric/NumericDifferenceAssertionsSpecs.cs#L336) covers `DoubleAssertions.CalculateDifferenceForFailureMessage`.
2. [The following test](https://github.com/iliashkolyar/fluentassertions/blob/display-diff-numeric-assertions/Tests/FluentAssertions.Specs/Numeric/NumericDifferenceAssertionsSpecs.cs#L353) covers `NullableDoubleAssertions.CalculateDifferenceForFailureMessage` (In .NET6.0 as well)

dennisdoomen:
Weird.

jnyrup:
Are you looking at the newest run Dennis? 
I recall seeing the missing coverage you refer to in an older coverage report.

dennisdoomen:
I've just requeued it and this is [the result](https://coveralls.io/builds/47943263):

image

iliashkolyar:
> I've just requeued it and this is [the result](https://coveralls.io/builds/47943263):
> 
> image

Since the actual code coverage is missing [from the file](https://coveralls.io/builds/47943263/source?filename=Src%2FFluentAssertions%2FNumeric%2FNullableDoubleAssertions.cs), I don't know which lines it refers as not covered.
As I indicated in the comment above, the `CalculateDifferenceForFailureMessage` is covered.
The only other option is the `IsNaN` method, but it is also covered by [the following test](https://github.com/fluentassertions/fluentassertions/blob/a0f8165b1a4a10328dbc3f5dd96cb0ff88d2f5dd/Tests/FluentAssertions.Specs/Numeric/NullableNumericAssertionSpecs.cs#L100).
So bottom line, I don't really know what's missing.

PR #1860, opened 2022-03-20
[WIP] Add WithStrictTyping options

### LINKED ISSUE #798: Add an option to BeEquivalentTo that compares types of objects recursively.
I am trying to compare AST and found that `BeEquivalentTo` has no option to instruct the library to check types on comparing.  For example, I expect the way to modify the following test code to fail because of class names do not match. This behavior should be recursive (so if property value does not have exactly the same type, comparison should fail). I see that stackoverflow has a [question on this](https://stackoverflow.com/questions/24156925/how-can-i-make-fluentassertions-shouldbeequivalentto-check-for-type-when-compari).

```c#
class OneThing
{
	public string Name => "Test";
}
class OtherThing
{
	public string Name => "Test";
}

[TestMethod]
public void MyTestMethod()
{
	new OneThing().Should().BeEquivalentTo(new OtherThing(), o => o.RespectingRuntimeTypes());
}
```

dennisdoomen:
Closes #798

## IMPORTANT 

Copilot:all

### DIFF
--- a/Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
@@ -161,6 +161,11 @@ namespace FluentAssertions.Equivalency
 
             return this;
         }
+
+        public EquivalencyAssertionOptions<TExpectation> WithStrictTyping()
+        {
+            return this;
+        }
     }
 
     /// <summary>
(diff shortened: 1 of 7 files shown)


dennisdoomen:
> @dennisdoomen just checking here to see if you think this could be implemented in time for v7.

I've been quite busy with other community activities the last two months, but I'll be back on it from the beginning of December. We have no plans for a v7 whatsoever, so plenty of time ;-)

Issue #1868, opened 2022-03-28
StringAssertions - Count regex matches


ITaluone:
### Description

I am using the FA string assertions `MatchRegex()` to check strings. But because the regex pattern have to be really generic to match all future strings, it turned out to be a test case that will mostly ever pass (and therefore is useless).

Now if I could restrict the assertion to a predictable count of regex matches, this could be (for me) a better solution.

### Complete minimal example reproducing the issue

```csharp

class A
{
    public string LongLongString = "Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est (100,00€)Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore (104,23€)magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.";
}

[Fact]
public void Test_Regex_matches()
{
    // Arrange
    var a = new A();
    
    // Act / Assert
    // Matching e.g. "(100,00€)"
    a.LongLongString.Should().MatchRegex(@"(.*\(\d+,\d{2}€\))+", Exactly.Twice());
}

```

### Expected behavior:

I think an overload like the `string.Contains("bla", AtLeast.Once())` should be good, so:
`a.LongLongString().Should().MatchRegex("bla", Exactly.Twice());`

### Actual behavior:

Cannot be done (except I am missing some way to achieve this)

### Versions

* Which version of Fluent Assertions are you using? -> 6.4.0
* Which .NET runtime and version are you targeting? .NET Core 3.1

### Additional Information

Unfortunately `string.Contains("bla", AtLeast.Once())` and similar assertions can not be done, because I am trying to match price values inside the text.

jnyrup:
You can achieve it using the `Match()` assertion, but it's not pretty and the failure message will be poor.

```cs
a.LongLongString.Should().Match(subject => Regex.Matches(subject, @"(.*\(\d+,\d{2}€\))+").Count == 2);
```

ITaluone:
Ok, thank you

[...]

PR #1869, opened 2022-03-28
Added a new overload to `MatchRegex()` to assert on the number of regex matches

IT-VBFK:
Ref. #1868 

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Primitives/StringAssertions.cs
+++ b/Src/FluentAssertions/Primitives/StringAssertions.cs
@@ -387,14 +387,40 @@ namespace FluentAssertions.Primitives
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
+        /// <summary>
+        /// Asserts that a string matches a regular expression with expected occurrence
+        /// </summary>
+        /// <param name="regularExpression">
+        /// The regular expression with which the subject is matched.
+        /// </param>
+        /// <param name="occurrenceConstraint">
+        /// The excpected occurences of regex matches
+        /// </param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
         public AndConstraint<TAssertions> MatchRegex([RegexPattern][StringSyntax("Regex")] string regularExpression,
             OccurrenceConstraint occurrenceConstraint, string because = "", params object[] becauseArgs)
         {
-            MatchRegex(regularExpression, because, becauseArgs);
+            Guard.ThrowIfArgumentIsNull(regularExpression, nameof(regularExpression), "Cannot match string against <null>. Provide a regex pattern or use the BeNull method.");
 
-            MatchRegex(new Regex(regularExpression), occurrenceConstraint, because, becauseArgs);
+            Regex regex;
+            try
+            {
+                regex = new Regex(regularExpression);
+            }
+            catch (ArgumentException)
+            {
+                Execute.Assertion.FailWith("Cannot match {context:string} against {0} because it is not a valid regular expression.",
+                    regularExpression);
+                return new AndConstraint<TAssertions>((TAssertions)this);
+            }
 
-            return new AndConstraint<TAssertions>((TAssertions)this);
+            return MatchRegex(regex, occurrenceConstraint, because, becauseArgs);
         }
 
         /// <summary>
@@ -430,17 +456,53 @@ namespace FluentAssertions.Primitives
             return MatchRegex(regex, because, becauseArgs);
         }
 
+        /// <summary>
+        /// Asserts that a string matches a regular expression with expected occurrence
+        /// </summary>
+        /// <param name="regularExpression">
+        /// The regular expression with which the subject is matched.
+        /// </param>
+        /// <param name="occurrenceConstraint">
+        /// The excpected occurences of regex matches
+        /// </param>
+        /// <param name="because">
+        /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+        /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+        /// </param>
+        /// <param name="becauseArgs">
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
         public AndConstraint<TAssertions> MatchRegex(Regex regularExpression,
             OccurrenceConstraint occurrenceConstraint, string because = "", params object[] becauseArgs)
         {
-            int actual = regularExpression.Matches(Subject).Count;
+            Guard.ThrowIfArgumentIsNull(regularExpression, nameof(regularExpression),
+                    "Cannot match string against <null>. Provide a regex pattern or use the BeNull method.");
 
-            Execute.Assertion
-                .ForConstraint(occurrenceConstraint, actual)
+            var regexStr = regularExpression.ToString();
+            if (regexStr.Length == 0)
+            {
+                throw new ArgumentException(
+                    "Cannot match string against an empty string. Provide a regex pattern or use the BeEmpty method.",
+                    nameof(regularExpression));
+            }
+
+            bool success = Execute.Assertion
+                .ForCondition(Subject is not null)
                 .UsingLineBreaks
                 .BecauseOf(because, becauseArgs)
-                .FailWith($"Expected {{context:string}} to match regex {{0}} {{expectedOccurrence}}{{reason}}, but found it {actual.Times()}.",
-                    regularExpression.ToString());
+                .FailWith("Expected {context:string} to match regex {0}{reason}, but it was <null>.", regexStr);
+
+            if (success)
+            {
+                int actual = regularExpression.Matches(Subject).Count;
+
+                Execute.Assertion
+                    .ForConstraint(occurrenceConstraint, actual)
+                    .UsingLineBreaks
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith($"Expected {{context:string}} to match regex {{0}} {{expectedOccurrence}}{{reason}}, but found it {actual.Times()}.",
+                        regularExpression.ToString());
+            }
 
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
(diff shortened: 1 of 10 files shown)


Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -387,6 +387,16 @@ public AndConstraint<TAssertions> NotMatch(string wildcardPattern, string becaus
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
+        public AndConstraint<TAssertions> MatchRegex([RegexPattern][StringSyntax("Regex")] string regularExpression,
+            OccurrenceConstraint occurrenceConstraint, string because = "", params object[] becauseArgs)
+        {
```
jnyrup, line 392: The two new overloads need to match closer to the existing `MatchRegex` methods in terms of input validation.

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -387,6 +387,16 @@ public AndConstraint<TAssertions> NotMatch(string wildcardPattern, string becaus
...
+            MatchRegex(regularExpression, because, becauseArgs);
```
jnyrup, line 393: We should not call the regular `MatchRegex` overload first as it would make this test fail.
`"a".Should().MatchRegex("b", Times.Exactly(0))`

docs/_pages/strings.md
```diff
@@ -108,3 +108,10 @@ someString.Should().MatchRegex(new System.Text.RegularExpressions.Regex("h.*\\sw
 subject.Should().NotMatchRegex(new System.Text.RegularExpressions.Regex(".*earth.*"));
 subject.Should().NotMatchRegex(".*earth.*");
 ```
+
+And if that's not enough, you can count the occurrence of a regular expression:
```
jnyrup, line 112:
```suggestion
And if that's not enough, you can assert on the number of matches of a regular expression:
```

docs/_pages/strings.md
```diff
@@ -108,3 +108,10 @@ someString.Should().MatchRegex(new System.Text.RegularExpressions.Regex("h.*\\sw
...
+
+```csharp
+someString.Should().MatchRegex("h.*\\sworld.$", Exactly.Once());
+someString.Should().MatchRegex(new System.Text.RegularExpressions.Regex("h.*\\sworld.$"), AtLeast.Twice());
```
jnyrup, line 116:
```suggestion
someString.Should().MatchRegex(new Regex("h.*\\sworld.$"), AtLeast.Twice());
```

docs/_pages/releases.md
```diff
@@ -14,6 +14,8 @@ sidebar:
 * Added support for .NET6 `DateOnly` struct - [#1844](https://github.com/fluentassertions/fluentassertions/pull/1844)
 * Added support for .NET6 `TimeOnly` struct - [#1848](https://github.com/fluentassertions/fluentassertions/pull/1848)
 * Added `NotBe` for nullable boolean values - [#1865](https://github.com/fluentassertions/fluentassertions/pull/1865)
+* Added ability to count regex matches - [#1869](https://github.com/fluentassertions/fluentassertions/pull/1869)
+
```
jnyrup, line 18: This line should somehow include `MatchRegex`

jnyrup (changes requested):

Src/FluentAssertions/Primitives/StringAssertions.cs
IT-VBFK: Ok, do you mean that `Guard.ThrowIfArumentNull` and the `try/catch` block? Is that not considered as code duplication, because it would be copy/paste.

jnyrup: Yes, these are the parts I mean.

I wouldn't consider copying that as code duplication.

IT-VBFK: Oh.. ok then I will treat this two overloads as "independant" from the other two `MatchRegex()` methods?

IT-VBFK: A possible solution could also be to implement a new `Guad` method e.g. `Guard.ThrowIfRegexIsInvalid()` to make the regex testing in one spot...

Src/FluentAssertions/Primitives/StringAssertions.cs
IT-VBFK: ok, I see.. working on a fix in approximately 3 hours..

jnyrup (commented):
Developing on mac should be fine, but I haven't tried.
For the best experience, you probably need to ensure that the version you're using has support for .net 6

Once the Code Coverage pipelines has run (~10 minutes after this comment), it'll probably point out some lines that aren't covered by any tests.

Every time there is some conditional logic, there should be a tests covering both paths.
E.g. there are no tests covering the paths where `Guard.ThrowIfArgumentIsNull` throws an exception or the assertion fails due to an invalid regular expression.

IT-VBFK:
> E.g. there are no tests covering the paths where Guard.ThrowIfArgumentIsNull throws an exception or the assertion fails due to an invalid regular expression.

Is it really necessary to test the `Guard` stuff in every new scenario? They are already covered by other tests and therefore I can assume it works?

But in case they are necessary I will add them.

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -420,6 +456,57 @@ public AndConstraint<TAssertions> NotMatch(string wildcardPattern, string becaus
...
+            if (success)
+            {
+                int actual = regularExpression.Matches(Subject).Count;
+
+                Execute.Assertion
+                    .ForConstraint(occurrenceConstraint, actual)
+                    .UsingLineBreaks
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith($"Expected {{context:string}} to match regex {{0}} {{expectedOccurrence}}{{reason}}, but found it {actual.Times()}.",
+                        regularExpression.ToString());
```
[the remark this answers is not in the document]
IT-VBFK: Oh... what a dumb fail 🙈

IT-VBFK:
Just for clarification: Both account are mine :)

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -394,7 +394,10 @@ public AndConstraint<TAssertions> NotMatch(string wildcardPattern, string becaus
         /// The regular expression with which the subject is matched.
         /// </param>
         /// <param name="occurrenceConstraint">
-        /// The excpected occurences of regex matches
+        /// A constraint specifying the expected amount of times a regex should match a string.
+        /// It can be created by invoking static methods Once, Twice, Thrice, or Times(int)
+        /// on the classes <see cref="Exactly"/>, <see cref="AtLeast"/>, <see cref="MoreThan"/>, <see cref="AtMost"/>, and <see cref="LessThan"/>.
+        /// For example, <see cref="Exactly.Times(int)"/> or <see cref="LessThan.Twice()"/>.
```
eNeRGy164, line 400: 🔁 Newlines in documentation will not be kept when rendering in the IDE. If you intend these lines as separate lines (I think that makes it more readable) surround each line with a `<para></para>` tag

ITaluone: Thanks for the hint.. will add this soon

ITaluone: image

Hmm..

eNeRGy164: Ok, maybe not every IDE, but at least Visual Studio:
image
_(I added the XML to another method, but just for a test)_

eNeRGy164: Ok, apparently Visual Studio only applies line breaks in tooltip to the _summary_, and not the _param_, even with `<para>`.

It does apply the line breaks in the object browser, but that's quite obscure.
image


So forget what I asked 😞  does not seem worth the trouble.

jnyrup:
> Is it really necessary to test the Guard stuff in every new scenario? They are already covered by other tests and therefore I can assume it works?

Yes. 
It's not about testing that `Guard` _itself_ works, but that `MatchRegex` throws the expected exception.
Without such a test I could remove the guard from `MatchRegex` and no test would fail.
You can more or less apply the "The Beyoncé rule" to every feature or line of code - "If you liked it, you should have put a CI test on it".

IT-VBFK:
Ok, this totally makes sense 👍🏻

jnyrup (approved):

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -420,6 +460,60 @@ public AndConstraint<TAssertions> NotMatch(string wildcardPattern, string becaus
...
+        /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+        /// </param>
+        public AndConstraint<TAssertions> MatchRegex(Regex regularExpression,
+            OccurrenceConstraint occurrenceConstraint, string because = "", params object[] becauseArgs)
+        {
+            Guard.ThrowIfArgumentIsNull(regularExpression, nameof(regularExpression),
+                "Cannot match string against <null>. Provide a regex pattern or use the BeNull method.");
+
+            var regexStr = regularExpression.ToString();
+            if (regexStr.Length == 0)
```
dennisdoomen, line 489: 🤔 Isn't this something that the `Regex` class verifies by itself?

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.MatchRegex.cs
```diff
@@ -199,6 +199,159 @@ public void When_a_string_is_matched_against_an_empty_regex_it_should_throw_with
                 .WithParameterName("regularExpression");
         }
 
+        [Fact]
+        public void When_a_string_is_matched_against_a_regex_and_the_count_of_matches_fits_into_the_expected_it_passes()
```
dennisdoomen, line 203: ⛏️ You don't really need to repeat the fact that you're testing against a regex. That's clear from the context.

dennisdoomen (approved):

Src/FluentAssertions/Primitives/StringAssertions.cs
ITaluone: hmm.. I have taken this from the two other MatchRegex overloads.. 

I am rechecking this in a moment

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.MatchRegex.cs
ITaluone: hmm.. true

Src/FluentAssertions/Primitives/StringAssertions.cs
ITaluone: The constructor of Regex checks if the patterns is null but not for empty

PR #1871, opened 2022-03-31
More code coverage

ITaluone:
This ref. #1823 

I don't know if these tests are important, but in case they are.. 

## IMPORTANT 


Src/FluentAssertions/Common/Guard.cs
```diff
@@ -8,10 +8,7 @@ internal static class Guard
     {
         public static void ThrowIfArgumentIsNull<T>([ValidatedNotNull] T obj, string paramName)
         {
-            if (obj is null)
-            {
-                throw new ArgumentNullException(paramName);
-            }
+            ThrowIfArgumentIsNull(obj, paramName, string.Empty);
```
jnyrup, line 11: Are you aware that this changes behavior?
I think the former, i.e. keeping it as it is, is better.
image

Also applies to the change to `ThrowIfArgumentIsNullOrEmpty`.

jnyrup (changes requested):
Besides the single comment I'm good with this.

ITaluone:
I don't know how to arrange a test for `ThrowIfArgumentIsOutOfRange`.. Because `T` is an `enum`

eNeRGy164:
> I don't know how to arrange a test for `ThrowIfArgumentIsOutOfRange`.. Because `T` is an `enum`

Cast any enum to a value that's not listed. For example `(CSharpAccessModifier)1000`

ITaluone:
ahh.. thank you :)

dennisdoomen:
To be honest, I see no value in those extra tests. This is internal code that is already covered by all other code and only exists because of that other code. In fact, `ThrowIfArgumentContainsNull` can even be removed since it is dead code anyway.

IT-VBFK:
Ok, removing dead code counts as "improve code coverage" IMO :)

dennisdoomen:
There's an overload.

IT-VBFK:
How I can see the code coverage on coveralls for the latest CI build?

jnyrup:
I've _just_ started the CI pipeline.
Once it's done in ~10min the Coveralls bot will post a comment in this PR with a summary of the results.
To view the individual files you'll have to login to Coveralls with your GitHub account.

jnyrup:
And if you can't wait for the bot to post, here are the change for `Guard.cs`
https://coveralls.io/builds/47932197/source?filename=Src%2FFluentAssertions%2FCommon%2FGuard.cs

jnyrup:
Instead of adding tests of `ThrowIfArgumentContainsNull<T>(IEnumerable<T> values, string paramName, string message)` we could delete it, as it is dead code.

According to https://coveralls.io/builds/47932179/source?filename=Src%2FFluentAssertions%2FCommon%2FGuard.cs#L58
all the other lines are already covered by other tests.

dennisdoomen:
> Instead of adding tests of `ThrowIfArgumentContainsNull<T>(IEnumerable<T> values, string paramName, string message)` we could delete it, as it is dead code.

That's exactly what I said ;-)

However, I don't get why Coverall did not post an update here. @eNeRGy164 any idea why it sometimes works and sometimes doesn't?

IT-VBFK:
> There's an overload.

Here is an example of what I have thought about to be able to remove `ThrowIfArgumentContainsNull<T>()`.

This applies to all four assertions..

```csharp

public MethodInfoSelector(IEnumerable<Type> types)
{
    types.ToList()
        .ForEach(type => Guard.ThrowIfArgumentIsNull(type, nameof(type)));
    Guard.ThrowIfArgumentIsNull(types, nameof(types));

    selectedMethods = types.SelectMany(t => t
        .GetMethods(BindingFlags.DeclaredOnly | BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic)
        .Where(method => !HasSpecialName(method)));
}

```

because removing the overload would be a breaking change IMO

jnyrup:
It's non-breaking to remove any unused members of `Guard` since the class is `internal`.

So `public static void ThrowIfArgumentContainsNull<T>(IEnumerable<T> values, string paramName)` should be kept as it is used.

`public static void ThrowIfArgumentContainsNull<T>(IEnumerable<T> values, string paramName, string message)` can be removed as it unused and not public because `Guard` is `internal`.

dennisdoomen:
Maybe you're using an IDE that doesn't allow you to find usages, but my IDE (Rider) clearly tells me that the overload of `ThrowIfArgumentContainsNull` that takes three parameters isn't used at all. 

image

ITaluone:
ok, that was completely my fault 🙈 

I have written the tests and therefore it showes up "2 usages".. (the tests)

I should have checked that before writing the tests :D

eNeRGy164:
> However, I don't get why Coverall did not post an update here. @eNeRGy164 any idea why it sometimes works and sometimes doesn't?

At first, there was no run CI initiated by default because it's @ITaluone's first contribution.
But why there was no post after the actual run, I also don't know. I can only guess that maybe the API between Coveralls and GitHub is overloaded or something like that. Can't see anything specific.

dennisdoomen:
> I should have checked that before writing the tests :D

Those tests can be removed anyway.

ITaluone:
ok, got it

jnyrup (approved):

PR #1875, opened 2022-04-03
Improve difference calculation of overflowing integrals

jnyrup:
## IMPORTANT 

Currently we avoid reporting incorrect differences for 32 and 64 bit integrals by doing a `checked` subtraction, catch any `OverflowException` and omit the difference in the failure message.

Instead we can do a widening conversion of `subject` and `expected` and _then_ perform a regular subtraction.
This removes the need for `checked` and try/catch and returns the difference.

### DIFF
--- a/Src/FluentAssertions/Numeric/DecimalAssertions.cs
+++ b/Src/FluentAssertions/Numeric/DecimalAssertions.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
+using System.Globalization;
 
 namespace FluentAssertions.Numeric
 {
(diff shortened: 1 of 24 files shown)


Src/FluentAssertions/Numeric/DecimalAssertions.cs
```diff
@@ -14,12 +15,12 @@ internal DecimalAssertions(decimal value)
...
 
-        private protected override decimal? CalculateDifferenceForFailureMessage(decimal expected)
+        private protected override string CalculateDifferenceForFailureMessage(decimal subject, decimal expected)
         {
             try
             {
-                var difference = checked(Subject - expected);
-                return difference != 0 ? difference : null;
+                decimal difference = checked(subject - expected);
+                return difference != 0 ? difference.ToString(CultureInfo.InvariantCulture) : null;
```
dennisdoomen, line 23: I'm a bit concerned that we're no longer using the value formatters that are specifically tailored for that and add

jnyrup: FYI, we don't do that currently with `Invariant($" (difference of {difference}).")`.

The thing is that we cannot return a difference in type `uint` for two `uints` as the signed difference can be negative.

dennisdoomen (approved):

PR #1889, opened 2022-04-14
Add missing tests for matching `null` with a wildcard

IT-VBFK:
This refs #1823 

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Primitives/StringWildcardMatchingValidator.cs
+++ b/Src/FluentAssertions/Primitives/StringWildcardMatchingValidator.cs
@@ -49,11 +49,6 @@ namespace FluentAssertions.Primitives
 
         private string CleanNewLines(string input)
         {
-            if (input is null)
-            {
-                return null;
-            }
-
             return IgnoreNewLineDifferences ? input.RemoveNewLines() : input;
         }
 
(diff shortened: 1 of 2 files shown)


dennisdoomen (approved):

IT-VBFK:
Since the tests for overridden `Equals` are covered in PR #1891 remove them here

jnyrup:
Was this test intended to add coverage for https://coveralls.io/builds/48309149/source?filename=Src%2FFluentAssertions%2FPrimitives%2FStringWildcardMatchingValidator.cs#L54 ?
As currently it does not add any extra coverage.

In any case, I'm fine with having this test as it acts as documentation of how `Match` should be behave.

IT-VBFK:
Hmm.. true 🤔

That was the intention of this test.. I have no idea how to set up the scenario to get into this path? 

Removing is not an option? 🫢

jnyrup:
The null check in `CleanNewLines` can be removed as it is unreachable.
Line 33 and 35 together validates that both `Subject` and `Expected` are non-null.
https://github.com/fluentassertions/fluentassertions/blob/9e03620a904861a7c5aa66de00283a36155cc8e4/Src/FluentAssertions/Primitives/StringValidator.cs#L33-L35

IT-VBFK:
Yeah.. that what I meant :)

Removing this as well..

jnyrup (approved):

PR #1893, opened 2022-04-16
DataSet Collections: ContainItemWithName

logiclrd:
This PR adds functionality on top of the `DataSet` assertions which was originally written for #1812 but which got moved out for its own independent pull request/evaluation. The changes apply to `DataTableCollection` and `DataColumnCollection`, which have elements that are logically named, and include unit tests and documentation.

## IMPORTANT 


jnyrup:
Also please move the API discussion to an issue.
Note how I edited #1896 with the proposed API.

logiclrd:
API proposal discussion issue: #1897

dennisdoomen:
As discussed in #1897, we decided not to make this API part of the library.

logiclrd:
If I wanted to make my own extension library and make this a part of it, I would run into the problem that the code needs to identify the type `ReadOnlyNonGenericCollectionWrapper<TCollection, TItem>`, and that type is `internal`. What is the recommendation for this situation?

* Reflection? Slow and not type-safe.
* Make the type public? Leaky.
* `[InternalsVisibleTo]` in the main library?
* Other?

Thanks :-)

dennisdoomen:
I can't think of any other way then just making that wrapper public. You can submit a PR for that if you want

logiclrd:
Alright, will do. Thanks :-)

logiclrd:
Oh, huh, the wrapper doesn't exist any more in the mainline. The code in this branch isn't compatible with 6.0 now that non-generic collection support got removed.

logiclrd:
So, I never added unit tests explicitly asserting that basic collection functionality worked on `DataTableCollection`, `DataColumnCollection` and `DataRowCollection`. It "just worked" because these types implement `IEnumerable`. The lack of tests meant that when non-generic collection support was removed, it introduced an undetected regression in this area. I should have explicitly tested those types. Sigh. :-P

PR #1902, opened 2022-04-18
Try to stabilize UIFact tests by running them sequentially

jnyrup:
Everyone's tired of `[UIFact]` tests (semi-)randomly failing and I don't have a clue what's causing the issue.
I initially wanted to use https://github.com/JoshKeegan/xRetry to simply retry them, but it doesn't seem combinable with `UIFact`.

So instead I went with running the `[UIFact]` tests sequentially and crossing fingers it'll improve things.

Disable whitespace while reviewing.

### DIFF
--- a/Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
@@ -84,17 +84,21 @@ namespace FluentAssertions.Specs.Exceptions
             await act2.Should().ThrowAsync<XunitException>();
         }
 
-        [UIFact]
-        public async Task When_async_method_throws_an_empty_AggregateException_on_UI_thread_it_should_fail()
+        [Collection("UIFacts")]
+        public partial class UIFacts
         {
-            // Arrange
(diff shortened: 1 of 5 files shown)


Tests/FluentAssertions.Specs/Exceptions/AsyncFunctionExceptionAssertionSpecs.cs
```diff
@@ -84,17 +84,21 @@ public async Task When_async_method_throws_an_empty_AggregateException_it_should
             await act2.Should().ThrowAsync<XunitException>();
         }
 
-        [UIFact]
-        public async Task When_async_method_throws_an_empty_AggregateException_on_UI_thread_it_should_fail()
+        [Collection("UIFacts")]
```
dennisdoomen, line 87: I'm [pretty sure](https://xunit.net/docs/running-tests-in-parallel) that tests within the same class will never run in parallel. So moving all those tests in the same (partial) class should be enough.

jnyrup: This change should also put tests across different classes in the same test collection to avoid _any_ `UIFact`s being run in parallel.

dennisdoomen (approved):

PR #1913, opened 2022-04-26
Fix the failure message for occurrence regex

ITaluone:
## IMPORTANT 

In my daily work, I was really pissed about the fact, that I have forgotten to include the actual subject in the failure message.
This PR fixes my own fault. :)

### DIFF
--- a/Src/FluentAssertions/Primitives/StringAssertions.cs
+++ b/Src/FluentAssertions/Primitives/StringAssertions.cs
@@ -507,8 +507,9 @@ namespace FluentAssertions.Primitives
                     .ForConstraint(occurrenceConstraint, actual)
                     .UsingLineBreaks
                     .BecauseOf(because, becauseArgs)
-                    .FailWith($"Expected {{context:string}} to match regex {{0}} {{expectedOccurrence}}{{reason}}, but found it {actual.Times()}.",
-                        regexStr);
+                    .FailWith($"Expected {{context:string}} {{0}} to match regex {{1}} {{expectedOccurrence}}{{reason}}, " +
+                        $"but found it {actual.Times()}.",
+                        Subject, regexStr);
             }
 
             return new AndConstraint<TAssertions>((TAssertions)this);
--- a/Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.MatchRegex.cs
+++ b/Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.MatchRegex.cs
@@ -225,7 +225,7 @@ namespace FluentAssertions.Specs.Primitives
 
                 // Assert
                 act.Should().Throw<XunitException>()
-                    .WithMessage($"Expected subject to match regex*\"Lorem.*\" exactly 2 times, but found it 1 time.");
+                    .WithMessage($"Expected subject*Lorem*to match regex*\"Lorem.*\" exactly 2 times, but found it 1 time*");
             }
 
             [Fact]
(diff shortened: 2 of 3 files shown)


Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.MatchRegex.cs
```diff
@@ -252,7 +252,7 @@ public void When_a_string_is_matched_and_the_expected_count_is_zero_and_string_m
 
                 // Assert
                 act.Should().Throw<XunitException>()
-                    .WithMessage($"Expected subject to match regex*\"a\" exactly 0 times, but found it 1 time.");
+                    .WithMessage($"Expected subject to match regex*\"a\" exactly 0 times, but found it 1 time in *{subject}*");
```
jnyrup, line 255: Wouldn't it be shorter to just write `\"a\"` instead of `*{subject}*`.

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -507,8 +507,9 @@ public AndConstraint<TAssertions> NotMatch(string wildcardPattern, string becaus
                     .ForConstraint(occurrenceConstraint, actual)
                     .UsingLineBreaks
                     .BecauseOf(because, becauseArgs)
-                    .FailWith($"Expected {{context:string}} to match regex {{0}} {{expectedOccurrence}}{{reason}}, but found it {actual.Times()}.",
-                        regexStr);
+                    .FailWith($"Expected {{context:string}} to match regex {{0}} {{expectedOccurrence}}{{reason}}, " +
+                    $"but found it {actual.Times()} in {{1}}.",
+                        regexStr, Subject);
```
jnyrup, line 512: Should we use the same pattern for including `Subject` as `Contain(string expected, OccurrenceConstraint occurrenceConstraint` and `ContainEquivalentOf(string expected, OccurrenceConstraint occurrenceConstraint` does?

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.MatchRegex.cs
ITaluone: Yes..

Src/FluentAssertions/Primitives/StringAssertions.cs
ITaluone: Hmm.. fair point 🤔

But what if you have a really long subject? I think this will mess up the failure message more than just appending the subject.

ITaluone: Another posibility is to insert a line break before and/or after subjects contents.

dennisdoomen: Maybe we should consider to update `StringValueFormatter` to add two newlines when the string exceeds a certain length?

ITaluone: Should this be part of this PR?

dennisdoomen: Maybe not. Probably not.

jnyrup: I recall our failure messages in general could use an overhaul to break long lines.
But I agree, that should not be part of this PR.
Let's align with the other methods for now in this PR.

jnyrup (approved):

dennisdoomen (approved):

PR #1469, opened 2021-01-14
Limit formatting to a maximum number of lines

dennisdoomen:
* Formatter.ToString will take an options object that allow you to specify the max depth and the max number of lines to produce
Both will have defaults
* AssertionScope will have properties to override this in case you need it
* Also considering to have the defaults come from AssertionOptions
* Limit formatting to a maximum number of lines

Fixes #1014

**Can be reviewed commit by commit**

PR #1916, opened 2022-04-28
Fix `CA1062`

ITaluone:
https://docs.microsoft.com/de-de/dotnet/fundamentals/code-analysis/quality-rules/ca1062

## IMPORTANT 

Fix some VS informations..

Maybe better to disable globally (because there are 2560 more 🙈 )?

### DIFF
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -3035,6 +3035,7 @@ namespace FluentAssertions.Collections
         protected void AssertCollectionEndsWith<TActual, TExpectation>(IEnumerable<TActual> actual, ICollection<TExpectation> expected, Func<TActual, TExpectation, bool> equalityComparison, string because = "", params object[] becauseArgs)
         {
             Guard.ThrowIfArgumentIsNull(equalityComparison, nameof(equalityComparison));
+            Guard.ThrowIfArgumentIsNull(expected, nameof(expected));
 
             Execute.Assertion
                 .BecauseOf(because, becauseArgs)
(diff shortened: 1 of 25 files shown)


dennisdoomen:
The fact that your changes did not affect any tests proves that the changes were unnecessary. We already ensured that the relevant paths were covered correctly.

jnyrup:
It's mostly important to guard the parts of the public API that are called _directly_ by users - we could call it the "public public API".
E.g. `AggregateExceptionValueFormatter.Format` should only be called from `Formatter.ToString` should be used.
For inspiration see how [this commit](https://github.com/fluentassertions/fluentassertions/pull/1469/files#diff-38976e8d9f6336ed83c4cc8bb5d3ec1dd7d27790e25f8411ac7240f75881c125) changed a test to call `Formatter.ToString` instead of `DateTimeOffsetValueFormatter`.
Also the analyzer can't see that we guard calls to `IValueFormatter.Format` with `IValueFormatter.CanHandle`.
So I don't think we should add any of these null guards.

I'm not sure we should disable the warning as I recall having used it to find places that we definitely needed to guard.

> The fact that your changes did not affect any tests proves that the changes were unnecessary.

That we don't have a test exercising that a null path, does not prove that a null guard is unnecessary.

Here's the logic (I knew that CS degree would come handy).
Let `p` and `q` be the propositions "we have observed an NRE" and "we should have a null guard", respectively.
 
Observing an NRE implies we should have a null guard, so `p → q`.
What I read you stating is because we have _not_ seen an NRE, we should _not_ have a null guard - or in formal logic `¬p → ¬q`.

This truth table show that the two statements are not logically equivalent as they are not true for all combinations of values for `p` and `q`.

| P Q | (P → Q) ⟷ (¬P → ¬Q)  |
|-----|:-:|
| 1 1 | 1 |
| 1 0 | 0 |
| 0 1 | 0 |
| 0 0 | 1 |

jnyrup:
Anyways, we agree that these null guards are not necessary, as in the analyzer is over-cautious, and any nulls passed in to those methods should either be null checked in higher level APIs.

PR #1917, opened 2022-04-28
Add ContainSingle.Where to collection examples

timabell:
This if very useful and not at all obvious from the docs that this is possible.


dennisdoomen:
Thanks

PR #1920, opened 2022-05-03
Do not add all arguments of type `T` to the matching events, if one is found

### LINKED ISSUE #1915: EventRaisingExtensions WithArgs matches arguments with wrong type
### Description

When using the `WithArgs` method in `EventRaisingExtensions`, all occured events are searched for a parameter of the given type. Once a matching parameter is found, all following occured events are added to the `eventsMatchingPredicate`, even if they don't have a matching parameter of the given type

Complete means the code snippet can be copied into a unit test method in a fresh C# project and run.
Minimal means it is stripped from code not related to reproducing the issue.

E.g.

```csharp
using System;
using FluentAssertions;
using FluentAssertions.Events;
using Xunit;

public class A
{
    public event EventHandler<object> Event;

    public void OnEvent(object o)
    {
        Event.Invoke(nameof(A), o);
    }
}

public class B {}

public class C {}

public class FluentAssertionsEventRaisingExtensions
{
    [Fact]
    public void Test()
    {
        A a = new A();
        using var aMonitor = a.Monitor();
        
        a.OnEvent(new B());
        a.OnEvent(new C());

        IEventRecording filteredEvents = aMonitor.GetRecordingFor(nameof(A.Event)).WithArgs<B>();
        filteredEvents.Should().HaveCount(1);
    }
}
```

### Expected behavior:

bool hasArgumentOfRightType gets reset at the end of the foreach loop

### Actual behavior:

Once any parameter matches the type, all following following events are added to the filtered recording.

### Versions

6.6.0
* Which .NET runtime and version are you targeting? E.g. .NET framework 4.6.1 or .NET Core 2.1.
.NET 6

### Additional Information

ITaluone:
This closes #1915

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/EventRaisingExtensions.cs
+++ b/Src/FluentAssertions/EventRaisingExtensions.cs
@@ -62,35 +62,27 @@ namespace FluentAssertions
 
             Func<T, bool> compiledPredicate = predicate.Compile();
 
-            bool hasArgumentOfRightType = false;
-            var eventsMatchingPredicate = new List<OccurredEvent>();
+            var eventsWithMatchingPredicate = new List<OccurredEvent>();
 
             foreach (OccurredEvent @event in eventRecording)
             {
                 var typedParameters = @event.Parameters.OfType<T>().ToArray();
-                if (typedParameters.Any())
-                {
-                    hasArgumentOfRightType = true;
-                }
 
                 if (typedParameters.Any(parameter => compiledPredicate(parameter)))
                 {
-                    eventsMatchingPredicate.Add(@event);
+                    eventsWithMatchingPredicate.Add(@event);
                 }
             }
 
-            if (!hasArgumentOfRightType)
-            {
-                throw new ArgumentException("No argument of event " + eventRecording.EventName + " is of type <" + typeof(T) + ">.");
-            }
+            bool foundMatchingEvent = eventsWithMatchingPredicate.Any();
 
-            if (!eventsMatchingPredicate.Any())
-            {
-                Execute.Assertion
-                    .FailWith("Expected at least one event with arguments matching {0}, but found none.", predicate.Body);
-            }
+            Execute.Assertion
+                .ForCondition(foundMatchingEvent)
+                .FailWith("Expected at least one event which arguments are of type <{0}> and matches {1}, but found none.",
+                    typeof(T),
+                    predicate.Body);
 
-            return new FilteredEventRecording(eventRecording, eventsMatchingPredicate);
+            return new FilteredEventRecording(eventRecording, eventsWithMatchingPredicate);
         }
 
         /// <summary>
@@ -104,16 +96,12 @@ namespace FluentAssertions
         {
             Func<T, bool>[] compiledPredicates = predicates.Select(p => p?.Compile()).ToArray();
 
-            bool hasArgumentOfRightType = false;
-            var eventsMatchingPredicate = new List<OccurredEvent>();
+            var eventsWithMatchingPredicate = new List<OccurredEvent>();
 
             foreach (OccurredEvent @event in eventRecording)
             {
                 var typedParameters = @event.Parameters.OfType<T>().ToArray();
-                if (typedParameters.Any())
-                {
-                    hasArgumentOfRightType = true;
-                }
+                bool hasArgumentOfRightType = typedParameters.Any();
 
                 if (predicates.Length > typedParameters.Length)
                 {
@@ -121,7 +109,7 @@ namespace FluentAssertions
                         $"Expected the event to have at least {predicates.Length} parameters of type {typeof(T)}, but only found {typedParameters.Length}.");
                 }
 
-                bool isMatch = true;
+                bool isMatch = hasArgumentOfRightType;
                 for (int index = 0; index < predicates.Length && isMatch; index++)
                 {
                     isMatch = compiledPredicates[index]?.Invoke(typedParameters[index]) ?? true;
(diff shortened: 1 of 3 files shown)


docs/_pages/releases.md
```diff
@@ -18,6 +18,7 @@ sidebar:
 
 ### Fixes
 * Fix the failure message for regex matches (occurrence overload) to include the missing subject - [#1913](https://github.com/fluentassertions/fluentassertions/pull/1913)
+* Fixed `WithArgs` matching arguments with wrong type, if at least one argument type is matching - [#1920](https://github.com/fluentassertions/fluentassertions/pull/1920)
```
dennisdoomen, line 21: ⛏️ `Fixed `WithArgs` matching too many events when at least one argument matched the expected type`

dennisdoomen (changes requested):
Thanks for fixing this. I only have nitpicks and small suggestions.

Src/FluentAssertions/EventRaisingExtensions.cs
```diff
@@ -62,33 +62,25 @@ public static IEventRecording WithArgs<T>(this IEventRecording eventRecording, E
...
+            bool hasFoundArgumentsBefore = eventsMatchingPredicate.Any();
 
-            if (!eventsMatchingPredicate.Any())
-            {
-                Execute.Assertion
-                    .FailWith("Expected at least one event with arguments matching {0}, but found none.", predicate.Body);
-            }
+            Execute.Assertion
+                .ForCondition(hasFoundArgumentsBefore)
+                .FailWith($"Expected at least one event with arguments matching either the type <{{0}}> or parameters {{1}}, but found none.",
```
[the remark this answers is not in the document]
IT-VBFK: Ah?

That don't get into my mind, because a parameterless call of `WithArgs` with just a type is also matching. That is an or in this case, isn't it?

Might be that I am on the wrong track here..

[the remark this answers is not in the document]
dennisdoomen: That's true as well. So it expects either the type of the argument to match, and optionally, the predicate.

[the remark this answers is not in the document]
IT-VBFK: ok, got it :) thanks for clarification

dennisdoomen (approved):

Src/FluentAssertions/EventRaisingExtensions.cs
```diff
@@ -53,49 +53,43 @@ public static IEventRecording WithSender(this IEventRecording eventRecording, ob
         }
 
         /// <summary>
-        /// Asserts that at least one occurrence of the events had at least one of the arguments matching a predicate. Returns
-        /// only the events that matched that predicate.
+        /// Asserts that at least one of the occurred events is of the expected type <typeparamref name="T"/>
+        /// and optionally has a matching predicate. Returns
+        /// only the events that matched those type and predicate.
```
jnyrup, line 58: Some thoughts:
We don't know if any events occured.
The arguments should be of type `T`, not the event itself.
It's one the arguments that should match a predicate.

```suggestion
        /// Asserts that at least one occurence of the events had one or more arguments of the expected type <typeparamref name="T"/>
        /// any of which matched a predicate. Returns
        /// only the events that matched both type and predicate.
```

Src/FluentAssertions/EventRaisingExtensions.cs
```diff
@@ -53,49 +53,43 @@ public static IEventRecording WithSender(this IEventRecording eventRecording, ob
...
+            bool hasFoundArgumentsBefore = eventsMatchingPredicate.Any();
 
-            if (!eventsMatchingPredicate.Any())
-            {
-                Execute.Assertion
-                    .FailWith("Expected at least one event with arguments matching {0}, but found none.", predicate.Body);
-            }
+            Execute.Assertion
+                .ForCondition(hasFoundArgumentsBefore)
+                .FailWith($"Expected at least one event which arguments are of type <{{0}}> and which matches {{1}}, but found none.",
```
jnyrup, line 82:
```suggestion
                .FailWith($"Expected at least one event which arguments are of type <{{0}}> and matches {{1}}, but found none.",
```

Src/FluentAssertions/EventRaisingExtensions.cs
```diff
@@ -53,49 +53,43 @@ public static IEventRecording WithSender(this IEventRecording eventRecording, ob
...
 
             return new FilteredEventRecording(eventRecording, eventsMatchingPredicate);
         }
 
         /// <summary>
-        /// Asserts that at least one of the occurred events has arguments the match the predicates in the same order. Returns
-        /// only the events that matched those predicates.
+        /// Asserts that at least one of the occurred events is of the expected type <typeparamref name="T"/>
+        /// and optionally has predicates matching in the same order. Returns
+        /// only the events that matched those type and predicates.
```
jnyrup, line 92:
```suggestion
        /// Asserts that at least one occurence of the events had one or more arguments of the expected type <typeparamref name="T"/>
        /// which matched the predicates in the same order. Returns
        /// only the events that matched both type and predicates.
```

Src/FluentAssertions/EventRaisingExtensions.cs
```diff
@@ -133,18 +123,13 @@ public static IEventRecording WithArgs<T>(this IEventRecording eventRecording, p
                 }
             }
 
-            if (!hasArgumentOfRightType)
-            {
-                throw new ArgumentException("No argument of event " + eventRecording.EventName + " is of type <" + typeof(T) + ">.");
-            }
+            bool hasFoundArgumentsBefore = eventsMatchingPredicate.Any();
 
-            if (!eventsMatchingPredicate.Any())
```
jnyrup, line 141: This construct was to avoid calling `string.Join` when the assertion passes.

Src/FluentAssertions/EventRaisingExtensions.cs
```diff
@@ -53,49 +53,43 @@ public static IEventRecording WithSender(this IEventRecording eventRecording, ob
...
                 {
                     eventsMatchingPredicate.Add(@event);
                 }
             }
 
-            if (!hasArgumentOfRightType)
-            {
-                throw new ArgumentException("No argument of event " + eventRecording.EventName + " is of type <" + typeof(T) + ">.");
-            }
+            bool hasFoundArgumentsBefore = eventsMatchingPredicate.Any();
```
jnyrup, line 78: How about naming this `foundMatchingEvent` or `foundEventMatchingPredicate`?

jnyrup (changes requested):

Src/FluentAssertions/EventRaisingExtensions.cs
```diff
@@ -133,18 +123,13 @@ public static IEventRecording WithArgs<T>(this IEventRecording eventRecording, p
...
 
-            if (!eventsMatchingPredicate.Any())
-            {
-                Execute
-                    .Assertion
-                    .FailWith("Expected at least one event with arguments matching {0}, but found none.",
+            Execute.Assertion
+                .ForCondition(hasFoundArgumentsBefore)
+                .FailWith($"Expected at least one event which arguments are of type <{{0}}> and matches {{1}}, but found none.",
```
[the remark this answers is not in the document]
ITaluone: You mean the $-sign? yeah... fair point :) Fixed

Src/FluentAssertions/EventRaisingExtensions.cs
ITaluone: Yeah.. can agree with that

jnyrup: I'm sorry, I only meant renaming `hasFoundArgumentsBefore` to either `foundMatchingEvent` _or_ `foundEventMatchingPredicate`.

`eventsMatchingPredicate` shouldn't be renamed.

PR #1929, opened 2022-05-18
More specifically clarify the intentions of `WithArgs`

jnyrup:
## IMPORTANT 

Adjust the XML summaries of `WithArgs` to more closely clarify the implementation.
* For the overload taking a single predicate only a single argument of type `T` has to match the predicate.
  * So for three arguments of type `T` it's enough for e.g. the second argument to match.
* I found "some argument" preferable to "an argument" to signal that the event can have other arguments besides the matching one(s).
* Be explicit about that it's the arguments and not the event that have to match the predicate.

I also discovered that the summaries for the entire type was in past tense where all others (as far as I could see) are written in present tense.

### DIFF
--- a/Src/FluentAssertions/EventRaisingExtensions.cs
+++ b/Src/FluentAssertions/EventRaisingExtensions.cs
@@ -14,9 +14,11 @@ namespace FluentAssertions
     public static class EventRaisingExtensions
     {
         /// <summary>
-        /// Asserts that all occurrences of the event originated from the <param name="expectedSender"/> and
-        /// returns only the events that came from that sender.
+        /// Asserts that all occurrences of the event originates from the <param name="expectedSender"/>.
         /// </summary>
+        /// <returns>
+        /// Returns only the events that comes from that sender.
+        /// </returns>
         public static IEventRecording WithSender(this IEventRecording eventRecording, object expectedSender)
         {
             var eventsForSender = new List<OccurredEvent>();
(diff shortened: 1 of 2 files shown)


dennisdoomen (approved):
Sorry for the late review. Has been a very busy week.

Src/FluentAssertions/EventRaisingExtensions.cs
```diff
@@ -53,10 +55,12 @@ public static IEventRecording WithSender(this IEventRecording eventRecording, ob
...
-        ///  Returns only the events that matched both type and optionally a predicate.
+        /// Asserts that at least one occurrence of the events has some argument of the expected
+        /// type <typeparamref name="T"/> that matches the given predicate.
         /// </summary>
+        /// <returns>
+        /// Returns only the events having some argument matching both type and predicate.
+        /// </returns>
         public static IEventRecording WithArgs<T>(this IEventRecording eventRecording, Expression<Func<T, bool>> predicate)
         {
             Guard.ThrowIfArgumentIsNull(predicate, nameof(predicate));
```
julealgon, line 66: I see you are not documenting these kinds of exceptions (contract exceptions) today with `<exception` elements. Have you considered doing so to improve the API documentation?

I know this is not directly related to this PR here (which I found by accident by following a link from the latest release notes), but since it is related to documentation, I figured I'd mention here.

jnyrup: Occurrences:
* `<exception`: 166
* `Guard.`: 224

So we mostly (~75%) do guard calls to `Guard.`, but sure, let's get that number closer to 100%.

julealgon: Ahhh I see, thought this was indicative of the trend but didn't really check it on my end, thanks for clarifying!

PR #1950, opened 2022-07-02
Extend IObjectInfo with info about DeclaringType

### LINKED ISSUE #1949: How do I get DeclaringType from a given IObjectInfo
### Discussed in https://github.com/fluentassertions/fluentassertions/discussions/1944

<div type='discussions-op-text'>

<sup>Originally posted by **AlexCrosby** June 23, 2022</sup>
I'm refactoring a bunch of code that I have upgraded FluentAssertions from v5 to v6 and have come across an issue where `IMemberInfo` has been replaced with `IObjectInfo` (in this case the info variable).
```cs
.When(info => info != null
           && info.SelectedMemberInfo != null
           && info.SelectedMemberInfo.DeclaringType == foo);
```
I know that in `IMemberInfo` `SelectedInfo` has been flattened down, but `IObjectInfo` does not have a `DeclaringType` and I can see no way to get to it from `IObjectInfo` and I'm not entirely sure whether `CompileTimeType` is equivalent.</div>

dennisdoomen:
Adds `IObjectInfo.ParentType` to help determining the declaring type in a call to `Using`/`When` constructs.

Also marks `IObjectInfo.Type` as obsolete since its purpose is so confusing.

Fixes #1949

### DIFF
--- a/Src/FluentAssertions/Equivalency/Execution/ObjectInfo.cs
+++ b/Src/FluentAssertions/Equivalency/Execution/ObjectInfo.cs
@@ -7,12 +7,15 @@ internal class ObjectInfo : IObjectInfo
     public ObjectInfo(Comparands comparands, INode currentNode)
     {
         Type = currentNode.Type;
+        ParentType = currentNode.ParentType;
         Path = currentNode.PathAndName;
         CompileTimeType = comparands.CompileTimeType;
         RuntimeType = comparands.RuntimeType;
     }
 
     public Type Type { get; }
+
+    public Type ParentType { get; }
 
     public string Path { get; set; }
 
--- a/Src/FluentAssertions/Equivalency/INode.cs
+++ b/Src/FluentAssertions/Equivalency/INode.cs
@@ -1,9 +1,11 @@
 using System;
+using JetBrains.Annotations;
 
 namespace FluentAssertions.Equivalency;
 
 /// <summary>
-/// Represents a node in the object graph as it is expected in a structural equivalency check.
+/// Represents a node in the object graph that is being compared as part of a structural equivalency check.
+/// This can be the root object, a collection item, a dictionary element, a property or a field.
 /// </summary>
 public interface INode
 {
@@ -22,9 +24,18 @@ public interface INode
     string Name { get; set; }
 
     /// <summary>
-    /// Gets the type of this node.
+    /// Gets the type of this node, e.g. the type of the field or property, or the type of the collection item. 
     /// </summary>
     Type Type { get; }
+
+    /// <summary>
+    /// Gets the type of the parent node, e.g. the type that declares a property or field.
+    /// </summary>
+    /// <value>
+    /// Is <c>null</c> for the root object.
+    /// </value>
+    [CanBeNull]
+    Type ParentType { get; }
 
     /// <summary>
     /// Gets the path from the root object UNTIL the current node, separated by dots or index/key brackets.
--- a/Src/FluentAssertions/Equivalency/IObjectInfo.cs
+++ b/Src/FluentAssertions/Equivalency/IObjectInfo.cs
@@ -1,4 +1,5 @@
 using System;
+using JetBrains.Annotations;
 
 namespace FluentAssertions.Equivalency;
 
@@ -8,9 +9,19 @@ namespace FluentAssertions.Equivalency;
 public interface IObjectInfo
 {
     /// <summary>
-    /// Gets the type of this node.
+    /// Gets the type of the object 
     /// </summary>
+    [Obsolete("Use CompileTimeType or RuntimeType instead")]
     Type Type { get; }
+
+    /// <summary>
+    /// Gets the type of the parent, e.g. the type that declares a property or field.
+    /// </summary>
+    /// <value>
+    /// Is <c>null</c> for the root object.
+    /// </value>
+    [CanBeNull]
+    Type ParentType { get; }
 
     /// <summary>
     /// Gets the full path from the root object until the current node separated by dots.
--- a/Src/FluentAssertions/Equivalency/Node.cs
+++ b/Src/FluentAssertions/Equivalency/Node.cs
@@ -1,15 +1,11 @@
 using System;
 using System.Collections;
-using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
 using FluentAssertions.Common;
 
 namespace FluentAssertions.Equivalency;
 
-/// <summary>
-/// Represents a node in the object graph that is being compared as part of a structural equivalency check.
-/// </summary>
 public class Node : INode
 {
     private static readonly Regex MatchFirstIndex = new(@"^\[\d+\]$");
@@ -21,6 +17,8 @@ public class Node : INode
     public GetSubjectId GetSubjectId { get; protected set; } = () => string.Empty;
 
     public Type Type { get; protected set; }
+
+    public Type ParentType { get; protected set; }
 
     public string Path
     {
@@ -82,6 +80,7 @@ public class Node : INode
             Name = string.Empty,
             Path = string.Empty,
             Type = typeof(T),
+            ParentType = null,
             RootIsCollection = IsCollection(typeof(T))
         };
     }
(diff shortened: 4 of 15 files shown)


docs/_pages/releases.md
```diff
@@ -10,6 +10,7 @@ sidebar:
 ## Unreleased
 
 ### What's new
+* Adds a `ParentType` to `IObjectInfo` to help determining the parent in a call to `Using`/`When` constructs - [#2000](https://github.com/fluentassertions/fluentassertions/pull/2000)
```
jnyrup, line 13:
```suggestion
* Added a `ParentType` to `IObjectInfo` to help determining the parent in a call to `Using`/`When` constructs - [#1950](https://github.com/fluentassertions/fluentassertions/pull/1950)
```

Src/FluentAssertions/Equivalency/IObjectInfo.cs
```diff
@@ -8,9 +9,19 @@ namespace FluentAssertions.Equivalency;
 public interface IObjectInfo
 {
     /// <summary>
-    /// Gets the type of this node.
+    /// Gets the type of the object 
     /// </summary>
+    [Obsolete("Use CompileTimeType or RunTimeType instead")]
```
jnyrup, line 14:
```suggestion
    [Obsolete("Use CompileTimeType or RuntimeType instead")]
```

Src/FluentAssertions/Equivalency/INode.cs
```diff
@@ -22,9 +24,18 @@ public interface INode
...
     Type Type { get; }
+    
+    /// <summary>
+    /// Gets the type of the parent node, e.g. the type that declares a property or field.
+    /// </summary>
+    /// <value>
+    /// Is <c>null</c> for the root object.
+    /// </value>
+    [CanBeNull]
+    Type ParentType { get; }
```
jnyrup, line 38: We should be explicit about whether `ParentType` is the compile-time or runtime type of the parent.

dennisdoomen: It's the `DeclaringType`, so that shouldn't matter.

jnyrup: Ohh.. it's the declaring type.
From the name I got the impression that it was either the compile-time or runtime type of the parent _instance_ carrying the value.

How about naming it `DeclaringType` then?

dennisdoomen: Because a node can be something else than a field or property.

dennisdoomen: So what do you think?

dennisdoomen: @jnyrup ?

Src/FluentAssertions/Equivalency/Execution/ObjectInfo.cs
```diff
@@ -7,12 +7,15 @@ internal class ObjectInfo : IObjectInfo
     public ObjectInfo(Comparands comparands, INode currentNode)
     {
         Type = currentNode.Type;
+        ParentType = currentNode.ParentType;
```
jnyrup, line 10:
```suggestion
        ParentType = currentNode.ParentType;
```

Src/FluentAssertions/Equivalency/Execution/ObjectInfo.cs
```diff
@@ -7,12 +7,15 @@ internal class ObjectInfo : IObjectInfo
...
     {
         Type = currentNode.Type;
+        ParentType = currentNode.ParentType;    
         Path = currentNode.PathAndName;
         CompileTimeType = comparands.CompileTimeType;
         RuntimeType = comparands.RuntimeType;
     }
 
     public Type Type { get; }
+
```
jnyrup, line 17:
```suggestion

```

Src/FluentAssertions/Equivalency/INode.cs
```diff
@@ -22,9 +24,18 @@ public interface INode
     string Name { get; set; }
 
     /// <summary>
-    /// Gets the type of this node.
+    /// Gets the type of this node, e.g. the type of the field or property, or the type of the collection item. 
     /// </summary>
     Type Type { get; }
+
```
jnyrup, line 30:
```suggestion

```

Src/FluentAssertions/Equivalency/Node.cs
```diff
@@ -21,6 +17,8 @@ public class Node : INode
     public GetSubjectId GetSubjectId { get; protected set; } = () => string.Empty;
 
     public Type Type { get; protected set; }
+
```
jnyrup, line 20:
```suggestion

```

Src/FluentAssertions/Equivalency/IObjectInfo.cs
```diff
@@ -8,9 +9,19 @@ namespace FluentAssertions.Equivalency;
 public interface IObjectInfo
 {
     /// <summary>
-    /// Gets the type of this node.
+    /// Gets the type of the object 
     /// </summary>
+    [Obsolete("Use CompileTimeType or RuntimeType instead")]
     Type Type { get; }
+
```
jnyrup, line 16:
```suggestion

```

Src/FluentAssertions/Equivalency/INode.cs
jnyrup: Now the penny finally dropped, I was missing that the `ParentType` can also be e.g. a collection.

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
```diff
@@ -926,6 +926,7 @@ namespace FluentAssertions.Equivalency
         FluentAssertions.Equivalency.GetSubjectId GetSubjectId { get; }
         bool IsRoot { get; }
         string Name { get; set; }
+        System.Type ParentType { get; }
```
jnyrup, line 929: We have previously _technically_ broken `INode` by adding a member to it, so it's fine by me to do it again.

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
```diff
@@ -934,8 +935,10 @@ namespace FluentAssertions.Equivalency
     public interface IObjectInfo
     {
         System.Type CompileTimeType { get; }
+        System.Type ParentType { get; }
```
jnyrup, line 938: I think we can apply the same reasoning for `IObjectInfo` as to `INode`, that this interface is only read and never re-implemented, so I'm also fine with this change.

jnyrup (approved):

Src/FluentAssertions/Equivalency/Execution/ObjectInfo.cs
dennisdoomen: You nitpicker ;-)

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
dennisdoomen: Yeah, I was pondering on this same issue, but settled my own doubts with the same assumption.

jnyrup:
Can we add a test to exercise the missed path? 
i.e. the getter for `CollectionMemberObjectInfo.ParentType`.

I tried out the added test `Can_exclude_all_properties_of_the_parent_type` locally, and omitting `si.ParentType == expectation.GetType()` still makes the test pass as there is only a single `Path` ending with `Id`.

If we instead have two properties named `Id` `ParentType` comes to its right to distinguish the inner and outer `Id`.

```cs
[Fact]
public void Can_exclude_all_properties_of_the_parent_type()
{
    // Arrange
    var subject = new
    {
        Id = new
        {
            Id = "foo"
        }
    };

    var expectation = new
    {
        Id = new
        {
            Id = "bar"
        }
    };

    // Act
    subject.Should().BeEquivalentTo(expectation,
        o => o
            .Using<string>(c => c.Subject.Should().HaveLength(c.Expectation.Length))
            .When(si => si.ParentType == expectation.Id.GetType() && si.Path.EndsWith("Id", StringComparison.InvariantCulture)));
}
```

dennisdoomen:
Sure. I will take care of it.

PR #1960, opened 2022-07-21
Avoid allocating sub-arrays in ContainInOrder

jnyrup:
There's no need to repeatedly allocating sub-arrays.
In the worst case, when subject and expected are identical, we will allocate N sub-arrays of a total length of O(N^2).

Instead we simply keep an index of how far we have scanned the subject.

The added tests are not meant as authoritative but only to ensure behavior has not changed.

### DIFF
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -920,16 +920,15 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> :
             IList<T> expectedItems = expected.ConvertOrCastToList();
             IList<T> actualItems = Subject.ConvertOrCastToList();
 
+            int subjectIndex = 0;
+
             Func<T, T, bool> areSameOrEqual = ObjectExtensions.GetComparer<T>();
             for (int index = 0; index < expectedItems.Count; index++)
             {
                 T expectedItem = expectedItems[index];
-                actualItems = actualItems.SkipWhile(actualItem => !areSameOrEqual(actualItem, expectedItem)).ToArray();
-                if (actualItems.Any())
-                {
-                    actualItems = actualItems.Skip(1).ToArray();
-                }
-                else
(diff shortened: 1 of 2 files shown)


Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -2261,43 +2271,38 @@ public AndConstraint<TAssertions> NotContainInOrder(params T[] unexpected)
...
-            actualItems = actualItems.SkipWhile(actualItem =>
-            {
-                actualItemsSkipped++;
-                return !areSameOrEqual(actualItem, unexpectedItem);
-            }).ToArray();
+            IList<T> actualItems = Subject.ConvertOrCastToList();
+            int subjectIndex = 0;
 
-            if (actualItems.Any())
+            Func<T, T, bool> areSameOrEqual = ObjectExtensions.GetComparer<T>();
```
dennisdoomen, line 2279: The bulk of this method and that of `NotContainInOrder` is the same. Considering the size of the body, I think it might be worthwhile to extract that into an extension method or private method. It'll make the algorithm clearer.

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.ContainInOrder.cs
```diff
@@ -100,6 +100,20 @@ public void When_passing_in_null_while_checking_for_ordered_containment_it_shoul
...
 
+        [Fact]
+        public void Collections_contain_the_empty_sequence()
+        {
+            // Assert
+            new[] { 1 }.Should().ContainInOrder(new int[0]);
+        }
+
+        [Fact]
+        public void Collections_do_not_not_contain_the_empty_sequence()
```
jnyrup, line 111: I struggled finding a satisfying name for this test, so all ears if you can come up with a more fitting name.

dennisdoomen: `Every_collection_contains_an_empty_collection`?

dennisdoomen (approved):

PR #1963, opened 2022-07-26
Extend collection assertions with `ContainInConsecutiveOrder` and `NotContainInConsecutiveOrder`

### LINKED ISSUE #1956: Extend `GenericCollectionAssertions` to include a `ContainInConsecutiveOrder`
### Description

We were writing tests where we needed to check if one collection contained another in order. For this, we used `ContainInOrder`.

However, we found that because `ContainInOrder` doesn't check that the items are consecutive we could introduce mutations in the code and our tests would still pass.

To solve this we have made a new method in the same class, `GenericCollectionAssertions`, called `ContainInConsecutiveOrder`. This is a stricter version of the current check, adding the check that the expected items all appear in consecutive order. For completeness, we also introduced the opposite `NotContainInConsecutiveOrder`.

The API change is an extension to the API, with, as far as I can see, no breaking changes.

We have implemented these changes, with associated tests, in a branch and would like to submit a PR.

StacyCash:
Adds `ContainInConsecutiveOrder` and `NotContainInConsecutiveOrder` to check that a collection is contained within another in order and with no gaps.

This will close #1956 

## IMPORTANT 

* [X] If the PR changes the public API the changes needs to be included by running [AcceptApiChanges.ps1](../tree/develop/AcceptApiChanges.ps1) or [AcceptApiChanges.sh](../tree/develop/AcceptApiChanges.sh).
* [X] If the PR affects [the documentation](../tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).


Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -944,6 +944,98 @@ public AndConstraint<TAssertions> ContainInOrder(params T[] expected)
...
+                    actualItems = actualItems.Skip(1).ToArray();
+                }
+                else
+                {
+                    Execute.Assertion
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith(
+                        "Expected {context:collection} {0} to contain items {1} in order{reason}" +
+                        ", but {2} (index {3}) did not appear (in the right order).",
+                        Subject, expected, expectedItem, highestIndex + 1);
```
eNeRGy164, line 1031: 🔍 nitpick: missing indent

```suggestion
                        .BecauseOf(because, becauseArgs)
                        .FailWith(
                            "Expected {context:collection} {0} to contain items {1} in order{reason}" +
                            ", but {2} (index {3}) did not appear (in the right order).",
                            Subject, expected, expectedItem, highestIndex + 1);
```

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.ContainInConsecutiveOrder.cs
```diff
@@ -0,0 +1,293 @@
...
+using Xunit.Sdk;
+
+namespace FluentAssertions.Specs.Collections;
+
+/// <content>
+/// The [Not]ContainInOrder specs.
+/// </content>
+public partial class CollectionAssertionSpecs
+{
+    public class ContainInExplicitOrder
```
jnyrup, line 13: explicit -> consecutive.
here and more places in the PR.

docs/_pages/releases.md
```diff
@@ -10,6 +10,7 @@ sidebar:
 ## Unreleased
 
 ### What's new
+* Add `ContainInExplicitOrder` and `NotContainInExplicitOrder` assertions to check if a collection contains or not items in a specific order and to be consecutive. - [#](https://github.com/fluentassertions/fluentassertions/pull/)
```
jnyrup, line 13:
```suggestion
* Added `ContainInConsecutiveOrder` and `NotContainInConsecutiveOrder` assertions to check if a collection contains items in a specific order and to be consecutive - [#1963](https://github.com/fluentassertions/fluentassertions/pull/1963)
```

eNeRGy164, line 13:
```suggestion
* Add `ContainInExplicitOrder` and `NotContainInExplicitOrder` assertions to check if a collection contains or not items in a specific order and to be consecutive. - [#1963](https://github.com/fluentassertions/fluentassertions/pull/1963)
```

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.ContainInConsecutiveOrder.cs
StacyCash: Sorry, I thought I caught them all. I'll double check

docs/_pages/releases.md
StacyCash: Fixed

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -944,6 +944,98 @@ public AndConstraint<TAssertions> ContainInOrder(params T[] expected)
...
+    /// <exception cref="ArgumentNullException"><paramref name="expected"/> is <c>null</c>.</exception>
+    public AndConstraint<TAssertions> ContainInConsecutiveOrder(IEnumerable<T> expected, string because = "",
+        params object[] becauseArgs)
+    {
+        Guard.ThrowIfArgumentIsNull(expected, nameof(expected), "Cannot verify ordered containment against a <null> collection.");
+
+        bool success = Execute.Assertion
+            .BecauseOf(because, becauseArgs)
+            .ForCondition(Subject is not null)
+            .FailWith("Expected {context:collection} to contain {0} in order{reason}, but found <null>.", expected);
```
jnyrup, line 980:
```suggestion
            .FailWith("Expected {context:collection} to contain {0} in consecutive order{reason}, but found <null>.", expected);
```

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -944,6 +944,98 @@ public AndConstraint<TAssertions> ContainInOrder(params T[] expected)
...
+                if (actualItems.Any())
+                {
+                    actualItems = actualItems.Skip(1).ToArray();
+                }
+                else
+                {
+                    Execute.Assertion
+                        .BecauseOf(because, becauseArgs)
+                        .FailWith(
+                            "Expected {context:collection} {0} to contain items {1} in order{reason}" +
```
jnyrup, line 1029:
```suggestion
                            "Expected {context:collection} {0} to contain items {1} in consecutive order{reason}" +
```

jnyrup (commented):
The code coverage reports two branch misses.
If you login to coveralls.io with with your github account, you can inspect the report there.
Alternatively you can run `./build.ps1` locally and look in`TestResults\reports\index.html`.

Do you intend to apply the optimizations I suggested in https://github.com/fluentassertions/fluentassertions/issues/1956#issuecomment-1194471717?
I should not be necessary to do any calls `ToArray()`.
If you prefer, I can also have a look at that. Just let me know.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -944,6 +944,98 @@ public AndConstraint<TAssertions> ContainInOrder(params T[] expected)
...
+        Guard.ThrowIfArgumentIsNull(expected, nameof(expected), "Cannot verify ordered containment against a <null> collection.");
+
+        bool success = Execute.Assertion
+            .BecauseOf(because, becauseArgs)
+            .ForCondition(Subject is not null)
+            .FailWith("Expected {context:collection} to contain {0} in order{reason}, but found <null>.", expected);
+
+        if (success)
+        {
+            IList<T> expectedItems = expected.ConvertOrCastToList();
```
dennisdoomen, line 984: I'm struggling to understand this implementation. I was expecting a much simpler algorithm (preferably using separate methods), probably because it's too long. Something like:

1. Scan the subject collection until you find the first item from the expected collection
2. From that point on, verify whether all successive items match. Remember the index of the first mismatching item
3. If not, restart scanning from the last index you found in 1
4. If no full match is found, use the index found in 2 to report a proper error

dennisdoomen (changes requested):
Thanks for the contribution.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
StacyCash: I was trying to keep the code in a similar style as the existing function.

I've rewritten the code based on the suggestion. I think that it reads better now, and found a problem in a test case, and a missing test case in the process. I hope that it's what you had in mind.

I do think that I need to take another look at the not version of the function, but as I am about to go away on vacation for a week I'll do that when I return.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -2303,6 +2395,116 @@ public AndConstraint<TAssertions> NotContainInOrder(params T[] unexpected)
...
+
+            if (actualItems.Any())
+            {
+                if (index == unexpectedItemsCount - 1)
+                {
+                    Execute.Assertion
+                        .BecauseOf(because, becauseArgs)
+                        .FailWith(
+                            "Expected {context:collection} {0} to not contain items {1} in order{reason}, " +
+                            "but items appeared in order ending at index {2}.",
```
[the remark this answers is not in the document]
StacyCash: Done

StacyCash:
@[jnyrup](https://github.com/jnyrup) I've looked at your changes and rewritten both the `ContainsInConsecutitveOrder` and `NotContainsInConsecutiveOrder` based it. Thanks! They are both **much** easier to read now.

jnyrup:
Major improvement!
Now I only have a few styling comments.

I haven't looked deeply into the tests, but it seems we could use a test when calling with an empty `expected`/`unexpected`.

StacyCash:
There is an empty test for `NotContainsInConsecuritveOrder`, I've added the same for `ContainsInConsecutiveOrder`

If you let me know the style issues I'll either look at them in the morning or when I return.

jnyrup (commented):
> If you let me know the style issues I'll either look at them in the morning or when I return.

The style comments were still pending 🤦‍♂️

StacyCash:
> The style comments were still pending 🤦‍♂️

😅 Been there, done that 😊

I've committed the suggestions.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -2287,6 +2360,95 @@ public AndConstraint<TAssertions> NotContainInOrder(params T[] unexpected)
...
+                int previousSubjectIndex = subjectIndex;
+                subjectIndex = IndexOf(actualItems, unexpectedItem, subjectIndex, areSameOrEqual);
+                highestIndex = Math.Max(index, highestIndex);
+
+                if (subjectIndex == -1)
+                {
+                    return new AndConstraint<TAssertions>((TAssertions)this);
+                }
+
+                if (index > 0 && subjectIndex - previousSubjectIndex > 1)
```
jnyrup, line 2434:
```suggestion
                if (index > 0 && subjectIndex != previousSubjectIndex + 1)
```

jnyrup (approved):

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -943,6 +943,79 @@ public AndConstraint<TAssertions> ContainInOrder(params T[] expected)
...
+
+            Func<T, T, bool> areSameOrEqual = ObjectExtensions.GetComparer<T>();
+            int index;
+            int highestIndex = 0;
+            for (index = 0; index < expectedItems.Count; index++)
+            {
+                T expectedItem = expectedItems[index];
+                int previousSubjectIndex = subjectIndex;
+                subjectIndex = IndexOf(actualItems, expectedItem, subjectIndex, areSameOrEqual);
+                highestIndex = Math.Max(index, highestIndex);
```
dennisdoomen, line 996: 🤔 What do you need the highest index for? Why not just use the `index` in the failure message?

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -2287,6 +2360,95 @@ public AndConstraint<TAssertions> NotContainInOrder(params T[] unexpected)
...
+            int index;
+            int highestIndex = 0;
+
+            for (index = 0; index < unexpectedItems.Count; index++)
+            {
+                T unexpectedItem = unexpectedItems[index];
+
+                int previousSubjectIndex = subjectIndex;
+                subjectIndex = IndexOf(actualItems, unexpectedItem, subjectIndex, areSameOrEqual);
+                highestIndex = Math.Max(index, highestIndex);
```
dennisdoomen, line 2427: 🔧 You're not using the `highestIndex`, so please remove it.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
dennisdoomen: 🤔 What does this statement do? Maybe wrapping it in a functional helper method will make it more obvious.

dennisdoomen (changes requested):
Some final notes and then I'll stop complaining ;-)

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
jnyrup: It checks if the found match was also consecutive.

dennisdoomen: My point was that this was not clear to me.... ;-)

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
jnyrup: The test on line 76 in the test class exercises this. 
If we didn't had `highestIndex` and used `index` the error message would contain "3 (index 0)" instead of the current "3 (index 2)". 
I think it can be described as the highest index of a found item that was not consecutive.

StacyCash: That's it exactly, when just looking for items in order we only search once, and so know what the highest index is. When searching for items in consecutive order we can search multiple times so need to keep track of the best match we can make

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
StacyCash: As I've moved the code to a wrapper helper the code suggestion is out of date.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
StacyCash: Have done

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -943,6 +943,79 @@ public AndConstraint<TAssertions> ContainInOrder(params T[] expected)
...
+
+        if (success)
+        {
+            IList<T> expectedItems = expected.ConvertOrCastToList();
+            IList<T> actualItems = Subject.ConvertOrCastToList();
+
+            int subjectIndex = 0;
+
+            Func<T, T, bool> areSameOrEqual = ObjectExtensions.GetComparer<T>();
+            int index;
```
dennisdoomen, line 989: ♻️ You can move the declaration in the `for` loop. It'll helps to make the method less long (and noisy).

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -2287,6 +2360,93 @@ public AndConstraint<TAssertions> NotContainInOrder(params T[] unexpected)
...
+
+            for (index = 0; index < unexpectedItems.Count; index++)
+            {
+                T unexpectedItem = unexpectedItems[index];
+
+                int previousSubjectIndex = subjectIndex;
+                subjectIndex = IndexOf(actualItems, unexpectedItem, subjectIndex, areSameOrEqual);
```
dennisdoomen, line 2425: ♻️ I think we can make this call more self-explanatory by doing two things:
1. Since all of the invocations pass the same compared, make this method generic and then have it call `ObjectExtensions.GetComparer`.
2. Rename the `index` parameter to `startIndex` and then use a named parameter on the calling side. I had to study the implementation to understand what it did.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -3308,4 +3468,6 @@ private static int IndexOf(IList<T> items, T item, int index, Func<T, T, bool> c
 
         return -1;
     }
+
+    private static bool SubjectIndexIsConsecutive(int subjectIndex, int previousSubjectIndex) => subjectIndex != previousSubjectIndex + 1;
```
dennisdoomen, line 3472: ♻️ Maybe by changing this into an extension method would make it more readable, e.g. `int1.IsConsecutiveWith(int2)`
🤔 Isn't this implementation doing the opposite? It is consecutive if the indexes only differ by 1, right?

dennisdoomen (changes requested):
Sorry for more comments, but I've trying to find ways to refactor (♻️ ) the code a bit to make the algorithm a bit more self-explanatory.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
StacyCash: Yes, that was me doing things after travelling all day 😅 Will be fixed in the next commit

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
StacyCash: I'm not sure what you mean here, the declarations here need to be outside of the for loop to work

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
StacyCash: For changing the function to an extension. How do you envision this? Looking at the codebase in the extension folder I see public extensions, but I'm assuming that this isn't something that you want to change the public interface?

Personally, I am not sure that I agree with making it an extension (for something that is used in two places in 1 class) - but will change if you let me know where you want it.

dennisdoomen: > For changing the function to an extension. How do you envision this? Looking at the codebase in the extension folder I see public extensions, but I'm assuming that this isn't something that you want to change the public interface?

As an `internal` extension under `Common`

> Personally, I am not sure that I agree with making it an extension (for something that is used in two places in 1 class) - but will change if you let me know where you want it.

I'm always looking for ways to make code more self-explanetory

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
dennisdoomen: Why? It's only needed within the block scope created by the `for` loop.

StacyCash: Compiler error AV1530 - if you declare the index variable in the loop you cannot update it in the loop itself. Which we need to do to reset it should we have to start searching from the start of the unexpected items again.

(Of course, that does raise the question of if the loop is the right way to solve the problem 🤔 - but I think that it is the cleanest)

The 'subjectIndex` and `highestIndex` need to be outside of the loop as they shouldn't be recreated/reset each loop.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
StacyCash: Thanks for the pointer of where to put it. I'm going to try and see how it looks.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
dennisdoomen: Ah, that wasn't obvious at all. Updating a loop variable is quite ugly indeed. But I see now how it works.

StacyCash: Would you rather that this is really obvious?

We could declare the index in the loop and disable VS1530 for the loop itself... I see that is done elsewhere for other reasons.

dennisdoomen: I think it would make it clearer that you're doing that.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
StacyCash: Changed

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
StacyCash: Having that code to disable the rule annoyed me so I spent some time thinking about it and figured out how to rewrite it to remove the index manipulation. I'm on the fence as to which is easier to read, let me know which you prefer (I'll revert if you want 😅)

dennisdoomen (approved):

dennisdoomen:
Thank you for your contribution @StacyCash !

PR #1967, opened 2022-08-05
Added NotCompleteWithinAsync for Task assertions

lg2de:
In addition to `CompleteWithAsync` in assertions for `Task` and `Task<T>`, and similar to assertions for `TaskCompletionSource`, I added `NotCompleteWithinAsync`.
Checking the existing tests I found some inconsistency which I've fixed (hopefully) in several separate commits. Maybe you want to review these commits one by one for better overview.

### DIFF
--- a/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
@@ -85,7 +85,7 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
         bool success = Execute.Assertion
             .ForCondition(Subject is not null)
             .BecauseOf(because, becauseArgs)
-            .FailWith("Expected {context:task} to not complete within {0}{reason}, but found <null>.", timeSpan);
+            .FailWith("Did not expect {context:task} to complete within {0}{reason}, but found <null>.", timeSpan);
 
         if (success)
         {
@@ -96,7 +96,7 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
                 Execute.Assertion
                     .ForCondition(!completesWithinTimeout)
                     .BecauseOf(because, becauseArgs)
-                    .FailWith("Expected {context:task} to not complete within {0}{reason}.", timeSpan);
+                    .FailWith("Did not expect {context:task} to complete within {0}{reason}.", timeSpan);
             }
         }
 
--- a/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
@@ -41,36 +41,63 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
     public async Task<AndConstraint<TAssertions>> CompleteWithinAsync(
         TimeSpan timeSpan, string because = "", params object[] becauseArgs)
     {
-        Execute.Assertion
+        bool success = Execute.Assertion
             .ForCondition(Subject is not null)
             .BecauseOf(because, becauseArgs)
             .FailWith("Expected {context:task} to complete within {0}{reason}, but found <null>.", timeSpan);
 
-        ITimer timer = Clock.StartTimer();
-        TTask task = Subject.Invoke();
-        TimeSpan remainingTime = timeSpan - timer.Elapsed;
+        if (success)
+        {
+            (TTask task, TimeSpan remainingTime) = InvokeWithTimer(timeSpan);
 
+            success = Execute.Assertion
+                .ForCondition(remainingTime >= TimeSpan.Zero)
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected {context:task} to complete within {0}{reason}.", timeSpan);
+
+            if (success)
+            {
+                bool completesWithinTimeout = await CompletesWithinTimeoutAsync(task, remainingTime);
+                Execute.Assertion
+                    .ForCondition(completesWithinTimeout)
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith("Expected {context:task} to complete within {0}{reason}.", timeSpan);
+            }
+        }
+
+        return new AndConstraint<TAssertions>((TAssertions)this);
+    }
+
+    /// <summary>
+    /// Asserts that the current <typeparamref name="TTask"/> will not complete within the specified time.
+    /// </summary>
+    /// <param name="timeSpan">The allowed time span for the operation.</param>
+    /// <param name="because">
+    /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+    /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+    /// </param>
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+    /// </param>
+    public async Task<AndConstraint<TAssertions>> NotCompleteWithinAsync(
+        TimeSpan timeSpan, string because = "", params object[] becauseArgs)
+    {
         bool success = Execute.Assertion
-            .ForCondition(remainingTime >= TimeSpan.Zero)
+            .ForCondition(Subject is not null)
             .BecauseOf(because, becauseArgs)
-            .FailWith("Expected {context:task} to complete within {0}{reason}.", timeSpan);
+            .FailWith("Expected {context:task} to not complete within {0}{reason}, but found <null>.", timeSpan);
 
         if (success)
         {
-            using var timeoutCancellationTokenSource = new CancellationTokenSource();
-            Task completedTask =
-                await Task.WhenAny(task, Clock.DelayAsync(remainingTime, timeoutCancellationTokenSource.Token));
-
-            if (completedTask == task)
+            (Task task, TimeSpan remainingTime) = InvokeWithTimer(timeSpan);
+            if (remainingTime >= TimeSpan.Zero)
             {
-                timeoutCancellationTokenSource.Cancel();
-                await completedTask;
+                bool completesWithinTimeout = await CompletesWithinTimeoutAsync(task, remainingTime);
+                Execute.Assertion
+                    .ForCondition(!completesWithinTimeout)
+                    .BecauseOf(because, becauseArgs)
+                    .FailWith("Expected {context:task} to not complete within {0}{reason}.", timeSpan);
             }
-
-            Execute.Assertion
-                .ForCondition(completedTask == task)
-                .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context:task} to complete within {0}{reason}.", timeSpan);
         }
 
         return new AndConstraint<TAssertions>((TAssertions)this);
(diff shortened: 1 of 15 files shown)


Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -41,38 +41,99 @@ public AsyncFunctionAssertions(Func<TTask> subject, IExtractExceptions extractor
...
+        if (completedTask == task)
+        {
+            timeoutCancellationTokenSource.Cancel();
+            await completedTask;
+        }
+
+        Execute.Assertion
+            .ForCondition(completedTask != task)
+            .BecauseOf(because, becauseArgs)
+            .FailWith("Expected {context:task} to not complete within {0}{reason}.", timeSpan);
```
dennisdoomen, line 135: 🔧 We prefer to phrase this as `Did not expect`. See also https://github.com/fluentassertions/fluentassertions/issues/1340

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -41,38 +41,99 @@ public AsyncFunctionAssertions(Func<TTask> subject, IExtractExceptions extractor
...
+            return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
+        ITimer timer = Clock.StartTimer();
+        TTask task = Subject.Invoke();
+        TimeSpan remainingTime = timeSpan - timer.Elapsed;
+        if (remainingTime < TimeSpan.Zero)
+        {
+            // expected timeout reached, done
+            return new AndConstraint<TAssertions>((TAssertions)this);
```
dennisdoomen, line 119: 🤔 I personally don't like all these intermediate `return` statements, but I don't hold this opinion strongly.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -41,38 +41,99 @@ public AsyncFunctionAssertions(Func<TTask> subject, IExtractExceptions extractor
...
+            await Task.WhenAny(task, Clock.DelayAsync(remainingTime, timeoutCancellationTokenSource.Token));
+
+        if (completedTask == task)
+        {
+            timeoutCancellationTokenSource.Cancel();
+            await completedTask;
+        }
+
+        Execute.Assertion
+            .ForCondition(completedTask != task)
```
dennisdoomen, line 133: ♻️ You can skip this condition.

dennisdoomen (changes requested):
Thanks for the contribution. Just some improvements.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
lg2de: Fixed, also in the TCS assertions.

lg2de: Do we need to mention this explicitly in the release notes?

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
lg2de: I would like to know why do you prefer higher indention instead?
For me, the complexity is lower with early return.
https://docs.codeclimate.com/docs/cognitive-complexity

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
lg2de: Sorry, but I do not understand why,
The `if` before only checks whether the operation needs to be cancelled.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
dennisdoomen: Specify what?

lg2de: That the failure text has been changed.

lg2de:
I'm wondering why there is no "comment" from coveralls with the coverage report!?

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
dennisdoomen: No, I actually prefer to keep the methods small by breaking them down into smaller parts with clear names (without making them too small):

```csharp
    public async Task<AndConstraint<TAssertions>> NotCompleteWithinAsync(
        TimeSpan timeSpan, string because = "", params object[] becauseArgs)
    {
            .FailWith("Did not expect {context:task} to complete within {0}{reason}, but found <null>.", timeSpan);

        if (success)
        {
            (Task task, TimeSpan executionTime) = InvokeWithTimer(Subject, Clock);
            TimeSpan remainingTime = timeSpan - executionTime;
            if (remainingTime >= TimeSpan.Zero)
            {
                Execute.Assertion
                    .ForCondition(!await CompletesWithinTimeoutAsync(task, remainingTime, Clock))
                    .FailWith("Did not expect {context:task} to complete within {0}{reason}.", timeSpan);
            }
        }

    }
    
    private static (T result, TimeSpan executionTime) InvokeWithTimer<T>(Func<T> func, IClock clock)
    {
        ITimer timer = clock.StartTimer();
        T result = func.Invoke();
        
        return (result, timer.Elapsed);
    }
    
    private static async Task<bool> CompletesWithinTimeoutAsync(Task target, TimeSpan remainingTime, IClock clock)
    {
        using var timeoutCancellationTokenSource = new CancellationTokenSource();
              
        Task completedTask =
            await Task.WhenAny(target, clock.DelayAsync(remainingTime, timeoutCancellationTokenSource.Token));

        if (completedTask == target)
        {
            timeoutCancellationTokenSource.Cancel();
            await completedTask;
            return true;
        }
        
        return false;
    }
```

And I would make `InvokeWithTimer` and `CompletesWithinTimeoutAsync` extension methods.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
dennisdoomen: If you move it within the `if` block above, it would work as expected. But also check my rewritten version in the other comment.

dennisdoomen:
> I'm wondering why there is no "comment" from coveralls with the coverage report!?

It sometimes doesn't update the PR, but the [analysis did run](https://coveralls.io/builds/51462329).

Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
```diff
@@ -308,8 +308,8 @@ await act.Should().ThrowAsync<ArgumentOutOfRangeException>()
                 .WithMessage("* value of pollInterval must be non-negative*");
         }
 
-        [Fact]
-        public async Task When_no_exception_should_be_thrown_async_for_null_after_wait_time_it_should_throw()
+        [Fact] // TODO What is the relevant difference to When_subject_is_null_it_should_fail?
```
jnyrup, line 311: I added both methods in #1289, but with today's eyes I cannot see what purpose _this_ method serves.
`When_subject_is_null_it_should_fail` also verifies that `because` uses the `objectArgs`.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
jnyrup: Not in this case as it simply a choice of words and we consider the message an implementation detail that can change any time.
If we had added substantial new information to the failure message it could be worth mentioning.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
jnyrup: Personally I also praise return-early-return-often, but for Fluent Assertions I try to follow Dennis' preference of single returns in the name of consistency.

The few exceptions (that I can easily find) are places where using single return would require additional variables/work.
E.g. when returning early also act as breaking early out of a loop.

https://github.com/fluentassertions/fluentassertions/blob/de9faf3057b261ce4658a4579d8a62a2e811e2f3/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs#L2248-L2288

In any case I would really prefer to keep style changes separated from functional changes 🙏

dennisdoomen: @jnyrup did you see my suggested refactoring bove?

jnyrup: I did and it looks fine to me. 
It attacks some underlying problems from other angles.

1. Having too many active variables at a time.
2. A method being so lengthy that it requires long vertical eye travel or even scrolling to get an overview of.

lg2de: Based on your suggestion, Dennis, I reworked the methods using new helper methods.
I did not create them as extension methods because:
What kind if type should be extended? The relevant types are very generic. Extension methods (should be public for extensibility) must not accidentally used in customers code.

I understand the idea, Dennis, to keep methods small. But this is not an argument for single-return.
So far I will go with Jonas and accept the environment style.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
lg2de: obsolete, now

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -41,38 +41,99 @@ public AsyncFunctionAssertions(Func<TTask> subject, IExtractExceptions extractor
...
+        Execute.Assertion
+            .ForCondition(completedTask == task)
+            .BecauseOf(because, becauseArgs)
+            .FailWith("Expected {context:task} to complete within {0}{reason}.", timeSpan);
+
+        return new AndConstraint<TAssertions>((TAssertions)this);
+    }
+
+    /// <summary>
+    /// Asserts that the current <typeparamref name="TTask"/> will complete within the specified time.
```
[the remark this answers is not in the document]
lg2de: done, thanks

Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
lg2de: removed

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
dennisdoomen: Great to see you liked my suggestions. And for the record, returning out of the method at the beginning, based on some precondition, is totally fine by me as well. @jnyrup and I are completely aligned on this.

dennisdoomen (approved):

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -266,6 +293,38 @@ async Task<AndConstraint<TAssertions>> AssertionTaskAsync()
         }
     }
 
+    /// <summary>
+    ///     Invokes the subject and measures the sync execution time.
+    /// </summary>
+    protected (TTask result, TimeSpan remainingTime) InvokeWithTimer(TimeSpan timeSpan, IClock clock)
```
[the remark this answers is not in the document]
lg2de: Done.
But keep in mind that these methods are now unavailable for extensions.
Sure, we can make them non-private on request.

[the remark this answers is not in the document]
lg2de: Yep.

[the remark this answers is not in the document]
jnyrup: `CompletesWithinTimeoutAsync` is still `protected`.

I'm open to exposing the methods, I just like postpone that until someone asks for it and another PR.

lg2de:
I found another bug...

lg2de:
Would be great to merge this PR without squashing because of the refactorings.

dennisdoomen:
> Would be great to merge this PR without squashing because of the refactorings.

If you would squash some of the commits that belong together and keep the refactoring and renames separate, we can.

lg2de:
> If you would squash some of the commits that belong together and keep the refactoring and renames separate, we can.

Hope, this is sufficient.

jnyrup (approved):

PR #1971, opened 2022-08-14
Make `CompletesWithinTimeoutAsync` private before release

lg2de:
In #1967 I added two protected methods. In review I made only one private.
After suggestion of @jnyrup I made both private.
For better review I split it into two commits.
Updating release notes should not be necessary.


jnyrup:
Thanks. 
I suspect the pipeline will fail due to approval tests needing to be updated.

jnyrup (approved):

dennisdoomen (approved):

PR #1972, opened 2022-08-14
Prepare constructors of `AsyncFunctionAssertions` to make them protected in V7

lg2de:
While implementing #1967 I found Rider hint on the second constructor of `AsyncFunctionAssertions` that it can be made protected.
Thinking deeper I found that the this class is constructed like a base class but not implemented like that.

I recommend to make this class a real base class, maybe also with renaming, but at least making used ctor protected and removing unused ctor. Further some methods can be moved to one base class because it is over written in the other one (with `new`).
All these are breaking changes, I guess. This is why I recommend to make the ctors obsolete today.
What do you think?

### DIFF
--- a/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
@@ -15,11 +15,13 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
     where TTask : Task
     where TAssertions : AsyncFunctionAssertions<TTask, TAssertions>
 {
+    [Obsolete("This class is intended as base class. This ctor is accidentally public and will be removed in Version 7.")]
     public AsyncFunctionAssertions(Func<TTask> subject, IExtractExceptions extractor)
         : this(subject, extractor, new Clock())
     {
     }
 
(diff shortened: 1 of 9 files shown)


dennisdoomen (approved):

jnyrup (approved):

lg2de:
Oh, thank for fast merge.
I guess we need an information in the release notes, isn't it?`It could be possible that the obsolete ctors are used.
I hadn't written the release notes yet because I wanted to know your opinion first.
And how about a label for the automatic release information?
Please advice if I should create release notes.

jnyrup:
>  What do you think?

I completely missed that line 😄 

I see you previously added this comment.
> Many projects have activated "Warnung as Error". So, `Obsolete` will cause build errors. Therefore it should not be used here.

https://github.com/fluentassertions/fluentassertions/pull/1673#issuecomment-909982753

In #1950 we added `[Obsolete]` which together with the ones added from this PR are the only ones we have (exception on guarding `Should`s).

I don't really have any gut feeling about the best approach here.

dennisdoomen:
It is _very_ unlikely that somebody would use those constructors.

lg2de:
> I see you previously added this comment.

You are right, this looks inconsistent. But there it was regarding "normal customer" API, here it is regarding "extension developer" API. I guess there are much less users accessing the FA classes and there are much less possible errors to be fixed.

PR #1974, opened 2022-08-17
Added `ThrowWithinAsync` assertions

lg2de:
Any thoughts before I continue, @dennisdoomen, @jnyrup?

### DIFF
--- a/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
@@ -112,9 +112,7 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
     /// <summary>
     /// Asserts that the current <see cref="Func{Task}"/> throws an exception of the exact type <typeparamref name="TException"/> (and not a derived exception type).
     /// </summary>
-    /// <typeparam name="TException">
-    /// The type of the exception it should throw.
-    /// </typeparam>
+    /// <typeparam name="TException">The type of exception expected to be thrown.</typeparam>
     /// <param name="because">
     /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
     /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
@@ -156,6 +154,7 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
     /// <summary>
     /// Asserts that the current <see cref="Func{Task}"/> throws an exception of type <typeparamref name="TException"/>.
     /// </summary>
+    /// <typeparam name="TException">The type of exception expected to be thrown.</typeparam>
     /// <param name="because">
     /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
     /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
@@ -316,6 +315,7 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
     /// <summary>
     /// Asserts that the current <see cref="Func{Task}"/> does not throw an exception of type <typeparamref name="TException"/>.
     /// </summary>
+    /// <typeparam name="TException">The type of exception expected to not be thrown.</typeparam>
     /// <param name="because">
     /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
     /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
--- a/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions.Common;
@@ -181,6 +182,106 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
     }
 
     /// <summary>
+    /// Asserts that the current <see cref="Func{Task}"/> throws an exception of type <typeparamref name="TException"/>
+    /// within a specific timeout.
+    /// </summary>
+    /// <typeparam name="TException">The type of exception expected to be thrown.</typeparam>
+    /// <param name="timeSpan">The allowed time span for the operation.</param>
+    /// <param name="because">
+    /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+    /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+    /// </param>
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+    /// </param>
+    public async Task<ExceptionAssertions<TException>> ThrowWithinAsync<TException>(
+        TimeSpan timeSpan, string because = "", params object[] becauseArgs)
+        where TException : Exception
+    {
+        bool success = Execute.Assertion
+            .ForCondition(Subject is not null)
+            .BecauseOf(because, becauseArgs)
+            .FailWith("Expected {context} to throw {0} within {1}{reason}, but found <null>.",
+                typeof(TException), timeSpan);
+
+        if (success)
+        {
+            Exception caughtException = await InvokeWithInterceptionAsync(timeSpan);
+            return AssertThrows<TException>(caughtException, timeSpan, because, becauseArgs);
+        }
+
+        return new ExceptionAssertions<TException>(Array.Empty<TException>());
+    }
+
+    private ExceptionAssertions<TException> AssertThrows<TException>(
+        Exception exception, TimeSpan timeSpan, string because, object[] becauseArgs)
+        where TException : Exception
+    {
+        TException[] expectedExceptions = Extractor.OfType<TException>(exception).ToArray();
+
+        Execute.Assertion
+            .BecauseOf(because, becauseArgs)
+            .WithExpectation("Expected a <{0}> to be thrown within {1}{reason}, ",
+                typeof(TException), timeSpan)
+            .ForCondition(exception is not null)
+            .FailWith("but no exception was thrown.")
+            .Then
+            .ForCondition(expectedExceptions.Any())
+            .FailWith("but found <{0}>: {1}{2}.",
+                exception?.GetType(),
+                Environment.NewLine,
+                exception)
+            .Then
+            .ClearExpectation();
+
+        return new ExceptionAssertions<TException>(expectedExceptions);
+    }
+
+    private async Task<Exception> InvokeWithInterceptionAsync(TimeSpan timeout)
+    {
+        try
+        {
+            // For the duration of this nested invocation, configure CallerIdentifier
+            // to match the contents of the subject rather than our own call site.
+            //
+            //   Func<Task> action = async () => await subject.Should().BeSomething();
+            //   await action.Should().ThrowAsync<Exception>();
+            //
+            // If an assertion failure occurs, we want the message to talk about "subject"
+            // not "await action".
+            using (CallerIdentifier.OnlyOneFluentAssertionScopeOnCallStack()
+                       ? CallerIdentifier.OverrideStackSearchUsingCurrentScope()
+                       : default)
+            {
+                (TTask task, TimeSpan remainingTime) = InvokeWithTimer(timeout);
+                if (remainingTime < TimeSpan.Zero)
+                {
+                    // timeout reached without exception
+                    return null;
+                }
+
+                if (task.IsFaulted)
+                {
+                    // exception in synchronous portion
+                    return task.Exception!.GetBaseException();
+                }
+
+                // Start monitoring the task regarding timeout.
+                // Here we do not need to know whether the task completes (successfully) in timeout
+                // or does not complete. We are only interested in the exception which is thrown, not returned.
+                // So, we can ignore the result.
+                _ = await CompletesWithinTimeoutAsync(task, remainingTime);
+            }
+
+            return null;
+        }
+        catch (Exception exception)
+        {
+            return exception;
+        }
+    }
+
+    /// <summary>
     /// Asserts that the current <see cref="Func{Task}"/> does not throw any exception.
     /// </summary>
     /// <param name="because">
@@ -330,18 +431,25 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
     /// </summary>
     private protected async Task<bool> CompletesWithinTimeoutAsync(Task target, TimeSpan remainingTime)
     {
-        using var timeoutCancellationTokenSource = new CancellationTokenSource();
+        using var delayCancellationTokenSource = new CancellationTokenSource();
 
         Task completedTask =
-            await Task.WhenAny(target, Clock.DelayAsync(remainingTime, timeoutCancellationTokenSource.Token));
+            await Task.WhenAny(target, Clock.DelayAsync(remainingTime, delayCancellationTokenSource.Token));
+
+        if (completedTask.IsFaulted)
+        {
+            // Throw the inner exception.
+            await completedTask;
+        }
 
         if (completedTask != target)
         {
+            // The monitored task did not complete.
             return false;
         }
 
-        // cancel the clock
-        timeoutCancellationTokenSource.Cancel();
+        // The monitored task is completed, we shall cancel the clock.
+        delayCancellationTokenSource.Cancel();
         return true;
     }
 
--- a/Src/FluentAssertions/Specialized/DelegateAssertionsBase.cs
+++ b/Src/FluentAssertions/Specialized/DelegateAssertionsBase.cs
@@ -17,21 +17,22 @@ public abstract class DelegateAssertionsBase<TDelegate, TAssertions>
     where TDelegate : Delegate
     where TAssertions : DelegateAssertionsBase<TDelegate, TAssertions>
 {
-    private readonly IExtractExceptions extractor;
+    private protected IExtractExceptions Extractor { get; }
 
     private protected DelegateAssertionsBase(TDelegate @delegate, IExtractExceptions extractor, IClock clock)
         : base(@delegate)
     {
-        this.extractor = extractor ?? throw new ArgumentNullException(nameof(extractor));
+        Extractor = extractor ?? throw new ArgumentNullException(nameof(extractor));
         Clock = clock ?? throw new ArgumentNullException(nameof(clock));
     }
 
     private protected IClock Clock { get; }
 
-    protected ExceptionAssertions<TException> ThrowInternal<TException>(Exception exception, string because, object[] becauseArgs)
+    protected ExceptionAssertions<TException> ThrowInternal<TException>(
+        Exception exception, string because, object[] becauseArgs)
         where TException : Exception
     {
-        TException[] expectedExceptions = extractor.OfType<TException>(exception).ToArray();
+        TException[] expectedExceptions = Extractor.OfType<TException>(exception).ToArray();
 
         Execute.Assertion
             .BecauseOf(because, becauseArgs)
@@ -63,7 +64,7 @@ public abstract class DelegateAssertionsBase<TDelegate, TAssertions>
     protected AndConstraint<TAssertions> NotThrowInternal<TException>(Exception exception, string because, object[] becauseArgs)
         where TException : Exception
     {
-        IEnumerable<TException> exceptions = extractor.OfType<TException>(exception);
+        IEnumerable<TException> exceptions = Extractor.OfType<TException>(exception);
 
         Execute.Assertion
             .ForCondition(!exceptions.Any())
--- a/Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
+++ b/Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
@@ -2272,6 +2272,8 @@ namespace FluentAssertions.Specialized
             where TException : System.Exception { }
         public System.Threading.Tasks.Task<FluentAssertions.Specialized.ExceptionAssertions<TException>> ThrowExactlyAsync<TException>(string because = "", params object[] becauseArgs)
             where TException : System.Exception { }
+        public System.Threading.Tasks.Task<FluentAssertions.Specialized.ExceptionAssertions<TException>> ThrowWithinAsync<TException>(System.TimeSpan timeSpan, string because = "", params object[] becauseArgs)
+            where TException : System.Exception { }
     }
     public abstract class DelegateAssertionsBase<TDelegate, TAssertions> : FluentAssertions.Primitives.ReferenceTypeAssertions<TDelegate, FluentAssertions.Specialized.DelegateAssertionsBase<TDelegate, TAssertions>>
         where TDelegate : System.Delegate
--- a/Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
@@ -1,9 +1,12 @@
 ﻿using System;
 using System.Threading;
 using System.Threading.Tasks;
+using FluentAssertions.Execution;
 using FluentAssertions.Extensions;
 using Xunit;
 using Xunit.Sdk;
+
+using static FluentAssertions.FluentActions;
 
 namespace FluentAssertions.Specs.Specialized;
 
@@ -44,7 +47,7 @@ public static class TaskAssertionSpecs
         }
 
         [Fact]
-        public async Task When_task_completes_fast_t_should_succeed()
+        public async Task When_task_completes_fast_it_should_succeed()
         {
             // Arrange
             var timer = new FakeClock();
@@ -96,7 +99,6 @@ public static class TaskAssertionSpecs
             // Act
             Func<Task> action = () =>
                 taskFactory.Awaiting(t => (Task)t.Task).Should(timer).CompleteWithinAsync(100.Milliseconds());
-
             timer.Complete();
 
             // Assert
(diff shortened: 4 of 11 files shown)


jnyrup:
Sorry for the silence.
I would prefer to discuss new APIs in issues before heading to implementations as it makes it easier for us to _only_ focus on the shape of the API and not the implementation details.
Please open an issue and if possible provide some thoughts/examples about the use cases.

This reminds me that I have to finish some documentation on the procress of proposing new APIs.
It will be heavily inspired by https://github.com/dotnet/runtime/blob/main/docs/project/api-review-process.md, but scaled down to a two-person OSS project.

lg2de:
implements #2034

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
```diff
@@ -2273,7 +2275,7 @@ namespace FluentAssertions.Specialized
         protected FluentAssertions.AndConstraint<TAssertions> NotThrowInternal(System.Exception exception, string because, object[] becauseArgs) { }
         protected FluentAssertions.AndConstraint<TAssertions> NotThrowInternal<TException>(System.Exception exception, string because, object[] becauseArgs)
             where TException : System.Exception { }
-        protected FluentAssertions.Specialized.ExceptionAssertions<TException> ThrowInternal<TException>(System.Exception exception, string because, object[] becauseArgs)
+        protected FluentAssertions.Specialized.ExceptionAssertions<TException> ThrowInternal<TException>(System.Exception exception, System.TimeSpan timeSpan, string because, object[] becauseArgs)
```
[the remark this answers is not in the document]
jnyrup: We can avoid changing this overload, which I would also prefer since `InvokeWithInterceptionAsync(Func<Task> action, TimeSpan timeSpan)` only uses `action` in the true part and `timeSpan` in the false part.

https://github.com/jnyrup/fluentassertions/commit/9f5676cd7015674d62313a2c1a2239b7e07b6ea5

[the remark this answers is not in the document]
dennisdoomen: I don't understand what @jnyrup is suggesting, but I would just duplicate the implementation of `ThrowInternal`. In retrospect, it should have never been added to the base-class and should have been a local `private` method. IMO, base-classes should never be used to reuse code. Use composition instead.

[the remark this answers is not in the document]
jnyrup: I'm suggesting the same. 
Duplicating `ThrowInternal` and `InvokeWithInterceptionAsync` rather than introducing conditional logic to fit both flows into the same methods.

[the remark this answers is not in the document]
lg2de: I'm fine with optimizing the code, but I would like to understand your motivations.

Dennis wrote "IMO, base-classes should **never** be used to reuse code."
Why you are such strict? What is your intention for base-classes?

Jonas wrote "... and timeSpan (only) in the false part" which is not correct because it is part of the condition,
and "... rather than introducing conditional logic to fit both flows into the same methods." without any explanation.

I think having one method `InvokeWithInterceptionAsync` including some variations is easier to understand than two methods with same name and totally different signature and behavior (more than common for method overloads).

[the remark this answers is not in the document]
dennisdoomen: Let's just say that I've experienced the pain more than a few times. Using base-classes like that leads to classes that don't have a clear purpose other than reusing logic. This often creates unnecessary coupling that I'd like to avoid. I've seen my fair share of those becoming god classes that nobody really understands. In this case, if we would have kept the logic private to the consuming class, we would have never had a need for the 'protected` member in the first place and never had to deal with the potentially breaking change.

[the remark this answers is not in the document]
lg2de: Then please define and check such rules for FA globally. 
We have already several base classes with and without protected members.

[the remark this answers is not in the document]
dennisdoomen: We were discussing this in the context of this specific PR. And those rules are not that black and white. Sometimes a base class is fine and sometimes it isn't. It all depends on its purpose. Also, time will also change the way we look at our own code.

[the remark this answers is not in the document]
lg2de: But this makes it complicated for contributors. 
Things I did in the "right way" in the past are "wrong" next time.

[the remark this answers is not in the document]
dennisdoomen: Yep, that's possible. And that's why we review all PRs ;-)

dennisdoomen:
Now that the API is approved, I guess you can continue, right?

lg2de:
> Now that the API is approved, I guess you can continue, right?

Yes, so far I did not have sufficient time.

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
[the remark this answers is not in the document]
lg2de: I've restarted the PR with all the received input.
Please review again.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -167,6 +166,32 @@ public AsyncFunctionAssertions(Func<TTask> subject, IExtractExceptions extractor
...
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+    /// </param>
+    public async Task<ExceptionAssertions<TException>> ThrowWithinAsync<TException>(
+        TimeSpan timeSpan, string because = "", params object[] becauseArgs)
+        where TException : Exception
+    {
+        Execute.Assertion
+            .ForCondition(Subject is not null)
+            .BecauseOf(because, becauseArgs)
```
jnyrup, line 187: We're missing a test that excersizes when `Subject` is null.

Src/FluentAssertions/Specialized/DelegateAssertionsBase.cs
```diff
@@ -49,6 +50,30 @@ protected ExceptionAssertions<TException> ThrowInternal<TException>(Exception ex
...
+        TException[] expectedExceptions = extractor.OfType<TException>(exception).ToArray();
+
+        Execute.Assertion
+            .BecauseOf(because, becauseArgs)
+            .WithExpectation("Expected a <{0}> to be thrown within {1}{reason}, ",
+                typeof(TException), timeSpan)
+            .ForCondition(exception is not null)
+            .FailWith("but no exception was thrown.")
+            .Then
+            .ForCondition(expectedExceptions.Any())
```
jnyrup, line 66: We're missing a test where the an exception is thrown within the expected timeout, but of an unexpected type.

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
```diff
@@ -180,6 +181,117 @@ public async Task When_task_completes_late_it_should_succeed()
...
+            // Arrange
+            var timer = new FakeClock();
+            var taskFactory = new TaskCompletionSource<bool>();
+
+            // Act
+            Func<Task> action = () =>
+            {
+                return taskFactory
+                    .Awaiting(t => (Task)t.Task)
+                    .Should(timer).ThrowWithinAsync<InvalidOperationException>(100.Milliseconds());
```
jnyrup, line 284: This would be a fitting test to `because` and `becauseArgs` and verify them.

jnyrup (commented):
Looks really good, thanks!

Only a few comments about test coverage.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
lg2de: done

Src/FluentAssertions/Specialized/DelegateAssertionsBase.cs
lg2de: done

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
lg2de: done

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -167,6 +166,32 @@ public AsyncFunctionAssertions(Func<TTask> subject, IExtractExceptions extractor
...
+            .FailWith("Expected {context} to throw {0} within {1}{reason}, but found <null>.",
+                typeof(TException), timeSpan);
+
+        Exception exception = await InvokeWithInterceptionAsync(timeSpan);
```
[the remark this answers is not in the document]
lg2de: done

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -177,6 +175,37 @@ public AsyncFunctionAssertions(Func<TTask> subject, IExtractExceptions extractor
...
+        TimeSpan timeSpan, string because = "", params object[] becauseArgs)
+        where TException : Exception
+    {
+        bool success = Execute.Assertion
+            .ForCondition(Subject is not null)
+            .BecauseOf(because, becauseArgs)
+            .FailWith("Expected {context} to throw {0} within {1}{reason}, but found <null>.",
+                typeof(TException), timeSpan);
+
+        if (!success)
```
jnyrup, line 200: Let's follow the project style of not using early returns.
(despite our shared personal preference 😉)

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -10,7 +10,6 @@ namespace FluentAssertions.Specialized;
 /// <summary>
 /// Contains a number of methods to assert that an asynchronous method yields the expected result.
 /// </summary>
-[DebuggerNonUserCode]
```
jnyrup, line 13: revert?

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
dennisdoomen: > Let's follow the project style of not using early returns.

Early returns at the top of the method are fine by me too. Just not the ones hidden in the middle of the code base.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
lg2de: ups...

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
lg2de: Mmh, I just copied from Jonas' changes above...
I'll change it in the same way when discussion in #2105 is completed.

lg2de: done

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
```diff
@@ -180,6 +184,274 @@ public async Task When_task_completes_late_it_should_succeed()
...
+            {
+                return taskFactory
+                    .Awaiting(t => (Task)t.Task)
+                    .Should(timer).ThrowAsync<InvalidOperationException>();
+            };
+            taskFactory.SetException(new InvalidOperationException("foo"));
+
+            // Assert
+            await action.Should().CompleteWithinAsync(1.Seconds());
+        }
```
jnyrup, line 241: Disclaimer, I often struggle with mentally keeping track of everything our timing related tests...

But shouldn't we use `timer` for anything? 🤔 
Otherwise, would the snippet below suffice?

```cs
[Fact]
public async Task When_task_throws_it_should_succeed()
{
    // Arrange
    var taskFactory = new TaskCompletionSource<bool>();

    // Act
    Func<Task> action = () =>
    {
        return taskFactory
            .Awaiting(t => (Task)t.Task)
            .Should().ThrowAsync<InvalidOperationException>();
    };
    taskFactory.SetException(new InvalidOperationException("foo"));

    // Assert
    await action.Should().NotThrowAsync();
}
```

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
```diff
@@ -180,6 +184,274 @@ public async Task When_task_completes_late_it_should_succeed()
...
+            {
+                return Awaiting(() => (Task)taskFactory.Task)
+                    .Should(timer).ThrowWithinAsync<InvalidOperationException>(1.Seconds());
+            };
+            var _ = action.Invoke();
+            taskFactory.SetException(new InvalidOperationException("foo"));
+
+            // Assert
+            await action.Should().CompleteWithinAsync(1.Seconds());
+        }
```
jnyrup, line 362: Why do we begin executing the task here?
`var _ = action.Invoke();`

That seems odd to me, as if we're not actually testing what we think we're testing.

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
lg2de: `CompleteWithinAsync` and `NotThrowAsync` has the same meaning but `CompleteWithinAsync` defines a timeout for the test.
But you are right, the timer can be removed here.

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
lg2de: The test wants to validate the scenario, that the asynchronous method throws AFTER the synchronous portion has been completed. Therefore I need to "invoke".

lg2de:
Please note that I create my commits always to be merged without squashing.
If you disagree with my commits please drop me a note.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -366,4 +404,48 @@ private static async Task<Exception> InvokeWithInterceptionAsync(Func<Task> acti
             return exception;
         }
     }
+
+    private async Task<Exception> InvokeWithInterceptionAsync(TimeSpan timeSpan)
```
dennisdoomen, line 408: 🔧 This method should be placed directly under the calling `ThrowWithinAsync`
🔧 I think renaming `timeSpan` to `timeOut` makes a bit more self-explanatory

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -177,6 +176,37 @@ public AsyncFunctionAssertions(Func<TTask> subject, IExtractExceptions extractor
...
+    {
+        bool success = Execute.Assertion
+            .ForCondition(Subject is not null)
+            .BecauseOf(because, becauseArgs)
+            .FailWith("Expected {context} to throw {0} within {1}{reason}, but found <null>.",
+                typeof(TException), timeSpan);
+
+        if (success)
+        {
+            Exception exception = await InvokeWithInterceptionAsync(timeSpan);
```
dennisdoomen, line 203: 🔧 I think we need to call this variable `caughtException` to emphasize that it can be `null`.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -177,6 +176,37 @@ public AsyncFunctionAssertions(Func<TTask> subject, IExtractExceptions extractor
...
+            return ThrowInternal<TException>(exception, timeSpan, because, becauseArgs);
```
dennisdoomen, line 204: 🔧 This method implies it's throwing an exception, but I think the point is to assert that a particular exception was thrown. Let's name it accordingly.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -330,12 +361,19 @@ private protected async Task<bool> CompletesWithinTimeoutAsync(Task target, Time
...
+
         if (completedTask != target)
         {
+            // The monitored task did not complete.
             return false;
         }
 
-        // cancel the clock
+        // The monitored task is completed, we shall cancel the clock.
         timeoutCancellationTokenSource.Cancel();
```
dennisdoomen, line 377: 🏕️ This name is confusing. Would you mind renaming it to something that better describes you're canceling the delay?

Src/FluentAssertions/Specialized/DelegateAssertionsBase.cs
```diff
@@ -49,6 +50,30 @@ protected ExceptionAssertions<TException> ThrowInternal<TException>(Exception ex
         return new ExceptionAssertions<TException>(expectedExceptions);
     }
 
+    private protected ExceptionAssertions<TException> ThrowInternal<TException>(
```
dennisdoomen, line 53: 🔧 This method belongs in the `AsyncFunctionAssertions`,

dennisdoomen (changes requested):

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
IT-VBFK: I think `timeout` is the correct casing?

dennisdoomen: Yeah, since we're following the [US spelling](https://grammarist.com/spelling/timeout-time-out/), I guess you're right

lg2de: Ok, done.
So far I supposed ordering by member type ("public", "private", "static", etc.).

Src/FluentAssertions/Specialized/DelegateAssertionsBase.cs
lg2de: This does not work because it is accessing `private readonly IExtractExceptions extractor`.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
dennisdoomen: No, we order according to https://csharpcodingguidelines.com/layout-guidelines/#AV2406.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
dennisdoomen: It's `private protected`, so how would it be a breaking change?

Src/FluentAssertions/Specialized/DelegateAssertionsBase.cs
dennisdoomen: The `AsyncFunctionAssertions` class also has that `IExtractExceptions` as a parameter.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
lg2de: The new one is private, the old one is protected. The old one was invented and named long time ago.
Both have similar behavior, so both should have similar name, isn't it? 
This is why I recommend consistent renaming with V7.

dennisdoomen: When you move it back to `AsyncFunctionAssertions`, the naming becomes a local concern (as-in within the same class), so it becomes less important the naming is concistent.

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
```diff
@@ -185,6 +187,272 @@ public async Task When_task_completes_late_it_should_succeed()
...
+            var timer = new FakeClock();
+            var taskFactory = new TaskCompletionSource<bool>();
+
+            // Act
+            Func<Task> action = () =>
+            {
+                return Awaiting(() => (Task)taskFactory.Task)
+                    .Should(timer).ThrowWithinAsync<InvalidOperationException>(1.Seconds());
+            };
+            var _ = action.Invoke();
```
jnyrup, line 358:
```suggestion
            _ = action.Invoke();
```

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
```diff
@@ -185,6 +187,272 @@ public async Task When_task_completes_late_it_should_succeed()
...
```
jnyrup, line 446:
```suggestion
            _ = action.Invoke();
```

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
jnyrup: Roger.
From CC I see that this also exercises the `return exception;` statement on line 254.

I would have thought we could simplify this scenario with the following, but it fails when for me running the build script.
```cs
var action = async () =>
{
    await Task.Yield();
    throw new InvalidOperationException("foo");
};

// Act
Func<Task> testAction = () => action.Should().ThrowWithinAsync<InvalidOperationException>(1.Seconds());

// Assert
await testAction.Should().NotThrowAsync();
```

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
```diff
@@ -185,6 +187,272 @@ public async Task When_task_completes_late_it_should_succeed()
...
+                    .Should(timer).ThrowWithinAsync<InvalidOperationException>(
+                        100.Milliseconds(), "because we want to test the failure {0}", "message");
+            };
+            timer.Delay(200.Milliseconds());
+
+            // Assert
+            await action.Should().ThrowAsync<XunitException>().WithMessage(
+                "Expected a <System.InvalidOperationException> to be thrown within 100ms"
+                + " because we want to test the failure message,"
+                + " but no exception was thrown.");
```
jnyrup, line 386:
```suggestion
            // Arrange
            var timer = new FakeClock();
            var taskFactory = new TaskCompletionSource<bool>();

            // Act
            Func<Task> action = () =>
            {
                return taskFactory
                    .Awaiting(t => (Task)t.Task)
                    .Should(timer).ThrowWithinAsync<InvalidOperationException>(
                        100.Ticks(), "because we want to test the failure {0}", "message");
            };
            timer.Delay(101.Ticks());

            // Assert
            await action.Should().ThrowAsync<XunitException>().WithMessage(
                "Expected a <System.InvalidOperationException> to be thrown within 10.0µs"
                + " because we want to test the failure message,"
                + " but no exception was thrown.");
```

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
```diff
@@ -185,6 +187,272 @@ public async Task When_task_completes_late_it_should_succeed()
...
+            {
+                return taskFactory
+                    .Awaiting(t => (Task)t.Task)
+                    .Should(timer).ThrowWithinAsync<InvalidOperationException>(100.Milliseconds());
+            };
+            taskFactory.SetException(new InvalidOperationException("foo"));
+
+            // Assert
+            await action.Should().CompleteWithinAsync(1.Seconds());
+        }
```
jnyrup, line 343: To exercise the edge cases around the `timeSpan` limit.

```cs
[Theory]
[InlineData(99)]
[InlineData(100)]
public async Task When_task_throws_fast_it_should_succeed(int delay)
{
    // Arrange
    var timer = new FakeClock();
    var taskFactory = new TaskCompletionSource<bool>();

    // Act
    Func<Task> action = () =>
    {
        return taskFactory
            .Awaiting(t => (Task)t.Task)
            .Should(timer).ThrowWithinAsync<InvalidOperationException>(100.Ticks());
    };

    timer.Delay(delay.Ticks());
    taskFactory.SetException(new InvalidOperationException("foo"));

    // Assert
    await action.Should().NotThrowAsync();
}

[Fact]
public async Task When_task_throws_slow_it_should_fail()
{
    // Arrange
    var timer = new FakeClock();
    var taskFactory = new TaskCompletionSource<bool>();

    // Act
    Func<Task> action = () =>
    {
        return taskFactory
            .Awaiting(t => (Task)t.Task)
            .Should(timer).ThrowWithinAsync<InvalidOperationException>(100.Ticks());
    };

    timer.Delay(101.Ticks());
    taskFactory.SetException(new InvalidOperationException("foo"));
    timer.Complete();

    // Assert
    await action.Should().ThrowAsync<XunitException>();
}
```

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
```diff
@@ -185,6 +187,272 @@ public async Task When_task_completes_late_it_should_succeed()
...
+                    .Awaiting(t => (Task)t.Task)
+                    .Should(timer).ThrowAsync<InvalidOperationException>();
+            };
+            taskFactory.SetResult(true);
+            timer.Complete();
+
+            // Assert
+            await action.Should().ThrowAsync<XunitException>().WithMessage(
+                "Expected a <System.InvalidOperationException> to be thrown, but no exception was thrown.");
+        }
```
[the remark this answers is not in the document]
lg2de: In the first test you are USING `ThrowAsync` instead of TEST it.
In general you are right and I've updated the tests accordingly.

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
lg2de: Great work!
The only thing I've changed is using the assertions with timeout.
This is reason why I've introduced them: The tested code may cause deadlocks. In this case the test may never complete.
Asynchronous methods may always be tested with timeout.

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
```diff
@@ -39,12 +42,12 @@ public async Task When_subject_is_null_it_should_throw()
                 timeSpan, "because we want to test the failure {0}", "message");
 
             // Assert
-            await testAction.Should().ThrowAsync<XunitException>()
+            await testAction.Should().ThrowWithinAsync<XunitException>(1.Seconds())
```
jnyrup, line 45: Why are you changing many assertions to use `ThrowWithinAsync` and `CompleteWithinAsync`? 🤔

lg2de: See above.

> The only thing I've changed is using the assertions with timeout.
> This is reason why I've introduced them: The tested code may cause deadlocks. In this case the test may never complete.
> Asynchronous methods may always be tested with timeout.

jnyrup: In what case can we deadlock?
Adding timeouts to combat deadlocks sounds like symptomatic treatment to me.

lg2de: Due to bugs in the tested code. This will create early failing tests.

jnyrup: So we're talking about potential deadlocks?
I'd like to not add timeouts prematurely.

jnyrup: For certain test suites it's undoubtedly useful, but let's leave them out of the this test suite.
Our tests are small enough that if we face deadlocks _something_ will fail and we can diagnose it at that time.
For now we think it makes the tests less clear on _exactly_ what scenario they're testing.

dennisdoomen: I agree with @jnyrup here.

lg2de: done

dennisdoomen (approved):

dennisdoomen:
There's a tiny regression in test coverage. Not sure if it's really an issue.

Tests/FluentAssertions.Specs/Specialized/TaskAssertionSpecs.cs
```diff
@@ -181,7 +183,286 @@ public async Task When_task_completes_late_it_should_succeed()
...
+            Func<Task> action = () =>
+            {
+                return Awaiting(() => (Task)taskFactory.Task)
+                    .Should(timer).ThrowWithinAsync<InvalidOperationException>(1.Seconds());
+            };
+            _ = action.Invoke();
+            taskFactory.SetException(new NotSupportedException("foo"));
+
+            // Assert
+            await action.Should().ThrowWithinAsync<XunitException>(1.Seconds()).WithMessage(
```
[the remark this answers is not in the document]
lg2de: I've fixed this already.
Maybe you have reviewed the previous version?

lg2de:
> There's a tiny regression in test coverage. Not sure if it's really an issue.

I know that there are lines not covered. But they are edge cases. I tried already to catch them.
I can try again the next days...

dennisdoomen:
> I can try again the next days...

Unless @jnyrup thinks so, don't bother.

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -180,6 +180,106 @@ public AsyncFunctionAssertions(Func<TTask> subject, IExtractExceptions extractor
...
+                {
+                    // exception in synchronous portion
+                    return result.Exception!.GetBaseException();
+                }
+
+                if (await CompletesWithinTimeoutAsync(result, remainingTime))
+                {
+                    // completed without exception
+                    return null;
+                }
```
jnyrup, line 272: I think we can write it like this to all keep lines covered.
```cs
_ = await CompletesWithinTimeoutAsync(result, remainingTime);

// completed without exception
return null;
```

In the other usages `CompletesWithinTimeoutAsync(` we the Boolean result in the `ForCondition`, whereas here we use the returned exception from this method.

lg2de: You are right, but it's a bit hard explain, WHY we can ignore the result, at least while "quick reading the code"...
So, I tried to write an explanation comment.
Thanks!

jnyrup (approved):

julealgon:
> Fixes #2034

Looks like this only works in the PR description. The issue is still open.

PR #1980, opened 2022-08-26
Deduplicate target frameworks

0xced:
In the ApproveApi test, read the target frameworks directly from the FluentAssertions.csproj file instead of copying them all as InlineData.

This should future proof when new target frameworks are added/removed from the main project.


jnyrup (approved):
Thanks!

I also didn't know about the strongly typed `TheoryData<T>` :muscle:

0xced:
Yes, this is a nice little known feature of xUnit.net 😀

PR #1981, opened 2022-08-26
Add support for asserting [ReadOnly]Span<char> and [ReadOnly]Span<T>

### LINKED ISSUE #902: Add support for Span and ReadOnlySpan

0xced:
When writing a test for an implementation of `ISpanFormattable` I was surprised that my Span<char> did not have a Should() method. I figured that the solution proposed by @jnyrup in https://github.com/fluentassertions/fluentassertions/issues/902#issuecomment-1167769893 would be a welcome addition to the library.

Fixes #902

### DIFF
--- a/Src/FluentAssertions/AssertionExtensions.cs
+++ b/Src/FluentAssertions/AssertionExtensions.cs
@@ -344,6 +344,28 @@ public static class AssertionExtensions
         return new GenericCollectionAssertions<T>(actualValue);
     }
 
+#if !NETFRAMEWORK && !NETSTANDARD2_0
+    /// <summary>
+    /// Returns an <see cref="GenericCollectionAssertions{T}"/> object that can be used to assert the
+    /// current <see cref="Span{T}"/>.
+    /// </summary>
+    [Pure]
+    public static GenericCollectionAssertions<T> Should<T>(this Span<T> actualValue)
+    {
+        return new GenericCollectionAssertions<T>(actualValue.ToArray());
+    }
+
+    /// <summary>
+    /// Returns an <see cref="GenericCollectionAssertions{T}"/> object that can be used to assert the
+    /// current <see cref="ReadOnlySpan{T}"/>.
+    /// </summary>
+    [Pure]
+    public static GenericCollectionAssertions<T> Should<T>(this ReadOnlySpan<T> actualValue)
+    {
+        return new GenericCollectionAssertions<T>(actualValue.ToArray());
+    }
+#endif
+
     /// <summary>
     /// Returns an <see cref="StringCollectionAssertions"/> object that can be used to assert the
     /// current <see cref="IEnumerable{T}"/>.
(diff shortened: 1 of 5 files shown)


jnyrup:
Are there any assertions in `GenericCollectionAssertions`, `StringAssertions` or their base classes that would return inconsistent/meaningless results?

Some examples:
* `[Not]BeSameAs` as two value types can't have referential equality.
* `[Not]BeNull` as a value type is always non-null.

[DataTableCollectionAssertionExtensions](https://github.com/fluentassertions/fluentassertions/blob/develop/Src/FluentAssertions/DataTableCollectionAssertionExtensions.cs) works around this problem by adding an extension method which resolves better than the provided one.

Maybe rather make `SpanAssertions<T> : GenericCollectionAssertions<T>` and override the problematic overloads?

Maybe there are no problems at all as the methods will return the "right" result, but just be meaningless to call?

lg2de:
Would be great to see some unit tests with several usages of the new function.
This helps to identify whether this approach works as expected.

dennisdoomen:
@0xced are you planning to continue working on this?

0xced:
I recently used the extensions in an actual project to test an `ISpanFormattable` implementation and realized something is not yet quite right. 😕

Here's the test I wrote:

```csharp
[Theory]
[InlineData(null, "Hello")]
[InlineData("iv", "Hello")]
[InlineData("de", "Guten Tag")]
[InlineData("de-CH", "Guten Tag")]
[InlineData("en", "Hello")]
[InlineData("en-CH", "Hello")]
[InlineData("fr", "Bonjour")]
[InlineData("fr-CH", "Bonjour")]
[InlineData("it", "Ciao")]
[InlineData("it-CH", "Ciao")]
public void TryFormat_GivenCultureInfo_WritesCorrectString(string? cultureName, string expected)
{
    // Arrange
    var culture = cultureName is null ? null : new CultureInfo(cultureName);
    var localizedString = new LocalizedString(German: "Guten Tag", French: "Bonjour", Italian: "Ciao", English: "Hello");
    Span<char> destination = stackalloc char[16];

    // Act
    var result = localizedString.TryFormat(destination, out var charsWritten, "", culture);

    // Assert
    result.Should().BeTrue();
    charsWritten.Should().Be(expected.Length);
    destination[..charsWritten].Should().Be(expected);
}
```

I'm not really happy with how the assertions looks like, especially this part: `destination[..charsWritten].Should()`.

So to answer your question: yes, I'd think I may continue working on this (also working on @jnyrup suggestion of `SpanAssertions<T>`).

Also, I'm open to your suggestions (if you have any) on how this `destination[..charsWritten].Should()` could be improved.

dennisdoomen:
@0xced you gave up on this?

dennisdoomen:
I'm going to close this one for lack of activity.

PR #1990, opened 2022-09-06
Improves how BeEquivalentTo handles fields hiding base-class fields

### LINKED ISSUE #1982: Should().BeEquivalentTo() throws "Sequence contains more than one matching element" when comparing  classes that have a field that uses the "new" keyword to hide a field with the same name.
I have class `A` that inherits class `ABase`. `ABase` contains field `Id` of type `string`. `A` has its own `Id` string field declared as `new` to hide the `Id` member coming from `ABase`.

Scenario:
Calling `A.Should().BeEquivalentTo(A)`

Expected:
The two fields are compared correctly for equivalency and the proper result is returned.

Actual:
Throws "Sequence contains more than one matching element", thrown on line 144 (return list. SingleOrDefault()) in method `TypeExtensions.FindProperty()` in my decompiled code.

Note:
Does not happen if you compare the same class instance to itself.

dennisdoomen:
Fixes #1982 

If a derived class is used as the expectation and hides a base-class field of the same type using the `new` keyword, FA would get confused about which field to use. Now, it'll start at the compile- or run-time type (depending on the options and reference passed) and find the property or field from there. 

This does have a potential side-effect. Since we don't know the compile-time type of the subject-under-test, we now do the same thing for the subject's properties and fields. Before this PR, we tried to find the best match. 

**Note** Review by commit

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Common/TypeExtensions.cs
+++ b/Src/FluentAssertions/Common/TypeExtensions.cs
@@ -178,16 +178,19 @@ internal static class TypeExtensions
     /// <returns>
     /// Returns <see langword="null"/> if no such property exists.
     /// </returns>
-    public static PropertyInfo FindProperty(this Type type, string propertyName, Type preferredType)
+    public static PropertyInfo FindProperty(this Type type, string propertyName)
     {
-        List<PropertyInfo> properties =
-            type.GetProperties(AllInstanceMembersFlag)
-                .Where(pi => pi.Name == propertyName)
-                .ToList();
+        while (type != typeof(object))
+        {
+            if (type.GetProperty(propertyName, AllInstanceMembersFlag | BindingFlags.DeclaredOnly) is { } property)
+            {
+                return property;
+            }
 
-        return properties.Count > 1
-            ? properties.SingleOrDefault(p => p.PropertyType == preferredType)
-            : properties.SingleOrDefault();
+            type = type.BaseType;
+        }
+
+        return null;
     }
 
     /// <summary>
@@ -196,16 +199,19 @@ internal static class TypeExtensions
     /// <returns>
     /// Returns <see langword="null"/> if no such property exists.
     /// </returns>
-    public static FieldInfo FindField(this Type type, string fieldName, Type preferredType)
+    public static FieldInfo FindField(this Type type, string fieldName)
     {
-        List<FieldInfo> properties =
-            type.GetFields(AllInstanceMembersFlag)
-                .Where(pi => pi.Name == fieldName)
-                .ToList();
+        while (type != typeof(object))
+        {
+            if (type.GetField(fieldName, AllInstanceMembersFlag | BindingFlags.DeclaredOnly) is { } field)
+            {
+                return field;
+            }
 
-        return properties.Count > 1
-            ? properties.SingleOrDefault(p => p.FieldType == preferredType)
-            : properties.SingleOrDefault();
+            type = type.BaseType;
+        }
+
+        return null;
     }
 
     public static IEnumerable<MemberInfo> GetNonPrivateMembers(this Type typeToReflect, MemberVisibility visibility)
--- a/Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
@@ -531,15 +531,40 @@ public class SelectionRulesSpecs
         }
 
         [Fact]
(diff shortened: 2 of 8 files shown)


jnyrup:
We currently have the "preferred type" - See `When_a_property_is_hidden_in_a_derived_class_it_should_ignore_it` introduced in 3c0f594993f6c69e9e2c5c09d213edb44c771648.

The test added for this PR seems to add an orthogonal criterion for selecting the most matching field/property, namely the most derived member.

Note: `When_a_property_is_hidden_in_a_derived_class_it_should_ignore_it` also pass if we only select the most derived member as its most derived member also happens to have the preferred type.

One way to only select based on the most derived type could be
```cs
public static PropertyInfo FindProperty(this Type type, string propertyName) // <-- removed `preferredType` parameter
{
    while (type != typeof(object))
    {
        if (type.GetProperty(propertyName, AllInstanceMembersFlag | BindingFlags.DeclaredOnly) is { } property)
        {
            return property;
        }

        type = type.BaseType;
    }

    return null;
}
```

There are probably cases with compile-time typing vs runtime typing that I haven't thought about. 
While few people hopefully uses hidden members, we should ensure we have or add good coverage for the changes here.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -506,53 +506,175 @@ public void When_a_property_is_protected_it_should_be_ignored()
         act.Should().NotThrow();
     }
 
-    [Fact]
-    public void When_a_property_is_hidden_in_a_derived_class_it_should_ignore_it()
+    public class MemberHiding
```
dennisdoomen, line 509: @jnyrup can you please check the next five specs to see if they make sense to you too. Then I can mimic the same ones for the fields.

jnyrup: They all make sense to me.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -617,28 +642,172 @@ public void Excluding_the_property_hiding_the_base_class_one_does_not_reveal_the
...
+
+            var expectation = new SubclassBHidingField
+            {
+                Field = "DerivedValue"
+            };
+
+            ((AnotherBaseWithField)expectation).Field = "ExpectedBaseValue";
+
+            // Act / Assert
+            subject.Should().BeEquivalentTo(expectation, _ => _.IncludingFields());
```
jnyrup, line 697: Should we keep this consistent with the majority of the test suite, i.e. using `opt => opt` or `options => options` instead of `_ => _`?

Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -178,16 +178,19 @@ public static bool OverridesEquals(this Type type)
...
+            }
+
+            type = type.BaseType;
+        }
 
-        return properties.Count > 1
-            ? properties.SingleOrDefault(p => p.PropertyType == preferredType)
-            : properties.SingleOrDefault();
+        return null;
     }
```
jnyrup, line 194: Nice! 🚀 

I was fiddling with this method semi-recently as I didn't like how we called `ToList()` only the select a single member from it.

jnyrup (approved):
Remember to update the release notes.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
dennisdoomen: Yeah, I'm fine with that. I only recently learned through Nuke that you can use `_` multiple times.

Src/FluentAssertions/Common/TypeExtensions.cs
jnyrup: I had completely forgot that, way to many beers ago.
But that explains why I bonded so much with it 😂

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
jnyrup: It easily becomes hard to read for humans.

E.g. is this perfectly valid C# and the two `_` refers to different parameters.
```cs
Func<int, Func<int, int>> NestedUsageOfDiscards() => _ => _ => 42;
```

For more C# syntax horror, see https://www.youtube.com/watch?v=jaPk6Nt33KM 
or https://twitter.com/Lucas_Trz/status/1613644113814986767

PR #2009, opened 2022-10-10
Treat record structs as records

### LINKED ISSUE #1808: record struct with List or IEnumerable property fails Should BeEquivalentTo
**Before you file a bug, have you:**

* Tried upgrading to newest version of Fluent Assertions, to see if your issue has already been resolved and released? Yes
* Checked existing open *and* closed [issues](https://github.com/fluentassertions/fluentassertions/issues?utf8=%E2%9C%93&q=is%3Aissue), to see if the issue has already been reported? Yes
* Tried reproducing your problem in a new isolated project? Yes
* Read the [documentation](https://fluentassertions.com/introduction)? Yes
* Considered if this is a general question and not a bug?. For general questions please use [StackOverflow](https://stackoverflow.com/questions/tagged/fluent-assertions?mixed=1).

### Description

A `record struct` which contains a `List` or `IEnumberable` fails `BeEquivalentTo`

```csharp
using FluentAssertions;
using Xunit;

namespace fluent_assertion_record_struct_issue;

record struct Foo(List<int> Values);

[Fact]
public void TestRecordStructWithLists()
{
    var expected = new Foo(new() {1, 2, 3});
    var actual = new Foo(new() {1, 2, 3});
}
```

### Expected behavior:

The test should pass, by default, just like with a `record` / `record class`

### Actual behavior:

```
Xunit.Sdk.XunitException: Expected actual to be Foo { Values = System.Collections.Generic.List`1[System.Int32] }, but found Foo { Values = System.Collections.Generic.List`1[System.Int32] }.

With configuration:
- Compare enums by value

   at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message) in /_/Src/FluentAssertions/Execution/XUnit2TestFramework.cs:line 33
   at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message) in /_/Src/FluentAssertions/Execution/TestFrameworkProvider.cs:line 32
   at FluentAssertions.Execution.CollectingAssertionStrategy.ThrowIfAny(IDictionary`2 context) in /_/Src/FluentAssertions/Execution/CollectingAssertionStrategy.cs:line 42
   at FluentAssertions.Equivalency.EquivalencyValidator.AssertEquality(Comparands comparands, EquivalencyValidationContext context) in /_/Src/FluentAssertions/Equivalency/EquivalencyValidator.cs:line 27
   at FluentAssertions.Primitives.ObjectAssertions`2.BeEquivalentTo[TExpectation](TExpectation expectation, Func`2 config, String because, Object[] becauseArgs) in /_/Src/FluentAssertions/Primitives/ObjectAssertions.cs:line 145
   at FluentAssertions.Primitives.ObjectAssertions`2.BeEquivalentTo[TExpectation](TExpectation expectation, String because, Object[] becauseArgs) in /_/Src/FluentAssertions/Primitives/ObjectAssertions.cs:line 97
```

### Versions

* Which version of Fluent Assertions are you using? 6.5.1
* Which .NET runtime and version are you targeting? .net 6

### Additional Information

`record` and `class` works fine.
This is related to #1507 and #1451.

I tried `options => options.ComparingRecordsByMembers()`, but that did not work.
`options => options.ComparingByMembers<Foo>()` _does_ work.

salvois:
Fixed #1808

There appears to be no objective way to detect whether a type is indeed a record struct. This pull requests tries to make a best guess by looking for a value type having at least a pair of compiler-generated equality and inequality operators.


Tests/FluentAssertions.Specs/Types/TypeExtensionsSpecs.cs
```diff
@@ -124,6 +125,33 @@ public void When_getting_fake_implicit_conversion_operator_from_a_type_with_fake
...
+    [InlineData(typeof(string), false)]
+    public void IsRecord_should_detect_records_correctly(Type type, bool expected)
+    {
+        type.IsRecord().Should().Be(expected);
+    }
+
+    [Fact]
+    public void When_checking_if_anonymous_type_is_record_it_should_return_false()
+    {
+        new { Value = 42 }.GetType().IsRecord().Should().Be(false);
```
IT-VBFK, line 146: Maybe use `Should().BeFalse()`

Tests/FluentAssertions.Specs/Types/TypeExtensionsSpecs.cs
```diff
@@ -124,6 +125,33 @@ public void When_getting_fake_implicit_conversion_operator_from_a_type_with_fake
...
+    }
+
+    [Fact]
+    public void When_checking_if_class_with_multiple_equality_methods_is_record_it_should_return_false()
+    {
+        typeof(ImmutableArray<int>).IsRecord().Should().Be(false);
```
IT-VBFK, line 152: Maybe use `Should().BeFalse()`

Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -584,12 +584,18 @@ private static bool IsAnonymousType(this Type type)
...
     {
         return TypeIsRecordCache.GetOrAdd(type, static t =>
-            t.GetMethod("<Clone>$") is not null &&
-            t.GetTypeInfo()
-                 .DeclaredProperties
-                 .FirstOrDefault(p => p.Name == "EqualityContract")?
-                 .GetMethod?
-                 .GetCustomAttribute(typeof(CompilerGeneratedAttribute)) is not null);
+        {
+            bool isRecord = t.GetMethod("<Clone>$") is not null &&
```
dennisdoomen, line 588: 🔧 For readability I would prefer to clean-up the layout a little bit and keep an empty line between the two boolean statements
🤔 Any links to background material that we can add here for future reference?

Tests/FluentAssertions.Equivalency.Specs/RecordSpecs.cs
```diff
@@ -16,6 +16,16 @@ public void When_the_subject_is_a_record_it_should_compare_it_by_its_members()
         actual.Should().BeEquivalentTo(expected);
     }
 
+    [Fact]
+    public void When_the_subject_is_a_readonly_record_struct_it_should_compare_it_by_its_members()
+    {
+        var actual = new MyReadonlyRecordStruct("foo", new[] { "bar", "zip", "foo" });
```
dennisdoomen, line 22: ⛏️ (nitpick) Would you mind adding the `// Arrange`, `// Act` and `// Assert` elements?

Tests/FluentAssertions.Specs/Types/TypeExtensionsSpecs.cs
```diff
@@ -124,6 +125,33 @@ public void When_getting_fake_implicit_conversion_operator_from_a_type_with_fake
...
+    [InlineData(typeof(MyRecord), true)]
+    [InlineData(typeof(MyRecordStruct), true)]
+    [InlineData(typeof(MyRecordStructWithOverriddenEquality), true)]
+    [InlineData(typeof(MyReadonlyRecordStruct), true)]
+    [InlineData(typeof(MyStruct), false)]
+    [InlineData(typeof(MyStructWithOverriddenEquality), false)]
+    [InlineData(typeof(MyClass), false)]
+    [InlineData(typeof(int), false)]
+    [InlineData(typeof(string), false)]
+    public void IsRecord_should_detect_records_correctly(Type type, bool expected)
```
dennisdoomen, line 138: 🤔 Is there a way to _not_ directly exercise tests against `IsRecord` and instead use the real API surface (`BeEquivalentTo`)?
🌱 (thinking out loud/seed) In retrospect, we should have kept `IsRecord` extension method inside the `Equivalency` folder

dennisdoomen (commented):
🔧 You also need to update the `releases.md` file to mention this change
🤔 Anything to add to the `equivalency.md` file about support for record structs?

salvois:
> 🤔 Anything to add to the `equivalency.md` file about support for record structs?

I guess you mean objectgraphs.md. I'm unsure about it, because it depends on how broad the meaning of "record" is in that context. If "record" is an umbrella term to mean both reference-type-based records and value-type-based record structs (either readonly or not), nothing has to be changed, otherwise we could be more explicit. For example, I was quite surprised discovering that record structs *were not* treated as records.

Src/FluentAssertions/Common/TypeExtensions.cs
salvois: No real background material actually, besides the link already posted in the linked issue (https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/record-structs#open-questions) where it is stated that recognizing record structs is an open point.
This pull request is based on common sense and heuristic testing, apparently giving good results but not supported by official documentation.

Tests/FluentAssertions.Equivalency.Specs/RecordSpecs.cs
salvois: I don't mind, but none of the tests in that file have those marker comments. I really prefer to strive for consistency.

Tests/FluentAssertions.Specs/Types/TypeExtensionsSpecs.cs
salvois: I guess that really depends on personal preference and what is one's view about units and composition. In one case you would test a smaller unit which is notable its own (in this case a pure function, besides caching, which is always nice to test 😄), then give its functionality for granted in broader context. At the other end you would exercise the code indirectly, hiding what could be an implementation detail at the expenses of a more verbose test (usually in the arrange part), resulting in a more "end to end" approach, if you allow me to stretch the meaning.
My personal preference is guided from the cardinality and complexity of scenarios, and in this case it leans towards the former approach, but feel free to dispose of the donated code as you see fit!

dennisdoomen:
> I guess you mean objectgraphs.md. I'm unsure about it, because it depends on how broad the meaning of "record" is in that context. If "record" is an umbrella term to mean both reference-type-based records and value-type-based record structs (either readonly or not), nothing has to be changed, otherwise we could be more explicit. For example, I was quite surprised discovering that record structs _were not_ treated as records.

Yes, sorry, that's what I meant. I think it wouldn't hurt to mention all of them.

Src/FluentAssertions/Common/TypeExtensions.cs
dennisdoomen: I would mention exactly using in-line comments

Tests/FluentAssertions.Specs/Types/TypeExtensionsSpecs.cs
dennisdoomen: I think that's an acceptable trade-off.

dennisdoomen (approved):

docs/_pages/objectgraphs.md
```diff
@@ -39,7 +39,7 @@ orderDto.Should().BeEquivalentTo(order, options =>
 
 ### Value Types
 
-To determine whether Fluent Assertions should recurs into an object's properties or fields, it needs to understand what types have value semantics and what types should be treated as reference types. The default behavior is to treat every type that overrides `Object.Equals` as an object that was designed to have value semantics. Anonymous types, records and tuples also override this method, but because the community proved us that they use them quite often in equivalency comparisons, we decided to always compare them by their members.
+To determine whether Fluent Assertions should recurs into an object's properties or fields, it needs to understand what types have value semantics and what types should be treated as reference types. The default behavior is to treat every type that overrides `Object.Equals` as an object that was designed to have value semantics. Anonymous types, records, record structs, readonly record structs and tuples also override this method, but because the community proved us that they use them quite often in equivalency comparisons, we decided to always compare them by their members.
```
jnyrup, line 42: As with the other comment, I don't think mentioning `readonly` applies here. 
How about just expanding the umbrella term into "record classes" and "record structs".

```suggestion
To determine whether Fluent Assertions should recurs into an object's properties or fields, it needs to understand what types have value semantics and what types should be treated as reference types. The default behavior is to treat every type that overrides `Object.Equals` as an object that was designed to have value semantics. Anonymous types, record classes, record structs and tuples also override this method, but because the community proved us that they use them quite often in equivalency comparisons, we decided to always compare them by their members.
```

Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -584,12 +584,29 @@ private static bool IsAnonymousType(this Type type)
...
+            // and heuristic testing, apparently giving good results but not supported by official documentation.
+            var isRecordStruct = t.BaseType == typeof(ValueType) &&
+                                 t.GetMethods()
+                                     .Where(m => m.Name == "op_Inequality")
+                                     .SelectMany(m => m.GetCustomAttributes(typeof(CompilerGeneratedAttribute)))
+                                     .Any() &&
+                                 t.GetMethods()
+                                     .Where(m => m.Name == "op_Equality")
+                                     .SelectMany(m => m.GetCustomAttributes(typeof(CompilerGeneratedAttribute)))
+                                     .Any();
```
jnyrup, line 606: While this is probably the best we can currently do, I wonder a bit if this is still too broad a check?

docs/_pages/releases.md
```diff
@@ -25,6 +25,7 @@ sidebar:
 
 ### Enhancements
 * Included the time difference in the error message of `BeCloseTo` - [#2013](https://github.com/fluentassertions/fluentassertions/pull/2013)
+* Treated record structs and readonly record structs as records, thus comparing them by member by default - [#2009](https://github.com/fluentassertions/fluentassertions/pull/2009)
```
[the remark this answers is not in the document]
salvois: I like yours more :smile: Though, I believe "record class" is not the official term to define them: in the official documentation they are just "records".

Tests/FluentAssertions.Equivalency.Specs/RecordSpecs.cs
```diff
@@ -16,6 +16,16 @@ public void When_the_subject_is_a_record_it_should_compare_it_by_its_members()
         actual.Should().BeEquivalentTo(expected);
     }
 
+    [Fact]
+    public void When_the_subject_is_a_readonly_record_struct_it_should_compare_it_by_its_members()
```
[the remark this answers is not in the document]
salvois: Definitely right. I've been influenced by my personal preference because in my opinion record structs should have been immutable in the first place!

Src/FluentAssertions/Common/TypeExtensions.cs
salvois: I'm afraid of this too. This is backed by MSDN too (see comment). I've tried my best to "crack" this test and was not able to, but please feel free to add counter-examples.

dennisdoomen:
@salvois any follow-up?

docs/_pages/objectgraphs.md
```diff
@@ -48,7 +48,7 @@ subject.Should().BeEquivalentTo(expected,
    options => options.ComparingByValue<IPAddress>());
 ```
 
-For records, this works like this:
+For records, record structs and readonly record structs this works like this:
```
jnyrup, line 51:
```suggestion
For records and record structs this works like this:
```

Src/FluentAssertions/Common/TypeExtensions.cs
jnyrup: Checking for one operator should be enough.
```cs
var isRecordStruct = t.BaseType == typeof(ValueType) &&
                     t.GetMethod("op_Equality", BindingFlags.Static | BindingFlags.Public | BindingFlags.DeclaredOnly, null, new[] { t, t }, null)?
                        .GetCustomAttribute(typeof(CompilerGeneratedAttribute)) is not null;
```

jnyrup (commented):
This was probably waiting for me.

My concerns boils down to how many false positives the heuristic will have and if we're willing to take that risk, or we want to wait until C# eventually resolves the open point about recognizing record structs.

dennisdoomen:
> My concerns boils down to how many false positives the heuristic will have and if we're willing to take that risk, or we want to wait until C# eventually resolves the open point about recognizing record structs.

I think it's fair for people to expect FA to properly handle record structs the way we treat other records. So although we're not fully confident we can cover all edge cases, I think it's worth the risk. Whatever comes after that, we'll fix.

Src/FluentAssertions/Common/TypeExtensions.cs
jnyrup: I had another look at creating a heuristic to identify record structs with few false positives.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/record-structs#printing-members-printmembers-and-tostring-methods

specifies that a `record struct` contains a synthesized (read has `[CompilerGenerated]`)
```cs
private bool PrintMembers(System.Text.StringBuilder builder)
```

The name `PrintMembers` is also a [public part of Roslyn](https://github.com/dotnet/roslyn/blob/d71b443067e30dbd935a8cffdacda2ca7ba32e7f/src/Compilers/Core/Portable/Symbols/WellKnownMemberNames.cs#L387-L390).

```cs
var isRecordStruct = t.BaseType == typeof(ValueType) &&
                     t.GetMethod("PrintMembers", BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.DeclaredOnly, null, new[] { typeof(StringBuilder) }, null)?
                        .GetCustomAttribute(typeof(CompilerGeneratedAttribute)) is { };
```

What do you think about this?

salvois:
Hi @dennisdoomen , @jnyrup , I hope the new year finds you well.
I was under the impression we were in a kind of Mexican standoff, so I did not insist on it :)
Personally, I agree with @dennisdoomen on the benefit-cost ratio of a best guess, but I'd be glad to study the last @jnyrup 's proposal.

Src/FluentAssertions/Common/TypeExtensions.cs
salvois: That  may be a good catch. I'll look into it!

dennisdoomen:
@salvois any progress?

salvois:
Hi @dennisdoomen , I should be able to find some spare time today. Thanks

Src/FluentAssertions/Common/TypeExtensions.cs
salvois: @jnyrup that looked promising, but unfortunately you can "crack" the check for a compiler generated PrintMembers, because the compiler does allow you to define your PrintMembers methods with the very same signature in the body of the record struct. I know that this is very unlikely, but nonetheless it is possible.
On the other hand, the compiler does not allow to define your operator==, and to my best knowledge record structs are (currently!) the only value types having a compiler generated equality operator, so I'd go for it. I have used the simplified check you have proposed.

jnyrup: Not sure I understand you here 🤔 
As a record struct currently doesn't have any unspeakable parts, it's _always_ possible to write a struct in plain C# that looks like a record struct.

E.g. this is valid C# code, [SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABADAAgwRgDoAlAVwDsAXASwFsZCBhCWgB2oBsYoBlbgN2pgYAZwDcAWABQ0jAGYcIylFJhKOAILSA3tJz6cAbWZtO3AOIxy3AIaUYAEwC6eg/Lz4kOYBAgccEKy2lNA4ALxhABQaOAgANJo4AJ4AlOEAfDgAZjYcIjCSUgY4rvrGLOxcUJbWUHaOLkVuCgRePn4BQXUhUDgAhFEx8YmpGdm5+WIlUgC+QA===)
```cs
public struct A
{
    [CompilerGenerated]
    public static bool operator ==(A x, A y) => true;
    
    [CompilerGenerated]
    public static bool operator !=(A x, A y) => false 
}
```

I tried another thing.

I made a `record struct Auto {}` and copied its output back into regular C# code.
The _only_ thing I wasn't allowed to was using `[IsReadOnly]` as that gave a compiler error.

```
error CS8335: Do not use 'System.Runtime.CompilerServices.IsReadOnlyAttribute'. This is reserved for compiler usage.
```

`PrintMembers` is annotated with `[IsReadOnly]`

So what do you think about checking for this?

```cs
var isRecordStruct = t.BaseType == typeof(ValueType) &&
                     t.GetMethod("PrintMembers", BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.DeclaredOnly, null, new[] { typeof(StringBuilder) }, null)?
                        .GetCustomAttribute(typeof(IsReadOnlyAttribute)) is { };
```

I don't know if an arbitrary source generator is allowed to use `[IsReadOnly]`.
IL weaving would most likely be able to circumvent CS8335, but we aren't looking for a bullet proof solution.

[SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABATARgLABQGADAAQY4B0ASgK4B2ALgJYC2MlAwhKwA7MAbGFADKwgG7MwMAM4BuQiXJU6TNhwCSTYRF5iok6fMVkKlACowEjBQUKxIUACakZjKLTCNSAQVqMILABvAF9CQgBtEXdPRgAZAEMATwh/AApElP8AaWZ6J0oxAEdaGDUEgQAaUhFmAC8YUgBeUhwASgBdQjzGYXoK1xivX38IUhBSDQBREoTGBOAhAB4/AIA+QiDCUh3SAHo90giNGWoYBKcAeXoBJK6CXaPuPkFhAHEy4TmYJ3vHjAAzKQIOJhFBmE5GhQyOYINFwfQAOZpNrbXZbB6PXbwvKIgBCtEEkKgymIBKJwmapHoMAA7jV3LjyQJiSjbFjdtDmcTKD5eLwyk40gAiVYQYVtdkc0nc4S8/mCkWkIKkCVSjnMABmpDSAAUEYwALIwVjAYQyNJcwks4RtVGYjkY6VYq0UqDygX5NIAclI3slaI5YQdLpwZOtPL5nqF3pC/vVLoA7DKI3LYTikWzA6Rg9mDsdTucrjc7tmIs9+EIoB8aVBvr9s7xweJvqRgBAIAJSPqesbTea0hn8amScAR/bHk7pRhk5qKjIYAncyHyzxK+9PnXeg2Q4DlEg2x2u7ovgESQBCJppMWkISaxjVG/gxEAC0YE/R2f+yfPaTv3iaFpnzfAMQ2XR5VxeKsa1PH4/k5IEKAPdtO2BAUt2gZorxvf9H1GUhgPfbMpw5Gdbxge9KBmWh5zSQjQMecDdnzE4zgua5bngnZIPXatN3rLjyCBEEwQhRoelID5GAACQSGQX24SEUWIr9OWTYgl3CEMWMLdiS0EnjXj42sBOzPcRKgcFIUPVDqNoiBgAAKxgYYHMcj8dhIrEtR1NzSGYGQRgCDysS850yLsgQLWvUY2jchjpWDcLZ3nRdsyYnYdLY4tOLLCsjJgrc4LMoEUK7SKLRvCBGBfW0VJDb9SBiNKwMIEIgA==)

salvois: Interesting! I didn't know one could annotate with [CompilerGenerated] in source code; it looks a little silly to me that you can, but it is what it is I suppose :)
In this case, I definitely like your proposal with PrintMembers and IsReadOnly more. Let me try to give a couple of final touches and I'll be back with a new commit.

salvois: @jnyrup I have just tried the last proposal checking for a PrintMembers annotated with IsReadOnly. IsReadOnly is not available on NET47 and NETSTANDARD2_0, so it does not compile on those target, but here the problem is just my ignorance on properly handling multi-target.
The major problem for me is that the check fails for the MyReadOnlyRecordStruct and MyRecordStructWithCustomPrintMembers tests, for which PrintMembers has not the IsReadOnly attribute.
For this reason, at the end of the day (literally, for me 😄 ) I believe the lesser evil for now is checking for compiler generated equality, knowing that we can have a false positive if someone uses the [CompilerGenerated] attribute in source code (I have added a test to document this).

jnyrup: Aha!
I hadn't thought about the case in `MyRecordStructWithCustomPrintMembers` where one implements `PrintMembers` themselves such that the compiler _doesn't_ create one.
Well found 👍 

So it seems there will always be a `PrintMembers`, but we don't know who wrote it.

This code handles all your test cases and the false positive.

```cs
public static bool IsRecord(this Type type)
{
    return TypeIsRecordCache.GetOrAdd(type, static t => t.IsRecordClass() || t.IsRecordStruct());
}

private static bool IsRecordClass(this Type type)
{
    return type.GetMethod("<Clone>$", BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly) is { } &&
            type.GetProperty("EqualityContract", BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.DeclaredOnly)?
                .GetMethod?.IsDecoratedWith<CompilerGeneratedAttribute>() == true;
}

private static bool IsRecordStruct(this Type type)
{
    // As noted here: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/record-structs#open-questions
    // recognizing record structs from metadata is an open point. The following check is based on common sense
    // and heuristic testing, apparently giving good results but not supported by official documentation.
    return type.BaseType == typeof(ValueType) &&
            type.GetMethod("PrintMembers", BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.DeclaredOnly, null, new[] { typeof(StringBuilder) }, null) is { } &&
            type.GetMethod("op_Equality", BindingFlags.Static | BindingFlags.Public | BindingFlags.DeclaredOnly, null, new[] { type, type }, null)?
                .IsDecoratedWith<CompilerGeneratedAttribute>() == true;
}
```

I would be confident enough with this one.

salvois: Good idea: this way we rely on the two documented behaviors of record structs!
There is still a small possibility of a false positive, where one both implements a PrintMember and annotates equality with [CompilerGenerated] (I have added a test for reference), but I'd consider such code as deliberate sabotage 😄

jnyrup (approved):

dennisdoomen:
Hah, the spell check target fails on `struct` ;-)

docs/_pages/objectgraphs.md
```diff
@@ -39,7 +39,7 @@ orderDto.Should().BeEquivalentTo(order, options =>
 
 ### Value Types
 
-To determine whether Fluent Assertions should recurs into an object's properties or fields, it needs to understand what types have value semantics and what types should be treated as reference types. The default behavior is to treat every type that overrides `Object.Equals` as an object that was designed to have value semantics. Anonymous types, records and tuples also override this method, but because the community proved us that they use them quite often in equivalency comparisons, we decided to always compare them by their members.
+To determine whether Fluent Assertions should recurs into an object's properties or fields, it needs to understand what types have value semantics and what types should be treated as reference types. The default behavior is to treat every type that overrides `Object.Equals` as an object that was designed to have value semantics. Anonymous types, records, record structs and tuples also override this method, but because the community proved us that they use them quite often in equivalency comparisons, we decided to always compare them by their members.
```
IT-VBFK, line 42:
```suggestion
 determine whether Fluent Assertions should recurs into an object's properties or fields, it needs to understand what types have value semantics and what types should be treated as reference types. The default behavior is to treat every type that overrides `Object.Equals` as an object that was designed to have value semantics. Anonymous types, `record`s, `record struct`s and tuples also override this method, but because the community proved us that they use them quite often in equivalency comparisons, we decided to always compare them by their members.
```

IT-VBFK: This should work

Otherwise we have to add this to the ignore word list

ITaluone: @jnyrup missed that one? :)

IT-VBFK:
Try to finish #2093 today.. stay tuned 🙃

salvois:
Thanks @dennisdoomen , @IT-VBFK and @jnyrup , I really enjoyed the constructive exchange and I also got the opportunity to learn something new!

IT-VBFK:
You should never stop to improve yourself :)

PR #2010, opened 2022-10-11
Exposes the IMonitor object to extensions methods.

### LINKED ISSUE #2008: Expose IMonitor<T> in EventAssertions<T>
If I'm extending the fluent asertions library with some extension methods, it would be nice to be able to iterate through the underlying `IMonitor<>` of the `EventAssertions<>` class. 

**A Pseudo example:** 
```csharp
public class Fixture
{
    public void Test()
    {
        //Arrange 
        var parameter = "Hello";
        var sut = new MyService();
        using var monitor = sut.Monitor();
        
        //Act
        sut.DoTheThing();
        
        //Assert
        monitor
        .Should()
        .NotRaiseMyEventWithParameter(parameter);
    }
}


public static class MyExtension
{
    public static IEventRecording NotRaiseMyEventWithParameter<T>(this EventAssertions<T> assertion, string parameter)
    {
        // Monitor not exposed  in the EventAssertions!!!!!
        var recordings = assertion.**Monitor**.GetRecordingFor(nameof(IMyInterface.MyEvent));
        
        //Do the assertion here
    }
}
```
It is possible to make the extensions method directly on the monitor object, but then the `Should()` method is not invoked, and semantically it does not fit in with the fluent interface structure.

```csharp
monitor.ShouldNotRaiseMyEventWithParameter(parameter).
```

A-Kjeldgaard:
Changed the private field to a public property, thereby exposing the IMonitor object to extensions methods.

Closes #2008

- Added Unit test.
- Ran and updated accept api changes.
- Added release notes.

### DIFF
--- a/Tests/FluentAssertions.Specs/Events/EventAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Events/EventAssertionSpecs.cs
@@ -629,6 +629,20 @@ public class EventAssertionSpecs
             act.Should().Throw<ArgumentException>()
                 .WithParameterName("expression");
         }
+
+        [Fact]
+        public void Event_Assertions_should_expose_monitor()
+        {
+            // Arrange
+            var subject = new EventRaisingClass();
+            using var monitor = subject.Monitor();
+
+            // Act
+            var exposedMonitor = monitor.Should().Monitor;
+
+            // Assert
+            ((object)exposedMonitor).Should().BeSameAs(monitor);
+        }
     }
 
     public class Metadata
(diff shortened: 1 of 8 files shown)


Tests/FluentAssertions.Specs/Events/EventAssertionSpecs.cs
```diff
@@ -629,6 +629,20 @@ public void When_monitoring_an_object_with_invalid_property_expression_it_should
             act.Should().Throw<ArgumentException>()
                 .WithParameterName("expression");
         }
+
+        [Fact]
+        public void Event_Assertions_should_expose_monitor()
```
dennisdoomen, line 634: 🔧 Please keep names functional, so `Event_assertions_should_expose_the_monitor`

dennisdoomen (changes requested):
Thanks. Only minor comments.

jnyrup (approved):
I've got nothing extra to add. 
Always happy to improve the extensibility.

A-Kjeldgaard:
Requested changes hereby fixed 👍

dennisdoomen (approved):

PR #2013, opened 2022-10-14
Calculate the difference between the subject and the expected nearby time

### LINKED ISSUE #2012: Update error message at `DateTimeAssertions.BeCloseTo()` to include the calculated time range
**Before you file a bug, have you:**

* Tried upgrading to newest version of Fluent Assertions, to see if your issue has already been resolved and released?
* Checked existing open *and* closed [issues](https://github.com/fluentassertions/fluentassertions/issues?utf8=%E2%9C%93&q=is%3Aissue), to see if the issue has already been reported?
* Tried reproducing your problem in a new isolated project?
* Read the [documentation](https://fluentassertions.com/introduction)?
* Considered if this is a general question and not a bug?. For general questions please use [StackOverflow](https://stackoverflow.com/questions/tagged/fluent-assertions?mixed=1).

### Description

Now the error message looks like this:
`Expected logs.First().ChangedAt to be within 15s from <2022-10-13 06:44:06.8277343>, but found <2022-10-13 06:43:51>.`

But with this line I have to calculate the difference between the two `DateTime`s in case I want to know the exact difference.

This is not an issue inside FA, but a proposal to state more precisely what the time differnece was.

### Expected behavior:

Something like:
`Expected logs.First().ChangedAt to be within 15s from <2022-10-13 06:44:06.8277343>, but found <2022-10-13 06:43:51>, which is XXXs.`

### Actual behavior:

The calculated difference is not displayed.

### Versions

* Which version of Fluent Assertions are you using? 6.7.0
* Which .NET runtime and version are you targeting? .net6.0 and .net7.0

### Additional Information

-

IT-VBFK:
Closes #2012 

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Primitives/DateTimeAssertions.cs
+++ b/Src/FluentAssertions/Primitives/DateTimeAssertions.cs
@@ -168,12 +168,18 @@ public class DateTimeAssertions<TAssertions>
         long distanceToMaxInTicks = (DateTime.MaxValue - nearbyTime).Ticks;
         DateTime maximumValue = nearbyTime.AddTicks(Math.Min(precision.Ticks, distanceToMaxInTicks));
 
+        TimeSpan? difference = (Subject - nearbyTime)?.Duration();
+
         Execute.Assertion
-            .ForCondition((Subject >= minimumValue) && (Subject.Value <= maximumValue))
             .BecauseOf(because, becauseArgs)
-            .FailWith("Expected {context:the date and time} to be within {0} from {1}{reason}, but found {2}.",
-                precision,
-                nearbyTime, Subject);
+            .WithExpectation("Expected {context:the date and time} to be within {0} from {1}{reason}", precision, nearbyTime)
+            .ForCondition(Subject is not null)
+            .FailWith(", but found <null>.")
+            .Then
+            .ForCondition((Subject >= minimumValue) && (Subject <= maximumValue))
+            .FailWith(", but {0} was off by {1}.", Subject, difference)
+            .Then
+            .ClearExpectation();
 
         return new AndConstraint<TAssertions>((TAssertions)this);
     }
--- a/Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
+++ b/Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
@@ -304,12 +304,18 @@ public class DateTimeOffsetAssertions<TAssertions>
         long distanceToMaxInTicks = (DateTimeOffset.MaxValue - nearbyTime).Ticks;
         DateTimeOffset maximumValue = nearbyTime.AddTicks(Math.Min(precision.Ticks, distanceToMaxInTicks));
 
+        TimeSpan? difference = (Subject - nearbyTime)?.Duration();
+
         Execute.Assertion
-            .ForCondition(Subject >= minimumValue && (Subject <= maximumValue))
             .BecauseOf(because, becauseArgs)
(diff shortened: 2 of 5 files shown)


Src/FluentAssertions/Primitives/DateTimeAssertions.cs
```diff
@@ -168,12 +168,18 @@ public AndConstraint<TAssertions> Be(DateTime? expected, string because = "", pa
         long distanceToMaxInTicks = (DateTime.MaxValue - nearbyTime).Ticks;
         DateTime maximumValue = nearbyTime.AddTicks(Math.Min(precision.Ticks, distanceToMaxInTicks));
 
+        var difference = (Subject - nearbyTime)?.Duration();
```
IT-VBFK, line 171: `...Duration()`: 
I was not sure what fits better: the "real" difference (meaning this also could be negative) or the absolute values..

dennisdoomen: In this case, the value can be off in both directions, so an absolute value is enough.

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
```diff
@@ -304,12 +304,18 @@ public DateTimeOffsetAssertions(DateTimeOffset? value)
         long distanceToMaxInTicks = (DateTimeOffset.MaxValue - nearbyTime).Ticks;
         DateTimeOffset maximumValue = nearbyTime.AddTicks(Math.Min(precision.Ticks, distanceToMaxInTicks));
 
+        var differnece = (Subject - nearbyTime)?.Duration();
```
dennisdoomen, line 307: 🔧 Please use the explicit type here.

dennisdoomen (changes requested):
Can you also update the `releases.md` file?

Src/FluentAssertions/Primitives/DateTimeAssertions.cs
IT-VBFK: 👍🏻

IT-VBFK:
> Can you also update the `releases.md` file?

Ok
Wasn't sure wheter this change is enough to be mentioned in the official release notes :)

Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs
jnyrup: differnece -> difference.

dennisdoomen (approved):

Src/FluentAssertions/Primitives/DateTimeAssertions.cs
```diff
@@ -168,12 +168,18 @@ public AndConstraint<TAssertions> Be(DateTime? expected, string because = "", pa
...
             .BecauseOf(because, becauseArgs)
-            .FailWith("Expected {context:the date and time} to be within {0} from {1}{reason}, but found {2}.",
-                precision,
-                nearbyTime, Subject);
+            .WithExpectation("Expected {context:the date and time} to be within {0} from {1}{reason}", precision, nearbyTime)
+            .ForCondition(Subject is not null)
+            .FailWith(", but found <null>.")
+            .Then
+            .ForCondition((Subject >= minimumValue) && (Subject <= maximumValue))
+            .FailWith(", but {0} is off by {1}.", Subject, difference)
```
jnyrup, line 180: Should this be in past tense, i.e. "was off by"? 🤔

IT-VBFK: Hmm.. 

fair point :)

jnyrup (approved):

PR #2021, opened 2022-10-22
Enable CI unit tests on linux and mac

jnyrup:
An extra layer of testing to ensure we also work on other operating systems. 

Until now I've once in a while run our tests on my local Linux machine or in WSL.


dennisdoomen (approved):

dennisdoomen:
Nice!

PR #2023, opened 2022-10-27
Add missing exceptions to xml summaries

jnyrup:
@julealgon [pointed out](https://github.com/fluentassertions/fluentassertions/pull/1929#pullrequestreview-1153185757) that we were missing some `<exception>` in xml summaries.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -176,6 +176,7 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> :
     /// <param name="becauseArgs">
     /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
     /// </param>
+    /// <exception cref="ArgumentNullException"><paramref name="config"/> is <see langword="null"/>.</exception>
     public AndConstraint<TAssertions> AllBeEquivalentTo<TExpectation>(TExpectation expectation,
         Func<EquivalencyAssertionOptions<TExpectation>, EquivalencyAssertionOptions<TExpectation>> config,
         string because = "",
@@ -347,6 +348,7 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> :
     /// <param name="becauseArgs">
     /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
     /// </param>
+    /// <exception cref="ArgumentNullException"><paramref name="config"/> is <see langword="null"/>.</exception>
     public AndConstraint<TAssertions> BeEquivalentTo<TExpectation>(IEnumerable<TExpectation> expectation,
         Func<EquivalencyAssertionOptions<TExpectation>, EquivalencyAssertionOptions<TExpectation>> config, string because = "",
         params object[] becauseArgs)
@@ -1185,6 +1187,7 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> :
     /// <param name="becauseArgs">
     /// Zero or more objects to format using the placeholders in <paramref name="because" />.
     /// </param>
+    /// <exception cref="ArgumentNullException"><paramref name="expectation"/> is <see langword="null"/>.</exception>
     public AndConstraint<TAssertions> EndWith<TExpectation>(
         IEnumerable<TExpectation> expectation, Func<T, TExpectation, bool> equalityComparison, string because = "", params object[] becauseArgs)
     {
@@ -1690,6 +1693,7 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> :
     /// <param name="becauseArgs">
     /// Zero or more objects to format using the placeholders in <paramref name="because" />.
     /// </param>
+    /// <exception cref="ArgumentNullException"><paramref name="unexpected"/> is <see langword="null"/>.</exception>
     public AndConstraint<TAssertions> NotBeEquivalentTo<TExpectation>(IEnumerable<TExpectation> unexpected, string because = "",
         params object[] becauseArgs)
     {
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -756,10 +756,7 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> :
         Guard.ThrowIfArgumentIsNull(expected, nameof(expected), "Cannot verify containment against a <null> collection");
 
         ICollection<T> expectedObjects = expected.ConvertOrCastToCollection();
-        if (!expectedObjects.Any())
-        {
-            throw new ArgumentException("Cannot verify containment against an empty collection", nameof(expected));
-        }
+        Guard.ThrowIfArgumentIsEmpty(expectedObjects, nameof(expected), "Cannot verify containment against an empty collection");
 
         bool success = Execute.Assertion
             .BecauseOf(because, becauseArgs)
@@ -2153,10 +2150,7 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> :
         Guard.ThrowIfArgumentIsNull(unexpected, nameof(unexpected), "Cannot verify non-containment against a <null> collection");
 
         ICollection<T> unexpectedObjects = unexpected.ConvertOrCastToCollection();
-        if (!unexpectedObjects.Any())
-        {
-            throw new ArgumentException("Cannot verify non-containment against an empty collection", nameof(unexpected));
-        }
+        Guard.ThrowIfArgumentIsEmpty(unexpectedObjects, nameof(unexpected), "Cannot verify non-containment against an empty collection");
 
         bool success = Execute.Assertion
             .BecauseOf(because, becauseArgs)
@@ -2920,10 +2914,7 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> :
         Guard.ThrowIfArgumentIsNull(expected, nameof(expected), "Cannot verify against a <null> collection of inspectors");
 
         ICollection<Action<T>> elementInspectors = expected.ConvertOrCastToCollection();
-        if (!elementInspectors.Any())
-        {
-            throw new ArgumentException("Cannot verify against an empty collection of inspectors", nameof(expected));
-        }
+        Guard.ThrowIfArgumentIsEmpty(elementInspectors, nameof(expected), "Cannot verify against an empty collection of inspectors");
 
         bool success = Execute.Assertion
             .BecauseOf(because, becauseArgs)
@@ -3009,10 +3000,7 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> :
         Guard.ThrowIfArgumentIsNull(predicates, nameof(predicates), "Cannot verify against a <null> collection of predicates");
 
         IList<Expression<Func<T, bool>>> predicatesList = predicates.ConvertOrCastToList();
-        if (predicatesList.Count == 0)
-        {
-            throw new ArgumentException("Cannot verify against an empty collection of predicates", nameof(predicates));
-        }
+        Guard.ThrowIfArgumentIsEmpty(predicatesList, nameof(predicates), "Cannot verify against an empty collection of predicates");
 
         bool success = Execute.Assertion
             .BecauseOf(because, becauseArgs)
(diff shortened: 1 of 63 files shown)


Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -417,6 +419,7 @@ public AndConstraint<TAssertions> BeEmpty(string because = "", params object[] b
     public AndConstraint<SubsequentOrderingAssertions<T>> BeInAscendingOrder(
         IComparer<T> comparer, string because = "", params object[] becauseArgs)
     {
+        Guard.ThrowIfArgumentIsNull(comparer, nameof(comparer), "Cannot assert collection ordering without specifying a comparer.");
```
julealgon, line 422: The custom message doesn't seem to be adding much value here from my perspective (as a consumer). I'd suggest removing in cases like this to reduce noise in the codebase.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -750,10 +756,7 @@ public AndConstraint<TAssertions> Contain(IEnumerable<T> expected, string becaus
         Guard.ThrowIfArgumentIsNull(expected, nameof(expected), "Cannot verify containment against a <null> collection");
 
         ICollection<T> expectedObjects = expected.ConvertOrCastToCollection();
-        if (!expectedObjects.Any())
-        {
-            throw new ArgumentException("Cannot verify containment against an empty collection", nameof(expected));
-        }
+        Guard.ThrowIfArgumentIsEmpty(expectedObjects, nameof(expected), "Cannot verify containment against an empty collection");
```
julealgon, line 759: I know this is unrelated to the PR, but I just wanted to mention that this validation/exception is very unexpected to see: if feels ilogical.

This would be equivalent to [`string.Contains(string)` throwing if you pass an empty string as an argument. Instead, as expected, it just returns `true`](https://learn.microsoft.com/en-us/dotnet/api/system.string.contains?view=net-7.0#system-string-contains(system-string)), which IMHO is what this method here should do as well: "an empty collection is always contained in any other non-null collection" using the same logic.

Src/FluentAssertions/Common/Guard.cs
```diff
@@ -63,6 +63,22 @@ public static void ThrowIfArgumentIsEmpty<T>(IEnumerable<T> values, string param
         }
     }
 
+    public static void ThrowIfArgumentIsEmpty(string str, string paramName, string message)
+    {
+        if (str.Length == 0)
```
julealgon, line 68: I'd suggest using constant pattern matching for constant comparisons in new code:
```suggestion
        if (str.Length is 0)
```

Src/FluentAssertions/Common/Guard.cs
```diff
@@ -63,6 +63,22 @@ public static void ThrowIfArgumentIsEmpty<T>(IEnumerable<T> values, string param
         }
     }
 
+    public static void ThrowIfArgumentIsEmpty(string str, string paramName, string message)
```
julealgon, line 66: Since these are public facing, they should be validating their arguments (validation inception lol).

In all seriousness, you don't want to see null reference exceptions being thrown by this guard method when `str` is `null` for example. That would become incredibly confusing. Same for other parameters.

Src/FluentAssertions/Common/Guard.cs
```diff
@@ -63,6 +63,22 @@ public static void ThrowIfArgumentIsEmpty<T>(IEnumerable<T> values, string param
...
+    {
+        if (str.Length == 0)
+        {
+            throw new ArgumentException(message, paramName);
+        }
+    }
+
+    public static void ThrowIfArgumentIsNegative(TimeSpan timeSpan, string paramName, string message)
+    {
+        if (timeSpan < TimeSpan.Zero)
```
julealgon, line 76: You might want to watch out for the special case of the "infinite timespan" here. IIRC it is set to something like "-1 ticks" or whatever (can use [`Timeout.InfiniteTimespan`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.timeout.infinitetimespan?view=net-7.0) to get it).

That's usually considered a valid negative value even when you only want to deal with positive values.

Src/FluentAssertions/Equivalency/ConversionSelector.cs
```diff
@@ -46,6 +46,11 @@ public void IncludeAll()
         inclusions.Add(new ConversionSelectorRule(_ => true, "Try conversion of all members. "));
     }
 
+    /// <summary>
+    /// Instructs the equivalency comparison to try to convert the value of
+    /// a specific member on the expectation object before running any of the other steps.
+    /// </summary>
+    /// <exception cref="ArgumentNullException"><paramref name="predicate"/> is <c>null</c>.</exception>
```
julealgon, line 53: Is the lack of a `<param` doc intentional here (just double checking to be sure since I see parameter docs elsewhere)?

Src/FluentAssertions/Equivalency/Matching/MappedPathMatchingRule.cs
```diff
@@ -12,6 +12,11 @@ internal class MappedPathMatchingRule : IMemberMatchingRule
     private readonly MemberPath expectationPath;
     private readonly MemberPath subjectPath;
 
+    /// <summary>
+    /// Initializes a new instance of the <see cref="MappedPathMatchingRule"/> class.
+    /// </summary>
+    /// <exception cref="ArgumentNullException"><paramref name="expectationMemberPath"/> is <c>null</c> or empty.</exception>
```
julealgon, line 18: Using `ArgumentNullException` for "empty" is considered a bad practice. `ArgumentException` should be used instead.

Src/FluentAssertions/ObjectAssertionsExtensions.cs
```diff
@@ -112,6 +113,7 @@ public static class ObjectAssertionsExtensions
     /// <param name="becauseArgs">
     /// Zero or more objects to format using the placeholders in <paramref name="because" />.
     /// </param>
+    /// <exception cref="ArgumentNullException"><paramref name="options"/> is <c>null</c>.</exception>
```
julealgon, line 116: `assertions` should also be guard-checked ideally. (goes for any other extension method target like this)

Src/FluentAssertions/Primitives/DateTimeAssertions.cs
```diff
@@ -157,10 +158,7 @@ public AndConstraint<TAssertions> Be(DateTime? expected, string because = "", pa
     public AndConstraint<TAssertions> BeCloseTo(DateTime nearbyTime, TimeSpan precision, string because = "",
         params object[] becauseArgs)
     {
-        if (precision < TimeSpan.Zero)
-        {
-            throw new ArgumentOutOfRangeException(nameof(precision), $"The value of {nameof(precision)} must be non-negative.");
-        }
+        Guard.ThrowIfArgumentIsNegative(precision, nameof(precision), "The value must be non-negative.");
```
julealgon, line 161: Have you considered making this `"The value must be non-negative."` text be the default for `ThrowIfArgumentIsNegative`?

julealgon (commented):
Just a few (sometimes unrelated to this PR) points that I've noticed. Not an extensive review.

dennisdoomen:
I'm wondering why we bother documenting those exceptions. It's not like we want people to be able to catch them. We usually only throw them when somebody is doing something unexpected or passing an invalid argument.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
jnyrup: I can't find the discussions right now, but we decided to not always supports [vacuous truths](https://en.wikipedia.org/wiki/Vacuous_truth).

Src/FluentAssertions/Equivalency/ConversionSelector.cs
jnyrup: Not intentionally left out, I simply copied the summary from `SelfReferenceEquivalencyAssertionOptions.WithAutoConversionFor`.

Src/FluentAssertions/Common/Guard.cs
IT-VBFK: I think this is the reason:
>  using constant pattern matching

Src/FluentAssertions/Common/Guard.cs
julealgon: Fair enough.

julealgon:
> I'm wondering why we bother documenting those exceptions. It's not like we want people to be able to catch them. 

Being able to catch exceptions, or encouraging people to catch exceptions were never core reasons for documenting exceptions, so I'd say this is a fairly weak argument for not doing it.

Most exceptions in core .NET libraries are of this type: "design exceptions", or exceptions that "should never ever happen unless you have a bug". Yet, they are all documented properly.

> We usually only throw them when somebody is doing something unexpected or passing an invalid argument.

Which is why it is good to document them in the first place. You don't want people to need to look into your codebase to know whether or not an argument is required for example. Sure, in most cases it might be obvious, but that's not always the case.

I can only speak for myself, but these exception docs are very useful to me as a consumer to reduce the need to move away from the IDE to look into documentation or code. And this goes for any library, of course.

I'd also argue that a library with properly documented exceptions appears more professional and polished, at least in my view.

dennisdoomen:
> Most exceptions in core .NET libraries are of this type: "design exceptions", or exceptions that "should never ever happen unless you have a bug". Yet, they are all documented properly.

Fluent Assertions is not a general purpose core .NET library and has a different audience, so that argument doesn't make sense to me. If it would be a library used in production code, I would agree with you.

> Which is why it is good to document them in the first place. You don't want people to need to look into your codebase to know whether or not an argument is required for example. Sure, in most cases it might be obvious, but that's not always the case.

They don't need to. The API should be self-explanatory. Again, they should _never_ need to catch any exception in FA, so I don't see any value in documenting them everywhere (and keep maintaining them).

But maybe @jnyrup has a different opinion.

jnyrup:
I've been thinking quite some about this topic.
https://github.com/dotnet/dotnet-api-docs/wiki/Exceptions describes reasons to (try to) document exceptions in general purpose libraries.

As Dennis mentions this is not a general-purpose library, so if you have deterministic tests, you can just run the test to see the outcome.
For the majority of cases I don't think they add any _real_ value, at least I would write the test and run them to get fast feedback.
I acknowledge that my style of working might be different from others.

I've found two areas where I can imagine explicit documentation _might_ help.
* If you're writing an extension library on top of FA and hence don't know what data is passed on to FA.
* If you're running service tests or some kind of tests where the input to FA might not be deterministic.


If the documentation provides real value to consumers I don't mind maintaining them, but I also don't think it devalues the library significantly to remove them.

julealgon:
@dennisdoomen 
> > Most exceptions in core .NET libraries are of this type: "design exceptions", or exceptions that "should never ever happen unless you have a bug". Yet, they are all documented properly.
> 
> Fluent Assertions is not a general purpose core .NET library and has a different audience, so that argument doesn't make sense to me.

We probably have different definitions of what constitutes a .NET library. I see FA as I see any other library. The fact that it has a specific use case (unit testing), doesn't change any of that IMHO. Every library has a specific use case.

> If it would be a library used in production code, I would agree with you.

I consider my tests production code. I treat my test code with the exact same care and follow the same best practices as I do in my application code. A bug in my test code is treated as a bug in the application code as that breaks the build.

I understand some people treat test code as if it was a "second-class citizen", but I assure you that's not the case for everybody.

> They don't need to. The API should be self-explanatory. Again, they should _never_ need to catch any exception in FA, so I don't see any value in documenting them everywhere (and keep maintaining them).

You keep using this "nobody needs to catch any FA exception" argument as justification for not having them. This is a really bad justification IMHO. "Design exceptions" (most of the `ArgumentException` inherited ones) are _never_ supposed to be caught no matter the source. That's not their purpose. They exist to "fail fast" under invalid situations as well as provide information for the consumer to fix their usages. They always represent bugs, which is why they should never be handled. And this very much applies to _any_ library, including test-related ones.

I get trying to have a "self-explanatory API", but that's actually not 100% possible to do at compile time. There are (many) situations where you wouldn't be able to tell whether or not some argument is required, can be null, etc, without actual documentation, and having the exception docs in place is by far the most standard way of documenting that kind of thing.

---

@jnyrup 
> If the documentation provides real value to consumers I don't mind maintaining them, but I also don't think it devalues the library significantly to remove them.

I can obviously only speak for myself here, but to me these exceptions, when properly documented, provide very valuable information. They are as important as other parts of the method signatures.

Not having them in place at all (again, from my perspective) makes your library look fairly unprofessional, as well as forcing developers to navigate to source to see how some parameters work in some cases, which is never a good experience.

dennisdoomen (approved):

PR #2028, opened 2022-11-01
Added `BeOneOf` for `IComparable`s and object comparisons.

### LINKED ISSUE #2027: ComparableTypeAssertions should have BeOneOf method
Because `ComparableTypeAssertions` doesn't contain `BeOneOf`, the following code doesn't work:

```
float testFloat = 12f;
var testTuple = (1, 2);

testFloat.Should().BeOneOf(12f, 13f); // Works
testTuple.Should().BeOneOf((1, 2), (1, 3)); // Doesn't work
```

I don't see any technical reason why the latter shouldn't work, so how about adding `BeOneOf` to `ComparableTypeAssertions`?

jez9999:
Fixes #2027.

### DIFF
--- a/Src/FluentAssertions/Numeric/ComparableTypeAssertions.cs
+++ b/Src/FluentAssertions/Numeric/ComparableTypeAssertions.cs
@@ -1,6 +1,8 @@
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.Linq;
 using FluentAssertions.Common;
 using FluentAssertions.Equivalency;
 using FluentAssertions.Execution;
@@ -354,6 +356,41 @@ public class ComparableTypeAssertions<T, TAssertions> : ReferenceTypeAssertions<
             .BecauseOf(because, becauseArgs)
             .FailWith("Expected {context:object} to not be between {0} and {1}{reason}, but found {2}.",
                 minimumValue, maximumValue, Subject);
+
+        return new AndConstraint<TAssertions>((TAssertions)this);
+    }
+
+    /// <summary>
+    /// Asserts that a value is one of the specified <paramref name="validValues"/>.
+    /// </summary>
+    /// <param name="validValues">
+    /// The values that are valid.
+    /// </param>
+    public AndConstraint<TAssertions> BeOneOf(params T[] validValues)
+    {
+        return BeOneOf(validValues, string.Empty);
+    }
+
+    /// <summary>
+    /// Asserts that a value is one of the specified <paramref name="validValues"/>.
+    /// </summary>
+    /// <param name="validValues">
+    /// The values that are valid.
+    /// </param>
+    /// <param name="because">
+    /// A formatted phrase as is supported by <see cref="string.Format(string,object[])"/> explaining why the assertion
+    /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+    /// </param>
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+    /// </param>
+    public AndConstraint<TAssertions> BeOneOf(IEnumerable<T> validValues, string because = "",
+        params object[] becauseArgs)
+    {
+        Execute.Assertion
+            .ForCondition(validValues.Any(val => Subject.CompareTo(val) == Equal))
+            .BecauseOf(because, becauseArgs)
+            .FailWith("Expected {context:object} to be one of {0}{reason}, but found {1}.", validValues, Subject);
 
         return new AndConstraint<TAssertions>((TAssertions)this);
     }
--- a/Src/FluentAssertions/Numeric/ComparableTypeAssertions.cs
+++ b/Src/FluentAssertions/Numeric/ComparableTypeAssertions.cs
@@ -388,7 +388,7 @@ public class ComparableTypeAssertions<T, TAssertions> : ReferenceTypeAssertions<
         params object[] becauseArgs)
     {
         Execute.Assertion
-            .ForCondition(validValues.Any(val => Subject.CompareTo(val) == Equal))
+            .ForCondition(validValues.Any(val => Equals(Subject, val)))
             .BecauseOf(because, becauseArgs)
             .FailWith("Expected {context:object} to be one of {0}{reason}, but found {1}.", validValues, Subject);
 
--- a/Tests/FluentAssertions.Specs/Numeric/ComparableSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Numeric/ComparableSpecs.cs
@@ -103,6 +103,52 @@ public class ComparableSpecs
         }
     }
 
+    public class BeOneOf
+    {
+        [Fact]
+        public void When_a_value_is_not_one_of_the_specified_values_it_should_throw()
+        {
+            // Arrange
+            var value = new ComparableOfInt(3);
+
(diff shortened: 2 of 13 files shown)


jnyrup:
Thanks for contributing this - I love how this aligns the API better across assertion types.

Would you be willing to also add examples to the documentation
https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/basicassertions.md
https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/numerictypes.md

and a line to the release notes?
https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/releases.md

Src/FluentAssertions/Numeric/ComparableTypeAssertions.cs
```diff
@@ -358,6 +360,41 @@ public AndConstraint<TAssertions> BeGreaterThanOrEqualTo(T expected, string beca
...
+    /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+    /// </param>
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+    /// </param>
+    public AndConstraint<TAssertions> BeOneOf(IEnumerable<T> validValues, string because = "",
+        params object[] becauseArgs)
+    {
+        Execute.Assertion
+            .ForCondition(validValues.Any(val => Subject.CompareTo(val) == Equal))
```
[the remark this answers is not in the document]
jez9999: Oh I see, it's only used with `BeRankedEquallyTo`.  Ok.

jez9999:
Have made the suggested changes in a new commit.

> Thanks for contributing this - I love how this aligns the API better across assertion types.
> 
> Would you be willing to also add examples to the documentation https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/basicassertions.md https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/numerictypes.md
> 
> and a line to the release notes? https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/releases.md

Sure, when do I need to do this?  Presumably after merge?

IT-VBFK:
> Sure, when do I need to do this? Presumably after merge?

Why not in this PR?

jez9999:
Ok so I add it to the "unreleased 6.8.0" section?

IT-VBFK:
No, between this two headlines ;)

jez9999:
OK, docs added.

jez9999:
@jnyrup Does this look good to you with the changes?

Tests/FluentAssertions.Specs/Numeric/ComparableSpecs.cs
```diff
@@ -103,6 +103,67 @@ public void When_two_unequal_objects_should_not_be_equal_it_should_not_throw()
...
+            Action act = () => value.Should().BeOneOf(new ComparableOfInt(4), new ComparableOfInt(5));
+
+            // Assert
+            act
+                .Should().Throw<XunitException>()
+                .WithMessage("Expected value to be one of {4, 5}, but found 4.");
+        }
+
+        [Fact]
+        public void When_a_value_is_the_same_instance_as_one_of_the_specified_values_it_should_succeed()
```
jnyrup, line 154: I was wondering what this test was doing.
It seems we have a hole in our test coverage as `EquatableOfInt.Should()` is picked up by `ObjectAssertions` and not `ComparableTypeAssertions` as it does not implement `IComparable<T>`.
So the tests in `ComparableSpecs` of `[Not]Be` actually tests `ObjectAssertions.[Not]Be`.

I suggest making these changes to `EquatableOfInt` and use it instead of `ComparableInt` for all the added tests.
```diff
- public class EquatableOfInt
+ public class EquatableOfInt : IComparable<EquatableOfInt>
  {
+     int IComparable<EquatableOfInt>.CompareTo(EquatableOfInt other) =>
+        throw new NotSupportedException("This type is meant for assertions using Equals()");
  }
```

Tests/FluentAssertions.Specs/Primitives/ObjectAssertionSpecs.cs
```diff
@@ -146,6 +147,52 @@ public void When_two_equal_objects_are_expected_not_to_be_equal_it_should_fail_a
         }
     }
 
+    public class BeOneOf
+    {
+        [Fact]
+        public void When_a_value_is_not_one_of_the_specified_values_it_should_throw()
+        {
+            // Arrange
+            var value = new EquatableOfInt(3);
```
jnyrup, line 156: Use `ClassWithCustomEqualMethod` in this class like the tests of `Be`/`NotBe`.

jnyrup (changes requested):

Tests/FluentAssertions.Specs/Numeric/ComparableSpecs.cs
jez9999: OK, would it not be better for you to make these changes directly after mine as you understand this bit of the code base better than I do?

jnyrup: Then I'll rather make the changes before you, so you're PR is correct on merge.
I'll make a PR for the fix of `EquatableOfInt`. 
It might take a little time before it is merged as @dennisdoomen is busy speaking at conferences.

jnyrup: #2036

Tests/FluentAssertions.Specs/Numeric/ComparableSpecs.cs
```diff
@@ -103,6 +103,52 @@ public void When_two_unequal_objects_should_not_be_equal_it_should_not_throw()
         }
     }
 
+    public class BeOneOf
+    {
+        [Fact]
+        public void When_a_value_is_not_one_of_the_specified_values_it_should_throw()
```
dennisdoomen, line 109: ℹ️ Technically this one is superfluous.

dennisdoomen (approved):

Tests/FluentAssertions.Specs/Primitives/ObjectAssertionSpecs.cs
```diff
@@ -7,6 +7,7 @@
 using FluentAssertions.Execution;
 using FluentAssertions.Extensions;
 using FluentAssertions.Primitives;
+using FluentAssertions.Specs.Numeric;
```
jnyrup, line 10: delete

Tests/FluentAssertions.Specs/Numeric/ComparableSpecs.cs
jnyrup: #2036 has been merged, so you can rebase on `develop` and change these tests to use `EquatableOfInt` instead and not reuse `value` in this test.

jez9999:
@jnyrup I don't understand what you're looking to achieve with the tests.  It's meant to use `value` twice because it's checking that `BeOneOf` succeeds when it's comparing the same instance, but fails when comparing two comparables that compare as equal.  You're wanting to use `EquatableOfInt` in the file `ComparableSpecs.cs` but the name would suggest that `ComparableOfInt` is more appropriate.  I'm pretty confused by it.

jnyrup:
See the tests of `Be`, they also use `EquatableInt`.

jnyrup (approved):

Issue #241, opened 2015-02-27
ArgumentOutOfRangeException when comparing DateTime.MaxValue or DateTime.MinValue with BeCloseTo()


Sandra121289:
Hi,

yesterday I found this bug:

`DateTime.MaxValue.Should().BeCloseTo(DateTime.MaxValue);`

throws `System.ArgumentOutOfRangeException`. Same goes for `DateTime.MinValue`.

For now I use a custom compare method to work around the issue.

dennisdoomen:
Hi @Sandra121289, thanks for reporting this. Which version of FA have you been using?

Sandra121289:
Hi @dennisdoomen ,

we are using Version 3.2.2 with .NET 4.0.

Sandra121289:
just updated the nuget-Package to version 3.3.0, bug stays.

dennisdoomen:
Hmm, I suppose this is just an example, isn't it? You're not actually comparing `DateTime.MaxValue` with itself?

Sandra121289:
Well, yes and no. This is of course a simplified example. At some point in our software DateTime.MaxValue is representing infinity, so we compare some objects in the unit tests, that have these values.

vossad01:
I can confirm the given example fails for me.  If it is failing for you, it is probably another time zone dependent issue.

I wonder if @konstmsu was still thinking about working on `DateTime` and `DateTimeOffset` handling?

dennisdoomen:
Tonight, I was trying to fix this, but I'm afraid we're running out of options to work around all those implicit `DateTime` to `DateTimeOffset` conversions. I'm afraid we really need to start separating them.

dennisdoomen:
Will be part of release 3.4, but if you want to verify the fix, you can get pre-release package `FluentAssertions 3.4.0-unstable0004` or later from the MyGet feed here: https://www.myget.org/F/fluentassertions_ci/

PR #2030, opened 2022-11-07
Add `BeCloseTo()` / `NotBeCloseTo()` to TimeOnlyAssertions

### LINKED ISSUE #2029: Add `.BeCloseTo()/.NotBeCloseTo()` to TimeOnlyAssertions and DateOnlyAssertions
### Description

Similar to the `DateTimeAssertions`, the method `BeCloseTo()` should be available for TimeOnly and DateOnly.

E.g.

```csharp
// Arrange
SetTime('10:00:12");
var expected = new TimeOnly(10, 0, 12);

// Act
TimeOnly time = GetTime();

// Assert
result.Should().BeCloseTo(expected, TimeSpan.FromSeconds(1));
```

### Expected behavior:

Validation, if the variable `expected` is `>= 10:00:11` and `<= 10:00:13`

IT-VBFK:
## IMPORTANT 

Closes #2029

### DIFF
--- a/Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
+++ b/Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
@@ -48,6 +48,14 @@ public static class FluentTimeSpanExtensions
     }
 
     /// <summary>
+    /// Returns a <see cref="TimeSpan" /> with an added amount of ticks.
+    /// </summary>
+    public static TimeSpan AddTicks(this TimeSpan timeSpan, int ticks)
+    {
+        return TimeSpan.FromTicks(timeSpan.Ticks + ticks);
+    }
+
+    /// <summary>
     /// Gets the nanoseconds component of the time interval represented by the current <see cref="TimeSpan" /> structure.
     /// </summary>
     public static int Nanoseconds(this TimeSpan self)
--- a/Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
+++ b/Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
@@ -52,7 +52,7 @@ public static class FluentTimeSpanExtensions
     /// </summary>
     internal static TimeSpan AddTicks(this TimeSpan timeSpan, int ticks)
     {
-        return TimeSpan.FromTicks(timeSpan.Ticks + ticks);
+        return timeSpan + ticks.Ticks();
     }
 
     /// <summary>
--- a/Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
+++ b/Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
@@ -48,14 +48,6 @@ public static class FluentTimeSpanExtensions
     }
 
     /// <summary>
-    /// Returns a <see cref="TimeSpan" /> with an added amount of ticks.
-    /// </summary>
-    internal static TimeSpan AddTicks(this TimeSpan timeSpan, int ticks)
-    {
-        return timeSpan + ticks.Ticks();
-    }
-
-    /// <summary>
     /// Gets the nanoseconds component of the time interval represented by the current <see cref="TimeSpan" /> structure.
     /// </summary>
     public static int Nanoseconds(this TimeSpan self)
--- a/Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
+++ b/Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
@@ -50,7 +50,7 @@ public static class FluentTimeSpanExtensions
     /// <summary>
     /// Returns a <see cref="TimeSpan" /> with an added amount of ticks.
     /// </summary>
-    public static TimeSpan AddTicks(this TimeSpan timeSpan, int ticks)
+    internal static TimeSpan AddTicks(this TimeSpan timeSpan, int ticks)
     {
         return TimeSpan.FromTicks(timeSpan.Ticks + ticks);
     }
--- a/Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
+++ b/Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
@@ -152,14 +152,12 @@ public class TimeOnlyAssertions<TAssertions>
             throw new ArgumentOutOfRangeException(nameof(precision), $"The value of {nameof(precision)} must be non-negative.");
         }
 
-        long distanceToMinInTicks = (nearbyTime - TimeOnly.MinValue).Ticks;
-        TimeOnly minimumValue = nearbyTime.Add(-TimeSpan.FromTicks(Math.Min(precision.Ticks, distanceToMinInTicks)));
+        TimeOnly minimumValue = nearbyTime.Add(-precision);
+        TimeOnly maximumValue = nearbyTime.Add(TimeSpan.FromTicks(precision.Ticks + 1));
 
-        long distanceToMaxInTicks = (TimeOnly.MaxValue - nearbyTime).Ticks;
-        TimeOnly maximumValue = nearbyTime.Add(TimeSpan.FromTicks(Math.Min(precision.Ticks, distanceToMaxInTicks)));
-
-        long? ticksDifference = Subject?.Ticks - nearbyTime.Ticks;
-        TimeSpan? difference = (ticksDifference != null) ? TimeSpan.FromTicks(Math.Abs(ticksDifference.Value)) : null;
+        TimeSpan? difference = (Subject != null)
+            ? new[] { Subject.Value - nearbyTime, nearbyTime - Subject.Value }.Min()
+            : null;
 
         Execute.Assertion
             .BecauseOf(because, becauseArgs)
@@ -167,8 +165,8 @@ public class TimeOnlyAssertions<TAssertions>
             .ForCondition(Subject is not null)
             .FailWith("but found <null>.")
             .Then
-            .ForCondition((Subject >= minimumValue) && (Subject <= maximumValue))
-            .FailWith("but {0} was off by {1}.", Subject, difference)
+            .ForCondition(Subject.Value.IsBetween(minimumValue, maximumValue))
+            .FailWith("but {0} was off by {1}.", Subject, difference)            
             .Then
             .ClearExpectation();
 
@@ -200,11 +198,8 @@ public class TimeOnlyAssertions<TAssertions>
             throw new ArgumentOutOfRangeException(nameof(precision), $"The value of {nameof(precision)} must be non-negative.");
         }
 
-        long distanceToMinInTicks = (distantTime - TimeOnly.MinValue).Ticks;
-        TimeOnly minimumValue = distantTime.Add(TimeSpan.FromTicks(-Math.Min(precision.Ticks, distanceToMinInTicks)));
-
-        long distanceToMaxInTicks = (TimeOnly.MaxValue - distantTime).Ticks;
-        TimeOnly maximumValue = distantTime.Add(TimeSpan.FromTicks(Math.Min(precision.Ticks, distanceToMaxInTicks)));
+        TimeOnly minimumValue = distantTime.Add(-precision);
+        TimeOnly maximumValue = distantTime.Add(TimeSpan.FromTicks(precision.Ticks + 1));
 
         Execute.Assertion
             .BecauseOf(because, becauseArgs)
@@ -212,7 +207,7 @@ public class TimeOnlyAssertions<TAssertions>
             .ForCondition(Subject is not null)
             .FailWith("but found <null>.")
             .Then
-            .ForCondition((Subject < minimumValue) || (Subject > maximumValue))
+            .ForCondition(!Subject.Value.IsBetween(minimumValue, maximumValue))
             .FailWith("but it was {0}.", Subject)
             .Then
             .ClearExpectation();
--- a/Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
+++ b/Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
@@ -171,7 +171,7 @@ public class TimeOnlyAssertions<TAssertions>
             .ForCondition(Subject is not null)
             .FailWith("but found <null>.")
             .Then
-            .ForCondition(Subject.Value.IsBetween(minimumValue, maximumValue))
+            .ForCondition(Subject?.IsBetween(minimumValue, maximumValue) == true)
             .FailWith("but {0} was off by {1}.", Subject, difference)
             .Then
             .ClearExpectation();
@@ -226,7 +226,7 @@ public class TimeOnlyAssertions<TAssertions>
             .ForCondition(Subject is not null)
             .FailWith("but found <null>.")
             .Then
-            .ForCondition(!Subject.Value.IsBetween(minimumValue, maximumValue))
+            .ForCondition(!Subject?.IsBetween(minimumValue, maximumValue) == true)
             .FailWith("but it was {0}.", Subject)
             .Then
             .ClearExpectation();
--- a/Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
+++ b/Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
@@ -155,11 +155,11 @@ public class TimeOnlyAssertions<TAssertions>
 
         if (precision == TimeSpan.MaxValue)
         {
-            precision = TimeSpan.FromTicks(precision.Ticks - 1);
+            precision = precision.AddTicks(-1);
         }
 
         TimeOnly minimumValue = nearbyTime.Add(-precision);
-        TimeOnly maximumValue = nearbyTime.Add(TimeSpan.FromTicks(precision.Ticks + 1));
+        TimeOnly maximumValue = nearbyTime.Add(precision.AddTicks(1));
 
         TimeSpan? difference = (Subject != null)
             ? MinimumDifference(Subject.Value, nearbyTime)
@@ -212,8 +212,13 @@ public class TimeOnlyAssertions<TAssertions>
             throw new ArgumentOutOfRangeException(nameof(precision), $"The value of {nameof(precision)} must be non-negative.");
         }
 
+        if (precision == TimeSpan.MaxValue)
+        {
+            precision = precision.AddTicks(-1);
+        }
+
         TimeOnly minimumValue = distantTime.Add(-precision);
-        TimeOnly maximumValue = distantTime.Add(TimeSpan.FromTicks(precision.Ticks + 1));
+        TimeOnly maximumValue = distantTime.Add(precision.AddTicks(1));
 
         Execute.Assertion
             .BecauseOf(because, becauseArgs)
--- a/Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
+++ b/Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
@@ -158,9 +158,8 @@ public class TimeOnlyAssertions<TAssertions>
         long distanceToMaxInTicks = (TimeOnly.MaxValue - nearbyTime).Ticks;
         TimeOnly maximumValue = nearbyTime.Add(TimeSpan.FromTicks(Math.Min(precision.Ticks, distanceToMaxInTicks)));
 
-        TimeSpan? difference = (Subject >= nearbyTime
-            ? Subject - nearbyTime
-            : nearbyTime - Subject)?.Duration();
+        long? ticksDifference = Subject?.Ticks - nearbyTime.Ticks;
+        TimeSpan? difference = (ticksDifference != null) ? TimeSpan.FromTicks(Math.Abs(ticksDifference.Value)) : null;
 
         Execute.Assertion
             .BecauseOf(because, becauseArgs)
--- a/Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
+++ b/Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
@@ -128,6 +128,59 @@ public class TimeOnlyAssertions<TAssertions>
     }
 
     /// <summary>
+    /// Asserts that the current <see cref="TimeOnly"/>  is within the specified time
+    /// from the specified <paramref name="nearbyTime"/> value.
+    /// </summary>
+    /// <remarks>
+    /// Use this assertion when, for example the database truncates datetimes to nearest 20ms. If you want to assert to the exact datetime,
+    /// use <see cref="Be(TimeOnly, string, object[])"/>.
+    /// </remarks>
+    /// <param name="nearbyTime">
+    /// The expected time to compare the actual value with.
+    /// </param>
+    /// <param name="precision">
+    /// The maximum amount of time which the two values may differ.
+    /// </param>
+    /// <param name="because">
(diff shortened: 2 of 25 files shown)


Tests/FluentAssertions.Specs/Primitives/TimeOnlyAssertionSpecs.cs
```diff
@@ -170,6 +171,327 @@ public void Should_succeed_when_asserting_timeonly_value_is_not_equal_to_a_diffe
...
+            // Act
+            Action act = () => time.Should().NotBeCloseTo(nearbyTime, 35.Milliseconds());
+
+            // Assert
+            act.Should().Throw<XunitException>()
+                .WithMessage("Did not expect time to be within 35ms from <12:15:31.000>, but it was <12:15:31.035>.");
+        }
+
+        [Fact]
+        public void When_asserting_subject_null_time_is_not_close_to_another_it_should_throw()
```
jnyrup, line 450: This method is identical to `A_null_time_is_never_unclose_to_an_other_time`.

jnyrup (changes requested):
I can't figure out what kind of test branch coverage thinks we're missing 🤔

Tests/FluentAssertions.Specs/Primitives/TimeOnlyAssertionSpecs.cs
```diff
@@ -170,6 +171,327 @@ public void Should_succeed_when_asserting_timeonly_value_is_not_equal_to_a_diffe
         }
     }
 
+    public class BeCloseTo
+    {
+        [Fact]
+        public void When_asserting_that_time_is_close_to_a_negative_precision_it_should_throw()
+        {
+            // Arrange
+            var dateTime = TimeOnly.FromDateTime(DateTime.UtcNow);
```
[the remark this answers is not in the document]
IT-VBFK: ups..

IT-VBFK:
> I can't figure out what kind of test branch coverage thinks we're missing 🤔

Me too 😅

jnyrup (approved):

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
```diff
@@ -127,6 +127,100 @@ public AndConstraint<TAssertions> Be(TimeOnly? expected, string because = "", pa
...
+    /// </param>
+    public AndConstraint<TAssertions> BeCloseTo(TimeOnly nearbyTime, TimeSpan precision, string because = "",
+        params object[] becauseArgs)
+    {
+        if (precision < TimeSpan.Zero)
+        {
+            throw new ArgumentOutOfRangeException(nameof(precision), $"The value of {nameof(precision)} must be non-negative.");
+        }
+
+        long distanceToMinInTicks = (nearbyTime - TimeOnly.MinValue).Ticks;
```
dennisdoomen, line 155: 🤔 I'm wondering why the implementation looks so complicated. Wouldn't this be enough?
```
        long? ticksDifference = Subject?.Ticks - nearbyTime.Ticks;
        TimeSpan? difference = (ticksDifference != null) ? TimeSpan.FromTicks(Math.Abs(ticksDifference.Value)) : null;
```

Tests/FluentAssertions.Specs/Primitives/TimeOnlyAssertionSpecs.cs
```diff
@@ -170,6 +171,312 @@ public void Should_succeed_when_asserting_timeonly_value_is_not_equal_to_a_diffe
         }
     }
 
+    public class BeCloseTo
+    {
+        [Fact]
+        public void When_asserting_that_time_is_close_to_a_negative_precision_it_should_throw()
```
dennisdoomen, line 177: 🔧 Please avoid words like _asserting_

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
IT-VBFK: Hmm.. I like that approach :)

Tests/FluentAssertions.Specs/Primitives/TimeOnlyAssertionSpecs.cs
IT-VBFK: Ok I see..

Fixed

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
jnyrup: This is to handle positive and negative overflow.
See #241

IT-VBFK: I think this belongs to the difference calculation ~5 lines below? If not, I misundertand this completely 🙈

jnyrup: I think I'm the one who got confused.

IT-VBFK: 😅

jnyrup: ... But I also got curious 😉 

and this [generates less code](https://sharplab.io/#v2:D4AQTAjAsAULDOAXATgVwMaIAQEFYG9YtisQBmLAFQEsBbAUwGUAHAQwDsB+LAIXoGEANgHt49SsIAUNBgHl2ggJ7dGqAEYArepgA0VOvXlKs7eq2RrFM+nussOWZsm3V41Yez0gIABixrtVlQxLABeLAAiCL02ZFZaeCxhTW1EAG0AXX9A4PocZABzeABKIhJCGBIqrBF2Au5EanQAa3gAEWoAM076Z3Z0ejCsVRTMTgA6GhbEgFoTMwsrA0mm1oBuMur9BnsuLAATLp6+gaHJRumO7t76fsGAQnD2VEFBYqxuOzZ2cYAxZGEtCmrUkAFlWIgABbjHBqeDnVbtI43O7jABqrEEqHoxXeAC4TC9BBtKlssJtqiAAOwHZEnegkqoAXwpFPI2yY324fCEonEwjA0gMRmUw3UWl0HJF83MlmstgMu0cznQrncnlIvmy6CCIXCURi5niiWSEvSWQCOty+SKpVJxAqZI5u24h2u9LOIzNWEehNe70+iu+fwBQMRYIh0Nh8K9qXRmOxK2mWDmpllSwYSdauKwBOer0ZZIpVWptPdtwGheILJgTKAA=)
```cs
TimeSpan? difference = (Subject != null) ? TimeSpan.FromTicks(Math.Abs(Subject.Value.Ticks - nearbyTime.Ticks)) : null;
```

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
```diff
@@ -127,6 +127,99 @@ public AndConstraint<TAssertions> Be(TimeOnly? expected, string because = "", pa
...
+        TimeOnly minimumValue = nearbyTime.Add(-TimeSpan.FromTicks(Math.Min(precision.Ticks, distanceToMinInTicks)));
+
+        long distanceToMaxInTicks = (TimeOnly.MaxValue - nearbyTime).Ticks;
+        TimeOnly maximumValue = nearbyTime.Add(TimeSpan.FromTicks(Math.Min(precision.Ticks, distanceToMaxInTicks)));
+
+        long? ticksDifference = Subject?.Ticks - nearbyTime.Ticks;
+        TimeSpan? difference = (ticksDifference != null) ? TimeSpan.FromTicks(Math.Abs(ticksDifference.Value)) : null;
```
dennisdoomen, line 162: Do you still need `minimumValue` and `maximumValue`?

jnyrup:
I noticed that while `TimeSpan`, `DateTime` and `DateTimeOffset` are linear in time, `TimeOnly` is circular like a clock on the wall.
So it does not have overflow like the other, but will wrap around.

This bring the question, is 23:59 is close to 00:01 ±2 minutes?
I found out that [`TimeOnly.IsBetween`](https://learn.microsoft.com/en-us/dotnet/api/system.timeonly.isbetween?view=net-6.0) exists for this.

IT-VBFK:
> I found out that TimeOnly.IsBetween exists for this.

Yes, but this is _exclusive_ the end time. so `IsBetween(23:00, 01:00)` would fail if the precision includes 01:00. Is that what we want here, because `DateTime.BeCloseTo()` _includes_ the maximum end date / time. IMHO this behavior should be the same at all assertions.

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
IT-VBFK: By now: yes

I will wait for the discussion below :)

jnyrup:
Adding 1 tick was also my idea.
That we're using `IsBetween` is an implementation detail.

IT-VBFK:
Ok.. this should be easy..

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
IT-VBFK: Ahh.. that's why I was handling this that way: The reason was, that the difference was always the longer differnence.

Like: Subject: 23:59:59, nearbyTime 00:01:00. The difference I want to see is 2 minutes and not 23:58:00 hours :)

Btw.. 23:58:00 hours difference is technically correct, but not that useful :)


Error with the old approach
`Expected time to be within 2m from <23:59:00.000>, but <00:01:00.000> was off by 2m.`
Error with the new approach
`Expected time to be within 2m from <23:59:00.000>, but <00:01:00.000> was off by 23h and 58m.`

dennisdoomen:
> This bring the question, is 23:59 is close to 00:01 ±2 minutes?

Given the behavior of `TimeOnly`, I think this is fair.

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
```diff
@@ -127,6 +127,94 @@ public AndConstraint<TAssertions> Be(TimeOnly? expected, string because = "", pa
...
+        if (precision < TimeSpan.Zero)
+        {
+            throw new ArgumentOutOfRangeException(nameof(precision), $"The value of {nameof(precision)} must be non-negative.");
+        }
+
+        TimeOnly minimumValue = nearbyTime.Add(-precision);
+        TimeOnly maximumValue = nearbyTime.Add(TimeSpan.FromTicks(precision.Ticks + 1));
+
+        TimeSpan? difference = (Subject != null)
+            ? new[] { Subject.Value - nearbyTime, nearbyTime - Subject.Value }.Min()
```
IT-VBFK, line 159: I am not sure about this line..

Was by far the least code :)

but it works like I would expect it to work.

IT-VBFK: The alternative would be:
`TimeSpan.FromTicks(Math.Min((Subject.Value - nearbyTime).Duration().Ticks, (nearbyTime - Subject.Value).Duration().Ticks))`

IT-VBFK:
> ## Pull Request Test Coverage Report for [Build 3517626725](https://coveralls.io/builds/54431086)
> * **0** of **0**   changed or added relevant lines in **0** files are covered.
> * **11** unchanged lines in **1** file lost coverage.
> * Overall coverage increased (+**0.3%**) to **95.928%**
> 
> Files with Coverage Reduction	New Missed Lines	%
> [Src/FluentAssertions/Data/DataEquivalencyAssertionOptions.cs](https://coveralls.io/builds/54431086/source?filename=Src%2FFluentAssertions%2FData%2FDataEquivalencyAssertionOptions.cs#L32)	11	86.73%
> Totals	[Coverage Status](https://coveralls.io/builds/54431086)
> Change from base [Build 3504387639](https://coveralls.io/builds/54385462):	0.3%
> Covered Lines:	8494
> Relevant Lines:	8741
> ##### 💛 - [Coveralls](https://coveralls.io)

What's the best way to generate random numbers: let coveralls calculate your coverage.. 😂

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
IT-VBFK: Just committed a proposal..

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
```diff
@@ -127,6 +127,94 @@ public AndConstraint<TAssertions> Be(TimeOnly? expected, string because = "", pa
...
+    public AndConstraint<TAssertions> BeCloseTo(TimeOnly nearbyTime, TimeSpan precision, string because = "",
+        params object[] becauseArgs)
+    {
+        if (precision < TimeSpan.Zero)
+        {
+            throw new ArgumentOutOfRangeException(nameof(precision), $"The value of {nameof(precision)} must be non-negative.");
+        }
+
+        TimeOnly minimumValue = nearbyTime.Add(-precision);
+        TimeOnly maximumValue = nearbyTime.Add(TimeSpan.FromTicks(precision.Ticks + 1));
```
jnyrup, line 156: If `precision` is `TimeSpan.MaxValue` this overflows.

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
jnyrup: I think this works

```cs
TimeSpan? difference = Subject.HasValue ? MinimumDifference(Subject.Value, nearbyTime) : null;
...
private static TimeSpan? MinimumDifference(TimeOnly a, TimeOnly b)
{
    var diff1 = a - b;
    var diff2 = b - a;

    return diff1 < diff2 ? diff1 : diff2;
}
```

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
IT-VBFK: True

The only way to get around this, is to define the borders from `TimeOnly.MinValue.Ticks`to `TimeOnly.MaxValue.Ticks - 1`

This is IMO "close" enough.. 😂

Tests/FluentAssertions.Specs/Primitives/TimeOnlyAssertionSpecs.cs
```diff
@@ -414,6 +414,20 @@ public void When_a_time_is_close_to_an_earlier_time_by_one_tick_it_should_fail()
...
+        {
+            // Arrange
+            TimeOnly time = new TimeOnly(23, 59, 0);
+            TimeOnly nearbyTime = new TimeOnly(0, 1, 0);
+
+            // Act
+            Act act = () => time.Should().NotBeCloseTo(nearbyTime, TimeSpan.MaxValue);
+
+            // Assert
+            act.Should().Throw<XunitException>();
```
IT-VBFK, line 428: This line is intentional, because a `TimeOnly` (which has 24h) couldn't ever be close to a maximum time span.

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
```diff
@@ -127,6 +128,112 @@ public AndConstraint<TAssertions> Be(TimeOnly? expected, string because = "", pa
...
+            throw new ArgumentOutOfRangeException(nameof(precision), $"The value of {nameof(precision)} must be non-negative.");
+        }
+
+        if (precision == TimeSpan.MaxValue)
+        {
+            precision = precision.AddTicks(-1);
+        }
+
+        TimeOnly minimumValue = nearbyTime.Add(-precision);
+        TimeOnly maximumValue = nearbyTime.Add(precision.AddTicks(1));
```
jnyrup, line 162: As the interval for `IsBetween` is left-_closed_ and right-_open_, we should only adjust for the ending time.

```cs
TimeOnly minimumValue = nearbyTime.Add(-precision);
TimeOnly maximumValue = nearbyTime.Add((precision == TimeSpan.MaxValue) ? precision : precision.AddTicks(1));
```

Alternatively we could just copy the implementation of [`IsBetween`](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/TimeOnly.cs,208) and adjust it to be closed in both ends.

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
```diff
@@ -127,6 +128,112 @@ public AndConstraint<TAssertions> Be(TimeOnly? expected, string because = "", pa
...
+            ? MinimumDifference(Subject.Value, nearbyTime)
+            : null;
+
+        Execute.Assertion
+            .BecauseOf(because, becauseArgs)
+            .WithExpectation("Expected {context:the time} to be within {0} from {1}{reason}, ", precision, nearbyTime)
+            .ForCondition(Subject is not null)
+            .FailWith("but found <null>.")
+            .Then
+            .ForCondition(Subject.Value.IsBetween(minimumValue, maximumValue))
```
jnyrup, line 174: If a call to `BeCloseTo` is wrapped inside an `AssertionScope` this chain of calls continues even `.ForCondition(Subject is not null)` fails.
```suggestion
            .ForCondition(Subject?.IsBetween(minimumValue, maximumValue) == true)
```

Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
```diff
@@ -47,6 +47,14 @@ public static TimeSpan Ticks(this long ticks)
         return TimeSpan.FromTicks(ticks);
     }
 
+    /// <summary>
+    /// Returns a <see cref="TimeSpan" /> with an added amount of ticks.
+    /// </summary>
+    public static TimeSpan AddTicks(this TimeSpan timeSpan, int ticks)
+    {
+        return TimeSpan.FromTicks(timeSpan.Ticks + ticks);
+    }
```
jnyrup, line 56: Let's keep this an internal method for now to keep the PR focused on a single feature.

Consumers can currently use `timeSpan + 1.Ticks()` or `timeSpan.Add(1.Ticks())`.

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
IT-VBFK: Ok.. don't get why? But cool to know :D

Src/FluentAssertions/Extensions/FluentTimeSpanExtensions.cs
IT-VBFK: Ok

Tests/FluentAssertions.Specs/Primitives/TimeOnlyAssertionSpecs.cs
IT-VBFK: @jnyrup Is this test necessary at all? Because this test does not "test" anything.. (is never failing). Maybe we should remove it?

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
```diff
@@ -127,6 +128,102 @@ public AndConstraint<TAssertions> Be(TimeOnly? expected, string because = "", pa
...
+            ? MinimumDifference(Subject.Value, nearbyTime)
+            : null;
+
+        Execute.Assertion
+            .BecauseOf(because, becauseArgs)
+            .WithExpectation("Expected {context:the time} to be within {0} from {1}{reason}, ", precision, nearbyTime)
+            .ForCondition(Subject is not null)
+            .FailWith("but found <null>.")
+            .Then
+            .ForCondition(Subject?.IsBetween(minimumValue, maximumValue) == true)
```
jnyrup, line 169: We should have a test wrapping `BeCloseTo` in an `AssertionScope` to exercise the null check on `Subject`.
Same for `NotBeCloseTo`.

Tests/FluentAssertions.Specs/Primitives/TimeOnlyAssertionSpecs.cs
jnyrup: Doesn't it exercise the `(precision == TimeSpan.MaxValue)` check?

jnyrup (commented):
Looks good!
I have no further comments on the functionality/implementation.
Thanks for taking this through the iteration of changes necessary to this result.

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
IT-VBFK: That's correct

Tests/FluentAssertions.Specs/Primitives/TimeOnlyAssertionSpecs.cs
IT-VBFK: Yes. What I meant was: This could not fail, because the maximum time span is something around 10000 days. 

But if you do not have this check it does not fail the test.. 

> Looks good!
> I have no further comments on the functionality/implementation.
> Thanks for taking this through the iteration of changes necessary to this result.

No problem :)

we all love good clean code.. That is a necessary "evil" 😉

IT-VBFK:
> Looks good!
> I have no further comments on the functionality/implementation.
> Thanks for taking this through the iteration of changes necessary to this result.

No problem :)

we all love good clean code.. That is a necessary "evil" 😉

Src/FluentAssertions/Primitives/TimeOnlyAssertions.cs
IT-VBFK: Just re-thought this.. and yes.. now I get it. This is just the "normal" use-case for an assertion scope

Tests/FluentAssertions.Specs/Primitives/TimeOnlyAssertionSpecs.cs
jnyrup: > But if you do not have this check it does not fail the test..

Ah... I somehow thought `AddTicks(this TimeSpan ts, int ticks)` was implemented as `ts + ticks.Ticks()` which for `TimeSpan.MaxValue + 1.Ticks()` would use `TimeSpan.Add` and throw an `OverflowException`.
https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs,234
but instead we would do an unchecked overflow on long addition.

I'm beginning to think we should just copy and adjust `TimeOnly.IsBetween` to check for the closed interval instead.
With that implementation I would be fine with removing this test.

IT-VBFK: To be honest: I think the first approach is less "magic" if you know what I mean. And the lesser change :D

jnyrup:
I'm sorry, I must have been unclear 😢 
See https://github.com/jnyrup/fluentassertions/commit/5a34585c283a3849c13986bb8822aa42f09a0de9 for how we can avoid both checked overflow (`TimeSpan.Add`) and unchecked overflow (`long.operator+`).

IT-VBFK:
Ahh.. now I get the point :)

jnyrup:
> Do you mind if I copy-paste this? :)
Not at all, it was tailor made to be copied.

IT-VBFK:
Ok.. just a moment please :D

dennisdoomen (approved):
Nice work

IT-VBFK:
Thanks :)

jnyrup:
> Is it intentional to not tag this PR with feature or enhancement? 

I simply forgot.
Thanks for noticing.

PR #2037, opened 2022-11-19
Improve coverage on data assertions

lg2de:
I've added new tests to improve code coverage according to #1823.
While analyzing the code I found that there was thrown an `Exception`, an exception type which shall not be used directly. I've changed it to `ArgumentException` which should compatible.

Further tests on `DataEquivalencyAssertionOptions` I could not create because I do not understand their intention.

Please note (and check?) the method `ExcludingColumns(string tableName, IEnumerable<string> columnNames)` which creates locally a variable named `excludeColumnNames` which hides similar member. I'm not sure whether this is correct or potential bug. Maybe @logiclrd can check this because you created this code in #1419.

### DIFF
--- a/Tests/FluentAssertions.Equivalency.Specs/DataColumnSpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/DataColumnSpecs.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Data;
 using Xunit;
 using Xunit.Sdk;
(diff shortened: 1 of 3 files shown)


jnyrup (approved):

logiclrd:
> Please note (and check?) the method ExcludingColumns(string tableName, IEnumerable<string> columnNames) which creates locally a variable named excludeColumnNames which hides similar member. I'm not sure whether this is correct or potential bug.

The code looks right to me. I'm not sure how to improve the code's comprehensibility, but I can explain what's going on: The object persists, in a dictionary keyed by table name, hash sets of excluded column names, and this method injects additional column names for a particular table. The caller supplies an enumeration of additional column names to be ignored, and the method finds the `HashSet<string>` that is storing the configuration for the table in question and injects those additional column names into it.

dennisdoomen (approved):

Tests/FluentAssertions.Equivalency.Specs/DataColumnSpecs.cs
```diff
@@ -77,6 +78,43 @@ public void When_DataColumn_has_changes_but_is_excluded_it_should_succeed()
             .ExcludingColumn(dataColumn2));
     }
 
+    [Fact]
+    public void When_DataColumn_has_changes_but_is_excluded_as_params_it_should_succeed()
```
dennisdoomen, line 82: 🌱 We really need to start improving those test names, but that's besides the point of this PR.

dennisdoomen (approved):

PR #2038, opened 2022-11-20
Bump minimum SDK to 6.0.400

jnyrup:
PolySharp requires a newer SDK to use source generators

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "6.0.100",
+        "version": "6.0.400",
         "rollForward": "latestMajor"
     }
 }


dennisdoomen (approved):

global.json
```diff
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "6.0.100",
+        "version": "6.0.400",
```
dennisdoomen, line 3: Are you sure it shouldn't be ` 6.0.403`?

dennisdoomen (approved):

global.json
jnyrup: Not sure, but here are some facts.

PolySharp 1.0.1 was released 2022-10-22, 6.0.403 was released on 2022-11-08.

https://github.com/Sergio0694/PolySharp/blob/main/src/PolySharp.SourceGenerators/PolySharp.targets#L53 mentions VS 17.3
VS 17.3 was released with 6.0.400

https://dotnet.microsoft.com/en-us/download/dotnet/6.0
image

jnyrup: I just tried out locally fixing to 6.0.302 and 6.0.400.
6.0.302 didn't work.
6.0.400 worked.

jnyrup:
Opened https://github.com/Sergio0694/PolySharp/pull/27 to help others hitting the same issue.

PR #2039, opened 2022-11-20
Pattern combinators

jnyrup:
1) Found some more places where patterns can be used.

2) The guards were introduced in #669, but this is only an optimization when passing in 0.
I estimate that most usages do not that, as why would you call `AddMicroSeconds(0)`.


dennisdoomen (approved):
Nice one!

PR #2042, opened 2022-11-22
Add unit tests for better coverage in Execution namespace

lg2de:
I've created some new tests (and minor refactoring) for better coverage in the namespace `Execution` according to #1823.

### DIFF
--- a/Src/FluentAssertions/Common/Services.cs
+++ b/Src/FluentAssertions/Common/Services.cs
@@ -46,6 +46,6 @@ public static class Services
 #else
         ConfigurationStore = new NullConfigurationStore();
 #endif
-        ThrowException = TestFrameworkProvider.Throw;
+        ThrowException = new TestFrameworkProvider(Configuration).Throw;
     }
 }
--- a/Tests/FluentAssertions.Specs/Execution/FallbackTestFrameworkTests.cs
+++ b/Tests/FluentAssertions.Specs/Execution/FallbackTestFrameworkTests.cs
@@ -1,0 +1,24 @@
+using FluentAssertions.Execution;
+using Xunit;
+
+namespace FluentAssertions.Specs.Execution;
+
+public class FallbackTestFrameworkTests
+{
+    [Fact]
+    public void The_fallback_test_framework_is_available()
+    {
+        var sut = new FallbackTestFramework();
+
+        sut.IsAvailable.Should().BeTrue();
+    }
+
+    [Fact]
+    public void Throwing_with_messages_throws_the_exception()
+    {
+        var sut = new FallbackTestFramework();
+
+        sut.Invoking(x => x.Throw("test message")).Should().ThrowExactly<AssertionFailedException>()
+            .WithMessage("test message");
+    }
+}
--- a/Tests/FluentAssertions.Specs/Execution/TestFrameworkProviderTests.cs
+++ b/Tests/FluentAssertions.Specs/Execution/TestFrameworkProviderTests.cs
@@ -2,8 +2,7 @@ using System;
 using FluentAssertions.Common;
 using FluentAssertions.Execution;
 using Xunit;
-
-using static FluentAssertions.FluentActions;
+using Xunit.Sdk;
 
 namespace FluentAssertions.Specs.Execution;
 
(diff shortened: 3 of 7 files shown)


jnyrup:
I'm wondering if we can test this through the public API instead of accessing internals.

E.g. could we add a test project with an custom/unsupported test framework to exercise when `!framework.IsAvailable`?

lg2de:
> I'm wondering if we can test this through the public API instead of accessing internals.

Currently the analysis of test framework is not public. So it cannot be tested with public API.
Should we change this?

> E.g. could we add a test project with an custom/unsupported test framework to exercise when `!framework.IsAvailable`?

I guess this is also currently not possible because the API for "custom test framework" is not possible.

jnyrup:
> Currently the analysis of test framework is not public. So it cannot be tested with public API.
> Should we change this?

By public API I intended to mean through some part of the existing public API that can reach `TestFrameworkProvider`, which seems (I'm not at my development) to be `Services.Throw`.

I can recommend reading [Unit testing internals](https://blog.ploeh.dk/2015/09/22/unit-testing-internals/) for some interesting thoughts on this subject.

lg2de:
Within really existing testing frameworks (which are used in the FA solution) the `FallbackFramework` is never instantiated.
It is not possible to **change** the configuration to use the fallback or any dummy because `Services.Configuration` is getter only.
Finally `FallbackFramework` is `internal`.
So, how to test it without `InternalsVisibleTo`?
It would require "major" refactoring to test these classes.

dennisdoomen:
> E.g. could we add a test project with an custom/unsupported test framework to exercise when

That's exactly what we should be doing instead.

lg2de:
What is you preferred unsupported framework?

lg2de:
Mmh, `AssertionOptions` is a static class which cannot savely be changed for a single test only.
I guess it will get quite complicated to get progress here.

lg2de:
I've removed the tests which requires (additional) `InternalsVisibleTo`.
Refactoring `AssertionOption` needs more discussion. I guess this could be a target for v7.

Tests/FluentAssertions.Specs/Execution/TestFrameworkProviderTests.cs
```diff
@@ -0,0 +1,78 @@
...
+        result.Should().BeOfType<XUnit2TestFramework>();
+    }
+
+    [Fact]
+    public void When_running_test_with_unknown_test_framework_it_should_throw()
+    {
+        var configurationStore = new TestConfigurationStore { Data = { [Configuration.TestFrameworkConfigurationKey] = "foo" } };
+        var configuration = new Configuration(configurationStore);
+
+        this.Invoking(_ => TestFrameworkProvider.AttemptToDetectUsingAppSetting(configuration))
```
jnyrup, line 38: If you add
```cs
using static FluentAssertions.FluentActions;
```

You can write this as
```cs
Invoking(() => TestFrameworkProvider.AttemptToDetectUsingAppSetting(configuration))
```

In the end it is functionally equivalent to the existing, but to keep the majority of the test suite a prime example of how to write unit tests using Fluent Assertions.

Tests/FluentAssertions.Specs/Execution/TestFrameworkProviderTests.cs
```diff
@@ -0,0 +1,78 @@
...
+    private class TestConfigurationStore : IConfigurationStore
+    {
+        public IDictionary<string, string> Data { get; } = new Dictionary<string, string>();
+
+        public string GetSetting(string name)
+        {
+            Data.TryGetValue(name, out var setting);
+            return setting ?? string.Empty;
+        }
+    }
```
jnyrup, line 77: To avoid relying on `Configuration.TestFrameworkConfigurationKey` and removes the need to extract it to a constant.
```cs
private sealed class TestConfigurationStore : IConfigurationStore
{
    private readonly string testFramework;

    public TestConfigurationStore(string testFramework)
    {
        this.testFramework = testFramework;
    }

    public string GetSetting(string name) => testFramework;
}
```

Tests/FluentAssertions.Specs/Execution/FallbackTestFrameworkTests.cs
```diff
@@ -0,0 +1,24 @@
+using FluentAssertions.Execution;
+using Xunit;
+
+namespace FluentAssertions.Specs.Execution;
+
+public class FallbackTestFrameworkTests
+{
+    [Fact]
+    public void When_accessing_fallback_test_framework_it_is_always_available()
```
jnyrup, line 9: We're trying to migrate (at least new test classes) to an in our opinion less noisy naming schema.
So instead of the sometimes inflexible `when_foo_then_bars` format, this method could e.g. be named `The_fallback_test_framework_is_available()`.

See https://enterprisecraftsmanship.com/posts/you-naming-tests-wrong/

dennisdoomen (commented):
🤔 I have to admit that I don't like this approach where we make private stuff `internal`. Instead, I prefer that we make the list of test frameworks configurable. In other words, we allow people to modify the `Frameworks` collection through `AssertionOptions` (in the `AssertionOptionSpecs`. Then you just empty the collection and observe the effect.

Tests/FluentAssertions.Specs/Execution/TestFrameworkProviderTests.cs
lg2de: What is the problem in defining the constant?
I think this is the better way at all instead if using literals, isn't it?

Tests/FluentAssertions.Specs/Execution/FallbackTestFrameworkTests.cs
lg2de: For me it gets more and more difficult to name the tests according your thoughts...
Is your version above a general suggestion or a requirement?
According to the linked article it is correct, I guess.

lg2de:
> 🤔 I have to admit that I don't like this approach where we make private stuff `internal`. Instead, I prefer that we make the list of test frameworks configurable. In other words, we allow people to modify the `Frameworks` collection through `AssertionOptions` (in the `AssertionOptionSpecs`. Then you just empty the collection and observe the effect.

As I already wrote above, making used testing framework configurable is very difficult, because `AssertionOptions` is a static class. Multiple tests running in parallel cannot configure different testing frameworks. I guess this requires major rework in V7. (I'll make a separate issue therefore.)

The current version of the PR does not change the public API. So, what is the problem with the proposed changes?

dennisdoomen:
> As I already wrote above, making used testing framework configurable is very difficult, because `AssertionOptions` is a static class. Multiple tests running in parallel cannot configure different testing frameworks. I guess this requires major rework in V7. (I'll make a separate issue therefore.)

The `AssertionOptionsSpecs` don't run in parallel with any other test and are a safe place to add these kinds of tests.

Tests/FluentAssertions.Specs/Execution/TestFrameworkProviderTests.cs
jnyrup: The problem is tests using internals when not needed.
I had an extra look at this and we can use `Configuration.TestFrameworkName` instead.

```cs
[Fact]
public void When_running_xunit_test_explicitly_it_should_be_detected()
{
    var configuration = new Configuration(new TestConfigurationStore())
    {
        TestFrameworkName = "xunit2"
    };
    var result = TestFrameworkProvider.DetectFramework(configuration);

    result.IsAvailable.Should().BeTrue();
    result.Should().BeOfType<XUnit2TestFramework>();
}

private sealed class TestConfigurationStore : IConfigurationStore
{
    string IConfigurationStore.GetSetting(string name) => string.Empty;
}
```

Tests/FluentAssertions.Specs/Execution/TestFrameworkProviderTests.cs
lg2de: done

Tests/FluentAssertions.Specs/Execution/TestFrameworkProviderTests.cs
lg2de: done

Tests/FluentAssertions.Specs/Execution/FallbackTestFrameworkTests.cs
lg2de: done

lg2de:
The mentioned tests in the context of `AssertionOptions` I'll take care separately.

Hope the PR is acceptable now.

Tests/FluentAssertions.Specs/Execution/TestFrameworkProviderTests.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        var configuration = new Configuration(new TestConfigurationStore())
+        {
+            TestFrameworkName = "foo"
+        };
+
+        Invoking(() => TestFrameworkProvider.AttemptToDetectUsingAppSetting(configuration))
+            .Should().Throw<InvalidOperationException>()
+            .WithMessage("FluentAssertions was configured to use the test framework 'foo' but this is not supported."
+                         + " Please use one of the supported frameworks: mspec, nspec3, nunit, mstestv2, xunit2.");
```
jnyrup, line 46: Just a suggestion to rely a little less on the _exact_ failure message.

```cs
.WithMessage("*the test framework 'foo' but this is not supported*");
```

jnyrup (approved):

dennisdoomen:
Instead of making certain members `internal`, I thought we agreed to do the following:
* Make the test framework configurable through the `AssertionScope`
* Move the tests to the `AssertionScopeSpecs`?

lg2de:
I made internal member of an internal class. Is this that bad?
Configuration using `AssertionScope` is required for the tests I already removed from this PR.

Tests/FluentAssertions.Specs/Execution/TestFrameworkProviderTests.cs
lg2de: done

dennisdoomen:
> I made internal member of an internal class. Is this that bad?

You're testing the internals of a class. That's something we should avoid at all costs. If we would change the internals, it should not break existing tests. That's my golden rule of thumb...

jnyrup:
> Make the test framework configurable through the `AssertionScope`

You still mean `AssertionOptions`, right?
`AssertionScope`s are constructed over and over, but you don't want/can't to change test framework provider in the middle of a test run.

Currently if a user wants to provide their own throwing mechanism, they can assign one to `Services.ThrowException`

dennisdoomen:
> > Make the test framework configurable through the `AssertionScope`
> 
> You still mean `AssertionOptions`, right? `AssertionScope`s are constructed over and over, but you don't want/can't to change test framework provider in the middle of a test run.
> 
> Currently if a user wants to provide their own throwing mechanism, they can assign one to `Services.ThrowException`

Yes, I meant `AssertionOptions`.... 🤦‍♂️

lg2de:
Now I have removed the last commit too.
The effort to create code coverage on `TestFrameworkProvider` without accessing internals is IMHO too high.

jnyrup:
@dennisdoomen How about this approach to make `TestFrameworkProvider` testable without exposing private methods or changing the public API.

https://github.com/jnyrup/fluentassertions/commit/eea17860f2d0c9135ca73a73641e1faa7e57e485

dennisdoomen:
> @dennisdoomen How about this approach to make `TestFrameworkProvider` testable without exposing private methods or changing the public API.
> 
> [jnyrup@eea1786](https://github.com/jnyrup/fluentassertions/commit/eea17860f2d0c9135ca73a73641e1faa7e57e485)

Love it!

jnyrup:
@lg2de do you want to incorporate the changes from my commit above?
I don't want to steal your hard work.

Src/FluentAssertions/Common/Services.cs
```diff
@@ -46,6 +46,6 @@ public static void ResetToDefaults()
 #else
         ConfigurationStore = new NullConfigurationStore();
 #endif
-        ThrowException = TestFrameworkProvider.Throw;
+        ThrowException = new TestFrameworkProvider(Configuration).Throw;
```
lg2de, line 49: @jnyrup Did you think about the outcome of creating the provide with every thrown exception?
The class was "designed" before for e.g. one-time detection of the framework.

jnyrup: This line is only invoked once from the static constructor of `Services`, unless you explicitly class `Services.ResetToDefaults()`.

jnyrup (approved):

dennisdoomen (approved):

PR #1594, opened 2021-05-31
Guard more assertions against null subject variables.

eNeRGy164:
Working on #1039

**ReferenceTypeAssertions**
- [x] `BeOfType<T>`
- [x] `BeOfType`
- [x] `NotBeOfType<T>`
- [x] `NotBeOfType`
- [x] `BeAssignableTo<T>`
- [x] `BeAssignableTo`
- [x] `NotBeAssignableTo<T>`
- [x] `NotBeAssignableTo`

**StringCollectionAssertions**
- [x] `ContainMatch`
- [x] `NotContainMatch`

**GenericCollectionAssertions**
- [x] `AllBeAssignableTo`
- [x] `AllBeAssignableTo<T>`
- [x] `AllBeOfType`
- [x] `AllBeOfType<T>`
- [x] `BeEmpty`
- [x] `BeInAscendingOrder`
- [x] `BeInDescendingOrder`
- [x] `BeSubsetOf`
- [x] `Contain` (3x)
- [x] `ContainEquivalentOf`
- [x] `ContainInOrder`
- [x] `ContainItemsAssignableTo<T>`
- [x] `ContainSingle`
- [x] `HaveCount`
- [x] `HaveCountGreaterOrEqualTo`
- [x] `HaveCountGreaterThan`
- [x] `HaveCountLessOrEqualTo`
- [x] `HaveCountLessThan`
- [x] `HaveElementAt`
- [x] `HaveElementPreceding`
- [x] Receive Microsoft MVP award
- [x] `HaveElementSucceeding`
- [x] `HaveSameCount`
- [x] `IntersectWith`
- [x] `NotBeEmpty`
- [x] `NotBeInAscendingOrder`
- [x] `NotBeInDescendingOrder`
- [x] `NotBeNullOrEmpty`
- [x] `NotBeSubsetOf`
- [x] `NotContain`
- [x] `NotContainEquivalentOf`
- [x] `NotContainNulls`
- [x] `NotEqual`
- [x] `NotHaveCount`
- [x] `NotHaveSameCount`
- [x] `NotIntersectWith`
- [x] `OnlyContain`
- [x] `OnlyHaveUniqueItems`
- [x] `SatisfyRespectively`
- [x] `Satisfy`

Also small updates to documentation tags where appropriate.

PR #2044, opened 2022-11-25
Nested `AssertionScope`s do not print inner scope reportables

94sedighi:
## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -395,6 +395,7 @@ public sealed class AssertionScope : IAssertionScope
                 parent.assertionStrategy.HandleFailure(failureMessage);
             }
 
+            parent.contextData.Add(contextData);
             parent.AppendTracing(tracing.ToString());
 
             parent = null;
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -263,7 +263,7 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> :
             .FailWith("but found a null element.")
             .Then
             .ForCondition(subject => subject.All(x => expectedType == GetType(x)))
-            .FailWith("but found {0}.", subject => $"but found [{string.Join(", ", subject.Select(x => GetType(x).FullName))}].")
+            .FailWith("but found {0}.", subject => $"[{string.Join(", ", subject.Select(x => GetType(x).FullName))}]")
             .Then
             .ClearExpectation();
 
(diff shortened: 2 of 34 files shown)


Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -395,6 +395,16 @@ public void Dispose()
...
 
+            IDictionary<string, object> reportables = contextData.GetReportable();
+
+            if (reportables.Count > 0)
+            {
+                foreach (KeyValuePair<string, object> reportable in reportables)
+                {
+                    parent.AddReportable(reportable.Key, reportable.Value.ToString());
+                }
+            }
```
jnyrup, line 406: I haven't dug all the way into this to know all the implications, but we might be able to use `parent.contextData.Add(contextData);` instead.

This should also ensure that:
* `reportable.Value` is properly formatted with `Formatter.ToString` when `ToString()` is insufficient.
* `reportable.Value` is not formatted when the test passed.
* non-reportables are handled as well.
  That seems to benefit e.g. `NotBeEquivalentTo` to include the non-reportable section about the used equivalency settings.
  See `When_asserting_collections_not_to_be_equivalent_with_options_but_subject_collection_is_null_it_should_throw`

94sedighi: may i ask what is the purpose of adding an AsserionScope here in: When_asserting_collections_not_to_be_equivalent_with_options_but_subject_collection_is_null_it_should_throw()
{
  Action act = () =>
  {
      actual.Should().NotBeEquivalentTo(new[] { 1, 2, 3 }, opt => opt, "we want to test the failure {0}", "message");
  };
}

94sedighi: It actually throws the exception, but with a different message:

    Expected exception message to match the equivalent of "Expected actual not to be equivalent *failure message*, but found <null>.", but "Expected actual not to be equivalent because we want to test the failure message, but found <null>.

With configuration:
- Compare enums by value
" does not.

jnyrup: > May I ask what is the purpose of adding an AssertionScope here in:

As an assertion don't fail immediately when wrapped in an `AssertionScope`, we want to exercise the case, where even though we have null checked `Subject` we cannot safely assume it to be non-null afterwards.

This particular test was added in #1594.

jnyrup: > It actually throws the exception, but with a different message:

Yes, and at first glimpse that extended failure message looks more correct to me, as it includes the "equivalency configuration".
The test should then be adapted to this, by adding a trailing `*` to the assertion.

dennisdoomen (approved):

jnyrup (approved):

PR #2046, opened 2022-11-28
Comparable type assertions referential equality

jnyrup:
`SameInstanceIsNotEqualClass` did not implement `IComparable<T>` so `Should` resolved to `ObjectAssertions` instead of `ComparableTypeAssertions`.

The PR is structured into three commits for easier reviewing.

This is a follow-up to #2036 and #2028

### DIFF
--- a/Tests/FluentAssertions.Specs/Numeric/ComparableSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Numeric/ComparableSpecs.cs
@@ -21,20 +21,18 @@ public class ComparableSpecs
         }
 
         [Fact]
-        public void When_two_instances_are_the_same_reference_but_are_not_considered_equal_it_should_not_succeed()
+        public void When_two_instances_are_the_same_reference_but_are_not_considered_equal_it_should_succeed()
         {
             // Arrange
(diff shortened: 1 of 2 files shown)


Tests/FluentAssertions.Specs/Numeric/ComparableSpecs.cs
```diff
@@ -21,20 +21,18 @@ public void When_two_instances_are_equal_it_should_succeed()
         }
 
         [Fact]
-        public void When_two_instances_are_the_same_reference_but_are_not_considered_equal_it_should_not_succeed()
+        public void When_two_instances_are_the_same_reference_but_are_not_considered_equal_it_should_succeed()
```
dennisdoomen, line 24: 🤔 I can't get my head around this one ;-)

dennisdoomen (approved):

Tests/FluentAssertions.Specs/Numeric/ComparableSpecs.cs
jnyrup: Yeah it's quite unlikely that someone would implement a type that is not reflexive.

PR #2055, opened 2022-12-03
Add first rough attempt of improving string assertion failure messages

### LINKED ISSUE #2050: Improve failure message for `StringAssertions.Be`
### Discussed in https://github.com/fluentassertions/fluentassertions/discussions/2048 and https://github.com/fluentassertions/fluentassertions/discussions/1941

<div type='discussions-op-text'>

<sup>Originally posted by **thomOrbelius** November 30, 2022</sup>
Hi.

In my department we have been starting to use FluentAssertions. One of the selling point I had was the improved failure messages. A colleague of mine pointed out that the Xunit failure messages for strings is much more helpful (see example below) than the FluentAssertion equivalent. The example shows a case where we use a string with a newline separator.

Do any one else think that a message more similar to the one Xunit has is more helpful than the current? If so, should we update the message used?

```csharp
using FluentAssertions;
using Xunit;

namespace Sectra.Courier.RisAdapter.Tests.Integrations.SectraRis;

public class FluentComparison {
    private const string string1 = "String with \rnewline";
    private const string string2 = "String with \ra newline";

    [Fact]
    public void FluentAssert() {
        // Fails with:
        // newline" has a length of 20, differs near "new" (index 13).
        string1.Should().Be(string2);
    }

    [Fact]
    public void XUnitAssert() {
        // Fails with:
        // Assert.Equal() Failure
        //                         ↓ (pos 13)
        // Expected: String with \rnewline
        // Actual:   String with \ra newline
        //                         ↑ (pos 13)
        Assert.Equal(string1, string2);
    }
}
```
</div>

IT-VBFK:
This closes #2050

## IMPORTANT 

To do:

- [ ] Decide what to do with really long strings

### DIFF
--- a/Src/FluentAssertions/Primitives/StringStartValidator.cs
+++ b/Src/FluentAssertions/Primitives/StringStartValidator.cs
@@ -45,9 +45,21 @@ internal class StringStartValidator : StringValidator
         {
             int indexOfMismatch = Subject.IndexOfFirstMismatch(Expected, stringComparison);
 
+            string subjectIndexMarker = $" ↓ (index {indexOfMismatch})";
+            string expectedIndexMarker = $" ↑ (index {indexOfMismatch})";
+            string subjectDescription = "Subject:  ";
+            string expectationDescription = "Expected: ";
+
             Assertion.FailWith(
-                ExpectationDescription + "{0}{reason}, but {1} differs near " + Subject.IndexedSegmentAt(indexOfMismatch) + ".",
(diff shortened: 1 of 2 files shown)


dennisdoomen:
What do we want with this PR?

IT-VBFK:
I am not sure if this change could be considered as a "breaking change", because this requires a lot of internal changes I think.

I will close this for now.

PR #2056, opened 2022-12-05
Add `ThatAre[Not]Structs` in `TypeSelector.cs`

94sedighi:
## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Types/TypeSelector.cs
+++ b/Src/FluentAssertions/Types/TypeSelector.cs
@@ -184,6 +184,32 @@ public class TypeSelector : IEnumerable<Type>
     }
 
     /// <summary>
+    /// Determines wheter the type is a struct (a value type but not an Enum)
+    /// </summary>
+    /// <returns></returns>
+    public TypeSelector ThatAreStruct()
+    {
+        // A Type is a struct when it is a valueType, but Enums and primitives are also
+        // valueTypes so they must be excluded
+        types = types.Where(t => t.IsValueType && !t.IsEnum
+                                    && !t.IsPrimitive && !t.IsEquivalentTo(typeof(decimal)))
+                                    .ToList();
+        return this;
+    }
+
+    /// <summary>
+    /// Determines wheter the type is not a struct
+    /// </summary>
+    /// <returns></returns>
+    public TypeSelector ThatAreNotStruct()
+    {
+        types = types.Where(t => !t.IsValueType || t.IsEnum
+                                    || t.IsPrimitive || t.IsEquivalentTo(typeof(decimal)))
+                                    .ToList();
+        return this;
+    }
+
+    /// <summary>
     /// Determines whether the type is static
     /// </summary>
     public TypeSelector ThatAreStatic()
--- a/Src/FluentAssertions/Types/TypeSelector.cs
+++ b/Src/FluentAssertions/Types/TypeSelector.cs
@@ -202,6 +202,24 @@ public class TypeSelector : IEnumerable<Type>
     }
 
     /// <summary>
+    /// Filters and returns the types that are sealed
+    /// </summary>
+    public TypeSelector ThatAreSealed()
+    {
+        types = types.Where(t => t.IsSealed).ToList();
+        return this;
+    }
+
+    /// <summary>
+    /// Filters and returns the types that are not sealed
+    /// </summary>
+    public TypeSelector ThatAreNotSealed()
+    {
+        types = types.Where(t => !t.IsSealed).ToList();
+        return this;
+    }
+
+    /// <summary>
     /// Filters and returns the types that are structs (a value type but not an Enum)
     /// </summary>
     public TypeSelector ThatAreStructs()
--- a/Src/FluentAssertions/Types/TypeSelector.cs
+++ b/Src/FluentAssertions/Types/TypeSelector.cs
@@ -184,10 +184,27 @@ public class TypeSelector : IEnumerable<Type>
     }
 
     /// <summary>
-    /// Determines wheter the type is a struct (a value type but not an Enum)
+    /// Filters and returns the types that are abstract
     /// </summary>
-    /// <returns></returns>
-    public TypeSelector ThatAreStruct()
+    public TypeSelector ThatAreAbstract()
+    {
+        types = types.Where(t => t.IsCSharpAbstract()).ToList();
+        return this;
+    }
+
+    /// <summary>
+    /// Filters and returns the types that are not abstract
+    /// </summary>
+    public TypeSelector ThatAreNotAbstract()
+    {
+        types = types.Where(t => !t.IsCSharpAbstract()).ToList();
+        return this;
+    }
+
+    /// <summary>
+    /// Filters and returns the types that are structs (a value type but not an Enum)
+    /// </summary>
+    public TypeSelector ThatAreStructs()
     {
         // A Type is a struct when it is a valueType, but Enums and primitives are also
         // valueTypes so they must be excluded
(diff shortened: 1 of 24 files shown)


docs/_pages/releases.md
```diff
@@ -10,6 +10,7 @@ sidebar:
 ## Unreleased
 
 ### What's new
+* Added `ThatAre[Not]Struct method for Types Filter  - [#2056](https://github.com/fluentassertions/fluentassertions/pull/2056)
```
IT-VBFK, line 13:
```suggestion
* Added `ThatAre[Not]Struct` method for Types Filter  - [#2056](https://github.com/fluentassertions/fluentassertions/pull/2056)
```

IT-VBFK (changes requested):

Src/FluentAssertions/Types/TypeSelector.cs
```diff
@@ -183,6 +183,32 @@ public TypeSelector ThatAreNotClasses()
         return this;
     }
 
+    /// <summary>
+    /// Determines wheter the type is a struct (a value type but not an Enum)
```
dennisdoomen, line 187: 🔧 It doesn't determine w_h_ether a type is struct. It filters them. Same with the other methods.

Tests/FluentAssertions.Specs/Types/TypeSelectorSpecs.cs
```diff
@@ -495,6 +496,42 @@ public void When_selecting_types_that_are_not_classes_it_should_return_the_corre
                 .And.Contain(typeof(NotOnlyClassesEnumeration));
         }
 
+        [Fact]
+        public void When_selecting_types_that_are_structs_it_should_return_the_correct_types()
```
dennisdoomen, line 500: 🤔 What about record structs?

dennisdoomen (changes requested):

Src/FluentAssertions/Types/TypeSelector.cs
```diff
@@ -183,6 +183,32 @@ public TypeSelector ThatAreNotClasses()
...
+    /// </summary>
+    /// <returns></returns>
+    public TypeSelector ThatAreStruct()
```
IT-VBFK, line 190: What about renaming to `ThatAreStructs`?

94sedighi: make sense

Tests/FluentAssertions.Specs/Types/TypeSelectorSpecs.cs
94sedighi: i will extend the tests for record structs and i think it should work fine with record structs as well.

docs/_pages/releases.md
```diff
@@ -11,8 +11,11 @@ sidebar:
 
 ### What's new
 * Added `ThatAre[Not]Abstract` method for filtering the types - [#2058](https://github.com/fluentassertions/fluentassertions/pull/2058)
-* Added `ThatAre[Not]Abstract` methods to `MethodInfoSelector.cs` for filtering the methods - [#2060](https://github.com/fluentassertions/fluentassertions/pull/2060)
+* Added `ThatAre[Not]Struct` method for Types Filter  - [#2056](https://github.com/fluentassertions/fluentassertions/pull/2056)
```
dennisdoomen, line 14: Typo ;-)

jnyrup:
From my comment in https://github.com/fluentassertions/fluentassertions/issues/2061#issuecomment-1338106025
> E.g. in https://github.com/fluentassertions/fluentassertions/pull/2056 you're excluding `decimal` for the check if a `Type` is to be considered a struct.
> I don't immediately know why that is, but it looks like a concern that could influence whether an API should be added and hence should be discussed in the issue instead of a PR.

In IL there is no direct concept the C# keyword `struct` but only value types which also includes structs.
https://learn.microsoft.com/en-us/dotnet/standard/base-types/common-type-system?redirectedfrom=MSDN#structures

```csharp
public struct Foo{}
```
```
.class public sequential ansi sealed beforefieldinit Foo
    extends [System.Runtime]System.ValueType
{
    .pack 0
    .size 1

} // end of class Foo
```
[SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEBmAAgGcMoBXMDUgMQggGUrb6BvAXwFgAoEqUIAmRswDCuAIblypHgKHYAdhhhQAZtLAxxEAJJqN23Qr6CyMFdQC2+gKI37iy6QAmMXDADm09cIo+gAiXr7+MAAUAJQA3EA)

So like `IsCSharpAbstract` and `IsCSharpSealed` we cannot exactly detect a 1:1 mapping between C# code written and the IL we have available.
But I think this case is more complicated, and I'm not convinced we should add this API.

Some points of the implementation
* Primitive types are not considered structs despite e.g. `float` being an alias for the struct `System.Single`
* `System.Decimal` is specially treated as a non-struct - I can only assume to behave like the other primitive types?
* The newly added `System.Half` does _not_ have special treatment. I assume because `System.Half` was added in .NET 5 after https://stackoverflow.com/a/62727198 was written?
  * This is a simple case, but then how about any other custom numeric type? E.g. [`System.Numerics`](https://learn.microsoft.com/en-us/dotnet/api/system.numerics.complex?view=net-7.0)
  * [SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmARgFgAoIgBgAIiyA6AORwFsYoBLMAZwDc1agAVeeDAEk8AMwgAKDAE8ADjAiyFsgDYQAhhgCUR4VXE9JM+UrUatAExhgeHfTpNmLVuYpXrNBQAJd1lPMQlpX1sArQBhCA5VHRgEcPNI6z87QIBZZQBRPE50kRoUem8omwAVO3p/GCNqAG9qeg7GMgBOBQASACJWfS4QehbGthGYAF8B03bOpl7BqX4ANXccGDr1MYm7ZjXNnW3d2fmzTq6VgbWLDh4MHgA3GH3J+95H57e5hao12W/Tu/AAyhhcGAsOM1hCoRgYk1/lclj0FACZmVgBAIDp6HDIThoQpzg0jPQALwAPnoiw6GCOGy2Ozs9PoADIOfQAISMtZFTjsrm8/n8B5PV4wOmAzoivlMgoARxwr3cMEkNWysQUThcbg8ALKRAAzPQNZx6PlBRxxjMgA===)
  
  
To summarize:
There are several pitfalls which makes me think this API does not pull its own weight.
Not to sound (too) grumpy, but is why I like to discuss APIs in issues before rushing to implementation.

94sedighi:
I see, there is really a problem with the half and complex types.
I totally agree and will close the PR

94sedighi:
> On the other hand, a lot of developers may not even be aware of all those special cases and just want to know which of their types in their code base are `struct`. They will never expect to find primitive types, `decimal` or `Half` in the results, so they will never run into those specific edge cases. So I still see value in having this method, provided that we add clear documentation on these edge cases.

You are right, i also didn't know about these special types that C# offer, but to be honest i'm also new in C# .NET technologies and coming from java.

jnyrup:
> They will never expect to find primitive types, decimal or Half in the results

As we don't know all future or custom numeric types, we cannot filter them out. 

To avoid any the ambiguities we could have `ThatAreValueTypes`.

dennisdoomen:
> To avoid any the ambiguities we could have `ThatAreValueTypes`.

Yes, let's do that. @94sedighi what do you think?

94sedighi:
> > To avoid any the ambiguities we could have `ThatAreValueTypes`.
> 
> Yes, let's do that. @94sedighi what do you think?

@dennisdoomen It will cover a wider range of types, but we can do it, and there are no edge cases, as far as I know the C#.

PR #2068, opened 2022-12-19
Use TheoryData

jnyrup:
After I learned about `TheoryData` in #1980 I searched for more places we could apply this to increase type-safety and remove some cruft.


dennisdoomen (approved):
Nice. I didn't know that either.

IT-VBFK:
> After I learned about TheoryData in https://github.com/fluentassertions/fluentassertions/pull/1980

Didn't know that the dotnet universe is that old 😆

PR #2070, opened 2022-12-21
Use langword instead of <c> for C# keywords

jnyrup:
[Recommended XML tags for C# documentation comments](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#see) recommends using `<see langword="keyword"/>` when referring to language keywords.

Split of from #2023


dennisdoomen (approved):
Oh nice. Does it make a difference in the way it is presented?

jnyrup:
The difference in VS:

`<c>yield</c>`
image

`<see langword="yield"/>`
image


Some stats grep'ing https://github.com/dotnet/runtime

* `<c>(null|true|false)`: 646
* `langword="(null|true|false)`: 1965

PR #2074, opened 2022-12-28
Add `Imply()` to `BooleanAssertions`

### LINKED ISSUE #1886: Extend BooleanAssertions with Implies
### Description

Add logical implication to the available boolean assertions.

That is, `a.Should().Imply(b)` asserts that `a → b`.

This would be useful in asserting, for example, that object equality implies hash-code equality without additional logic and a better built-in "because".

### Complete minimal example

Looking for something like this.

```csharp
[DataTestMethod]
[DataRow(false, false)]
[DataRow(false, true)]
// failure case: [DataRow(true, false)]
[DataRow(true, true)]
public void A_Implies_B(bool a, bool b) => a.Should().Imply(b);
```

Note this is logically equivalent to `!a || b`.

### Additional Information

Also volunteering to add this.

IT-VBFK:
Closes #1886 

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Primitives/BooleanAssertions.cs
+++ b/Src/FluentAssertions/Primitives/BooleanAssertions.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 using FluentAssertions.Execution;
 
 namespace FluentAssertions.Primitives;
@@ -129,12 +130,15 @@ public class BooleanAssertions<TAssertions>
     /// <param name="becauseArgs">
     /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
     /// </param>
-    public AndConstraint<TAssertions> Imply(bool implicator, string because = "", params object[] becauseArgs)
+    public AndConstraint<TAssertions> Imply(bool implicator,
+        [CallerArgumentExpression("implicator")] string implicatorMessage = "",
+        string because = "",
+        params object[] becauseArgs)
     {
         Execute.Assertion
             .ForCondition(Subject is not null)
             .BecauseOf(because, becauseArgs)
-            .WithExpectation("Expected {context:boolean} ({0}) to imply {1} ({2}){reason}, ", Subject, nameof(implicator), implicator)
+            .WithExpectation("Expected {context:boolean} ({0}) to imply {1} ({2}){reason}, ", Subject, implicatorMessage, implicator)
             .FailWith("but found null.")
             .Then
             .ForCondition(!Subject.Value || implicator)
--- a/Src/FluentAssertions/Primitives/BooleanAssertions.cs
+++ b/Src/FluentAssertions/Primitives/BooleanAssertions.cs
@@ -118,15 +118,26 @@ public class BooleanAssertions<TAssertions>
         return new AndConstraint<TAssertions>((TAssertions)this);
     }
 
-    public AndConstraint<TAssertions> Imply(bool expectation, string because = "", params object[] becauseArgs)
+    /// <summary>
+    /// Asserts that the value implies the specified <paramref name="implicator"/> value.
+    /// </summary>
+    /// <param name="implicator">The second value for the implication</param>
+    /// <param name="because">
+    /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+    /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+    /// </param>
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
+    /// </param>
+    public AndConstraint<TAssertions> Imply(bool implicator, string because = "", params object[] becauseArgs)
     {
         Execute.Assertion
             .ForCondition(Subject is not null)
             .BecauseOf(because, becauseArgs)
-            .WithExpectation("Expected {context:boolean} ({0}) to imply {1} ({2}){reason}, ", Subject, nameof(expectation), expectation)
+            .WithExpectation("Expected {context:boolean} ({0}) to imply {1} ({2}){reason}, ", Subject, nameof(implicator), implicator)
             .FailWith("but found null.")
             .Then
-            .ForCondition(!Subject.Value || expectation)
+            .ForCondition(!Subject.Value || implicator)
             .FailWith("but it did not.")
             .Then
             .ClearExpectation();
--- a/Src/FluentAssertions/Primitives/BooleanAssertions.cs
+++ b/Src/FluentAssertions/Primitives/BooleanAssertions.cs
@@ -120,10 +120,16 @@ public class BooleanAssertions<TAssertions>
 
     public AndConstraint<TAssertions> Imply(bool expectation, string because = "", params object[] becauseArgs)
     {
-        Execute.Assertion            
-            .ForCondition(!Subject.Value || expectation)
+        Execute.Assertion
+            .ForCondition(Subject is not null)
             .BecauseOf(because, becauseArgs)
-            .FailWith("Expected {context:boolean} ({0}) to imply {1} ({2}){reason}, but it did not.", Subject, nameof(expectation), expectation);
+            .WithExpectation("Expected {context:boolean} ({0}) to imply {1} ({2}){reason}, ", Subject, nameof(expectation), expectation)
+            .FailWith("but found null.")
+            .Then
+            .ForCondition(!Subject.Value || expectation)
+            .FailWith("but it did not.")
+            .Then
+            .ClearExpectation();
         
         return new AndConstraint<TAssertions>((TAssertions)this);
     }
--- a/Src/FluentAssertions/Primitives/BooleanAssertions.cs
+++ b/Src/FluentAssertions/Primitives/BooleanAssertions.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Diagnostics;
-using System.Runtime.CompilerServices;
 using FluentAssertions.Execution;
 
 namespace FluentAssertions.Primitives;
@@ -122,7 +121,7 @@ public class BooleanAssertions<TAssertions>
     /// <summary>
     /// Asserts that the value implies the specified <paramref name="consequent"/> value.
     /// </summary>
-    /// <param name="consequent">The right hand side for the implication</param>
+    /// <param name="consequent">The right hand side of the implication</param>
     /// <param name="because">
     /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
     /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
--- a/Src/FluentAssertions/Primitives/BooleanAssertions.cs
+++ b/Src/FluentAssertions/Primitives/BooleanAssertions.cs
@@ -123,6 +123,7 @@ public class BooleanAssertions<TAssertions>
     /// Asserts that the value implies the specified <paramref name="consequent"/> value.
     /// </summary>
     /// <param name="consequent">The right hand side for the implication</param>
+    /// <remarks>If you want to use the `because` / `becauseArgs` construct, it is **necessary** to use the caller argument expression, because this would lead to unexpected failure messages, when omitted!</remarks>
     /// <param name="because">
     /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
     /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
--- a/Src/FluentAssertions/Primitives/BooleanAssertions.cs
+++ b/Src/FluentAssertions/Primitives/BooleanAssertions.cs
@@ -123,7 +123,6 @@ public class BooleanAssertions<TAssertions>
     /// Asserts that the value implies the specified <paramref name="consequent"/> value.
     /// </summary>
     /// <param name="consequent">The right hand side for the implication</param>
-    /// <remarks>If you want to use the `because` / `becauseArgs` construct, it is **necessary** to use the caller argument expression, because this would lead to unexpected failure messages, when omitted!</remarks>
     /// <param name="because">
     /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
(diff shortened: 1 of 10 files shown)


Src/FluentAssertions/Primitives/BooleanAssertions.cs
```diff
@@ -118,6 +118,16 @@ public AndConstraint<TAssertions> NotBe(bool unexpected, string because = "", pa
         return new AndConstraint<TAssertions>((TAssertions)this);
     }
 
+    public AndConstraint<TAssertions> Imply(bool expectation, string because = "", params object[] becauseArgs)
+    {
+        Execute.Assertion            
+            .ForCondition(!Subject.Value || expectation)
+            .BecauseOf(because, becauseArgs)
+            .FailWith("Expected {context:boolean} ({0}) to imply {1} ({2}){reason}, but it did not.", Subject, nameof(expectation), expectation);
```
IT-VBFK, line 126: I am not very happy with this solution, but currently the only possible solution..

Maybe we should keep this in mind until `[CallerArgumentExpression]`

docs/_pages/booleans.md
```diff
@@ -23,3 +23,9 @@ Obviously the above assertions also work for nullable booleans, but if you reall
 theBoolean.Should().NotBeFalse();
 theBoolean.Should().NotBeTrue();
 ```
+
+Implication: see [here](https://mathworld.wolfram.com/Implies.html)
```
IT-VBFK, line 27: Not sure if this is worthy to be mentioned in the docs, since it is a niche

Src/FluentAssertions/Primitives/BooleanAssertions.cs
```diff
@@ -118,6 +118,16 @@ public AndConstraint<TAssertions> NotBe(bool unexpected, string because = "", pa
         return new AndConstraint<TAssertions>((TAssertions)this);
     }
 
+    public AndConstraint<TAssertions> Imply(bool expectation, string because = "", params object[] becauseArgs)
```
jnyrup, line 121: Don't we use `expected` most places?

Tests/FluentAssertions.Specs/Primitives/BooleanAssertionSpecs.cs
```diff
@@ -6,151 +6,198 @@ namespace FluentAssertions.Specs.Primitives;
...
+            // Act / Assert
+            a.Should().Imply(b);
+        }
+
+        [Theory]
+        [MemberData(nameof(NonPassingImplications), MemberType = typeof(BooleanAssertionSpecs.Imply))]
+        public void A_does_not_imply_B(bool a, bool b)
+        {
+            // Act
+            Action act = () => a.Should().Imply(b);
```
jnyrup, line 196: Pass in a `because` and `becauseArgs` and verify them.

Tests/FluentAssertions.Specs/Primitives/BooleanAssertionSpecs.cs
```diff
@@ -6,151 +6,198 @@ namespace FluentAssertions.Specs.Primitives;
...
+        };
+
+        public static TheoryData<bool, bool> NonPassingImplications => new()
+        {
+            { true, false }
+        };
+
+        [Theory]
+        [MemberData(nameof(PassingImplications), MemberType = typeof(BooleanAssertionSpecs.Imply))]
+        public void A_implies_B(bool a, bool b)
```
jnyrup, line 185: How about `Subject_implies_expected` and similarly renaming `a` and `b`?

Src/FluentAssertions/Primitives/BooleanAssertions.cs
jnyrup: How about
`"Expected {context:boolean} to imply {0}{reason}, but found {1}.", expectation, Subject`

Src/FluentAssertions/Primitives/BooleanAssertions.cs
jnyrup, line 121: This is missing an XML summary.

jnyrup (changes requested):

Src/FluentAssertions/Primitives/BooleanAssertions.cs
IT-VBFK: True

Tests/FluentAssertions.Specs/Primitives/BooleanAssertionSpecs.cs
IT-VBFK: Funny thing: I first had this solution and swithc to the current one 😆 

But sure.. I can rename it

Src/FluentAssertions/Primitives/BooleanAssertions.cs
IT-VBFK: Hmm.. This would result in e.g. `"Expected subject to imply True, but found False.", expectation, Subject`

This does not help IMO.

Src/FluentAssertions/Primitives/BooleanAssertions.cs
IT-VBFK: But: this is not the expectation.. this is the second parameter for the implication.. so "operand" would fit better IMO

Tests/FluentAssertions.Specs/Primitives/BooleanAssertionSpecs.cs
IT-VBFK: Same thought as in https://github.com/fluentassertions/fluentassertions/pull/2074#discussion_r1058316427

Src/FluentAssertions/Primitives/BooleanAssertions.cs
IT-VBFK: I have called it "implicator" by now.. Feel free to disagree :D

Tests/FluentAssertions.Specs/Primitives/BooleanAssertionSpecs.cs
```diff
@@ -6,151 +6,200 @@ namespace FluentAssertions.Specs.Primitives;
...
+        public static TheoryData<bool?, bool> NonPassingImplications => new()
+        {
+            { null, true },
+            { null, false },
+            { true, false }
+        };
+
+        [Theory]
+        [MemberData(nameof(PassingImplications), MemberType = typeof(BooleanAssertionSpecs.Imply))]
+        public void Subject_implies_implicator(bool? subject, bool implicator)
```
jnyrup, line 187: Apparently the left hand side and right hand side of an implication are named "antecedent" and "consequent", respectively

https://en.wikipedia.org/wiki/Material_conditional

Tests/FluentAssertions.Specs/Primitives/BooleanAssertionSpecs.cs
```diff
@@ -6,151 +6,200 @@ namespace FluentAssertions.Specs.Primitives;
...
+
+        public static TheoryData<bool?, bool> NonPassingImplications => new()
+        {
+            { null, true },
+            { null, false },
+            { true, false }
+        };
+
+        [Theory]
+        [MemberData(nameof(PassingImplications), MemberType = typeof(BooleanAssertionSpecs.Imply))]
```
jnyrup, line 186: `MemberType` can be left out.
If you got it from `StreamAssertionSpecs` it is necessary there as e.g. `GetLengthExceptions` resides in the parent type of the tests referring to it.

Tests/FluentAssertions.Specs/Primitives/BooleanAssertionSpecs.cs
IT-VBFK: Ok.. will change that (at least the right hand side)

Tests/FluentAssertions.Specs/Primitives/BooleanAssertionSpecs.cs
IT-VBFK: Ok.. funnily enough it didn't work yesterday without it

Src/FluentAssertions/Primitives/BooleanAssertions.cs
```diff
@@ -118,6 +119,37 @@ public AndConstraint<TAssertions> NotBe(bool unexpected, string because = "", pa
...
+    public AndConstraint<TAssertions> Imply(bool consequent,
+        string because = "",
+        params object[] becauseArgs)
+    {
+        bool? antecedent = Subject;
+
+        Execute.Assertion
+            .ForCondition(antecedent is not null)
+            .BecauseOf(because, becauseArgs)
+            .WithExpectation($"Expected {{context:antecedent}} ({{0}}) to imply consequent ({{1}}){{reason}}, ", antecedent, consequent)
```
jnyrup, line 142:
```suggestion
            .WithExpectation("Expected {context:antecedent} ({0}) to imply consequent ({1}){reason}, ", antecedent, consequent)
```

Src/FluentAssertions/Primitives/BooleanAssertions.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
```
jnyrup, line 3:
```suggestion
```

Tests/FluentAssertions.Specs/Primitives/BooleanAssertionSpecs.cs
```diff
@@ -6,151 +6,200 @@ namespace FluentAssertions.Specs.Primitives;
...
+        public static TheoryData<bool?, bool> NonPassingImplications => new()
+        {
+            { null, true },
+            { null, false },
+            { true, false }
+        };
+
+        [Theory]
+        [MemberData(nameof(PassingImplications))]
+        public void Subject_implies_implicator(bool? antecedent, bool consequent)
```
jnyrup, line 187:
```suggestion
        public void antecedent_implies_consequent(bool? antecedent, bool consequent)
```

Tests/FluentAssertions.Specs/Primitives/BooleanAssertionSpecs.cs
```diff
@@ -6,151 +6,200 @@ namespace FluentAssertions.Specs.Primitives;
...
+        {
+            // Act / Assert
+            antecedent.Should().Imply(consequent);
+        }
+
+        [Theory]
+        [MemberData(nameof(NonPassingImplications))]
+        public void Subject_does_not_imply_implicator(bool? antecedent, bool consequent)
```
jnyrup, line 195:
```suggestion
        public void antecedent_does_not_imply_consequent(bool? antecedent, bool consequent)
```

Src/FluentAssertions/Primitives/BooleanAssertions.cs
```diff
@@ -118,6 +119,37 @@ public AndConstraint<TAssertions> NotBe(bool unexpected, string because = "", pa
         return new AndConstraint<TAssertions>((TAssertions)this);
     }
 
+    /// <summary>
+    /// Asserts that the value implies the specified <paramref name="consequent"/> value.
+    /// </summary>
+    /// <param name="consequent">The right hand side for the implication</param>
```
jnyrup, line 125:
```suggestion
    /// <param name="consequent">The right hand side of the implication</param>
```

jnyrup (approved):

Tests/FluentAssertions.Specs/Primitives/BooleanAssertionSpecs.cs
```diff
@@ -6,151 +6,200 @@ namespace FluentAssertions.Specs.Primitives;
...
+
+        public static TheoryData<bool?, bool> NonPassingImplications => new()
+        {
+            { null, true },
+            { null, false },
+            { true, false }
+        };
+
+        [Theory]
+        [MemberData(nameof(PassingImplications))]
```
dennisdoomen, line 186: 🤔 I think using `[InlineData]` is much clearer in this particular case and keeps the data with the test. Is there a reason why we can't use those?

IT-VBFK: There is no reason to not use `[InlineData]`, I just like the feature of `TheoryData`.. 

And I think it is (especially for new contributors) easier, if there is one "design" how to structure tests. 

But if @jnyrup agrees to the other approach I am willing to change that..

dennisdoomen: > And I think it is (especially for new contributors) easier, if there is one "design" how to structure tests.

I'm in the "use the right tool for the right job at the right time" camp ;-).

jnyrup: I had _completely_ forgotten about the simpler `InlineData` when reviewing this 😄

My thoughts on when to pick which:
* Aim for `[InlineData]` as it keeps the parameters closest to the method body.
* If the parameters are not compile-constants, defer to `TheoryData` defined right below the method using it.
* If the parameters have a _good_ reason to be shared between multiple tests, `TheoryData` might do the best job. There's a thin line here between DRY and over-DRY and we tend to prefer duplication over coupling between tests, that's why you'll see very few "helper" methods in the test suite.

Tests/FluentAssertions.Specs/Primitives/BooleanAssertionSpecs.cs
```diff
@@ -6,151 +6,190 @@ namespace FluentAssertions.Specs.Primitives;
...
-    {
-        // Act
-        Action action = () => true.Should().Equals(true);
-
-        // Assert
-        action.Should().Throw<NotSupportedException>().WithMessage("Equals is not part of Fluent Assertions. Did you mean Be() instead?");
+        [Theory]
+        [InlineData(false, false )]
+        [InlineData(false, true )]
+        [InlineData(true, true )]
```
jnyrup, line 174:
```suggestion
        [InlineData(false, false)]
        [InlineData(false, true)]
        [InlineData(true, true)]
```

Tests/FluentAssertions.Specs/Primitives/BooleanAssertionSpecs.cs
IT-VBFK: Ok... thanks for clarification :)

dennisdoomen (approved):
Great work!

PR #2081, opened 2023-01-01
Cleanups

jnyrup:
Various small cleanups.

PolySharp polyfills `Index` and `Range` for the net472 target.

Pattern matching vs (lifted) `==` is a complete minefield as they generate better/worse code in different cases.

[SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+B6DACAvL7AdwEsA7AZwFgAocgFygFcw7sBlBsgcxoG8bsg7AAEAzNmAQIAG2wBRAI6MAhtPIAKYQEYADNnpQAlHgB8+hgH4AdABkYpLnQAWeAjoDcAoV8FZsTujoAB3IQLC5iZ0ZgK0gAWwwAEwg6Uhg6DCgIcmkAT1IMYnJyRhhyDCQAVh0ATi0fEXFJGWwABWVAmChSAFkOsCdNXXMjUxHrOwdnbCLsDxoAXxoaPwByItWiMipaBmZWAGEnZSg+BrEJKVlFFTV1AZOLbDBjXDMwV2xVgEFVz2ohNhzk0rm0OnQur1+oMHlAni8xh9Zj8/otltQ/AA5CDYAD6YGkMBOuK2pDSp12TBY2AAkqQ6GcAUILs1rkpVBoyHQnsRXmZiJ8tP9AcDLi12p1un06AN1FyeXyZjNyNghWjqDQgA===)

The reduced CC is either because:
* `if (comparands.Expectation is not Array expectationAsArray || expectationAsArray?.Rank == 1)` had a redundant null check.
  * The file is still 100% line and branch covered.
* `QuotesParsingStrategy.cs` has fewer lines of code, but coverage of that files remains unchanged.

### DIFF
--- a/Src/FluentAssertions/CallerIdentification/MultiLineCommentParsingStrategy.cs
+++ b/Src/FluentAssertions/CallerIdentification/MultiLineCommentParsingStrategy.cs
@@ -11,7 +11,7 @@ internal class MultiLineCommentParsingStrategy : IParsingStrategy
     {
         if (isCommentContext)
         {
-            var isEndOfMultilineComment = symbol == '/' && commentContextPreviousChar == '*';
+            var isEndOfMultilineComment = symbol is '/' && commentContextPreviousChar is '*';
             if (isEndOfMultilineComment)
             {
                 isCommentContext = false;
@@ -26,9 +26,9 @@ internal class MultiLineCommentParsingStrategy : IParsingStrategy
         }
 
         var isStartOfMultilineComment =
-            symbol == '*'
+            symbol is '*'
             && statement.Length > 0
-            && statement[^1] == '/';
+            && statement[^1] is '/';
         if (isStartOfMultilineComment)
         {
             statement.Remove(statement.Length - 1, 1);
--- a/Src/FluentAssertions/CallerIdentification/MultiLineCommentParsingStrategy.cs
+++ b/Src/FluentAssertions/CallerIdentification/MultiLineCommentParsingStrategy.cs
@@ -28,7 +28,7 @@ internal class MultiLineCommentParsingStrategy : IParsingStrategy
         var isStartOfMultilineComment =
             symbol == '*'
             && statement.Length > 0
-            && statement[statement.Length - 1] == '/';
+            && statement[^1] == '/';
         if (isStartOfMultilineComment)
         {
             statement.Remove(statement.Length - 1, 1);
--- a/Src/FluentAssertions/CallerIdentification/QuotesParsingStrategy.cs
+++ b/Src/FluentAssertions/CallerIdentification/QuotesParsingStrategy.cs
@@ -1,5 +1,4 @@
-﻿using System.Linq;
-using System.Text;
+﻿using System.Text;
 
 namespace FluentAssertions.CallerIdentification;
 
@@ -54,17 +53,7 @@ internal class QuotesParsingStrategy : IParsingStrategy
 
     private bool IsVerbatim(StringBuilder statement)
     {
-        return
-            statement.Length >= 1
-            &&
-                new[]
-                {
-                    "$@",
-                    "@$",
-                }
-                .Any(verbatimStringOpener =>
-                    previousChar == verbatimStringOpener[1]
-                    && statement[statement.Length - 1] == verbatimStringOpener[1]
-                    && statement[statement.Length - 2] == verbatimStringOpener[0]);
+        return (previousChar == '@' && statement.Length >= 2 && statement[^2] == '$' && statement[^1] == '@')
+            || (previousChar == '$' && statement.Length >= 2 && statement[^2] == '@' && statement[^1] == '$');
     }
 }
--- a/Src/FluentAssertions/CallerIdentification/QuotesParsingStrategy.cs
+++ b/Src/FluentAssertions/CallerIdentification/QuotesParsingStrategy.cs
@@ -10,7 +10,7 @@ internal class QuotesParsingStrategy : IParsingStrategy
 
     public ParsingState Parse(char symbol, StringBuilder statement)
     {
-        if (symbol == '"')
+        if (symbol is '"')
         {
             if (isQuoteContext)
             {
@@ -53,7 +53,7 @@ internal class QuotesParsingStrategy : IParsingStrategy
 
     private bool IsVerbatim(StringBuilder statement)
     {
-        return (previousChar == '@' && statement.Length >= 2 && statement[^2] == '$' && statement[^1] == '@')
-            || (previousChar == '$' && statement.Length >= 2 && statement[^2] == '@' && statement[^1] == '$');
+        return (previousChar is '@' && statement.Length >= 2 && statement[^2] is '$' && statement[^1] is '@')
+            || (previousChar is '$' && statement.Length >= 2 && statement[^2] is '@' && statement[^1] is '$');
     }
 }
--- a/Src/FluentAssertions/Equivalency/MultiDimensionalArrayEquivalencyStep.cs
+++ b/Src/FluentAssertions/Equivalency/MultiDimensionalArrayEquivalencyStep.cs
@@ -13,7 +13,7 @@ internal class MultiDimensionalArrayEquivalencyStep : IEquivalencyStep
 {
     public EquivalencyResult Handle(Comparands comparands, IEquivalencyValidationContext context, IEquivalencyValidator nestedValidator)
     {
-        if (comparands.Expectation is not Array expectationAsArray || expectationAsArray?.Rank is 1)
+        if (comparands.Expectation is not Array expectationAsArray || expectationAsArray.Rank == 1)
         {
             return EquivalencyResult.ContinueWithNext;
         }
--- a/Src/FluentAssertions/Equivalency/MultiDimensionalArrayEquivalencyStep.cs
+++ b/Src/FluentAssertions/Equivalency/MultiDimensionalArrayEquivalencyStep.cs
@@ -13,7 +13,7 @@ internal class MultiDimensionalArrayEquivalencyStep : IEquivalencyStep
 {
     public EquivalencyResult Handle(Comparands comparands, IEquivalencyValidationContext context, IEquivalencyValidator nestedValidator)
     {
-        if (comparands.Expectation is not Array expectationAsArray || expectationAsArray?.Rank == 1)
+        if (comparands.Expectation is not Array expectationAsArray || expectationAsArray?.Rank is 1)
         {
             return EquivalencyResult.ContinueWithNext;
         }
--- a/Src/FluentAssertions/CallerIdentification/SemicolonParsingStrategy.cs
+++ b/Src/FluentAssertions/CallerIdentification/SemicolonParsingStrategy.cs
@@ -6,7 +6,7 @@ internal class SemicolonParsingStrategy : IParsingStrategy
 {
     public ParsingState Parse(char symbol, StringBuilder statement)
     {
-        if (symbol == ';')
+        if (symbol is ';')
         {
             statement.Clear();
             return ParsingState.Done;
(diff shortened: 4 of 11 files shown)


Src/FluentAssertions/CallerIdentification/QuotesParsingStrategy.cs
```diff
@@ -54,17 +53,7 @@ public void NotifyEndOfLineReached()
...
-                {
-                    "$@",
-                    "@$",
-                }
-                .Any(verbatimStringOpener =>
-                    previousChar == verbatimStringOpener[1]
-                    && statement[statement.Length - 1] == verbatimStringOpener[1]
-                    && statement[statement.Length - 2] == verbatimStringOpener[0]);
+        return (previousChar is '@' && statement.Length >= 2 && statement[^2] is '$' && statement[^1] is '@')
+            || (previousChar is '$' && statement.Length >= 2 && statement[^2] is '@' && statement[^1] is '$');
```
jnyrup, line 57: |  Method | previousChar | statement |       Mean |     Error |    StdDev | Ratio | Allocated |
|-------- |------------- |---------- |-----------:|----------:|----------:|------:|----------:|
| develop |         null |        "" |  5.7070 ns | 0.0368 ns | 0.0344 ns |  1.00 |      32 B |
|      PR |         null |        "" |  1.0103 ns | 0.0094 ns | 0.0088 ns |  0.18 |         - |
|         |              |           |            |           |           |       |           |
| develop |         null |        $@ | 38.2595 ns | 0.0580 ns | 0.0484 ns |  1.00 |     168 B |
|      PR |         null |        $@ |  1.0502 ns | 0.0056 ns | 0.0053 ns |  0.03 |         - |
|         |              |           |            |           |           |       |           |
| develop |            $ |        "" |  5.4029 ns | 0.0085 ns | 0.0075 ns |  1.00 |      32 B |
|      PR |            $ |        "" |  1.1945 ns | 0.0082 ns | 0.0069 ns |  0.22 |         - |
|         |              |           |            |           |           |       |           |
| develop |            $ |        $@ | 42.4361 ns | 0.4080 ns | 0.3817 ns |  1.00 |     168 B |
|      PR |            $ |        $@ |  2.5188 ns | 0.0063 ns | 0.0059 ns |  0.06 |         - |
|         |              |           |            |           |           |       |           |
| develop |            @ |        "" |  5.4685 ns | 0.0397 ns | 0.0372 ns |  1.00 |      32 B |
|      PR |            @ |        "" |  1.2068 ns | 0.0064 ns | 0.0057 ns |  0.22 |         - |
|         |              |           |            |           |           |       |           |
| develop |            @ |        $@ | 38.2881 ns | 0.4913 ns | 0.4595 ns |  1.00 |     168 B |
|      PR |            @ |        $@ |  3.5240 ns | 0.0278 ns | 0.0247 ns |  0.09 |         - |

eNeRGy164: Wow

dennisdoomen: Why is it so much faster?

jnyrup: [SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABADAAgwRgDoAZASwDsBHAbgFgAobPIgFRgQBc76HyBDALYwAzgAc+YGDgBiAGwCuMchwCCw4TCgdSEcsMIBhPrNmaAkgBMl2gGakwfbbu4MKHTf1l4ATDgCK8hDuwgAKfFDCFADmAMocUI4wUQCeDADeDDhZOKJQpABuiThgABbhAPw5sPk68sIGZVAu9NlVBUXAEBBeAPKyFgAUcXnkUQBC8qT9mjjCHIlCygCUmdkZLa3ZGADsq5ubcwvWJEpRHCU4AHwAvDj4e/vZAGRPD4+b5DAA7gDaALpvd5rQFAzYAIgAJBgMGCADQg0HZMHQiFwhHvAC+6MehBU5GSA3ymmAjlIAmG0R6oiUM2ul2xoNyMBqEDqDXCOGutyJUBJ2nJ8Up1M+UB++ABG0Rjxes3m7kWHB+h3lx2Ip3OODgdz+nO5xNJApGUSpNNF4oZQJlypgCqVcptqvVFy13h1XJwPL5ZIpoxNIp+WD+S24rSxkpwD1y7XcOE63RwADlvkNBaMJlMrFBZUdlg91u8djgBkyWWzGjhSMIcAByaHVnBW+0Kk6jDU3HC+Rs5xUAPVdFar1Yh9a7KuUPx74oHNbrK3D7zQaCLJdq9XLlZrw4bT2zY44LbOF3bnZ31ttfZ1G9rGBHp6b1gnU6vw+DDzDYaAA=)


image

jnyrup: It could even be a _tiny_ bit faster using `previousChar.GetValueOrDefault() is '@'`, as `previousChar is '@'` includes a redundant `.HasValue` check, but I found this code cleaner.

https://github.com/dotnet/roslyn/issues/60413 and https://github.com/dotnet/roslyn/issues/52629 track removing redundant `.HasValue` checks.

jnyrup: This change also fix a (potential) bug, since the old code checked for `statement.Length >= 1` but indexed into `statement.Length - 2`.

Src/FluentAssertions/CallerIdentification/MultiLineCommentParsingStrategy.cs
```diff
@@ -11,7 +11,7 @@ public ParsingState Parse(char symbol, StringBuilder statement)
     {
         if (isCommentContext)
         {
-            var isEndOfMultilineComment = symbol == '/' && commentContextPreviousChar == '*';
+            var isEndOfMultilineComment = symbol is '/' && commentContextPreviousChar is '*';
```
dennisdoomen, line 14: Why changing `==` to `is`? What's the added value? Or is this a matter of taste?

Src/FluentAssertions/Equivalency/MultiDimensionalArrayEquivalencyStep.cs
```diff
@@ -13,7 +13,7 @@ internal class MultiDimensionalArrayEquivalencyStep : IEquivalencyStep
 {
     public EquivalencyResult Handle(Comparands comparands, IEquivalencyValidationContext context, IEquivalencyValidator nestedValidator)
     {
-        if (comparands.Expectation is not Array expectationAsArray || expectationAsArray?.Rank is 1)
+        if (comparands.Expectation is not Array { Rank: not 1 } expectationAsArray)
```
dennisdoomen, line 16: Although Rider suggested the same fix, I really find it very hard to read. It's almost like a double negative.

Src/FluentAssertions/CallerIdentification/MultiLineCommentParsingStrategy.cs
IT-VBFK: I think @jnyrup first [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+B6DACAvL7AdwEsA7AZwFgAocgFygFcw7sBlBsgcxoG8bsg7AAEAzNmAQIAG2wBRAI6MAhtPIAKYQEYADNnpQAlHgB8+hgH4AdABkYpLnQAWeAjoDcAoV8FZsTujoAB3IQLC5iZ0ZgK0gAWwwAEwg6Uhg6DCgIcmkAT1IMYnJyRhhyDCQAVh0ATi0fEXFJGWwABWVAmChSAFkOsCdNXXMjUxHrOwdnbCLsDxoAXxoaPwByItWiMipaBmZWAGEnZSg+BrEJKVlFFTV1AZOLbDBjXDMwV2xVgEFVz2ohNhzk0rm0OnQur1+oMHlAni8xh9Zj8/otltQ/AA5CDYAD6YGkMBOuK2pDSp12TBY2AAkqQ6GcAUILs1rkpVBoyHQnsRXmZiJ8tP9AcDLi12p1un06AN1FyeXyZjNyNghWjqDQgA===) analysis should give a hint :)

dennisdoomen: Actually, it doesn't the `String`/`Equals` version looks simpler (and supposedly would be faster).

jnyrup: `commentContextPreviousChar` is a `char?` which is the second example in the SharpLab example.

For a plain `char` like `symbol` there's no difference in the generated code.

This [SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+BLAdgFxlLAQwBsACAAQCZSBZAV2JwwBlsYBhCAWy5lwAVCUAM7YA5gGUcUQnjEBPALAAoAN4rSm0gAcoGAG6yYpYBAhkMwzjz45OuGAhwBuDVt0GjpMAAshAfm9uXlx7PCd+WH0MCDorPyhXZS1SN01yAGYTMzIASWEAUSwAEwB5ADN6RgxiNmsQnAB9UuJigApfIVJheS5TYgBKNNJ1ZJStDHLSNst62zDHHCGx8ZHh1fSAdm7e/tIAXn3SAHIAemPSADJLoJtQiAcIqJi49gSDo+OAKmOkjdIAL7rLTArakcokYQwP5aIErVLwzLZcykfJFMqVBhMWpYDjBWyNAByMAA7h13j0+uZlilRv9JtNZvj7o8lqC1vD/uRtpS9pYTucrjdIHc7A9wjhIjBorF4l1+d9fuy4f92dzwZDocM4QCgA==)

Show the difference for a simplified version of this method.

Src/FluentAssertions/Equivalency/MultiDimensionalArrayEquivalencyStep.cs
jnyrup: I agree, went overboard with this.
reverted and only removed the redundant null check

Src/FluentAssertions/CallerIdentification/QuotesParsingStrategy.cs
jnyrup: Also for inspiration, with C#11 this can utilize [list patterns](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/patterns#list-patterns) and be written as

```cs
return (previousChar is '@' && statement is [.., '$', '@'])
    || (previousChar is '$' && statement is [.., '@', '$']);
```

The only _microscopic_ downside is that as far as I can tell, it will always generate a null check on `statement` which reduces the CC from 100%.

Src/FluentAssertions/CallerIdentification/MultiLineCommentParsingStrategy.cs
dennisdoomen: So `is` does result in a slightly faster implementation. Not sure I understand why there's a difference in the generated code though.

dennisdoomen (approved):

Src/FluentAssertions/CallerIdentification/MultiLineCommentParsingStrategy.cs
jnyrup: > Not sure I understand why there's a difference in the generated code though.

Normally I would guess that a [peephole optimization](https://en.wikipedia.org/wiki/Peephole_optimization) simply hasn't been implemented, but since it's the newer syntax that generates better code, I would put my money on that because an [optimizing compiler](https://en.wikipedia.org/wiki/Optimizing_compiler) might need to stay conservative to preserve correctness, there is some slight difference between the two operators which it hard to impossible to optimize in the general case.

PR #2076, opened 2022-12-29
Improve `Guard` helper

jnyrup:
Solves https://github.com/fluentassertions/fluentassertions/pull/1100#discussion_r310353598

PR #2087, opened 2023-01-05
Bump required .NET SDK to v7

jnyrup:
In #2076 I began using `nameof()` within an attribute which is technically a [C#11 feature](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/extended-nameof-scope), but seems to work as long as you use C#11 compatible SDK.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,6 +28,7 @@ jobs:
           2.1.x
           3.1.x
           6.0.x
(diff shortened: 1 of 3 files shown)


IT-VBFK:
Cool that the bugfix for bumping this to 7.0.101 is already merged :D

dennisdoomen (approved):

PR #2088, opened 2023-01-06
Add Building and Testing categories

jnyrup:
Add two new categories for the autogenerated release notes template.

A bit unsure if the titles are descriptive enough 🤔

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -24,6 +24,12 @@ changelog:
     - title: Documentation
       labels:
         - "documentation"
+    - title: Building
+      labels:
+        - "building"
+    - title: Testing
+      labels:
+        - "testing"
     - title: Others
       labels: 
         - "*"


IT-VBFK:
What about "building / composition"?

jnyrup:
What does composition mean in this context?

To be more specific about my question.
Should it be more explicit that the two categories relates to building and testing FA and hence should not affect code consuming FA.

IT-VBFK:
Ahh
Is it (from a consumer of FA) really necessary to know what changed on building and testing? I think not really first prio when reading the release notes.

You know what I mean?

jnyrup:
Just to be clear, we have two sets of release notes.

1) https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/releases.md which should list everything affecting consumers.

2) https://github.com/fluentassertions/fluentassertions/releases which is an autogenerated list of every PR between two releases.

This PR only affects 2) and is about moving more PRs from the fallback category "Other" and into named categories.

IT-VBFK:
Ok, then ignore all my noise here 🙈 

Sorry..

jnyrup:
No need to be sorry, I appreciate your inputs and the extra set of eyes ❤️

IT-VBFK:
Maybe add a description to the label and link to [here](https://github.com/fluentassertions/fluentassertions/labels) somehow?

dennisdoomen:
🤔 What does "testing" mean? And why do we even care to mention them separately. In the end, it's just internal improvements, right?

dennisdoomen:
I don't see much value in those, but if you do, knock yourself off ;-)

PR #2086, opened 2023-01-04
Add `cSpell` to NUKE build pipeline

IT-VBFK:
This ref. and finishes #2017 

## IMPORTANT 

* [ ] If the PR touches the public API, the changes have been approved in a separate issue with the "api-approved" label.
* [x] The code complies with the [Coding Guidelines for C#](https://www.csharpcodingguidelines.com/).
* [ ] The changes are covered by unit tests which follow the Arrange-Act-Assert syntax and the naming conventions such as is used [in these tests](../tree/develop/Tests/FluentAssertions.Equivalency.Specs/MemberMatchingSpecs.cs#L51-L430).
* [ ] If the PR adds a feature or fixes a bug, please update [the release notes](../tree/develop/docs/_pages/releases.md) with a functional description that explains what the change means to consumers of this library, which are published on the [website](https://fluentassertions.com/releases).
* [ ] If the PR changes the public API the changes needs to be included by running [AcceptApiChanges.ps1](../tree/develop/AcceptApiChanges.ps1) or [AcceptApiChanges.sh](../tree/develop/AcceptApiChanges.sh).
* [ ] If the PR affects [the documentation](../tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).
  * [x] Please also run `./build.sh --target spellcheck` or `.\build.ps1 --target spellcheck` before pushing and check the good outcome

PR #2089, opened 2023-01-07
Optimize NUKE spell check

IT-VBFK:
## IMPORTANT 

With friendly support from https://github.com/nuke-build/nuke/discussions/1098#discussioncomment-4616405 :)

This is a follow-up for #2086. Now NUKE checks if changes are made inside the `docs/` folder and running spell check depending on that.

~~Please note: this is only a first rough sketch how it can look like. So discussion is needed if this change is really necessary.~~

The only (possible) drawback is: You have to commit at least one change in the `docs/` folder.


Build/Build.cs
```diff
@@ -258,11 +272,36 @@ class Build : NukeBuild
...
     Target SpellCheck => _ => _
+        .OnlyWhenDynamic(() => DocumentationChangesDetected())
         .Executes(() =>
         {
             Node($"{YarnCli} install --silent", workingDirectory: RootDirectory);
             Node($"{YarnCli} --silent run cspell --no-summary", workingDirectory: RootDirectory,
                 customLogger: (_, msg) => Error(msg));
         });
+
+    bool DocumentationChangesDetected()
```
dennisdoomen, line 283: ⛏️ (nitpick) I prefer `HasDocumentationChanges` and `HasSourceChanges`
🔧 Make it properties

Build/Build.cs
```diff
@@ -258,11 +272,36 @@ class Build : NukeBuild
...
+    {
+        return Changes.Any(x => x.StartsWith("docs"));
+    }
+
+    bool SourceChangesDetected()
+    {
+        return Changes.Any(x => !x.StartsWith("docs"));
+    }
+
+    IEnumerable<string> Changes
```
dennisdoomen, line 293: 🔧 You're returning a lazily executed query here, so let's make it return an array instead and change the two methods into properties
🔧 The chained call was a bit misleading because of the newlines and the braces.

```
    string[] Changes
    {
        get
        {
            using var repo = new Repository(GitRepository.LocalDirectory);

            Tree targetBranch = repo.Branches["develop"].Tip.Tree;
            Tree workingDir = repo.Branches[repo.Head.FriendlyName].Tip.Tree;

            return repo.Diff
                .Compare<TreeChanges>(targetBranch, workingDir)
                .Where(x => x.Exists)
                .Select(x => x.Path)
                .ToArray();
        }
    }
```

Build/Build.cs
```diff
@@ -258,11 +272,36 @@ class Build : NukeBuild
...
+    {
+        get
+        {
+            using var repo = new Repository(GitRepository.LocalDirectory);
+
+            return repo.Diff.Compare<TreeChanges>(repo.Branches["develop"].Tip.Tree,
```
dennisdoomen, line 299: 🔧 Not all PRs target `develop`, so you'll need to get the target branch as a parameter from the GHA build.

Build/Build.cs
```diff
@@ -65,13 +70,15 @@ class Build : NukeBuild
     string YarnCli => ToolPathResolver.GetPackageExecutable("Yarn.MSBuild", "yarn.js", "1.22.19");
 
     Target Clean => _ => _
+        .OnlyWhenDynamic(() => SourceChangesDetected())
```
dennisdoomen, line 73: 🤔 I _think_ you only need the conditional on the `Push` target, but I'm not 100% sure.

dennisdoomen (changes requested):
Nice improvement. 

🔧 You also need to update the main build YAML and remove the new one you introduced in the other PR

Build/Build.cs
IT-VBFK: No, unfortunately not. I had similar confusions in my day job. The `OnlyWhen(Dynamic|Static)` only applies to the `Executes` part. Dependencies are still executed and have to be skipped.

Build/Build.cs
IT-VBFK: I had something in mind.. will check this out soon

dennisdoomen (approved):

Build/Build.cs
```diff
@@ -258,11 +275,38 @@ class Build : NukeBuild
...
+        get
+        {
+            using var repo = new Repository(GitRepository.LocalDirectory);
+
+            string baseRef = BaseRef ?? "develop";
+
+            Tree targetBranch = repo.Branches[baseRef].Tip.Tree;
+            Tree workingDir = repo.Branches[repo.Head.FriendlyName].Tip.Tree;
+
+            return repo.Diff
```
dennisdoomen, line 303: One more question. How does this behave during a local build?

IT-VBFK: I don't know if I get you correctly: 
If you have _uncommitted_ changes to the docs, it will ignore the spell check. Otherwise it will run the spell check before the build/test pipeline.

IT-VBFK: In [e90fe4d](https://github.com/fluentassertions/fluentassertions/pull/2089/commits/e90fe4d4c36e2f2d5f29598bfc3660c415f5f0bb) I committed 2 whitespace changes (will remove this commit afterwards)

Build/Build.cs
```diff
@@ -258,11 +275,38 @@ class Build : NukeBuild
...
+    bool HasSourceChanges =>
+        Changes.Any(x => !x.StartsWith("docs"));
+
+    string[] Changes
+    {
+        get
+        {
+            using var repo = new Repository(GitRepository.LocalDirectory);
+
+            string baseRef = BaseRef ?? "develop";
```
IT-VBFK, line 298: We have to possibilities here:

1) This solution
2) force the user to add the `BaseRef` if running local spell check `./build.sh SpellCheck --base-ref (develop|master|main|whatever)`

IT-VBFK: How do you usually create a release? 

Do you
- locally merge `develop` into `master`, set a tag and push `master`
or
- creating a pull request from `develop` to `master`, set a tag and merge online?

I am asking because I have the feelings that this `baseRef`-thing could cause a little side effect, where all NUKE targets are skipped and no release is created.

dennisdoomen: > How do you usually create a release?

I've done both and it shouldn't matter. I think that this `BaseRef` parameter should be optional. If you don't specify anything, just assume we want to run all targets. 

In fact, maybe we should rename that property to `PullRequestBase` or something like that.

IT-VBFK: Ok makes sense

Yeah... If a variable name also confuses the author.. you should probably give it a better name 🙈

Build/Build.cs
jnyrup: To make sure that this is intentional.

If I:
* check out this PR
* run `git clean -fdx`
* run `./build.ps1`

Then both `SpellCheck` and `Push` are executed.

IT-VBFK: Locally, running `build.ps1` spell check should execute but `Push` doesn't. Why `Push` is executed? There is no tag

jnyrup: Sorry, meant that both targets of `Main()` are executed.
`Push` _itself_ is properly skipped as always.

IT-VBFK: Ahh.. yes this is intentional.. I treat them as two individual pipelines, because I had no clue where I should add the spell check semantically

IT-VBFK: And: depending an other target on the documentation spell check.. is a bit odd :D

jnyrup (approved):
I'm learning a lot about NUKE from these PRs 🚀

IT-VBFK:
Should this PR be squashed with the previous "cSpell to NUKE" commit?

Technically it is no new feature, but an improvement of a none released feature.

dennisdoomen:
> Reason why I ask: Now (if I understand the release notes template correctly) there would be four entries about this one feature. Is this correct?

Those are technical release notes, so that's completely fine.

IT-VBFK:
Ahh.. sorry for the inconvenience.. But realizing now, that my latest commit to this PR fails the build (on develop branch).

Fixing that in a minute

PR #2093, opened 2023-01-09
Clean up `Build.cs`

IT-VBFK:
## IMPORTANT 

## This PR does following

- Switch to `GitHubActions` which will be injected by NUKE at build start
   - Therefore: Retire manually passed arguments/parameters
       - BranchSpec
       - BuildNumber
       - PullRequestBase
- Rename `ApiKey` to more specific `NuGetApiKey`
- Proceed with pipeline if spell check fails
- Move related code as near as possible to each other
- Clean up node commands


IT-VBFK:
Waiting for #2094...

IT-VBFK:
What about if we continue the pipeline, even if the spell check fails? @jnyrup or @dennisdoomen

IT-VBFK:
This is a really tricky task.. 

I am checking now if I can overcome this issue

jnyrup:
It seems we can simply use `.ProceedAfterFailure()` on both `SpellCheck` and `Push`.
image

IT-VBFK:
Oh... thats too easy.. 🙈

Build/Build.cs
```diff
@@ -32,17 +70,19 @@ class Build : NukeBuild
...
-    [Parameter("The target branch for the pull request")]
-    readonly string PullRequestBase;
+    string BranchSpec => GitHubActions?.Ref;
+    string BuildNumber => GitHubActions?.RunNumber.ToString();
+    string PullRequestBase => GitHubActions?.BaseRef;
 
     [Parameter("The key to push to Nuget")]
-    readonly string ApiKey;
+    [Secret]
+    readonly string Nugetapikey;
```
IT-VBFK, line 81: If this PR should be merged you have to rename the github secret `NUGETAPIKEY` either to `NUGET_API_KEY`or `API_KEY` to get the correct spelling when generating the yaml.

Build/Build.cs
```diff
@@ -3,24 +3,63 @@
...
 using static Nuke.Common.IO.PathConstruction;
+using static Nuke.Common.Tools.CoverallsNet.CoverallsNetTasks;
 using static Nuke.Common.Tools.DotNet.DotNetTasks;
 using static Nuke.Common.Tools.ReportGenerator.ReportGeneratorTasks;
 using static Nuke.Common.Tools.Xunit.XunitTasks;
 using static Serilog.Log;
 
 [UnsetVisualStudioEnvironmentVariables]
 [DotNetVerbosityMapping]
+[CustomGitHubActions(
```
dennisdoomen, line 27: We had an internal chat and we do like most of the changes, but we like to stay away from the code generation part. Although we don't like YAML, the generation code isn't much prettier and will make it more difficult to understand what's going on.

IT-VBFK:
Ok.. just to make sure we are in sync here:
"The most changes": includes _everything_ but the yaml generation?

Build/Build.cs
ITaluone: If you don't like to see the generation code, just put a `partial class Build {}` in a separate file and decorate this one ;)

So you don't see the yaml AND the generation code in future :D

dennisdoomen:
> Ok.. just to make sure we are in sync here:
> "The most changes": includes _everything_ but the yaml generation?

A quickly glanced at the PR and most seems legit, but we still need to do a thorough review.

Build/Build.cs
```diff
@@ -138,32 +145,37 @@ class Build : NukeBuild
...
         });
 
-    Target UnitTests => _ => _
+    Project[] Projects => new[]
+    {
+        Solution.Specs.FluentAssertions_Specs,
+        Solution.Specs.FluentAssertions_Equivalency_Specs
+    };
+
+    Target UnitTestsWin => _ => _
```
IT-VBFK, line 154: Advantages to split the one big target is, that you can see what is skipped and why in the build summary

Now you have to actually look at the code to see what is skipped

If you agree with me: maybe this should be also applied to the `TestFrameworks` target..

IT-VBFK: See here:
Windows:
image


 Unix:
image

Build/Build.cs
```diff
@@ -70,9 +68,18 @@ class Build : NukeBuild
...
-    string YarnCli => ToolPathResolver.GetPackageExecutable("Yarn.MSBuild", "yarn.js", "1.22.19");
+
+    Target Informations => _ => _
+        .OnlyWhenDynamic(() => !IsLocalBuild)
+        .Executes(() =>
+        {
+            Information(BranchSpec);
+            Information(BuildNumber);
+            Information(PullRequestBase);
+        });
```
IT-VBFK, line 79: Reason for this new target: I often had to look in the build logs (especially during the last PRs regarding the NUKE pipeline) to see the value of various variables.

Now... Put them in this target if you care.

IT-VBFK:
image


I don't know why, but after [e65159f](https://github.com/fluentassertions/fluentassertions/pull/2093/commits/e65159ffdd1920945a1942581d88b2be16fecab2) the pipeline is significantly faster?

Build/Build.cs
```diff
@@ -241,12 +263,15 @@ class Build : NukeBuild
...
-        .DependsOn(CalculateNugetVersion)
+        .DependsOn(
+            ApiChecks,
+            TestFrameworks,
+            UnitTests,
+            CodeCoverage,
+            CalculateNugetVersion
+        )
         .OnlyWhenDynamic(() => RunAllTargets || HasSourceChanges)
+        .Produces(ArtifactsDirectory / "*.nupkg")
```
jnyrup, line 274: What does this do?

Build/Build.cs
```diff
@@ -70,9 +68,18 @@ class Build : NukeBuild
...
+        .OnlyWhenDynamic(() => !IsLocalBuild)
+        .Executes(() =>
+        {
+            Information(BranchSpec);
+            Information(BuildNumber);
+            Information(PullRequestBase);
+        });
 
     Target Clean => _ => _
+        .Triggers(Informations)
```
jnyrup, line 82: Does `Informations` run independently on whether `SpellCheck` or `Push` is run?

Build/Build.cs
IT-VBFK: Oh.. nothing anymore..
If we would use the yaml code generation, this would generate the "upload artifacts" step

Build/Build.cs
IT-VBFK: No I don't think so.. is this a problem?

IT-VBFK:
Just as information: I will try coveralls.net today again.

Build/Build.cs
IT-VBFK: Ah.. got it.. will change that

IT-VBFK:
Fair point 👍🏻

Build/Build.cs
IT-VBFK: What about leaving this target, but do not call it at all? Just to invoke it via cli switch `build.sh --target informations`?

IT-VBFK:
Just have a question: How is the token to access coveralls called in this repo?

image
I think it isn't `GITHUB_TOKEN`?

IT-VBFK:
Upps wrong PR..

ITaluone:
We also need something like:

```yml
- name: Check file existence
  id: check_files
  uses: andstor/file-existence-action@v1
  with:
    files: "TestResults/reports/lcov.info"

- name: coveralls
      uses: coverallsapp/github-action@1.1.3
      if: steps.check_files.outputs.files_exists == 'true'
      with:
        github-token: ${{ secrets.GITHUB_TOKEN }}
        path-to-lcov: TestResults/reports/lcov.info
```

to prevent failing the pipeline when only documentation has changes

This could be part of this PR if you like..

Build/Build.cs
```diff
@@ -286,6 +285,8 @@ class Build : NukeBuild
                 customLogger: (_, msg) => Error(msg));
         });
 
+    string YarnCli => $"{ToolPathResolver.GetPackageExecutable("Yarn.MSBuild", "yarn.js", "1.22.19")} --silent";
```
dennisdoomen, line 316: 🤔 I actually liked the original place. Kept the "constants" like that at the top.

Build/Build.cs
```diff
@@ -288,10 +288,10 @@ class Build : NukeBuild
     string YarnCli => $"{ToolPathResolver.GetPackageExecutable("Yarn.MSBuild", "yarn.js", "1.22.19")} --silent";
     
     bool HasDocumentationChanges =>
-        Changes.Any(x => x.StartsWith("docs"));
+        Changes.Any(x => x.StartsWith("docs", StringComparison.InvariantCultureIgnoreCase));
```
dennisdoomen, line 319: 🤔 I don't think this is very important. It's just a build script and we don't have culture-sensitive folders.

Build/Build.cs
```diff
@@ -261,6 +261,7 @@ class Build : NukeBuild
     Target Push => _ => _
         .DependsOn(Pack)
         .OnlyWhenDynamic(() => IsTag)
+        .ProceedAfterFailure()
```
dennisdoomen, line 290: ❓ Since it's the last target in the chain, why do we need this one?

Build/Build.cs
```diff
@@ -240,12 +245,14 @@ class Build : NukeBuild
         });
 
     Target Pack => _ => _
-        .DependsOn(ApiChecks)
-        .DependsOn(TestFrameworks)
-        .DependsOn(UnitTests)
-        .DependsOn(CodeCoverage)
-        .DependsOn(CalculateNugetVersion)
+        .DependsOn(
```
dennisdoomen, line 267: ⛏️ It's a bit of taste thing, but I prefer the separate `DependsOn` lines.

Build/Build.cs
```diff
@@ -75,6 +76,9 @@ class Build : NukeBuild
         .OnlyWhenDynamic(() => RunAllTargets || HasSourceChanges)
         .Executes(() =>
         {
+            Information(BranchSpec);
```
dennisdoomen, line 77: 🤔 What does this have to do with the `Clean` target?

Build/Build.cs
```diff
@@ -32,14 +33,14 @@ class Build : NukeBuild
 
     public static int Main() => Execute<Build>(x => x.SpellCheck, x => x.Push);
 
-    [Parameter("A branch specification such as develop or refs/pull/1775/merge")]
-    readonly string BranchSpec;
+    GitHubActions GitHubActions => GitHubActions.Instance;
```
dennisdoomen, line 36: 🤔 @jnyrup do we want to tie ourselves to Github Actions like this? It kind of defies the idea of an independent build script. But we do assume there's a Github branch.

Build/Build.cs
```diff
@@ -69,13 +69,20 @@ class Build : NukeBuild
 
     string SemVer;
 
-    Target Clean => _ => _
-        .OnlyWhenDynamic(() => RunAllTargets || HasSourceChanges)
+    Target Informations => _ => _
```
dennisdoomen, line 73: 🔧 That's a normal English word...

Build/Build.cs
```diff
@@ -145,32 +145,37 @@ class Build : NukeBuild
...
         });
 
-    Target UnitTests => _ => _
+    Project[] Projects => new[]
+    {
+        Solution.Specs.FluentAssertions_Specs,
+        Solution.Specs.FluentAssertions_Equivalency_Specs
+    };
+
+    Target UnitTestsNetFramework => _ => _
```
dennisdoomen, line 155: 🤔 Why do you need this separation?

Build/Build.cs
IT-VBFK: As I understand the `.DependsOn()`: This works in opposite direction, meaning: we start at `Push` (in the main execute method) and then backwards until we reach the end (in this case `Clean`).

And because `Push` is the first "entry point" of the build pipeline, we need this at that place

Build/Build.cs
IT-VBFK: Correct :D

Mine would be that solution, but it's your project ;-)

Build/Build.cs
IT-VBFK: Nothing really.. I am unsure wether to keep or remove it.. I will remove this for now

Build/Build.cs
IT-VBFK: My intention was, that we get all variables free of charge just by using this injection.

Locally this won't make a difference by now, because we have default values in this cases (or we do not need them locally)

Build/Build.cs
IT-VBFK: Removed the target, since I only used it for debugging :)

Build/Build.cs
IT-VBFK: Hmm.. I don't like scrolling around to see the implementation very much.. (in any project)

And because splitting up the `Build` class I have thought about to keep related things together like this

Build/Build.cs
```diff
@@ -138,32 +136,35 @@ class Build : NukeBuild
...
+        });
 
+    Target UnitTestsNetCore => _ => _
+        .Unlisted()
+        .DependsOn(Compile)
+        .OnlyWhenDynamic(() => RunAllTargets || HasSourceChanges)
+        .Executes(() =>
+        {
             DotNetTest(s => s
                 .SetConfiguration("Debug")
```
IT-VBFK, line 169: I also would like to use `Configuration.Debug` (the enumeration class) to get type safety

Also applies to other appearances

Build/Build.cs
jnyrup: I like the less yaml fiddling and extra type safety.
It seems to have caught that https://github.com/fluentassertions/fluentassertions/commit/c56b55b7f08e9655fece50d3981495a2657956d5 accidentally added `BaseRef` in `.github/workflows/build.yml` for the linux/mac jobs.

While it does add coupling to a particular build system, namely GithubActions, I'm not sure it's _that bad_ because:
* It still runs locally
* The usage of GithubActions is limited to a single file.
* Should we ever want to change to AppVeyor, Azure DevOps, etc, it might be even easier to transition due to intellisense (given NUKE supports it), than adapting a yaml file to a new build system.

... _pendantically_ speaking we also have a coupling to git even though other source code control systems exists ;)

Build/Build.cs
jnyrup: Let's keep this at it's original place.
Then I don't have to scroll around when updating our dependencies.

Build/Build.cs
jnyrup: For the non-windows jobs it becomes explicit that the windows tests are skipped.

image

Build/Build.cs
jnyrup: If `public static int Main() => Execute<Build>(x => x.SpellCheck, x => x.Push);` runs the targets consecutively and in the specified order, we strictly don't need `ProceedAfterFailure` here as `Push` is the last target.
But I'm fine having it here.

Build/Build.cs
IT-VBFK: Ahh ok.. that'a fair point :)

Build/Build.cs
IT-VBFK: cool

Build/Build.cs
IT-VBFK: > It seems to have caught that https://github.com/fluentassertions/fluentassertions/commit/c56b55b7f08e9655fece50d3981495a2657956d5 accidentally added BaseRef in .github/workflows/build.yml for the linux/mac jobs.

IMO that fail to remove this variable shows the pitfall perfectly..

BTW: this is removed via this PR

Build/Build.cs
```diff
@@ -115,7 +115,7 @@ class Build : NukeBuild
         {
             DotNetBuild(s => s
                 .SetProjectFile(Solution)
-                .SetConfiguration("CI")
```
IT-VBFK, line 120: Just for interest: why is this configuration `CI` needed?

jnyrup: To reduce what projects we build during CI.
E.g. no need to build benchmarks projects as we don't run it during CI.

IT-VBFK: Ahh.. I missed this specific projects :)
thanks

jnyrup: It's easiest to get an overview of, if you look at it in your IDE's configuration manager.
In the CI configuration we also:
* Compile the library in release mode as that's the dll that's included in the nupkg, but 
* Compile the tests in debug mode as that's needed for CallerIdentification

IT-VBFK: Quite cool :)

Build/Build.cs
IT-VBFK: > Should we ever want to change to AppVeyor, Azure DevOps, etc, it might be even easier to transition due to intellisense (given NUKE supports it), than adapting a yaml file to a new build system.

If we would use the yaml generation part of NUKE, it would be even easier, because we only have to look at the attribute and would get a yaml which is correct for the "new" system

dennisdoomen: OK. Let's keep the specific Github Actions stuff

dennisdoomen (approved):
Nice work!

jnyrup (approved):

PR #2095, opened 2023-01-12
Retry converting GitHubAction `coveralls` to NUKE with `coveralls.net`

IT-VBFK:
## IMPORTANT 


ITaluone:
image

.github/workflows/build.yml
```diff
@@ -31,12 +31,7 @@ jobs:
         BuildNumber: ${{ github.run_number }}
         PullRequestBase: ${{ github.event.pull_request.base.ref }}
         ApiKey: ${{ secrets.NUGETAPIKEY }}
-
-    - name: coveralls
-      uses: coverallsapp/github-action@1.1.3
-      with:
-        github-token: ${{ secrets.GITHUB_TOKEN }}
-        path-to-lcov: TestResults/reports/lcov.info
+        CoverallsToken: ${{ secrets.GITHUB_TOKEN }}
```
ITaluone, line 34:
```suggestion
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

Build/Build.cs
```diff
@@ -44,6 +50,9 @@ class Build : NukeBuild
     [Parameter("The key to push to Nuget")]
     readonly string ApiKey;
 
+    [Parameter("Token for coveralls")]
+    readonly string CoverallsToken;
+
```
ITaluone, line 55:
```suggestion
```

Build/Build.cs
```diff
@@ -65,14 +74,33 @@ class Build : NukeBuild
 #endif
     Tool Node;
 
+    [PathExecutable("git")] Tool Git;
+
```
ITaluone, line 78:
```suggestion
```

Build/Build.cs
```diff
@@ -189,20 +217,76 @@ class Build : NukeBuild
...
+        .Consumes(CodeCoverage)
+        .Executes(() =>
+        {
+            CommitInfo commitInfo = GetLastCommit();
+
+            CoverallsNet(s => s
+                .SetDryRun(IsLocalBuild)
+                .SetProcessArgumentConfigurator(x => x.Add("--lcov"))
+                .SetInput(CoverageResultDirectory / "lcov.info")
+                .SetRepoToken(CoverallsToken)
```
ITaluone, line 250:
```suggestion
                .SetRepoToken(GitHubActions?.Token)
```

.github/workflows/build.yml
ITaluone, line 34:
```suggestion
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

Build/Build.cs
```diff
@@ -189,20 +214,76 @@ class Build : NukeBuild
...
+        .OnlyWhenDynamic(() => RunAllTargets || HasSourceChanges)
+        .DependsOn(CodeCoverage)
+        .Consumes(CodeCoverage)
+        .Executes(() =>
+        {
+            CommitInfo commitInfo = GetLastCommit();
+
+            CoverallsNet(s => s
+                .SetDryRun(IsLocalBuild)
+                .SetProcessArgumentConfigurator(x => x.Add("--lcov"))
```
ITaluone, line 245:
```suggestion                
                .SetProcessArgumentConfigurator(x => x.Add("--lcov"))
                .SetProcessArgumentConfigurator(arg => arg.Add("-v"))
```

Build/Build.cs
```diff
@@ -189,20 +214,77 @@ class Build : NukeBuild
...
+                .SetProcessArgumentConfigurator(arg => arg.Add("-v"))
```
ITaluone, line 246:
```suggestion
            .SetProcessArgumentConfigurator(x => x
                .Add("--lcov")
                .Add("-v"))
```

ITaluone, line 246:
```suggestion
                .SetProcessArgumentConfigurator(x => x
                    .Add("--lcov"))
```

IT-VBFK:
@jnyrup or @dennisdoomen 
Can you please recheck the integration of coveralls with the github token (or a different one)?
I cannot declare myself why else this pipeline is failing at coveralls..

dennisdoomen:
The token is correct and properly passed to the script. So something else is not right.

IT-VBFK:
Hmm.. that odd.. I can't figure out what the problem is..

jnyrup:
https://github.com/csMACnz/coveralls.net/issues/108 indicates that we're passing the wrong kind of token.

>  `--repoToken <repoToken>`                  The coveralls.io repository token.

https://github.com/csMACnz/coveralls.net/blob/main/src/csmacnz.Coveralls/MainArgs.cs

IT-VBFK:
Thank you for digging into this

I fear I cannot help with the token ;)

IT-VBFK:
> image

Just to make sure to not forget this check in NUKE build

dennisdoomen:
I've just added a new Coveralls-specific token under the name `COVERALLS_TOKEN`. Try that one instead.

IT-VBFK:
> I've just added a new Coveralls-specific token under the name `COVERALLS_TOKEN`. Try that one instead.

Is this a secret? Because it seems that there is nothing passed to NUKE:

image

dennisdoomen:
> Is this a secret? Because it seems that there is nothing passed to NUKE:

It's registered as a repository secret indeed.

IT-VBFK:
I cannot tell what the problem is.. if I try this locally with the token from my fork it works: https://coveralls.io/github/IT-VBFK/fluentassertions?branch=build/convert-coveralls

(ignore the coverage.. I skipped them, because of time issues)

jnyrup:
My guess is that the GH secret is not properly passed to Nuke somehow.

The call to `csmacnz.Coveralls.dll` doesn't include `--repoToken` which indicates that `Coveralls` passed to `SetRepoToken()` is null or whitespace, https://github.com/nuke-build/nuke/blob/afcfee51f667158044b241096c53ab079b4486b4/source/Nuke.Tooling/Arguments.cs#L55-L56

You could verify this by logging `string.IsNullOrWhiteSpace(Coveralls)`.

I have retried copying the repo token from coveralls into github secrets, without success.
And confirming that it is a repository secret named `COVERALLS_TOKEN`.

IT-VBFK:
There must be an issue with permissions or something, because if I add a repo secret in my fork and use that instead.. it works

see this latest workflow run: https://github.com/IT-VBFK/fluentassertions/actions/runs/3919802757/jobs/6701153117
and the coveralls-page: https://coveralls.io/github/IT-VBFK/fluentassertions

jnyrup:
New best guess about what's confusing us

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow

IT-VBFK:
uhh.. that's stupid.. that means we can't use that?

What a pity

jnyrup:
> uhh.. that's stupid..

It's for security.
https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions

Giving this a second thought, we didn't provide the `Coveralls` token before, so we shouldn't need it now.
Looking at the source for the currently used [github actions task](https://github.com/coverallsapp/github-action/blob/master/src/run.ts), it _does_ use `secrets.GITHUB_TOKEN`.

IT-VBFK:
> Giving this a second thought, we didn't provide the Coveralls token before, so we shouldn't need it now.
Looking at the source for the currently used [github actions task](https://github.com/coverallsapp/github-action/blob/master/src/run.ts), it does use secrets.GITHUB_TOKEN.

But this didn't work either.. See the very first commits/workflow runs of this PR

IT-VBFK:
What about using NUKE's secret parameter feature by storing the encrypted value in `Build/parameters.json`?

IT-VBFK:
Locally this works for me:

- Add `[Secret]` attribute to the coveralls variable in `Build.cs` and run `./build.sh` once to generate the `build.schema.json`
- adding my coveralls token as a NUKE secret (in root dir run following): 
  - `nuke :secrets`
  - choose the variable name
  - enter the secret
- run `./buil.sh` again
- it should be working now (at least on my machine

the only draw-back is, that you have to provide (or commit directly to this PR) the changes parameters.json, which holds the encrypted token

dennisdoomen:
> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

This means that the secret is not available from a build that is part of the fork. And that makes total sense. We're running builds from the main repo. And we're also passing the Nuget push key in the exact same way. And that works.

> Giving this a second thought, we didn't provide the `Coveralls` token before, so we shouldn't need it now.
> Looking at the source for the currently used [github actions task](https://github.com/coverallsapp/github-action/blob/master/src/run.ts), it _does_ use `secrets.GITHUB_TOKEN`.

That was my thought as well. I suspect that the GHA action for Coverall works in a different way than the .NET library

dennisdoomen:
> the only draw-back is, that you have to provide (or commit directly to this PR) the changed parameters.json, which holds the encrypted token

That is not an option.

IT-VBFK:
> That is not an option.

I have already guessed something like this :)

IT-VBFK:
> This means that the secret is not available from a build that is part of the fork. And that makes total sense. We're running builds from the main repo. And we're also passing the Nuget push key in the exact same way. And that works.

But this does not work either (see [latest workflow runs](https://github.com/fluentassertions/fluentassertions/actions/runs/3923361907/jobs/6706921792)). Both nuget api key and coveralls token are empty..

Have introduced the `Info` target again and check if `NuGetApiKey` and `CoverallsToken` are `!string.IsNullOrWhitespace`.

IT-VBFK:
I think I will close this PR soon, because I think we're stuck here..

Sorry

IT-VBFK:
Just to recap:

- Using GITHUB_TOKEN like in the coveralls action --> does not work
- Using a specific coveralls token works in the fork, but not in the main repo
- Passing secrets.Token to NUKE --> does not work (Null or whitespace)
- Passing token via NUKE secret parameters --> declined
- Storing the token as repo variable (not as secret) --> definitely not an option

I just realized that since the variable name changed from `ApiKey` to `NuGetApiKey` no key is passed to NUKE anymore

IT-VBFK:
An other possibility could be to use an external [key vault](https://www.vaultproject.io)

like proposed here: https://stackoverflow.com/a/67543604

jnyrup:
> An other possibility could be to use an external [key vault](https://www.vaultproject.io) or AzureKeyVault (which is also supported by NUKE)
> 
> like proposed here: https://stackoverflow.com/a/67543604

Not an option.
While we like to reduce the amount of yaml, introducing completely external systems introduces more complexity and fragility.

If we cannot find a way to inject some secret (github-token or coveralls token) into NUKE, then I say we keep it as it is right now.

dennisdoomen:
image

Maybe this only works if me or @jnyrup trigger the build....

IT-VBFK:
> image
> 
> Maybe this only works if me or @jnyrup trigger the build....

Is this what we really want? Seems like a bad "hack"

IT-VBFK:
As I don't have no ideas left.. I would close this.. 

Or do you have any ideas @jnyrup @dennisdoomen ?

jnyrup:
I also out of ideas.

IT-VBFK:
Ok.. seems we stick with yaml 🤔

dennisdoomen:
Something is off. I've used repository secrets many times before. But looking at the flaky behavior we see with Coveralls right now, I'm starting to think we're facing the same problem there. When a contributor creates a PR, that PR won't get the right token and doesn't produce a Coveralls report. It always requires us to manually rebuild it. 

But maybe I'm imagining things.

IT-VBFK:
I will finish this if github ever introduces a fix for this problem

PR #2101, opened 2023-01-14
Fix timeout check in WithResult extension

### LINKED ISSUE #1969: WithResult extension is always waiting for the Task to be completed
With #1478 I implemented the extension methods `WithResult` for assertions on asynchronous operations.
While playing around with this methods together with failed `Should().CompleteWithinAsync` I found that it could be unexpected, that these methods explicitly waiting for the completion of the `Task`.
The original intention of the assertions `CompleteWithinAsync` were ensuring tests will not hang forever. With awaiting the `Task` the "timeout" is skipped and the test may hang.

Today I think whether it would be better to not `await` the `Task` and instead check whether it is already completed.
In case the `Task` is not completed, further the result is not relevant anymore, isn't it?

lg2de:
## IMPORTANT 

Fixes #1969

### DIFF
--- a/Src/FluentAssertions/AsyncAssertionsExtensions.cs
+++ b/Src/FluentAssertions/AsyncAssertionsExtensions.cs
@@ -18,6 +18,12 @@ public static class AsyncAssertionsExtensions
     /// <param name="becauseArgs">
     /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
     /// </param>
+    /// <remarks>
+    /// Please note that this assertion cannot identify whether the previous assertion was successful or not.
+    /// In case it was not successful and it is running within an active <see cref="FluentAssertions.Execution.AssertionScope"/>
+    /// there is no current result to compare with.
+    /// So, this extension will compare with the default value.
+    /// </remarks>
     public static async Task<AndWhichConstraint<GenericAsyncFunctionAssertions<T>, T>> WithResult<T>(
         this Task<AndWhichConstraint<GenericAsyncFunctionAssertions<T>, T>> task,
         T expected, string because = "", params object[] becauseArgs)
--- a/Tests/FluentAssertions.Specs/FakeClock.cs
+++ b/Tests/FluentAssertions.Specs/FakeClock.cs
@@ -9,7 +9,7 @@ namespace FluentAssertions.Specs;
 /// Implementation of <see cref="IClock"/> for testing purposes only.
 /// </summary>
 /// <remarks>
-/// It allows you to control the "current" time.
+/// It allows you to control the "current" date and time for test purposes.
 /// </remarks>
 internal class FakeClock : IClock
 {
@@ -25,17 +25,26 @@ internal class FakeClock : IClock
 
     public ITimer StartTimer() => new TestTimer(() => elapsedTime);
 
+    /// <summary>
+    /// Advances the internal clock.
+    /// </summary>
     public void Delay(TimeSpan timeToDelay)
     {
         elapsedTime += timeToDelay;
     }
 
+    /// <summary>
+    /// Simulates the completion of the pending delay task.
+    /// </summary>
     public void Complete()
     {
         // the value is not relevant
         delayTask.SetResult(true);
     }
 
+    /// <summary>
+    /// Simulates the completion of the pending delay task after the internal clock has been advanced.
+    /// </summary>
     public void CompleteAfter(TimeSpan timeSpan)
     {
         Delay(timeSpan);
--- a/Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Diagnostics.CodeAnalysis;
 using System.Threading.Tasks;
 using FluentAssertions.Execution;
 using FluentAssertions.Extensions;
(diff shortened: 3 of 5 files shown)


Src/FluentAssertions/AsyncAssertionsExtensions.cs
```diff
@@ -22,6 +23,11 @@ public static class AsyncAssertionsExtensions
         this Task<AndWhichConstraint<GenericAsyncFunctionAssertions<T>, T>> task,
         T expected, string because = "", params object[] becauseArgs)
     {
+        if (AssertionScope.Current.HasFailures())
+        {
+            return null;
+        }
```
jnyrup, line 29: Other things than `CompleteWithinAsync` could have added a failure to the `AssertionScope`.

```cs
true.Should().BeFalse();
await func.Should().CompleteWithinAsync(100.Milliseconds()).WithResult(2);
```

lg2de: Yes, I know.
But what is the consequence? We skip one check while another already caused a failed test.

This is just a prove of concept. I'm thinking about a solution to prevent from test not completing.

jnyrup: The whole idea of `AssertionScope` is to continue asserting where possible.
So if `true.Should().BeFalse();` fails, but `CompleteWithinAsync` succeeds, `WithResult` should be executed.

But I get the case you want to solve, just wanted to drop a note.

Src/FluentAssertions/AsyncAssertionsExtensions.cs
```diff
@@ -18,6 +18,12 @@ public static class AsyncAssertionsExtensions
     /// <param name="becauseArgs">
     /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
     /// </param>
+    /// <remarks>
+    /// Please note that this assertion cannot identify whether the previous assertion was successful or not.
+    /// In case it was not successful and it is running within an active <see cref="FluentAssertions.Execution.AssertionScope"/>
+    /// there is not current result to compare.
```
dennisdoomen, line 24:
```suggestion
    /// there is no current result to compare.
```

Tests/FluentAssertions.Specs/FakeClock.cs
```diff
@@ -9,7 +9,7 @@ namespace FluentAssertions.Specs;
 /// Implementation of <see cref="IClock"/> for testing purposes only.
 /// </summary>
 /// <remarks>
-/// It allows you to control the "current" time.
+/// It allows you to control the "current" time for internal delays.
```
dennisdoomen, line 12:
```suggestion
/// It allows you to control the "current" date and time for test purposes.
```

Src/FluentAssertions/AsyncAssertionsExtensions.cs
dennisdoomen, line 24:
```suggestion
    ///  there is no current result to compare with.
```

Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
```diff
@@ -165,6 +164,28 @@ public async Task When_task_completes_late_it_in_assertion_scope_should_fail()
             await action.Should().ThrowAsync<XunitException>();
         }
 
+        [Fact]
+        public async Task When_task_does_not_completes_the_result_extension_does_not_hang()
```
dennisdoomen, line 168:
```suggestion
        public async Task When_task_does_not_complete_the_result_extension_does_not_hang()
```

docs/_pages/releases.md
```diff
@@ -12,6 +12,7 @@ sidebar:
 ### What's new
 
 ### Fixes
+* Fixes hanging of `CompleteWithinAsync` when using with `WithResult` and `AssertionScope`. - [#2101](https://github.com/fluentassertions/fluentassertions/pull/2101)
```
jnyrup, line 15:
```suggestion
* Fixed hanging of `CompleteWithinAsync` when used with `WithResult` and `AssertionScope` - [#2101](https://github.com/fluentassertions/fluentassertions/pull/2101)
```

Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
```diff
@@ -165,6 +164,28 @@ public async Task When_task_completes_late_it_in_assertion_scope_should_fail()
...
+                Func<Task<int>> func = () => taskFactory.Task;
+                using var _ = new AssertionScope();
+                return func.Should(timer).CompleteWithinAsync(100.Milliseconds()).WithResult(2);
+            };
+            timer.Complete();
+
+            // Assert
+            var assertionTask = action.Should().ThrowAsync<XunitException>()
+                .WithMessage("Expected*to complete within 100ms.*Expected return*to be 2, but found 0.");
+            await this.Awaiting(x => assertionTask).Should().CompleteWithinAsync(200.Seconds());
```
jnyrup, line 186: Adding `using static FluentAssertions.FluentActions;`

```suggestion
            await Awaiting(() => assertionTask).Should().CompleteWithinAsync(200.Seconds());
```

jnyrup (commented):
I simply love when such edge-cases has a simple and fitting code change.

jnyrup (approved):

dennisdoomen (approved):

dennisdoomen:
Needs a rebase...

PR #2103, opened 2023-01-15
Avoid exceptions when wrapping in `AssertionScope`

jnyrup:
## IMPORTANT 

#2101 inspired me to look for some more places, where we would continue assertion despite `Subject` being null, which would throw a `NullReferenceException` or `InvalidOperationException`.

### DIFF
--- a/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
@@ -127,10 +127,15 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
     {
         Type expectedType = typeof(TException);
 
-        Execute.Assertion
+        bool success = Execute.Assertion
             .ForCondition(Subject is not null)
             .BecauseOf(because, becauseArgs)
             .FailWith("Expected {context} to throw exactly {0}{reason}, but found <null>.", expectedType);
+
+        if (!success)
+        {
+            return new ExceptionAssertions<TException>(Array.Empty<TException>());
+        }
 
         Exception exception = await InvokeWithInterceptionAsync(Subject);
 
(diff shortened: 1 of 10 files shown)


dennisdoomen (approved):
Nice. You definitely have too much free time ;-)

dennisdoomen:
Weird that Coveralls didn't kick in again.

jnyrup:
FYI: I re-ran the pipeline as the coveralls step ran in the first pipeline, but I couldn't see it on coveralls.io

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -127,11 +127,16 @@ public AsyncFunctionAssertions(Func<TTask> subject, IExtractExceptions extractor
...
 
-        Execute.Assertion
+        bool success = Execute.Assertion
             .ForCondition(Subject is not null)
             .BecauseOf(because, becauseArgs)
             .FailWith("Expected {context} to throw exactly {0}{reason}, but found <null>.", expectedType);
 
+        if (!success)
+        {
+            return new ExceptionAssertions<TException>(Array.Empty<TException>());
```
lg2de, line 137: Wondering why this "early return" is accepted here.
In the past I was requested to NOT early return.

dennisdoomen: I agree with you. It makes the code more difficult to read.

lg2de: No, I (myself!!!) don't think so, but it is style you (!!) prefer.
So, I was wondering why do you approve such changes, @dennisdoomen ;)

dennisdoomen: Because we're just humans and sometimes miss things...

PR #2105, opened 2023-01-15
Revert early returns for project consistency

jnyrup:
Due to https://github.com/fluentassertions/fluentassertions/pull/2103#discussion_r1070650590

Review with whitespaces ignored

### DIFF
--- a/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
@@ -202,7 +202,7 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
             }
             catch (Exception exception)
             {
-                NotThrowInternal(exception, because, becauseArgs);
+                return NotThrowInternal(exception, because, becauseArgs);
             }
         }
 
@@ -235,7 +235,7 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
             }
             catch (Exception exception)
             {
-                NotThrowInternal<TException>(exception, because, becauseArgs);
+                return NotThrowInternal<TException>(exception, because, becauseArgs);
             }
         }
 
--- a/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
@@ -132,21 +132,21 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
             .BecauseOf(because, becauseArgs)
             .FailWith("Expected {context} to throw exactly {0}{reason}, but found <null>.", expectedType);
 
-        if (!success)
+        if (success)
         {
-            return new ExceptionAssertions<TException>(Array.Empty<TException>());
+            Exception exception = await InvokeWithInterceptionAsync(Subject);
+
+            Execute.Assertion
+                .ForCondition(exception is not null)
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected {0}{reason}, but no exception was thrown.", expectedType);
+
+            exception.Should().BeOfType(expectedType, because, becauseArgs);
+
+            return new ExceptionAssertions<TException>(new[] { exception as TException });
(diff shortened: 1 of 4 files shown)


dennisdoomen (approved):

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -132,21 +132,21 @@ public AsyncFunctionAssertions(Func<TTask> subject, IExtractExceptions extractor
...
+                .FailWith("Expected {0}{reason}, but no exception was thrown.", expectedType);
 
-        Execute.Assertion
-            .ForCondition(exception is not null)
-            .BecauseOf(because, becauseArgs)
-            .FailWith("Expected {0}{reason}, but no exception was thrown.", expectedType);
+            exception.Should().BeOfType(expectedType, because, becauseArgs);
 
-        exception.Should().BeOfType(expectedType, because, becauseArgs);
+            return new ExceptionAssertions<TException>(new[] { exception as TException });
```
lg2de, line 146: Somehow, this is an early return too, isn't it?
In this case I do not understand, why this is better read than before.
If there is only one return, ok. But here?

jnyrup: Every method that has N returns, will have N-1 that are earlier.
You can search for `^ *return (.*\r\n){1,3} *return` to find existing methods that have closely located return statements.
So it's not a religious "you cannot have multiple return statements".
Also preference change over time, so you'll most likely be able to find a spot where the code violate our currently desired style.

It's a preference of Dennis that he chose for this project and I respect that.
Please, let's not have a lengthy argument about what the universally best style is (spoiler: it does not exist)

Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
```diff
@@ -279,37 +275,37 @@ public Task<AndConstraint<TAssertions>> NotThrowAfterAsync(TimeSpan waitTime, Ti
...
-        async Task<AndConstraint<TAssertions>> AssertionTaskAsync()
+        if (success)
         {
-            TimeSpan? invocationEndTime = null;
-            Exception exception = null;
-            ITimer timer = Clock.StartTimer();
+            return AssertionTaskAsync();
 
-            while (invocationEndTime is null || invocationEndTime < waitTime)
+            async Task<AndConstraint<TAssertions>> AssertionTaskAsync()
```
julealgon, line 282: You might still want to keep local functions at the bottom of the parent method. Effectively, they work as "private methods in a method scope", meaning treating them as if they were private puts the focus on the public code of the method first, which can improve readability: people tend to want to look at the high level concepts first, and then drill down into the lower level ones as needed.

The justification would be similar as to why the recommendation is to keep `private` methods below `public` ones in classes.

Src/FluentAssertions/Specialized/TaskCompletionSourceAssertions.cs
```diff
@@ -124,18 +124,18 @@ public TaskCompletionSourceAssertions(TaskCompletionSource<T> tcs, IClock clock)
...
         }
 
-        bool completesWithinTimeout = await CompletesWithinTimeoutAsync(subject.Task, timeSpan);
-        Execute.Assertion
-            .ForCondition(completesWithinTimeout)
-            .BecauseOf(because, becauseArgs)
-            .FailWith("Expected {context:task} to complete within {0}{reason}.", timeSpan);
-        T result = subject.Task.IsCompleted ? subject.Task.Result : default;
-        return new AndWhichConstraint<TaskCompletionSourceAssertions<T>, T>(this, result);
+        return new AndWhichConstraint<TaskCompletionSourceAssertions<T>, T>(this, default(T));
```
julealgon, line 138: Have you considered creating a small factory method on `AndWhichConstraint` and similar generic classes to avoid having to declare all the redundant/noisy type information explicitly?

Something like this would be significantly cleaner:
```cshsarp
return AndWhichConstraint.For(this, default(T));
```

dennisdoomen: Oeh, I like that.

PR #2106, opened 2023-01-15
Regrouping and formatting of the selection rule specs

dennisdoomen:
**NOTE** Review while hiding whitespace changes.

### DIFF
--- a/Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
@@ -12,498 +12,1251 @@ namespace FluentAssertions.Equivalency.Specs;
 
 public class SelectionRulesSpecs
 {
-    private enum LocalOtherType : byte
+    public class Basic
     {
-        Default,
-        NonDefault
+        [Fact]
+        public void When_a_property_is_an_indexer_it_should_be_ignored()
+        {
+            // Arrange
+            var expected = new ClassWithIndexer
+            {
+                Foo = "test"
+            };
+
+            var result = new ClassWithIndexer
+            {
+                Foo = "test"
+            };
+
+            // Act
+            Action act = () => result.Should().BeEquivalentTo(expected);
+
+            // Assert
+            act.Should().NotThrow();
+        }
+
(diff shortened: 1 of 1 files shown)


Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -12,2045 +12,2318 @@ namespace FluentAssertions.Equivalency.Specs;
...
-        var subject = new
+        [Fact]
+        public void When_a_property_is_an_indexer_it_should_be_ignored()
         {
-            Age = 36,
-            Birthdate = new DateTime(1973, 9, 20),
-            Name = "John"
-        };
+            // Arrange
+            ClassWithIndexer expected = new()
```
jnyrup, line 21: Is it intentional that you switch from `var foo = new SomeType()` to `SomeType foo = new()`?
As far as I can tell we currently only use taget-type new expressions for local variables five places.

_If_ we want to switch completely to the newly supported syntax, I think we should:
* do it in a separate PR, 
* across the whole repo and
* enforce it via `.editorconfig`.


Using different syntax makes it difficult for contributors to stay consistent.

dennisdoomen: Undid the changes. Will look at that in a follow-up PR.

Tests/.editorconfig
```diff
@@ -1,149 +1,149 @@
 [*.cs]
 
 # IDE0051: Private member is unused
-dotnet_diagnostic.IDE0051.severity = none
+dotnet_diagnostic.ide0051.severity = none
```
jnyrup, line 4: Why this change?

dennisdoomen: Rider did that. Apparently that's the convention it used... 🤷🏻

jnyrup: Hmm... their [website](https://www.jetbrains.com/help/rider/Using_EditorConfig.html#what_is) uses uppercase like any other `.editorconfig` I've ever seen.

dennisdoomen: I removed the entire file from the PR.

jnyrup (approved):

PR #2108, opened 2023-01-16
Refactor/coveralls

dennisdoomen:
## IMPORTANT 


dennisdoomen:
@jnyrup @IT-VBFK So it seems that GHA will only pass those secrets to pull requests that originate from the current repo instead of a fork...

IT-VBFK:
Thanks for checking 👍🏻

Is this the behavior we are searching for?

dennisdoomen:
> Is this the behavior we are searching for?

That depends. Is it doing this because this PR includes changes to the YAML? Or is it never going to pass the secret to any repo?

And that makes me second-guess whether this token is really worth keeping a secret.

ITaluone:
> Or is it never going to pass the secret to any repo?

Yes, that's correct. Just look at a few workflow runs:

Latest run (triggered by @jnyrup):
image

Latest PR runs (triggered by @lg2de):
image

Latest release run (triggered by @dennisdoomen)
image

> And that makes me second-guess whether this token is really worth keeping a secret.
Is this really a "secret"? It only allowes to push coverage data.. Is this really that sensitive?

IT-VBFK:
> Latest run (triggered by @jnyrup):

Also keep in mind: this was not a PR run, but a merge into develop. So I think github considered this action as "action inside the same repo".

.github/workflows/build.yml
```diff
@@ -27,25 +27,15 @@ jobs:
...
-      with:
-        files: "TestResults/reports/lcov.info"
-
-    - name: coveralls
-      uses: coverallsapp/github-action@1.1.3
-      if: steps.check_files.outputs.files_exists == 'true'
-      with:
-        github-token: ${{ secrets.GITHUB_TOKEN }}
-        path-to-lcov: TestResults/reports/lcov.info
+        ApiKey: ${{ secrets.NUGETAPIKEY }}
```
IT-VBFK, line 30: This should be `NuGetApiKey`..
I have tested myself if this was the problem (apparently not :) )

IT-VBFK:
@dennisdoomen Can you please try this: make a change, but amend the latest commit (or squash a few commits, does not matter except you have to force-push). Then force push them. 

I my local tests coveralls was ignoring the coverage report from the new run.

ITaluone:
Hm.. thanks

Seems to be the same here..

Build/Build.cs
```diff
@@ -208,10 +230,33 @@ class Build : NukeBuild
...
+            CoverallsNet(s => s
+                .SetDryRun(IsLocalBuild)
+                .SetProcessArgumentConfigurator(x => x
+                    .Add("--lcov"))
+                .SetInput(CoverageResultDirectory / "lcov.info")
+                .SetRepoToken(CoverallsToken)
+                .When(IsPullRequest,
+                    s => s.SetPullRequest(GitHubActions?.PullRequestNumber)
+                )
+                .EnableUserRelativePaths()
```
IT-VBFK, line 252:
```suggestion
                .EnableUserRelativePaths()
                .SetJobId(GitHubActions?.RunId)
```

IT-VBFK: This should do the trick

IT-VBFK: Hmm.. `.SetJobId` accepts only an `int?` but currently the github run ids are beyond `int.MaxValue`..

Build/Build.cs
```diff
@@ -208,10 +230,33 @@ class Build : NukeBuild
...
+    Target Coveralls => _ => _
+        .OnlyWhenDynamic(() => !string.IsNullOrWhiteSpace(CoverallsToken) && (RunAllTargets || HasSourceChanges))
+        .DependsOn(CodeCoverage)
+        .Consumes(CodeCoverage)
+        .Executes(() =>
+        {
+            CoverallsNet(s => s
+                .SetDryRun(IsLocalBuild)
+                .SetProcessArgumentConfigurator(x => x
+                    .Add("--lcov"))
```
IT-VBFK, line 246:
```suggestion
                    .Add("--lcov")
                    .Add($"--jobId {GitHubActions?.RunId}"))
```

IT-VBFK: For now.. this works (until the coveralls tool in Nuke supports `long?`)

dennisdoomen:
So it's kind of all by design https://github.com/csMACnz/coveralls.net/wiki/Pull-Requests

dennisdoomen:
By the way, force pushing a commit did trigger the call to Coveralls, but there was no  update. I wonder if that only happens when the coverage changes. Will remove some tests to validate this.

IT-VBFK:
I had a play yesterday to mask a github variable (the plain text ones), but did not succeed.. see the latest workflow runs in my fork..

IT-VBFK:
Would this be bad: https://github.com/csMACnz/coveralls.net/wiki/Pull-Requests#insecure-coverallsio-token-for-pull-requests ?

What would be the consequences, if there are any..

ITaluone:
https://github.com/IT-VBFK/fluentassertions/actions/runs/3948328220/jobs/6758187257

IT-VBFK:
> By the way, force pushing a commit did trigger the call to Coveralls, but there was no  update. I wonder if that only happens when the coverage changes. Will remove some tests to validate this.

That's what I meant yesterday 🙃

IT-VBFK:
Hmm.. I don't get it :D

Do you mean the confusion (at coveralls) because of this two runs (same commit):
- https://coveralls.io/builds/56049700 (with action push)
- https://coveralls.io/builds/56049652 (with action pull request)

dennisdoomen:
> Do you mean the confusion (at coveralls) because of this two runs (same commit):

No. I meant to say that I pushed a force-pushed commit that affects the coverage level. And I already added the job ID part in an earlier PR. And I still didn't get an updated coverage report in the PR.

IT-VBFK:
Maybe somewhat related to this:

https://github.com/lemurheavy/coveralls-public/issues/1684

jnyrup:
> Would this be bad: https://github.com/csMACnz/coveralls.net/wiki/Pull-Requests#insecure-coverallsio-token-for-pull-requests ?
> 
> What would be the consequences, if there are any..

I'd would approach security with the opposite mind.
If I don't know what the consequences are, I'd rather not take any risk.

IT-VBFK:
> If I don't know what the consequences are, I'd rather not take any risk.

Usually true 👍 

But the real reason we discussing about this is the "secure" policy from github, not allowing secrets to be passed to fork-pipelines (even they are executed in the main repo).

IT-VBFK:
Had a quick conversation inside [NUKE discussions](https://github.com/nuke-build/nuke/discussions/1114) :)

Matthias has shared a link: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

Following this link, we have two options here:
1) using the trigger `pull_request_target` which will run every PR in target repo context (meaning: secrets are passed) OR
2) using the trigger `pull_request`, but using an additional workflow (which is a privileged one, or in target repo context) that consumes artifacts (in our case the PR number and the coverage artifacts) from the less priviledged one (the `pull_request`)

What do you think about that approach?

dennisdoomen:
I scanned that article as well but didn't get why the second workflow using `pull_request_target` would be safer. It's still a workflow that can be updated through a PR, right?

IT-VBFK:
If I understand the `workflow_run` trigger correctly: workflows with that trigger only rund if a non-priviledged previous workflow succeeds. And if this workflow just downloads the artifacts.. all good, isn't it?

Or do you mean, that one can update this particular workflow file and do "things"? If so, isn't this the "problem" for all workflow files?

dennisdoomen:
> Or do you mean, that one can update this particular workflow file and do "things"? If so, isn't this the "problem" for all workflow files?

Yes, that's why I don't get how that split is going to solve the real problem.

IT-VBFK:
In that case.. I have to admit I agree :D

PR #2110, opened 2023-01-20
Update target frameworks for .NET 6, .NET Framework 4.7, .NET Standard 2.0, .NET Standard 2.1

Romfos:
What's changed?
- Target frameworks now: .NET 6, .NET Framework 4.7, .NET Standard 2.0, .NET Standard 2.1

Why target frameworks were changed?

Microsoft recommend to target .NET 6, .NET Standard and minimal supported .NET Framework
https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/older-framework-versions-dropped

Notes:
.NET Framework 4.6.1 is out of support, .NET Framework 4.6.2 is the older version that is still supported
https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/

.NET 5 is out of support
https://devblogs.microsoft.com/dotnet/dotnet-5-end-of-support-update/

.NET Core 3.1 is out of support
https://devblogs.microsoft.com/dotnet/net-core-3-1-will-reach-end-of-support-on-december-13-2022/

.NET Core 2.1 is out of support
https://devblogs.microsoft.com/dotnet/net-core-2-1-will-reach-end-of-support-on-august-21-2021/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,50 +4,46 @@ on: [push, pull_request]
 
 jobs:
   build:
-
     runs-on: windows-latest
 
     env:
       DOTNET_NOLOGO: true
 
     steps:
-    - uses: actions/checkout@v3
-      with:
-        fetch-depth: 0
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
-    - name: Setup .NET SDKs
-      uses: actions/setup-dotnet@v3
-      with:
-        dotnet-version: |
-          2.1.x
-          3.1.x
-          6.0.x
-          7.0.x
+      - name: Setup .NET SDKs
+        uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: |
+            6.0.x
+            7.0.x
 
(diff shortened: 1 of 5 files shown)


lg2de:
Please check whether conditional compilation like `#if !NETSTANDARD2_0` can be removed (after maintainers has given general comment).

Romfos:
> 

But .NET Standard 2.0 is still supported :)

lg2de:
> But .NET Standard 2.0 is still supported =) no change for .NET Standard 2.0

Yes, you are right. This was the wrong example.
But there is a "lot" of conditional compilation which should be checked.

Romfos:
yes, but as I see - it was implemented in a right way:
```csharp
#if NETCOREAPP2_1_OR_GREATER || NETSTANDARD2_1
```

dennisdoomen:
Changing the target frameworks would be a breaking change and those are not allowed until we start working on version 7. So unfortunately we have to close this PR for now.

PR #2119, opened 2023-01-22
Add missing tests for `int.Seconds(TimeSpan)`

IT-VBFK:
In `FluentTimeSpanExtensions`

Main reason why I not simply deleted this method was, that this is part of the public API, and would be a potential "breaking" change.

## IMPORTANT 

### DIFF
--- a/Tests/FluentAssertions.Specs/Extensions/TimeSpanConversionExtensionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Extensions/TimeSpanConversionExtensionSpecs.cs
@@ -88,6 +88,26 @@ public class TimeSpanConversionExtensionSpecs
     }
 
     [Fact]
+    public void Add_time_span_to_given_seconds()
+    {
+        // Act
+        TimeSpan time = 4.Seconds(TimeSpan.FromSeconds(1));
+
+        // Assert
+        time.Should().Be(TimeSpan.FromSeconds(5));
+    }
+
+    [Fact]
+    public void Subtract_time_span_from_given_seconds()
+    {
+        // Act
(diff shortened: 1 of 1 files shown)


jnyrup (approved):

dennisdoomen (approved):

dennisdoomen:
Now why did GH decide to pass the GH token for this build?

ITaluone:
The `GITHUB_TOKEN` is the only secret github passes to PRs
image

That's the point I don't get: The coveralls github action works with the `GITHUB_TOKEN`, but coveralls.net don't.

dennisdoomen:
> That's the point I don't get: The coveralls github action works with the `GITHUB_TOKEN`, but coveralls.net doesn't.

Maybe the GHA is using a built-in token?

dennisdoomen:
> Maybe the GHA is using a built-in token?

Ah. That GITHUB_TOKEN [is specific](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret) to the repo _and_ the specific build. And [apparently](https://github.com/coverallsapp/github-action#inputs) the call to Coveralls.io is kind of unauthenticated. Coveralls uses the GH token to verify that the coverage data is actually coming from the right repo somehow.

IT-VBFK:
Aha. But why does coveralls.net does not work with `GITHUB_TOKEN`? If you dig into the source of [coveralls GHAction](https://github.com/coverallsapp/github-action/blob/3284643be2c47fb6432518ecec17f1255e8a06a6/src/run.ts#L18-L31)

Maybe it **must** have the `GITHUB_TOKEN` _and_ as service name `github` (line 26)
This results in following:
image

which is different in our coverall.net tests:
image

(Note the "Build Type" and "Committed by" fields

Issue #2120, opened 2023-02-02
[Feature]: Use reference to System.Net.Http using framework `Reference` instead of PackageReference 


jacekmlynek:
### Background and motivation

I hope this is the correct place to raise this.

Currently for .net47 target you are adding reference to `System.Net.Http` as package dependency: 

```
 <ItemGroup Condition="'$(TargetFramework)' == 'net47'">
    <PackageReference Include="System.Net.Http" Version="4.3.4" />
  </ItemGroup>
```
According to this [issue](https://github.com/dotnet/runtime/issues/26131) dot net team advise adding `System.Net.Http` standard framework reference instead. Just quoting them:

> if you're building a .NET Framework 4.7.2 class library, then you could just add a normal Framework reference to System.Net.Http.
> 
> In most cases, we don't advise people use the separate System.Net.Http NuGet package anymore. See https://github.com/dotnet/runtime/issues/18280 and https://github.com/dotnet/runtime/issues/20777.

The main reason why I am suggesting this change is that VisualStudio resolve 4.7.2 project dependency to `System.Net.Http` and all its System dependencies by adding references to .net framework installed on local machine instead of project nuget. I think this issue has been raise in multiple place e.g. [Broken System.Net.Http 4.1.1-4.3.0 post-mortem](https://github.com/dotnet/runtime/issues/20777#issuecomment-338418610) . It can also be problematic when you have reference to same assembly with both a single name reference and a package reference. For more details please take a look on [Single-name references are removed by the SDK when targeting .NET Framework 4.7.2 and referencing a package with the same assembly](https://github.com/Microsoft/dotnet/blob/main/releases/net472/KnownIssues/613745%20-%20Single-name%20references%20are%20removed%20by%20the%20SDK%20when%20targeting%204.7.2.md)

### Alternative Concerns

Could you consider changing above to something like:

```
<ItemGroup Condition=" '$(TargetFramework)' == 'net47' ">
    <Reference Include="System.Net.Http" />
</ItemGroup>
```

It looks like another few other projects decide to follow this advise e.g. [IdentityModel](https://github.com/IdentityModel/IdentityModel/issues/76), [Microsoft Graph](https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/312)

eNeRGy164:
Is this related to https://github.com/fluentassertions/fluentassertions/pull/1806?

jacekmlynek:
[...]

PR #2123, opened 2023-02-04
Get the System.Net.Http reference from the framework

### LINKED ISSUE #2121: Builds break because System.Net.Http NuGet and assembly reference do not agree
### Description

The net72 TargetFramework build of FluentAssertions in your NuGet package distribution is broken. The nuspec states this dependency:

```xml
        <dependency id="System.Net.Http" version="4.3.4" exclude="Build,Analyzers" />
```

This dependency produces a System.Net.Http reference with assembly version 4.1.1.3:
image

Yet your FluentAssertions assembly itself references a _higher_ assembly version of 4.2.0.0 for System.Net.Http:
image

This causes some consuming MSBuild projects to fail at the compiler with this error:

> CSC : error CS1705: Assembly 'FluentAssertions' with identity 'FluentAssertions, Version=6.8.0.0, Culture=neutral, PublicKeyToken=33f2691a05b67b6a' uses 'System.Net.Http, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' which has a higher version than referenced assembly 'System.Net.Http' with identity 'System.Net.Http, Version=4.1.1.3, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'

In fact no System.Net.Http package version offers an assembly version of 4.2.0.0. This leaves two options for fixing:

## Option 1

Actually compile against the System.Net.Http that your nuspec documents, giving your assembly reference a 4.1.1.3 assembly version instead of 4.2.0.0. This would remove the conflict in the compiler and resolve the error for your users.

## Option 2

Because the 4.2.0.0 assembly version comes from the .NET Framework targeting pack rather than nuget, remove the System.Net.Http nuget package reference from your net47 build and instead add a `frameworkAssemblies` reference to your nuspec:

```diff
    <dependencies>
      <group targetFramework=".NETFramework4.7">
-       <dependency id="System.Net.Http" version="4.3.4" exclude="Build,Analyzers" />
        <dependency id="System.Threading.Tasks.Extensions" version="4.5.0" exclude="Build,Analyzers" />
      </group>
 ...
    <frameworkAssemblies>
      <frameworkAssembly assemblyName="System.Configuration" targetFramework=".NETFramework4.7" />
      <frameworkAssembly assemblyName="System.Data" targetFramework=".NETFramework4.7" />
      <frameworkAssembly assemblyName="System.Xml" targetFramework=".NETFramework4.7" />
      <frameworkAssembly assemblyName="System.Xml.Linq" targetFramework=".NETFramework4.7" />
+     <frameworkAssembly assemblyName="System.Net.Http" targetFramework=".NETFramework4.7" />
    </frameworkAssemblies>
```

### Reproduction Steps

Unfortunately all the repro cases I have are in internal projects, and I haven't been able to repro on a File-> New Project to share with you.
I'm hoping that you can see from the Description how this is (at least arguably) faulty though and how the fix should hopefully be straightforward. We have a lot of projects that use FluentAssertions and working around this bug is quite tedious. 

### Expected behavior

The references that are produced by the nuget package in our consuming projects should match or exceed the assembly references in FluentAssertions.dll.

### Actual behavior

 some consuming MSBuild projects to fail at the compiler with this error:

> CSC : error CS1705: Assembly 'FluentAssertions' with identity 'FluentAssertions, Version=6.8.0.0, Culture=neutral, PublicKeyToken=33f2691a05b67b6a' uses 'System.Net.Http, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' which has a higher version than referenced assembly 'System.Net.Http' with identity 'System.Net.Http, Version=4.1.1.3, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

dennisdoomen:
Fixes #2121 and #2120

### DIFF
--- a/Src/FluentAssertions/FluentAssertions.csproj
+++ b/Src/FluentAssertions/FluentAssertions.csproj
@@ -91,11 +91,11 @@
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'net47'">
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.0" />
-    <PackageReference Include="System.Net.Http" Version="4.3.4" />
     <Reference Include="System.Configuration" />
     <Reference Include="System.Data" />
     <Reference Include="System.Xml" />
     <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Net.Http" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
     <PackageReference Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" Version="3.3.3">
--- a/Tests/Benchmarks/Benchmarks.csproj
+++ b/Tests/Benchmarks/Benchmarks.csproj
@@ -15,5 +15,8 @@
   </ItemGroup>
   <ItemGroup>
     <Reference Include="System.Data.DataSetExtensions" Condition="'$(TargetFramework)' == 'net472'" />
+    <Reference Include="System.Net.Http" Condition="'$(TargetFramework)' == 'net472'">
+      <HintPath>..\..\..\..\Program Files\dotnet\sdk\7.0.102\Microsoft\Microsoft.NET.Build.Extensions\net461\lib\System.Net.Http.dll</HintPath>
+    </Reference>
   </ItemGroup>
 </Project>


Tests/Benchmarks/Benchmarks.csproj
```diff
@@ -15,5 +15,8 @@
   </ItemGroup>
   <ItemGroup>
     <Reference Include="System.Data.DataSetExtensions" Condition="'$(TargetFramework)' == 'net472'" />
+    <Reference Include="System.Net.Http" Condition="'$(TargetFramework)' == 'net472'">
+      <HintPath>..\..\..\..\Program Files\dotnet\sdk\7.0.102\Microsoft\Microsoft.NET.Build.Extensions\net461\lib\System.Net.Http.dll</HintPath>
```
AArnott, line 19: This hintpath doesn't look ready for any machine where the repo was cloned at an arbitrary location.

Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -91,11 +91,11 @@
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'net47'">
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.0" />
-    <PackageReference Include="System.Net.Http" Version="4.3.4" />
     <Reference Include="System.Configuration" />
     <Reference Include="System.Data" />
     <Reference Include="System.Xml" />
     <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Net.Http" />
```
AArnott, line 98: FWIW, I found in testing that this change wasn't enough for the nuspec to express this as a framework dependency. That's why in my PR I add an additional item.

dennisdoomen: Let's continue with your PR

PR #2125, opened 2023-02-05
MSBuild improvements

AArnott:
## Add a default msbuild switches response file

- `/m` speeds up the build by building with multiple processors
- `verbosity: minimal` gives a clean, succinct console log that matches what VS prints it the output window by default.
- `clp:Summary` reminds you of any errors/warnings at the end of the console log in case it scrolled off screen during the build.
- `ForceNoAlign` prevents line breaks being forced into the console log, so that resizing the console window will allow proper word wrapping.

##  Add .binlog files to .gitignore

This is a relatively new msbuild log format that is very popular and easy to review. We should not check these in any more than we check in .log files.

### DIFF
--- a/Directory.Build.rsp
+++ b/Directory.Build.rsp
@@ -1,0 +1,15 @@
+#------------------------------------------------------------------------------
+# This file contains command-line options that MSBuild will process as part of
+# every build, unless the "/noautoresponse" switch is specified.
+#
+# MSBuild processes the options in this file first, before processing the
+# options on the command line. As a result, options on the command line can
+# override the options in this file. However, depending on the options being
+# set, the overriding can also result in conflicts.
+#
+# NOTE: The "/noautoresponse" switch cannot be specified in this file, nor in
+# any response file that is referenced by this file.
+#------------------------------------------------------------------------------
+/m
+/verbosity:minimal
+/clp:Summary;ForceNoAlign
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ packages/**
 *.tmp
 *.tmp_proj
 *.log
+*.binlog
 *.vspscc
 *.vssscc
 .builds


Directory.Build.rsp
```diff
@@ -0,0 +1,15 @@
...
+# MSBuild processes the options in this file first, before processing the
+# options on the command line. As a result, options on the command line can
+# override the options in this file. However, depending on the options being
+# set, the overriding can also result in conflicts.
+#
+# NOTE: The "/noautoresponse" switch cannot be specified in this file, nor in
+# any response file that is referenced by this file.
+#------------------------------------------------------------------------------
+/m
+/verbosity:minimal
```
dennisdoomen, line 14: I'd prefer that we put all these flags in the `build.cs`.

dennisdoomen (changes requested):
Thanks for the suggestions.

Directory.Build.rsp
AArnott: I'm not familiar with BuildNuke, which seems to be what build.cs is using. You're welcome to take these ideas from the PR and apply it yourself.
FWIW though, I never would have benefitted from the changes had you set these flags there, because as an OSS contributor I just used `dotnet build` and/or `msbuild` in your repo to build it. I tend to use the standard tools as much as possible and keep my repos working with those for others. The .rsp file applies to all msbuild CLI invocations. I don't know if it would impact your build.cs-spawned builds at all.

AArnott: All of these settings are defaults in VS for Windows too. This .rsp file addition is just to improve the CLI experience for those of us that use that (in addition to IDEs).

IT-VBFK:
That sounds like a job for me, isn't it :D

Directory.Build.rsp
ITaluone: I think this parameters are already the default (at least on my machine):

If I am running `dotnet build -clp` (or any other unknown/unfulfilled cli switch) I get the following output:
```cmd
MSBUILD : error MSB1037: Specify one or more parameters for the console logger if using the -consoleLoggerParameters switch
    Full command line: 'C:\Program Files\dotnet\sdk\7.0.102\MSBuild.dll -maxcpucount -verbosity:m -restore -consoleloggerparameters:Summary -clp'
```
-maxcpucount (which is the long term for `-m`)
-verbosity:minimal and
-consoleloggerparameters:Summary (which is the long term for -clp:Summary)

The only think which is missing is the `ForceNoAlign`

IT-VBFK: Can confirm: also on my mac this seems to be the default

AArnott: Yes, for `dotnet build` they are the default (which shows they're good!) but they are _not_ the default for `msbuild.exe`.

dennisdoomen: Is anybody still using `msbuild.exe`?

dennisdoomen:
Although I respect the work @AArnott did, it doesn't seem to make a difference to anybody except (supposedly) people that still use `msbuild.exe`. However, I've just tried to open a Visual Studio 2022 Command Prompt, and it can't build our `.sln` anymore, unless you upgrade it to a specific VS version. Using `dotnet build` works. In my opinion, if you want to contribute to FluentAssertions you either need to use Visual Studio, Rider or Visual Studio Code. If you really need to use a CLI, use `dotnet build` or `build.ps1/.cmd compile`

AArnott:
I understand. FWIW, this PR had two commits, the first one (which added .binlog to .gitignore) wasn't discussed, and I think is worthy of inclusion. Totally up to you though.

dennisdoomen:
Hmm, I missed that one. Where is that generated? I never saw these files appear in my working directory.

AArnott:
`dotnet build -bl` will generate the file. It's an _amazing_ file format for build logs. It contains more than text-based diagnostic logs, but doesn't slow down the build like diagnostic logging does, so it's excellent for diagnosing all kinds of build issues. I use it locally frequently, and have my cloud builds use it all the time and capture the binlog as an artifact (on build success and failure).

You can 'build' the binlog to playback the log with any text verbosity you want. But I almost never do that. Instead, I use a popular binlog viewer: https://www.msbuildlog.com/, which is written by a Microsoft employee is a private hobby. By far the very best experience looking through msbuild logs that I've ever seen.

Anyway, if you use binlogs and don't have it in your gitignore file, obviously you may inadvertently check in the binary file, which ain't great.

jnyrup:
MSBuild Log Viewer is indeed an amazing tool to diagnose build problems :heart: 
I recall using it to help speeding up building Fluent Assertions.

I'm completely fine adding it to `.gitignore`

AArnott:
Do you want to make that change yourself, or should I create a new PR with just the one commit?

IT-VBFK:
Should we add a new target that builds with that `-bl` switch? Or add this as default to the existing compile target?

Or can we assume that one who knows about binlog is able to generate one?

dennisdoomen:
I now remember that I've used it before. If you're using Rider like me, you'll get that suggestion whenever you run a diagnostic build from its IDE:

image

So my suggestion is to add a parameter to the `build.cs` that you can use to trigger this logging and to ignore the resulting output (or write it to the already ignored `artifacts` folder.

IT-VBFK:
Sounds good

Issue #1919, opened 2022-05-03
Excluding properties behind an array doesn't work in BeEquivalentTo on collections


Turnerj:
### Description

A bit of a weird title but basically if I call `myCollection.Should().BeEquivalent(expected, cfg => cfg.Excluding(i => i.SomeArray[0].SomeProperty))`, the property isn't correctly excluded and I get the follow assertion error:

> Expected property root[0].Items[0].Name to be "Foo", but "Bar" differs near "Bar"

### Complete minimal example reproducing the issue

```csharp
var expected = new A[] {
	new A { Items = new List<B> { new B { Name = "Foo" } } }
};

var actual = new A[] {
	new A { Items = new List<B> { new B { Name = "Bar" } } }
};

actual.Should().BeEquivalentTo(
	expected,
	cfg => cfg.Excluding(a => a.Items[0].Name)
);

class A
{
	public string Name { get; set; }
	public List<B> Items { get; set; }
}

class B
{
	public string Name { get; set; }
}
```

### Expected behavior:

The excluding rule to be used and the test to pass.

### Actual behavior:

The excluding rule is ignored and the test fails.

### Versions

* Which version of Fluent Assertions are you using?
v6.6.0

* Which .NET runtime and version are you targeting? E.g. .NET framework 4.6.1 or .NET Core 2.1.
.NET 6

### Additional Information

I believe this is a regression between v5.5.3 and v6.6.0 - I haven't (yet) nailed down which version but will update when I have.

**Edit**: The regression starts in v6.0.0. The last v5 has this behaving correctly.

Additionally, this only seems to affect properties behind the array in the excluded check. Properties on the same level as the array work as expected if excluded.

IT-VBFK:
Ahhm... this should be fixed in the next release. See #1782

Turnerj:
Ahhh OK. I did look for existing issues about this but must not have had the right keywords or something.

Looking forward to the next release then!

Turnerj:
I cloned the develop branch and tested this (as that PR you referenced was merged) but it doesn't seem to work:

```csharp
var expected = new A[]
{
    new A { Items = new List<B> { new B { Name = "Foo" } } }
};

var actual = new A[]
{
    new A { Items = new List<B> { new B { Name = "Bar" } } }
};

actual.Should().BeEquivalentTo(
    expected,
    cfg => cfg.For(a => a.Items).Exclude(a => a.Name)
);
```

This is the exception the test threw:
```
[...]

PR #2135, opened 2023-03-04
Make Excluding() work on nested collections if root is a collection

### LINKED ISSUE #1989: BeEquivalentTo For().Exclude() does not work for comparison of collections
### Description

I have an array or list of nested objects that I would like to check for equivalency. The child objects have some properties that I would like to exclude from the comparison. I'm using the new For().Exclude() syntax. That works fine when comparing individual objects, but it fails when comparing collections.

```csharp
using FluentAssertions;

public class ForExcludeTest
{
    [Fact]
    public void Test()
    {
        var actualContainer = new Container
        {
            Children = new[]
            {
                new Child
                {
                    ChildId = 20,
                    Name = "Child20"
                }
            }
        };

        var expectedContainer = new Container
        {
            Children = new[]
            {
                new Child
                {
                    ChildId = 20
                }
            }
        };

        // Comparing the actual objects works
        actualContainer.Should().BeEquivalentTo(expectedContainer,
            options => options
                .For(o => o.Children).Exclude(o => o.Name)
            );

        // Comparing the objects inside a list fails
        new List<Container> { actualContainer }.Should().BeEquivalentTo(new List<Container> { expectedContainer },
            options => options
                .For(o => o.Children).Exclude(o => o.Name)
            );
    }

    public class Container
    {
        public Child[] Children { get; set; }
    }

    public class Child
    {
        public int ChildId { get; set; }
    }
}
```

### Expected behavior:

Both assertions should succeed.

### Actual behavior:

The second assertion (on a list) fails with the following error:
```
Expected property root[0].Children[0].Name to be <null>, but found "Child20".

With configuration:
- Compare enums by value
- Exclude member Children[]Name


  Stack Trace: 
XUnit2TestFramework.Throw(String message)
TestFrameworkProvider.Throw(String message)
CollectingAssertionStrategy.ThrowIfAny(IDictionary`2 context)
EquivalencyValidator.AssertEquality(Comparands comparands, EquivalencyValidationContext context)
GenericCollectionAssertions`3.BeEquivalentTo[TExpectation](IEnumerable`1 expectation, Func`2 config, String because, Object[] becauseArgs)
ForExcludeTest.Test() line 38
```

### Versions

* FluentAssertions 6.7.0
* .NET 6.0

whymatter:
Fixes #1989 #1919

I removed the `!ContainsIndexingQualifiers(selectedPath)` check on this piece of code which was introduced here: https://github.com/fluentassertions/fluentassertions/commit/aab3303a1a227b6c35d66278ed88547dbc58f97a#diff-53539b5c2a4cdb7276ff5e63cbe410a164938d3620ebf08f20797810d22b394c. I am 100% unsure what it does and what the implications are of me removing it. However, there are no failing tests with the condition removed.

```c#
if (!ContainsIndexingQualifiers(selectedPath))  // <-- removed this
{
    currentPath = RemoveIndexQualifiers(currentPath);
}
```

If removing this condition is bad for whatever reason, please tell me which test case I can add to cover this.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Equivalency/Selection/SelectMemberByPathSelectionRule.cs
+++ b/Src/FluentAssertions/Equivalency/Selection/SelectMemberByPathSelectionRule.cs
@@ -11,7 +11,7 @@ internal abstract class SelectMemberByPathSelectionRule : IMemberSelectionRule
     public IEnumerable<IMember> SelectMembers(INode currentNode, IEnumerable<IMember> selectedMembers,
         MemberSelectionContext context)
     {
-        var currentPath = RemoveIndexQualifiers(currentNode.PathAndName);
+        var currentPath = RemoveRootIndexQualifier(currentNode.PathAndName);
         var members = selectedMembers.ToList();
         AddOrRemoveMembersFrom(members, currentNode, currentPath, context);
 
@@ -22,7 +22,7 @@ internal abstract class SelectMemberByPathSelectionRule : IMemberSelectionRule
         INode parent, string parentPath,
         MemberSelectionContext context);
 
-    private static string RemoveIndexQualifiers(string path)
+    private static string RemoveRootIndexQualifier(string path)
     {
         Match match = new Regex(@"^\[\d+]").Match(path);
 
--- a/Src/FluentAssertions/Equivalency/Selection/SelectMemberByPathSelectionRule.cs
+++ b/Src/FluentAssertions/Equivalency/Selection/SelectMemberByPathSelectionRule.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
@@ -7,32 +6,12 @@ namespace FluentAssertions.Equivalency.Selection;
 
 internal abstract class SelectMemberByPathSelectionRule : IMemberSelectionRule
 {
-    private string selectedPath;
-
-    protected SelectMemberByPathSelectionRule(string selectedPath)
-    {
-        this.selectedPath = selectedPath;
-    }
-
     public virtual bool IncludesMembers => false;
-
-    protected void SetSelectedPath(string path)
-    {
-        selectedPath = path;
-    }
 
     public IEnumerable<IMember> SelectMembers(INode currentNode, IEnumerable<IMember> selectedMembers,
         MemberSelectionContext context)
     {
-        string currentPath = currentNode.PathAndName;
-
-        // If we're part of a collection comparison, the selected path will not include an index,
-        // so we need to remove it from the current node as well.
-        if (!ContainsIndexingQualifiers(selectedPath))
-        {
-            currentPath = RemoveIndexQualifiers(currentPath);
-        }
-
+        var currentPath = RemoveIndexQualifiers(currentNode.PathAndName);
         var members = selectedMembers.ToList();
         AddOrRemoveMembersFrom(members, currentNode, currentPath, context);
 
@@ -42,11 +21,6 @@ internal abstract class SelectMemberByPathSelectionRule : IMemberSelectionRule
     protected abstract void AddOrRemoveMembersFrom(List<IMember> selectedMembers,
         INode parent, string parentPath,
         MemberSelectionContext context);
-
-    private static bool ContainsIndexingQualifiers(string path)
-    {
-        return path.Contains("[", StringComparison.Ordinal) && path.Contains("]", StringComparison.Ordinal);
-    }
 
     private static string RemoveIndexQualifiers(string path)
     {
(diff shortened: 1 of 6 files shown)


Src/FluentAssertions/Equivalency/Selection/SelectMemberByPathSelectionRule.cs
```diff
@@ -7,32 +6,12 @@ namespace FluentAssertions.Equivalency.Selection;
...
-    }
-
     public IEnumerable<IMember> SelectMembers(INode currentNode, IEnumerable<IMember> selectedMembers,
         MemberSelectionContext context)
     {
-        string currentPath = currentNode.PathAndName;
-
-        // If we're part of a collection comparison, the selected path will not include an index,
-        // so we need to remove it from the current node as well.
-        if (!ContainsIndexingQualifiers(selectedPath))
```
whymatter, line 31: This is the part where I am not sure about the implications, since all tests are green

whymatter:
Hmm, I don´t understand it. `RemoveIndexQualifiers` does only something if the root is a collection, right? So that the following is true 
```c#
RemoveIndexQualifiers("root[0].Foo[2].Bar") = "root.Foo[2].Bar" // root is collection
RemoveIndexQualifiers("root.Foo[2].Bar") => "root.Foo[2].Bar".  // root is no collection
```
The if condition I removed prevents the first case from happening if I add an exclusion like `x => x.Foo[2].Bar`. Why?
What is a test case for the if condition? I don't get it :(

dennisdoomen:
> Hmm, I don´t understand it. `RemoveIndexQualifiers` does only something if the root is a collection, right?

No, it was called when the path contained _any_ `[` and `]`, regardless of the root. This was used in _expressions_, not strings.

So `opt.Excluding(x => x.Foo[2].Bar)` would get normalized to `root.Foo[].Bar`, which is the old way of doing `opt.For(x => x.Foo).Excluding(y => .Bar)`.

whymatter:
So is my PR fine then?

dennisdoomen:
> So is my PR fine then?

I'm not sure anymore if my memory is serving me well. But the issue was that if we were using `BeEquivalentTo` on a collection, the caller could provide the property path for including or excluding a property without the first index (e.g. `root[0]`). Instead, it would specify the path that assumes an item of the collection. But the actual path of the node in the graph _will_ include `root[0]` or something like that.

whymatter:
> Also, if I remove that part, I have about 30 tests failing

I assume you removed the line `currentPath = RemoveIndexQualifiers(currentPath);` as well? I just removed the if condition so that `currentPath = RemoveIndexQualifiers(currentPath);` executes all the time

dennisdoomen:
> I assume you removed the line `currentPath = RemoveIndexQualifiers(currentPath);` as well? I just removed the if condition so that `currentPath = RemoveIndexQualifiers(currentPath);` executes all the time

Yeah, I removed the entire `if` statement.

whymatter:
> > I assume you removed the line `currentPath = RemoveIndexQualifiers(currentPath);` as well? I just removed the if condition so that `currentPath = RemoveIndexQualifiers(currentPath);` executes all the time
> 
> Yeah, I removed the entire `if` statement.

Okay, so because `currentPath = RemoveIndexQualifiers(currentPath);` is definitely required.
However, I do not understand if and why the `if` surrounding it is required? (Especially because removing the `if` resolves the bug I am working on)

dennisdoomen:
> However, I do not understand if and why the `if` surrounding it is required? (Especially because removing the `if` resolves the bug I am working on)

Me neither. So if all the tests succeed, I guess this should be good enough. I can't think of a scenario where we would not want to remove the starting `[0]`

dennisdoomen (approved):

Src/FluentAssertions/Equivalency/Selection/SelectMemberByPathSelectionRule.cs
```diff
@@ -7,32 +6,12 @@ namespace FluentAssertions.Equivalency.Selection;
...
-        {
-            currentPath = RemoveIndexQualifiers(currentPath);
-        }
-
+        var currentPath = RemoveIndexQualifiers(currentNode.PathAndName);
```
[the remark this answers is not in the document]
whymatter: I like it, however, renamed it to the singular `RemoveRootIndexQualifier`

docs/_pages/releases.md
```diff
@@ -15,6 +15,8 @@ sidebar:
 ### Fixes
 * Improved robustness of several assertions when they're wrapped in an `AssertionScope` - [#2133](https://github.com/fluentassertions/fluentassertions/pull/2133)
 
+* Fixed `.Excluding()`, `.For().Exclude()` not working if root is a collection - [#2101](https://github.com/fluentassertions/fluentassertions/pull/2135)
```
jnyrup, line 18:
```suggestion
* Fixed `.Excluding()` and `.For().Exclude()` not working if root is a collection - [#2135](https://github.com/fluentassertions/fluentassertions/pull/2135)
```

PR #2136, opened 2023-03-05
Preparing upgrade of analyzers

jnyrup:
## IMPORTANT 

I'm working on migrating our code from the now deprecated `Microsoft.CodeAnalysis.FxCopAnalyzers` and into [.NET analyzers](https://learn.microsoft.com/en-gb/visualstudio/code-quality/migrate-from-fxcop-analyzers-to-net-analyzers?view=vs-2019).

This PR focuses on the lowest hanging fruit as the remaining ones might require a discussion about whether we should disable the diagnostics globally or on a case by case basis.

### DIFF
--- a/Src/FluentAssertions/Equivalency/Steps/DataRowEquivalencyStep.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/DataRowEquivalencyStep.cs
@@ -195,9 +195,9 @@ public class DataRowEquivalencyStep : EquivalencyStep<DataRow>
     {
         var cacheKey = (comparands.CompileTimeType, comparands.RuntimeType, config);
 
-        if (!SelectedMembersCache.TryGetValue(cacheKey, out SelectedDataRowMembers selectedMembers))
+        if (!SelectedMembersCache.TryGetValue(cacheKey, out SelectedDataRowMembers selectedDataRowMembers))
         {
-            var members = Enumerable.Empty<IMember>();
+            IEnumerable<IMember> members = Enumerable.Empty<IMember>();
 
             foreach (IMemberSelectionRule rule in config.SelectionRules)
             {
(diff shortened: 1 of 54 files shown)


Src/FluentAssertions/Events/EventRecorder.cs
```diff
@@ -65,10 +65,10 @@ public void Attach(WeakReference subject, EventInfo eventInfo)
 
     public void Dispose()
     {
-        if (cleanup is not null)
+        if (this.cleanup is { } cleanup)
```
dennisdoomen, line 68: I actually think this makes it _less_ readable.

Src/FluentAssertions/Equivalency/Steps/DataRowEquivalencyStep.cs
```diff
@@ -205,15 +205,34 @@ private class SelectedDataRowMembers
                     new MemberSelectionContext(comparands.CompileTimeType, comparands.RuntimeType, config));
             }
 
-            selectedMembers = new SelectedDataRowMembers
-            {
-                HasErrors = members.Any(m => m.Name == nameof(DataRow.HasErrors)),
-                RowState = members.Any(m => m.Name == nameof(DataRow.RowState))
-            };
+            selectedMembers = GetSelectedDataRowMembers(members);
```
dennisdoomen, line 208: Wouldn't a simple `ToArray()` not be enough?

jnyrup: That forces a _complete_ enumeration of `members` where as this uses _at most_ a single enumeration, but I agree with your preference in using `ToArray`.

Src/FluentAssertions/Events/EventRecorder.cs
jnyrup: Would you prefer something like this?

```cs
Action localCleanup = cleanup;
if (localCleanup is not null)
{
    localCleanup();
    cleanup = null;
    EventObject = null;
    raisedEvents.Dispose();
}
```

dennisdoomen: Yeah, even though that's ugly too, it's _less_ difficult to read than comparing with `{ }`

Src/FluentAssertions/Equivalency/Steps/DataRowEquivalencyStep.cs
dennisdoomen: I reckon that's negligible with the kinds of assertions people would be doing.

PR #2145, opened 2023-03-11
Ensured the maximum depth is properly respecting the "official" definition of depth

### LINKED ISSUE #1993: MaxDepth in EquivalencyValidator.
https://github.com/fluentassertions/fluentassertions/blob/22b82e738c527e5e83edd80dd6a4794f6049e8d5/Src/FluentAssertions/Equivalency/EquivalencyValidator.cs#L48

MaxDepth is set to 10. 
However, when the current node is at depth 10, it would return false. 
Depth is however 0-based, so depth 9 is already 10 levels, but that is not clear in this line of code.

@dennisdoomen

dennisdoomen:
* Renamed some of the cyclic reference detection specs
* Removed two overlapping specs
* Moved two specs which had nothing to do with cyclic reference detection to `BasicSpecs`
* Added specs to properly cover the edge cases for deep object graphs.
* Ensured that the maximum depth is properly respecting the "official" definition of depth

Fixes #1993

### DIFF
--- a/Src/FluentAssertions/Equivalency/EquivalencyValidator.cs
+++ b/Src/FluentAssertions/Equivalency/EquivalencyValidator.cs
@@ -45,11 +45,11 @@ public class EquivalencyValidator : IEquivalencyValidator
     private static bool ShouldCompareNodesThisDeep(INode currentNode, IEquivalencyAssertionOptions options,
         AssertionScope assertionScope)
     {
-        bool shouldRecurse = options.AllowInfiniteRecursion || currentNode.Depth < MaxDepth;
+        bool shouldRecurse = options.AllowInfiniteRecursion || currentNode.Depth <= MaxDepth;
 
         if (!shouldRecurse)
         {
-            assertionScope.FailWith("The maximum recursion depth was reached.  ");
+            assertionScope.FailWith($"The maximum recursion depth of {MaxDepth} was reached.  ");
         }
 
         return shouldRecurse;
--- a/Tests/FluentAssertions.Equivalency.Specs/CyclicReferencesSpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/CyclicReferencesSpecs.cs
@@ -9,7 +9,7 @@ namespace FluentAssertions.Equivalency.Specs;
 public class CyclicReferencesSpecs
 {
     [Fact]
-    public void When_validating_nested_properties_that_have_cyclic_references_it_should_throw()
+    public void By_default_cyclic_references_are_not_valid()
     {
         // Arrange
         var cyclicRoot = new CyclicRoot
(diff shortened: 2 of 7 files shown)


Tests/FluentAssertions.Equivalency.Specs/CyclicReferencesSpecs.cs
```diff
@@ -310,79 +229,6 @@ IEnumerator IEnumerable.GetEnumerator()
...
-        public LogbookCode(string key)
-        {
-            Key = key;
-        }
-
-        public string Key { get; protected set; }
-    }
-
-    [Fact]
-    public void When_the_root_object_is_referenced_from_a_nested_object_it_should_treat_it_as_a_cyclic_reference()
```
jnyrup, line 336: Is this covered elsewhere?
It looks related to `When_the_root_object_is_referenced_from_an_object_in_a_nested_collection_it_should_treat_it_as_a_cyclic_reference` which is the only other test that references `MyCompany`, `MyUser` and `MyCompanyLogo`.

dennisdoomen: I couldn't think of a scenario where this test is doing something else than `Cyclic_references_can_be_ignored_in_equivalency_assertions` and the code coverage confirmed that.

jnyrup: There are limits to what information code coverage can provide 🙃 

`When_the_root_object_is_referenced_from_a_nested_object_it_should_treat_it_as_a_cyclic_reference` was introduced in 02e71b30159e34b0484dac40589633fe49256466 and is one of three tests to exercise `IgnoringCyclicReferences()`

`When_asserting_equivilence_on_objects_needing_high_recursion_depth_and_disabling_recursion_depth_limit_it_should_recurse_to_completion` was introduced in 36740ca412815e2c90a53632c88af00969fdb07e and is one of three tests to exercise `AllowingInfiniteRecursion()`

dennisdoomen: There's another `When_the_root_object_is_referenced_from_a_nested_object_it_should_treat_it_as_a_cyclic_reference` as well, which is doing the same thing.

dennisdoomen: Just to be on the safe side, I've restored that test, but renamed it to `Can_detect_cyclic_references_in_nested_objects_referring_to_the_root`

Src/FluentAssertions/Equivalency/EquivalencyValidator.cs
```diff
@@ -45,11 +45,11 @@ public void RecursivelyAssertEquality(Comparands comparands, IEquivalencyValidat
     private static bool ShouldCompareNodesThisDeep(INode currentNode, IEquivalencyAssertionOptions options,
         AssertionScope assertionScope)
     {
-        bool shouldRecurse = options.AllowInfiniteRecursion || currentNode.Depth < MaxDepth;
+        bool shouldRecurse = options.AllowInfiniteRecursion || currentNode.Depth <= MaxDepth;
```
jnyrup, line 48: Add a line to release notes about this?

jnyrup (approved):

PR #2147, opened 2023-03-20
Add support to convert into to enum

### LINKED ISSUE #2011: Comparing enum by value doesn't work even with auto conversion
**Comparing enum by value doesn't work even with auto conversion**

### Description

Testing object graph containing different enum type, but having the same value, it will not pass even auto conversion is enabled.

This is related with issue #1204.

### Reproducing the issue

Pay attention at `TemplateNo` when compared to `TemplateDto.Id` (`int`). I expect the old behavior (v5*) to work again, if `TemplateDto.Id` is 2, then it should be equal to `TemplateNo.AccountCreated`.

```csharp
public enum TemplateNo
{
    AccountCreated = 2,
    AccountDeactivated = 3
}

public class TemplateDomain
{
    public TemplateNo No { get; set; }
    public string Subject { get; set; } = string.Empty;
}

public class TemplateDto
{
    public int Id { get; set; }
    public string EmailSubject { get; set; } = string.Empty;
}

[Fact]
public void EnumTest() {
    // arrange
    var expected = new TemplateDomain { No = TemplateNo.AccountCreated, Subject = "Account created" };
    var actual = new TemplateDto { Id = 2, EmailSubject = "Account created" };

    // assert
    actual.Should().BeEquivalentTo(expected, opt => opt
        .WithMapping<TemplateDto>(domain => domain.No, dto => dto.Id)
        .WithMapping<TemplateDto>(domain => domain.Subject, dto => dto.EmailSubject)
        .ComparingEnumsByValue().WithAutoConversion()));
    // or
    actual.Should().BeEquivalentTo(expected, opt => opt
        .WithMapping<TemplateDto>(domain => domain.No, dto => dto.Id)
        .WithMapping<TemplateDto>(domain => domain.Subject, dto => dto.EmailSubject)
        .ComparingEnumsByValue().WithAutoConversionFor(o => o.Path.Contains(nameof(TemplateDomain.No))));
}
```

### Expected behavior:

Test should pass when enum is set to test by value and auto conversion is enabled

### Actual behavior:

```
Xunit.Sdk.XunitException
Expected property actual.Id to be equivalent to TemplateNo.AccountCreated {value: 2}, but found 2.

With configuration:
- Compare enums by value
- FluentAssertions.Equivalency.Matching.MappedPathMatchingRule
- FluentAssertions.Equivalency.Matching.MappedPathMatchingRule
```

### Versions

* Fluent Assertion 6.7.0
* .NET 6

### Additional info

It's breaking change for [v6.0](https://fluentassertions.com/upgradingtov6#enums). Currently I use equivalency step for workaround:
```cs
/// <summary>
/// See the problem https://fluentassertions.com/upgradingtov6#enums
/// and solution https://stackoverflow.com/questions/71131277/fluentassertions-6-objectgraph-compare-enum-to-string#answer-71134315
/// </summary>
public class EnumVersusIntegerStep : IEquivalencyStep
{
    public EquivalencyResult Handle(Comparands comparands, IEquivalencyValidationContext context, IEquivalencyValidator nestedValidator) {
        if (comparands.Subject is int subject && comparands.Expectation?.GetType().IsEnum == true) {
            AssertionScope.Current
                .ForCondition(subject == Convert.ToInt32(comparands.Expectation))
                .FailWith(() =>
                {
                    decimal? subjectsUnderlyingValue = ExtractDecimal(comparands.Subject);
                    decimal? expectationsUnderlyingValue = ExtractDecimal(comparands.Expectation);

                    string subjectsName = GetDisplayNameForEnumComparison(comparands.Subject, subjectsUnderlyingValue);
                    string expectationName = GetDisplayNameForEnumComparison(comparands.Expectation, expectationsUnderlyingValue);
                    return new($"Expected {{context:string}} to be equivalent to {expectationName}{{reason}}, but found {subjectsName}.");
                });

            return EquivalencyResult.AssertionCompleted;
        }

        if (comparands.Subject?.GetType().IsEnum == true && comparands.Expectation is int expectation) {
            AssertionScope.Current
                .ForCondition(Convert.ToInt32(comparands.Subject) == expectation)
                .FailWith(() =>
                {
                    decimal? subjectsUnderlyingValue = ExtractDecimal(comparands.Subject);
                    decimal? expectationsUnderlyingValue = ExtractDecimal(comparands.Expectation);

                    string subjectsName = GetDisplayNameForEnumComparison(comparands.Subject, subjectsUnderlyingValue);
                    string expectationName = GetDisplayNameForEnumComparison(comparands.Expectation, expectationsUnderlyingValue);
                    return new($"Expected {{context:enum}} to be equivalent to {expectationName}{{reason}}, but found {subjectsName}.");
                });

            return EquivalencyResult.AssertionCompleted;
        }

    }

    internal static string GetDisplayNameForEnumComparison(object o, decimal? v) {
        if (v is null) {
            return '\"' + o.ToString() + '\"';
        }

        string typePart = o.GetType().Name;
        string namePart = o.ToString()!.Replace(", ", "|", StringComparison.Ordinal);
        string valuePart = v.Value.ToString(CultureInfo.InvariantCulture);
        return $"{typePart}.{namePart} {{{{value: {valuePart}}}}}";
    }

    internal static decimal? ExtractDecimal(object o) => o.GetType().IsEnum
        ? Convert.ToDecimal(o, CultureInfo.InvariantCulture)
        : null;
}
```
then test it like this
```cs
// assert
actual.Should().BeEquivalentTo(expected, opt => opt
    .WithMapping<TemplateDto>(domain => domain.No, dto => dto.Id)
    .WithMapping<TemplateDto>(domain => domain.Subject, dto => dto.EmailSubject)
    .Using(new EnumVersusIntegerStep())));
```
I know I can make it pass right now. But it would be nice to see simpler solution. I will remove my workaround so I don't have to maintain any extra code and less explanation to my teammate for what I'm doing.

sdelarosbil:
Fixes #2011 

There's not really a way for ChangeType to handle this case so it is handled specifically.

### DIFF
--- a/Src/FluentAssertions/Equivalency/Steps/AutoConversionStep.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/AutoConversionStep.cs
@@ -56,6 +56,19 @@ public class AutoConversionStep : IEquivalencyStep
 
         try
         {
+            if (expectationType.IsEnum)
+            {
+                if (Enum.IsDefined(expectationType, subject))
+                {
+                    conversionResult = Enum.ToObject(expectationType, subject);
+                    return true;
+                }
+                else
+                {
+                    return false;
+                }
+            }
+
             conversionResult = Convert.ChangeType(subject, expectationType, CultureInfo.InvariantCulture);
             return true;
         }
--- a/Tests/FluentAssertions.Equivalency.Specs/MemberConversionSpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/MemberConversionSpecs.cs
@@ -52,6 +52,32 @@ public class MemberConversionSpecs
     }
 
     [Fact]
+    public void Numbers_can_be_converted_to_enums()
+    {
+        // Arrange
+        var expectation = new { Property = EnumFour.Three };
+        var subject = new { Property = 3 };
+
+        // Act / Assert
+        subject.Should().BeEquivalentTo(expectation, options => options.WithAutoConversion());
+    }
+
+    [Fact]
+    public void Numbers_that_are_out_of_range_cannot_be_converted_to_enums()
+    {
+        // Arrange
+        var expectation = new { Property = EnumFour.Three };
+        var subject = new { Property = 4 };
+
+        // Act
+        Action act = () => subject.Should().BeEquivalentTo(expectation, options => options.WithAutoConversion());
+
+        // Assert
+        act.Should().Throw<XunitException>()
+            .WithMessage("*subject*Property*EnumFour*");
+    }
+
+    [Fact]
     public void When_injecting_a_null_predicate_into_WithAutoConversionFor_it_should_throw()
     {
         // Arrange
--- a/docs/_pages/releases.md
+++ b/docs/_pages/releases.md
@@ -11,6 +11,7 @@ sidebar:
 
 ### What's new
 * Added `ThrowWithinAsync` for assertions on `Task` - [#1974](https://github.com/fluentassertions/fluentassertions/pull/1974)
+* Added support for converting integers to enums using `AutoConversion` - [#2147](https://github.com/fluentassertions/fluentassertions/pull/2147)
 
 ### Fixes
 * Improved robustness of several assertions when they're wrapped in an `AssertionScope` - [#2133](https://github.com/fluentassertions/fluentassertions/pull/2133)


Tests/FluentAssertions.Equivalency.Specs/MemberConversionSpecs.cs
```diff
@@ -51,6 +51,20 @@ public void When_an_int_is_compared_equivalent_to_a_string_representing_the_numb
         act.Should().NotThrow();
     }
 
+    [Fact]
+    public void When_an_int_is_compared_equivalent_to_an_enum_representing_the_int_it_should_pass()
```
dennisdoomen, line 55:
```suggestion
    public void Numbers_can_be_converted_to_enums()
```

Src/FluentAssertions/Equivalency/Steps/AutoConversionStep.cs
```diff
@@ -56,6 +56,18 @@ private static bool TryChangeType(object subject, Type expectationType, out obje
...
         {
+            if (expectationType.IsEnum)
+            {
+                if (Enum.IsDefined(expectationType, subject))
+                {
+                    subject = Enum.ToObject(expectationType, subject);
+                }
+                else
+                {
+                    return false;
```
dennisdoomen, line 67: 🔧 I miss a test that covers this path.

dennisdoomen (changes requested):
Thanks for the PR.

sdelarosbil:
I added a test that should cover this line

Tests/FluentAssertions.Equivalency.Specs/MemberConversionSpecs.cs
```diff
@@ -51,6 +51,35 @@ public void When_an_int_is_compared_equivalent_to_a_string_representing_the_numb
...
+    {
+        // Arrange
+        var expectation = new { Property = EnumFour.Three };
+        var subject = new { Property = 3 };
+
+        // Act
+        Action act = () => subject.Should().BeEquivalentTo(expectation, options => options.WithAutoConversion());
+
+        // Assert
+        act.Should().NotThrow();
```
IT-VBFK, line 65: You could omit the `Action` and call it directly in that case

like:
```csharp
// Act / Assert
subject.Should().BeEquivalentTo(expectation, options => options.WithAutoConversion());
```

dennisdoomen: Yes, that's true. We don't explicitly verify that an API does _not_ throw.

dennisdoomen (approved):

IT-VBFK:
@sdelarosbil Can you add a note in the [releases.md](https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/releases.md) ?

sdelarosbil:
> @sdelarosbil Can you add a note in the [releases.md](https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/releases.md) ?

Done.

Src/FluentAssertions/Equivalency/Steps/AutoConversionStep.cs
```diff
@@ -56,6 +56,18 @@ private static bool TryChangeType(object subject, Type expectationType, out obje
 
         try
         {
+            if (expectationType.IsEnum)
+            {
+                if (Enum.IsDefined(expectationType, subject))
+                {
+                    subject = Enum.ToObject(expectationType, subject);
```
jnyrup, line 63: Can we assign the result to `conversionResult` instead and return early?

jnyrup (approved):

PR #2150, opened 2023-03-23
Use ToString to format exceptions

### LINKED ISSUE #2015: InnerException information missing in output when using Should().NotThrow()
I have a unit test that exercises dependency resolution in Autofac. The test itself is simple:

```cs
// This enumerator ensures that this test case is invoked once per registered concrete type
[TestCaseSource(typeof(ConcreteTypeEnumerator))]
public void Service_should_be_instantiable(Type service)
{
    // This just sets up all the registrations from my app in a reusable way
    using var container = new CompositionRoot().Setup().Container;
    var act = () => container.Resolve(service);
    act.Should().NotThrow();
}
```

Then a specific iteration of this test *fails*, I get the below, truncated output:

```
Did not expect any exception, but found Autofac.Core.DependencyResolutionException with message "An exception was thrown while activating Recyclarr.Command.Helpers.CacheStoragePath."
     at Autofac.Core.Resolving.Middleware.ActivatorErrorHandlingMiddleware.Execute(ResolveRequestContext context, Action`1 next)
```

Now change the above example to this:

```cs
[TestCaseSource(typeof(ConcreteTypeEnumerator))]
public void Service_should_be_instantiable(Type service)
{
    using var container = new CompositionRoot().Setup(RegisterAdditionalServices).Container;
    container.Resolve(service);
}
```

This will throw the same exception, except this time it goes to NUnit3 untouched instead of through FluentAssertions. You now get this output:

```txt
Autofac.Core.DependencyResolutionException : An exception was thrown while activating Recyclarr.Command.Helpers.CacheStoragePath.
Data:
  ActivatorChain: Recyclarr.Command.Helpers.CacheStoragePath
  ----> Autofac.Core.DependencyResolutionException : None of the constructors found with 'Autofac.Core.Activators.Reflection.DefaultConstructorFinder' on type 'Recyclarr.Command.Helpers.CacheStoragePath' can be invoked with the available services and parameters:
Cannot resolve parameter 'Recyclarr.Command.IServiceCommand serviceCommand' of constructor 'Void .ctor(TrashLib.Startup.IAppPaths, Recyclarr.Command.IServiceCommand)'.
   at Autofac.Core.Resolving.Middleware.ActivatorErrorHandlingMiddleware.Execute(ResolveRequestContext context, Action`1 next)
```

As you can see, there's more detail in the second case. This additional detail is of particular importance for Autofac, because without it, I can't tell exactly why the resolution failed.

If I stop in the debugger, I see that `DependencyResolutionException.InnerException` seems to have the additional detail. But for some reason that's omitted and/or the exception output is truncated.

Please address the issue by ensuring that InnerException details are printed to the console in this case.

### Versions:

* Which version of Fluent Assertions are you using? `6.7.0`
* Which .NET runtime and version are you targeting? .NET 6

sdelarosbil:
Fixes #2015 

Also modified some tests so the new formatting matches. It turns out that the current formatting scheme was not as good as simply using ToString.

Formatter before:
```
System.Exception with message "OuterExceptionMessage"
```

After:
```
System.Exception: OuterExceptionMessage
 ---> System.InvalidOperationException: InnerExceptionMessage
   --- End of inner exception stack trace ---
```

### DIFF
--- a/Src/FluentAssertions/Formatting/ExceptionValueFormatter.cs
+++ b/Src/FluentAssertions/Formatting/ExceptionValueFormatter.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using static System.FormattableString;
 
 namespace FluentAssertions.Formatting;
 
@@ -19,16 +18,6 @@ public class ExceptionValueFormatter : IValueFormatter
 
     public void Format(object value, FormattedObjectGraph formattedGraph, FormattingContext context, FormatChild formatChild)
     {
-        var exception = (Exception)value;
-
-        formattedGraph.AddFragment(Invariant($"{exception.GetType().FullName} with message \"{exception.Message}\""));
-
-        if (exception.StackTrace is not null)
-        {
-            foreach (string line in exception.StackTrace.Split(new[] { Environment.NewLine }, StringSplitOptions.None))
-            {
-                formattedGraph.AddLine("  " + line);
-            }
-        }
+        formattedGraph.AddFragment(((Exception)value).ToString());
     }
 }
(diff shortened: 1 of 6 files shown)


sdelarosbil:
Added an example of the differences.

jnyrup (approved):

dennisdoomen (approved):

PR #2151, opened 2023-03-24
Add WithoutStrictOrderingFor overload with an expression

chrischu:
I hope it's okay that I circumvented the "separate issue with api-approved label" since the feature I'm adding is essentially already there for "WithStrictOrderingFor".

### DIFF
--- a/Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
@@ -75,6 +75,21 @@ public class EquivalencyAssertionOptions<TExpectation> :
     }
 
     /// <summary>
+    /// Causes the collection identified by <paramref name="expression"/> to be compared in the order
+    /// in which the items appear in the expectation.
+    /// </summary>
+    public EquivalencyAssertionOptions<TExpectation> WithoutStrictOrderingFor(
+        Expression<Func<TExpectation, object>> expression)
+    {
+        string expressionMemberPath = expression.GetMemberPath().ToString();
+        OrderingRules.Add(new PathBasedOrderingRule(expressionMemberPath)
+        {
+            Invert = true
+        });
+        return this;
+    }
+
+    /// <summary>
     /// Creates a new set of options based on the current instance which acts on a a collection of the <typeparamref name="TExpectation"/>.
     /// </summary>
     public EquivalencyAssertionOptions<IEnumerable<TExpectation>> AsCollection()
--- a/Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
+++ b/Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
@@ -75,7 +75,7 @@ public class EquivalencyAssertionOptions<TExpectation> :
     }
 
     /// <summary>
-    /// Causes the collection identified by <paramref name="expression"/> to be compared in the order
+    /// Causes the collection identified by <paramref name="expression"/> to be compared ignoring the order
     /// in which the items appear in the expectation.
     /// </summary>
     public EquivalencyAssertionOptions<TExpectation> WithoutStrictOrderingFor(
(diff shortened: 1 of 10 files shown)


Src/FluentAssertions/Equivalency/Ordering/PathBasedOrderingRule.cs
```diff
@@ -16,6 +16,8 @@ public PathBasedOrderingRule(string path)
         this.path = path;
     }
 
+    public bool Invert { get; set; }
```
lg2de, line 19: `Strict` ?

jnyrup: `Invert` matches how it's named in `PredicateBasedOrderingRule`.
https://github.com/fluentassertions/fluentassertions/blob/81d3fbf7154f8d9cf80ca0fa72759f863132cf94/Src/FluentAssertions/Equivalency/Ordering/PredicateBasedOrderingRule.cs#L17

Src/FluentAssertions/Equivalency/EquivalencyAssertionOptions.cs
```diff
@@ -74,6 +74,21 @@ public EquivalencyAssertionOptions<TExpectation> Including(Expression<Func<TExpe
         return this;
     }
 
+    /// <summary>
+    /// Causes the collection identified by <paramref name="expression"/> to be compared in the order
+    /// in which the items appear in the expectation.
```
jnyrup, line 79:
```suggestion
    /// Causes the collection identified by <paramref name="expression"/> to be compared ignoring the order
    /// in which the items appear in the expectation.
```

jnyrup (approved):
Thanks for adding this missing overload.
Please add a line to the [release notes](https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/releases.md)

Apparently I've known this shortcoming this since https://github.com/fluentassertions/fluentassertions/pull/974#issuecomment-441101713 but forgot to to anything about it.

jnyrup:
Code Coverage points out that we need a failing test for this to hit both cases of `Invert` in `ToString()`

chrischu:
> Code Coverage points out that we need a failing test for this to hit both cases of `Invert` in `ToString()`

I tried changing that but could not figure out where the coverage for the `Invert` property of `PredicateBasedOrderingRule` comes from (because I thought I could use that as a guide). 

It might even be impossible to cover the change, as the rule will only ever throw if `Invert` is `false`. Maybe the fix would be actually not vary the message based on the `Invert` property? I'll implement that change :).

Src/FluentAssertions/Equivalency/Ordering/PathBasedOrderingRule.cs
```diff
@@ -60,6 +63,7 @@ private string RemoveInitialIndexQualifier(string sourcePath)
 
     public override string ToString()
```
jnyrup, line 64: Something like this should hits the path.
```cs
[Fact]
public void Negated()
{
    // Arrange
    var subject = new
    {
        UnorderedCollection = new[] { 1 }
    };

    var expectation = new
    {
        UnorderedCollection = new[] { 2 }
    };

    // Act
    Action action = () => subject.Should().BeEquivalentTo(expectation, options => options
        .WithStrictOrdering()
        .WithoutStrictOrderingFor(x => x.UnorderedCollection));

    // Assert
    action.Should().Throw<XunitException>()
        .WithMessage("*not strict*");
}
```

Src/FluentAssertions/Equivalency/Ordering/PathBasedOrderingRule.cs
dennisdoomen, line 19: ♻️ You can change the setter to `init`

dennisdoomen (approved):
Only a suggestion from my IDE.

Src/FluentAssertions/Equivalency/Ordering/PathBasedOrderingRule.cs
jnyrup: I'll fix these in another PR

PR #2152, opened 2023-03-25
Ensure explicitly implemented members are discoverable in BeEquivalentTo

### LINKED ISSUE #2146: Explicitly implemented properties on the subject are not found during object graph comparison
```cs
interface IPerson
{
    string Name { get }
}

class Person : IPerson
{
    string IPerson.Name => "Bob";
}

IPerson person = new Person();

person.Should().BeEquivalentTo(new { Name = "Bob" });
```

Expected: Should not throw 
Actual: Throws with message "Expectation has property person.Name that the other object does not have."

dennisdoomen:
* When trying to find a property on the subject to match the one on the expectation, if a normal property exists next to an implicitly implemented one, the normal property will be selected.
* Some tests related to explicit interface implementations assumed a dependency on whether or not `RespectingDeclaredTypes` and `RespectingRuntimeTypes` was used. But those options don't affect the subject.
* The `WithMapping` construct also didn't support mapping to properties implemented through explicitly-implemented interfaces.
* Removed the "non-private" part names since visibility is now used to find members

**Note**   I had to extend `MemberVisibility` to optionally include explicitly implemented members. This _could_ be perceived as a breaking change. But it won't affect consumers, so for now, I treat it as non-breaking.

**Review**
Review by commit and with whitespace changes hidden

Fixes #2146 

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Common/TypeExtensions.cs
+++ b/Src/FluentAssertions/Common/TypeExtensions.cs
@@ -184,10 +184,10 @@ internal static class TypeExtensions
     /// </returns>
     public static PropertyInfo FindProperty(this Type type, string propertyName, MemberVisibility memberVisibility)
     {
-        var properties = type.GetNonPrivateProperties(memberVisibility);
+        var properties = type.GetProperties(memberVisibility);
 
         return Array.Find(properties, p =>
-            p.Name == propertyName || p.Name.EndsWith("." + propertyName, StringComparison.OrdinalIgnoreCase));
+            p.Name == propertyName || p.Name.EndsWith("." + propertyName, StringComparison.Ordinal));
     }
 
     /// <summary>
@@ -198,24 +198,24 @@ internal static class TypeExtensions
     /// </returns>
     public static FieldInfo FindField(this Type type, string fieldName, MemberVisibility memberVisibility)
     {
-        var fields = type.GetNonPrivateFields(memberVisibility);
+        var fields = type.GetFields(memberVisibility);
 
         return Array.Find(fields, p => p.Name == fieldName);
     }
 
-    public static MemberInfo[] GetNonPrivateMembers(this Type typeToReflect, MemberVisibility visibility)
+    public static MemberInfo[] GetMembers(this Type typeToReflect, MemberVisibility visibility)
     {
-        return GetTypeReflectorFor(typeToReflect, visibility).NonPrivateMembers;
+        return GetTypeReflectorFor(typeToReflect, visibility).Members;
     }
 
-    public static PropertyInfo[] GetNonPrivateProperties(this Type typeToReflect, MemberVisibility visibility)
+    public static PropertyInfo[] GetProperties(this Type typeToReflect, MemberVisibility visibility)
     {
-        return GetTypeReflectorFor(typeToReflect, visibility).NonPrivateProperties;
+        return GetTypeReflectorFor(typeToReflect, visibility).Properties;
     }
 
-    public static FieldInfo[] GetNonPrivateFields(this Type typeToReflect, MemberVisibility visibility)
+    public static FieldInfo[] GetFields(this Type typeToReflect, MemberVisibility visibility)
     {
-        return GetTypeReflectorFor(typeToReflect, visibility).NonPrivateFields;
+        return GetTypeReflectorFor(typeToReflect, visibility).Fields;
     }
 
     private static TypeMemberReflector GetTypeReflectorFor(Type typeToReflect, MemberVisibility visibility)
--- a/Src/FluentAssertions/Common/TypeExtensions.cs
+++ b/Src/FluentAssertions/Common/TypeExtensions.cs
@@ -15,9 +15,6 @@ internal static class TypeExtensions
     private const BindingFlags PublicInstanceMembersFlag =
         BindingFlags.Public | BindingFlags.Instance;
 
-    private const BindingFlags AllInstanceMembersFlag =
-        BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance;
-
     private const BindingFlags AllStaticAndInstanceMembersFlag =
         PublicInstanceMembersFlag | BindingFlags.NonPublic | BindingFlags.Static;
 
@@ -176,45 +173,34 @@ internal static class TypeExtensions
     }
 
     /// <summary>
-    /// Finds the property by a case-sensitive name.
+    /// Finds the property by a case-sensitive name and with a certain visibility.
     /// </summary>
+    /// <remarks>
+    /// If both a normal property and one that was implemented through an explicit interface implementation with the same name exist,
+    /// then the normal property will be returned.
+    /// </remarks>
     /// <returns>
     /// Returns <see langword="null"/> if no such property exists.
     /// </returns>
-    public static PropertyInfo FindProperty(this Type type, string propertyName)
+    public static PropertyInfo FindProperty(this Type type, string propertyName, MemberVisibility memberVisibility)
     {
-        while (type != typeof(object))
-        {
-            if (type.GetProperty(propertyName, AllInstanceMembersFlag | BindingFlags.DeclaredOnly) is { } property)
-            {
-                return property;
-            }
+        var properties = type.GetNonPrivateProperties(memberVisibility);
 
-            type = type.BaseType;
-        }
-
-        return null;
+        return Array.Find(properties, p =>
+            p.Name == propertyName || p.Name.EndsWith("." + propertyName, StringComparison.OrdinalIgnoreCase));
     }
 
     /// <summary>
     /// Finds the field by a case-sensitive name.
     /// </summary>
     /// <returns>
-    /// Returns <see langword="null"/> if no such property exists.
+    /// Returns <see langword="null"/> if no such field exists.
     /// </returns>
-    public static FieldInfo FindField(this Type type, string fieldName)
+    public static FieldInfo FindField(this Type type, string fieldName, MemberVisibility memberVisibility)
     {
-        while (type != typeof(object))
-        {
-            if (type.GetField(fieldName, AllInstanceMembersFlag | BindingFlags.DeclaredOnly) is { } field)
-            {
-                return field;
-            }
+        var fields = type.GetNonPrivateFields(memberVisibility);
 
-            type = type.BaseType;
-        }
-
-        return null;
+        return Array.Find(fields, p => p.Name == fieldName);
     }
 
     public static MemberInfo[] GetNonPrivateMembers(this Type typeToReflect, MemberVisibility visibility)
@@ -318,8 +304,11 @@ internal static class TypeExtensions
 
     public static ConstructorInfo GetConstructor(this Type type, IEnumerable<Type> parameterTypes)
     {
+        const BindingFlags allInstanceMembersFlag =
+            BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance;
+
         return type
-            .GetConstructors(AllInstanceMembersFlag)
+            .GetConstructors(allInstanceMembersFlag)
             .SingleOrDefault(m => m.GetParameters().Select(p => p.ParameterType).SequenceEqual(parameterTypes));
     }
 
--- a/Src/FluentAssertions/Common/TypeMemberReflector.cs
+++ b/Src/FluentAssertions/Common/TypeMemberReflector.cs
@@ -16,22 +16,22 @@ internal sealed class TypeMemberReflector
 
     public TypeMemberReflector(Type typeToReflect, MemberVisibility visibility)
     {
-        NonPrivateProperties = LoadNonPrivateProperties(typeToReflect, visibility);
-        NonPrivateFields = LoadNonPrivateFields(typeToReflect, visibility);
-        NonPrivateMembers = NonPrivateProperties.Concat<MemberInfo>(NonPrivateFields).ToArray();
+        Properties = LoadProperties(typeToReflect, visibility);
+        Fields = LoadFields(typeToReflect, visibility);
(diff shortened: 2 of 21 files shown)


Src/FluentAssertions/Common/TypeMemberReflector.cs
```diff
@@ -148,9 +157,17 @@ private static IEnumerable<FieldInfo> GetFieldsFromHierarchy(Type typeToReflect,
         return members;
     }
 
-    private static bool HasNonPrivateGetter(PropertyInfo propertyInfo)
+    private bool HasNonPrivateGetter(PropertyInfo propertyInfo)
```
jnyrup, line 160:
```suggestion
    private static bool HasNonPrivateGetter(PropertyInfo propertyInfo, MemberVisibility visibility)
```

Src/FluentAssertions/Common/TypeMemberReflector.cs
```diff
@@ -40,19 +40,23 @@ where HasNonPrivateGetter(propertyInfo)
...
         });
     }
 
-    private static FieldInfo[] LoadNonPrivateFields(Type typeToReflect, MemberVisibility visibility)
+    private static bool IsInternal(PropertyInfo property)
+    {
+        return property.GetMethod is { IsAssembly: true } or { IsFamilyOrAssembly: true };
+    }
+
+    private void LoadNonPrivateFields()
```
jnyrup, line 59:
```suggestion
    private static FieldInfo[] LoadNonPrivateFields(Type typeToReflect, MemberVisibility visibility)
```

Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -180,19 +180,12 @@ public static bool OverridesEquals(this Type type)
...
-                return property;
-            }
-
-            type = type.BaseType;
-        }
+        var properties = type.GetNonPrivateProperties(visibility);
 
-        return null;
+        return properties.FirstOrDefault(p =>
+            p.Name == propertyName || p.Name.EndsWith("." + propertyName, StringComparison.OrdinalIgnoreCase));
```
jnyrup, line 188:
```suggestion
        return Array.Find(properties, p =>
            p.Name == propertyName || p.Name.EndsWith("." + propertyName, StringComparison.OrdinalIgnoreCase));
```

Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -201,19 +194,12 @@ public static PropertyInfo FindProperty(this Type type, string propertyName)
...
-                return field;
-            }
-
-            type = type.BaseType;
-        }
+        var fields = type.GetNonPrivateFields(visibility);
 
-        return null;
+        return fields.FirstOrDefault(p =>
+            p.Name == fieldName || p.Name.EndsWith("." + fieldName, StringComparison.OrdinalIgnoreCase));
```
jnyrup, line 202:
```suggestion
        return Array.Find(fields, p =>
            p.Name == fieldName || p.Name.EndsWith("." + fieldName, StringComparison.OrdinalIgnoreCase));
```

Src/FluentAssertions/Common/TypeExtensions.cs
Corniel: I would even suggest to define aan extension Method for this, so that you ca do: properties.Find()

dennisdoomen: Nah, I would only do that when there's multiple uses and it would significantly improve readability.

Corniel: @dennisdoomen In some of my repos, I had multiple places where I did this. If that is not the case here, I fully agree.

Src/FluentAssertions/Common/TypeMemberReflector.cs
```diff
@@ -31,27 +31,45 @@ private static PropertyInfo[] LoadNonPrivateProperties(Type typeToReflect, Membe
...
-                .Where(property => includeInternals || property.GetMethod is { IsAssembly: false, IsFamilyOrAssembly: false })
+                .Where(property => includeInternal || !IsInternal(property))
+                .Select(property => new
+                {
+                    Property = property,
+                    IsExplicit = IsExplicitImplementation(property)
+                })
+                .OrderBy(x => x.IsExplicit)
+                .Select(x => x.Property)
                 .ToArray();
```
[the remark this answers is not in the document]
dennisdoomen: `IsExplicit` is a bool and initialized by the `Select`. So what's the problem here?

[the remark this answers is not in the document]
jn-foreflight: There's no problems in terms of corrects, but we can remove two `Select`s and still only invoke `IsExplicitImplementation` `n` times.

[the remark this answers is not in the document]
jnyrup: There's no problems in terms of correctness, but we can remove the two `Select`s and still only invoke `IsExplicitImplementation` `n` times.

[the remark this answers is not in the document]
dennisdoomen: Fixed it.

Src/FluentAssertions/Common/TypeMemberReflector.cs
```diff
@@ -31,27 +31,39 @@ private static PropertyInfo[] LoadNonPrivateProperties(Type typeToReflect, Membe
...
+    private static IEnumerable<PropertyInfo> GetPropertiesFromHierarchy(Type typeToReflect, MemberVisibility memberVisibility)
     {
-        bool includeInternals = memberVisibility.HasFlag(MemberVisibility.Internal);
+        bool includeInternal = memberVisibility.HasFlag(MemberVisibility.Internal);
 
         return GetMembersFromHierarchy(typeToReflect, type =>
         {
             return type
                 .GetProperties(AllInstanceMembersFlag | BindingFlags.DeclaredOnly)
-                .Where(property => property.GetMethod?.IsPrivate == false)
```
jnyrup, line 49: This also filter out properties without a getter.
How about we change it to just check if `property.GetMethod` is not null?

Src/FluentAssertions/Common/TypeMemberReflector.cs
```diff
@@ -31,27 +31,39 @@ private static PropertyInfo[] LoadNonPrivateProperties(Type typeToReflect, Membe
...
+    {
+        return property.GetMethod is { IsAssembly: true } or { IsFamilyOrAssembly: true };
+    }
+
+    private static bool IsExplicitImplementation(PropertyInfo property)
+    {
+        return property.GetMethod?.IsPrivate == true &&
+            property.SetMethod?.IsPrivate == true &&
+            property.Name.Contains('.', StringComparison.Ordinal);
+    }
```
jnyrup, line 65: Needs to be `property.SetMethod?.IsPrivate != false` to account for explicitly implemented properties without a setter.

If we're okay with letting this method know that we're filtered out properties without getters earlier (the commend above), we can remove the `?` from `GetMethod`

jnyrup (commented):
Here a my findings so far - I'll stop thinking for today.
I'm wondering if we should move this PR to V7 as it is changing quite a bit on the internals.
It would be

Src/FluentAssertions/Equivalency/Matching/TryMatchByNameRule.cs
```diff
@@ -10,14 +10,17 @@ internal class TryMatchByNameRule : IMemberMatchingRule
 {
     public IMember Match(IMember expectedMember, object subject, INode parent, IEquivalencyAssertionOptions options)
     {
-        PropertyInfo property = subject.GetType().FindProperty(expectedMember.Name);
+        PropertyInfo property = subject.GetType().FindProperty(expectedMember.Name,
+            options.IncludedProperties | MemberVisibility.ExplicitlyImplemented);
```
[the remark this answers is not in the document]
dennisdoomen: Good point. Apparently we don't have any tests for that. Will add them.

Src/FluentAssertions/Common/TypeMemberReflector.cs
```diff
@@ -123,10 +140,10 @@ private static List<FieldInfo> GetFieldsFromHierarchy(Type typeToReflect, Member
             members.InsertRange(0, newPropertyInfos);
         }
 
-        return members;
+        return members.ToArray();
```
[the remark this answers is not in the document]
dennisdoomen: Because Qodana wanted me to use a more abstract type than `List<TMemberINfo` as the return value of `GetInterfaceMembers`. I could try to change it to `IEnumerablt<T>`

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1779,10 +1779,33 @@ public void When_respecting_declared_types_explicit_interface_member_on_interfac
             subject.VehicleId = 1; // interface member
 
             // Act
-            Action action = () => subject.Should().BeEquivalentTo(expected, opt => opt.RespectingDeclaredTypes());
+            Action action = () => subject.Should().BeEquivalentTo(expected);
 
             // Assert
-            action.Should().NotThrow();
+            action.Should().Throw<XunitException>();
```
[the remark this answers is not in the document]
dennisdoomen: Because of the changed/improved behavior, the _normal_ property `VehicleId` will now be taking into account instead of the one that was explicitly implemented. And since the _normal_ `VehicleId` of the subject is 2, and `1` is expected, it _should_ throw.

dennisdoomen:
> I'm wondering if we should move this PR to V7 as it is changing quite a bit on the internals.

Given that you've found some more edge cases for which we don't even have tests yet, I'm fine moving this to v7 so we can expedite the final release. As far as I'm concerned, it's your call.

Src/FluentAssertions/Common/TypeMemberReflector.cs
dennisdoomen: Are you saying that we _should_ filter out properties without a getter?

dennisdoomen: What would be the added value?

Src/FluentAssertions/Common/TypeMemberReflector.cs
dennisdoomen: Added an extra test to cover this scenario.

Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -202,34 +196,27 @@ public static PropertyInfo FindProperty(this Type type, string propertyName)
...
-                return field;
-            }
+        var fields = type.GetFields(memberVisibility);
 
-            type = type.BaseType;
-        }
-
-        return null;
+        return Array.Find(fields, p =>
+            p.Name == fieldName || p.Name.EndsWith("." + fieldName, StringComparison.OrdinalIgnoreCase));
```
IT-VBFK, line 204: Should this also be `.OrdinalIgnoreCase` -> `.Ordinal`

dennisdoomen:
@jnyrup would be nice if you could give this one some attention soon...🙏

jnyrup:
I must admit I have postponed further reviewing this PR partly because it wants to solve multiple problems when using `BeEquivalentTo` together with explicitly implemented members and I've found it difficult to see what changes related to which problem.
Is it possible to split it further up?

E.g. `Can_find_explicitly_implemented_property_on_the_subject` from #2146 works if I rewrite `FindProperty` to

```cs
public static PropertyInfo FindProperty(this Type type, string propertyName)
{
    while (type != typeof(object))
    {
        PropertyInfo[] properties = type.GetProperties(AllInstanceMembersFlag | BindingFlags.DeclaredOnly);
        if (Array.Find(properties, e => e.Name == propertyName) is { } property)
        {
            return property;
        }

        if (Array.Find(properties, e => e.Name.EndsWith("." + propertyName, StringComparison.Ordinal)) is { } explicitProperty)
        {
            return explicitProperty;
        }

        type = type.BaseType;
    }

    return null;
}
```

Src/FluentAssertions/Common/TypeMemberReflector.cs
jnyrup: If `includeInternal` is true, we can now return a member without a getter, whereas currently `Where(property => property.GetMethod?.IsPrivate == false)` requires the property to have a getter.

But I see this doesn't matter in practice since `HasGetter` will discard any property without a getter .
image

So I withdraw my comments on this one.

Src/FluentAssertions/Common/TypeMemberReflector.cs
```diff
@@ -145,12 +162,18 @@ private static List<FieldInfo> GetFieldsFromHierarchy(Type typeToReflect, Member
             typeToReflect = typeToReflect.BaseType;
         }
 
-        return members;
+        return members.ToArray();
```
jnyrup, line 165: 🙃 Doesn't seem necessary

docs/_pages/releases.md
```diff
@@ -46,6 +46,7 @@ sidebar:
 
 ### Fixes
 * `because` and `becauseArgs` were not included in the error message when collections of enums were not equivalent - [#2214](https://github.com/fluentassertions/fluentassertions/pull/2214)
+* `BeEquivalentTo` will now find and can map subject properties that are implemented through an explicitly-implemented interface - [2152](https://github.com/fluentassertions/fluentassertions/pull/2152)
```
jnyrup, line 49:
```suggestion
* `BeEquivalentTo` will now find and can map subject properties that are implemented through an explicitly-implemented interface - [#2152](https://github.com/fluentassertions/fluentassertions/pull/2152)
```

jnyrup (approved):

PR #2154, opened 2023-03-27
Prevent `NullReferenceException`

jnyrup:
## IMPORTANT 

When `FailWith` does not cause immediate termination we must guard against it.
On line 33 we checked if `subjectMember` was null, but we might continue execution, so this PR guards against executing `subjectMember.IsBrowsable`.

The test _triggers_ the problem, but I failed to find a way to properly verify it in the assertion part 🤔

### DIFF
--- a/Src/FluentAssertions/Equivalency/Matching/MustMatchByNameRule.cs
+++ b/Src/FluentAssertions/Equivalency/Matching/MustMatchByNameRule.cs
@@ -35,12 +35,15 @@ internal class MustMatchByNameRule : IMemberMatchingRule
             Execute.Assertion.FailWith(
                 $"Expectation has {expectedMember.Description} that the other object does not have.");
         }
-
-        if (options.IgnoreNonBrowsableOnSubject && !subjectMember.IsBrowsable)
+        else if (options.IgnoreNonBrowsableOnSubject && !subjectMember.IsBrowsable)
         {
             Execute.Assertion.FailWith(
                 $"Expectation has {expectedMember.Description} that is non-browsable in the other object, and non-browsable " +
                 "members on the subject are ignored with the current configuration");
+        }
+        else
+        {
+            // Everything is fine
         }
 
         return subjectMember;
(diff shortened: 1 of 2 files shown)


Src/FluentAssertions/Equivalency/Matching/MustMatchByNameRule.cs
```diff
@@ -35,8 +35,7 @@ public IMember Match(IMember expectedMember, object subject, INode parent, IEqui
             Execute.Assertion.FailWith(
                 $"Expectation has {expectedMember.Description} that the other object does not have.");
         }
-
-        if (options.IgnoreNonBrowsableOnSubject && !subjectMember.IsBrowsable)
+        else if (options.IgnoreNonBrowsableOnSubject && !subjectMember.IsBrowsable)
```
dennisdoomen, line 38: 🔧 I also expected an additional (empty) `else` here.

dennisdoomen (changes requested):

Src/FluentAssertions/Equivalency/Matching/MustMatchByNameRule.cs
jnyrup: Not sure what you're saying here?
Something related to AV1537? (Finish every `if`-`else`-`if` statement with an `else` clause)

jnyrup: Like this?

```cs
if (subjectMember is null)
{
    Execute.Assertion.FailWith(
        $"Expectation has {expectedMember.Description} that the other object does not have.");
}
else if (options.IgnoreNonBrowsableOnSubject && !subjectMember.IsBrowsable)
{
    Execute.Assertion.FailWith(
        $"Expectation has {expectedMember.Description} that is non-browsable in the other object, and non-browsable " +
        "members on the subject are ignored with the current configuration");
}
else
{
    // Everything is fine
}
```

dennisdoomen: Yep. To make it explicit that you didn't forget about that scenario.

dennisdoomen:
> Whoops, too fast on the merge button - could have sworn I saw an approval somewhere.

That's what you get when shooting so many PRs from the hip ;-)

PR #2157, opened 2023-03-27
await `Task` when inside using block

jnyrup:
## IMPORTANT 

Some analyzer I played around with some time ago pointed out that we weren't awaiting these `Task`s.

### DIFF
--- a/Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Specialized/TaskOfTAssertionSpecs.cs
@@ -40,11 +40,11 @@ public static class TaskOfTAssertionSpecs
             Func<Task<int>> action = null;
 
             // Act
-            Func<Task> testAction = () =>
+            Func<Task> testAction = async () =>
             {
                 using var _ = new AssertionScope();
 
-                return action.Should().CompleteWithinAsync(
+                await action.Should().CompleteWithinAsync(
                     timeSpan, "because we want to test the failure {0}", "message");
             };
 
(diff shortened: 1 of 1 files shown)


lg2de:
I think the analyzers you mentioned created false positives.
In this casees returning the task and awaiting it creates same result.

jnyrup:
>  In this casees returning the task and awaiting it creates same result.

Not sure what you mean by "same result" here.

I guess we can agree that in _general_ awaiting a task or not within a using statement can make a difference.
<details>

```cs
async Task CompleteWithinAsync()
{
    Console.WriteLine("starting");
    await Task.Delay(1000);
    Console.WriteLine("done");
}

sealed class AssertionScope : IDisposable
{
    public void Dispose()
    {
        Console.WriteLine("disposing");
    }
}
```

```cs
var returning = async () =>
{
    await CompleteWithinAsync();
};

await returning();
```
```
starting
disposing
done
```

```cs
var awaiting = async () =>
{
    await CompleteWithinAsync();
};

await awaiting();
```

```
starting
done
disposing
```

</details>

I agree that it doesn't seem necessary here to extend the lifetime of the constructed `AssertionScope` to include the async part of `CompleteWithinAsync`, but it shouldn't harm and it brings the code closer to how an end-user would use an `AssertionScope`.

dennisdoomen:
> I think the analyzers you mentioned created false positives.
> In this casees returning the task and awaiting it creates same result.

You're right. It doesn't make a difference. I suspect that technically adding `async` makes it even slower because of the state machine it generates.

dennisdoomen (approved):

lg2de:
When looking on the code last night on my phone I did not realize the `AssertionScope`...
Yes, in this case there may be a difference.

jnyrup:
> > I think the analyzers you mentioned created false positives.
> > In this casees returning the task and awaiting it creates same result.
> 
> You're right. It doesn't make a difference. I suspect that technically adding `async` makes it even slower because of the state machine it generates.

@dennisdoomen 
Did you see that there's a `using var _ = new AssertionScope()` right before returning a `Task`?

dennisdoomen:
> Did you see that there's a `using var _ = new AssertionScope()` right before returning a `Task`?

Damnit, I did, but didn't realize the subtle change in behavior. 

Was it https://www.meziantou.net/enforcing-asynchronous-code-good-practices-using-a-roslyn-analyzer.htm#ma0100-await-tasks-b?

jnyrup:
I just had a note lying around, so not sure, but I think it might have been [Microsoft.CodeAnalysis.NetAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.NetAnalyzers/) which is the successor to `Microsoft.CodeAnalysis.FxCopAnalyzers`

PR #2165, opened 2023-04-01
Extracts the type reflection logic into a dedicated class

dennisdoomen:
As preparation for #2152, I've extracted all the logic to identify the members of a type into a dedicated class.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Common/TypeExtensions.cs
+++ b/Src/FluentAssertions/Common/TypeExtensions.cs
@@ -24,11 +24,8 @@ internal static class TypeExtensions
     private static readonly ConcurrentDictionary<Type, bool> HasValueSemanticsCache = new();
     private static readonly ConcurrentDictionary<Type, bool> TypeIsRecordCache = new();
 
-    private static readonly ConcurrentDictionary<(Type Type, MemberVisibility Visibility), IEnumerable<PropertyInfo>>
-        NonPrivatePropertiesCache = new();
-
-    private static readonly ConcurrentDictionary<(Type Type, MemberVisibility Visibility), IEnumerable<FieldInfo>>
-        NonPrivateFieldsCache = new();
+    private static readonly ConcurrentDictionary<(Type Type, MemberVisibility Visibility), TypeMemberReflector>
+        TypeMemberReflectorsCache = new();
 
     public static bool IsDecoratedWith<TAttribute>(this Type type)
         where TAttribute : Attribute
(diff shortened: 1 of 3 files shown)


Src/FluentAssertions/Common/TypeMemberReflector.cs
```diff
@@ -0,0 +1,163 @@
...
+    private void LoadNonPrivateProperties()
+    {
+        IEnumerable<PropertyInfo> query =
+            from propertyInfo in GetPropertiesFromHierarchy(typeToReflect, visibility)
+            where HasNonPrivateGetter(propertyInfo)
+            where !propertyInfo.IsIndexer()
+            select propertyInfo;
+
+        NonPrivateProperties = query.ToArray();
+    }
```
jnyrup, line 46: If we return the `PropertyInfo[]` here and assign it to `NonPrivateProperties` in the constructor, then:
* `NonPrivateProperties` can be `{ get }` ,
* `LoadNonPrivateProperties` can be `static` with `typeToReflect` and `visibility` as parameters  and
* `typeToReflect` and `visibility` are no longer need as members on this class.


In general I strive to only call static methods from constructors to avoid risking using members before they have been initialized.

Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -221,143 +218,23 @@ public static FieldInfo FindField(this Type type, string fieldName)
 
     public static IEnumerable<MemberInfo> GetNonPrivateMembers(this Type typeToReflect, MemberVisibility visibility)
```
jnyrup, line 219: FYI, If the return type is changed to `MemberInfo[]` then `DefaultValueFormatter.GetMembers` can have it's `ToArray()` removed.

jnyrup (changes requested):

Src/FluentAssertions/Common/TypeExtensions.cs
dennisdoomen: Nice catch

Src/FluentAssertions/Common/TypeMemberReflector.cs
dennisdoomen: I think it's a matter of taste, and I kind of liked having those two fields. But I have adopted your suggestions anyhow.

jnyrup (approved):
Really nice improvement from several perspectives 🚀

PR #2167, opened 2023-04-03
Improve the DateOnly and TimeOnly code coverage

sdelarosbil:
Mostly NotBe and BeNull that were missing coverage.

Part of #1823

### DIFF
--- a/Tests/FluentAssertions.Specs/Primitives/DateOnlyAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Primitives/DateOnlyAssertionSpecs.cs
@@ -27,6 +27,16 @@ public class DateOnlyAssertionSpecs
         dateOnly.Should().NotBeNull();
     }
 
+    [Fact]
+    public void Should_succeed_when_asserting_nullable_dateonly_value_with_null_to_be_null()
+    {
+        // Arrange
+        DateOnly? dateOnly = null;
+
+        // Act/Assert
+        dateOnly.Should().BeNull();
+    }
+
     public class Be
     {
         [Fact]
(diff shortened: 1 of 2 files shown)


jnyrup:
Thanks for looking into this ❤️ 

I just recently became aware that coveralls.io (or the way we send CC reports) not only doesn't display the CC information on files with multiple types, but also doesn't report any CC stats for those files.

dennisdoomen:
Delaying the review a bit until https://github.com/fluentassertions/fluentassertions/pull/2168 has been done (so I won't have to repeat the same comments ;-))

sdelarosbil:
I pushed applying the recommendations of the other pr.

dennisdoomen (approved):

Tests/FluentAssertions.Specs/Primitives/DateOnlyAssertionSpecs.cs
```diff
@@ -602,6 +612,65 @@ public void When_asserting_subject_null_dateonly_should_not_have_month_should_th
         }
     }
 
+    public class NotBe
+    {
+        [Fact]
+        public void Different_dateonly_values_are_valid()
```
jnyrup, line 618: Please let me know if you rather just have my raw review instead of giving small hints and thinking aloud.

In addition to Dennis' blog post, I can recommend https://freecontent.manning.com/making-better-unit-tests/

An unusual part about this way of naming methods is that it wants to _avoid_ dictating a strict naming pattern as one easily ends up bending over backwards to fit the scheme.

While `NotBe` should not be directly included the naming, my gut feeling is that `valid` is a little to vague even though we can infer that context from the name of the nested class `NotBe`.

The two first tests could maybe be named
* `different_values_are_unequal`
* `different_nullable_values_are_unequal`

Or is those just tautologies? 🤔

sdelarosbil: Yeah I have problems with the tautologies too and I feel valid gets us closer.

How about "considered" valid?

jnyrup: Nah... I don't think it helps adding `considered`.

Unless @dennisdoomen thinks this is worth digging into, I'll approve and merge it as it is.

PR #2168, opened 2023-04-03
Improve the ExpressionExtensions coverage

sdelarosbil:
I put the test there, but I don't know if there was a better place to put them, let me know if there is a better place to put these. It tests untested expression nodes and also the case of not using a supported one.

Part of #1823


Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.BeInAscendingOrder.cs
```diff
@@ -153,6 +153,71 @@ public void When_asserting_single_element_collection_by_property_expression_orde
             act.Should().NotThrow();
         }
 
+        [Fact]
+        public void When_asserting_single_element_collection_by_converted_property_expression_ordered_in_ascending_it_should_succeed()
```
dennisdoomen, line 157:
```suggestion
        public void Can_use_a_cast_expression_in_the_ordering_expression()
```

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.BeInAscendingOrder.cs
```diff
@@ -153,6 +153,71 @@ public void When_asserting_single_element_collection_by_property_expression_orde
...
+            var collection = new SomeClass[]
+            {
+                new() { Text = "a", Number = 1 }
+            };
+
+            // Act
+            Action act = () => collection.Should().BeInAscendingOrder(o => (float)o.Number);
+
+            // Assert
+            act.Should().NotThrow();
```
dennisdoomen, line 169: 🔧 For scenarios that are not supposed to throw, we just call the API and that's it.

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.BeInAscendingOrder.cs
```diff
@@ -153,6 +153,71 @@ public void When_asserting_single_element_collection_by_property_expression_orde
...
+        }
+
+        [Fact]
+        public void When_asserting_single_element_collection_by_indexed_List_property_expression_ordered_in_ascending_it_should_succeed()
```
dennisdoomen, line 173:
```suggestion
        public void Can_use_an_index_into_a_list_in_the_ordering_expression()
```

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.BeInAscendingOrder.cs
```diff
@@ -153,6 +153,71 @@ public void When_asserting_single_element_collection_by_property_expression_orde
...
+
+            // Act
+            Action act = () => collection.Should().BeInAscendingOrder(o => o[0].Number);
+
+            // Assert
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_asserting_single_element_collection_by_indexed_Array_property_expression_ordered_in_ascending_it_should_succeed()
```
dennisdoomen, line 189: ❓ Is this scenario really covering a different path compared to the previous test case?

dennisdoomen, line 189:
```suggestion
        public void Can_use_an_index_into_an_array_in_the_ordering_expression()
```

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.BeInAscendingOrder.cs
```diff
@@ -153,6 +153,71 @@ public void When_asserting_single_element_collection_by_property_expression_orde
...
+
+            // Act
+            Action act = () => collection.Should().BeInAscendingOrder(o => o[0].Number);
+
+            // Assert
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_asserting_single_element_collection_with_unsupported_operator_in_ascending_it_should_throw()
```
dennisdoomen, line 205:
```suggestion
        public void Unsupported_ordering_expressions_are_invalid()
```

dennisdoomen (changes requested):
We are actively working on simplifying the naming of our tests so they are shorter and more concise. Therefor I've suggested some improvements. I've tried the capture the background of this in https://www.continuousimprover.com/2023/03/test-naming.html

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.BeInAscendingOrder.cs
jnyrup: Yes, there's a difference between an array indexer and a List indexer.
See the currently uncovered `Convert`, `ArrayIndex` and `Call` cases

https://coveralls.io/builds/58517312/source?filename=Src%2FFluentAssertions%2FCommon%2FExpressionExtensions.cs

dennisdoomen: I should have checked that myself. I can always count on the thoroughness of @jnyrup 😜

jnyrup (approved):

dennisdoomen (approved):

PR #935, opened 2018-10-02
EquivalencyValidator performance fixes

pentp:
This fixes `GenericDictionaryEquivalencyStep` and `GenericEnumerableEquivalencyStep` performance problems caused by the use of uncached `LambdaCompiler.Compile` which is very expensive.
Related to #475.

This makes `ObjectAssertions.BeEquivalentTo` about 4.3x faster in our test projects.

PR #2180, opened 2023-04-12
Code coverage improvements

sdelarosbil:
This improves coverage in 3 small cases.

For the generic dictionary one, it turns out that the other version of the method was length checking already which made the check redundant as it wasn't possible to hit that line in general.

Part of #1823

### DIFF
--- a/Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
@@ -114,11 +114,6 @@ public class GenericDictionaryEquivalencyStep : IEquivalencyStep
         // the AssertSameLength method.
         where TExpectedKey : TSubjectKey
     {
-        if (expectation.Count == subject.Count)
-        {
-            return true;
-        }
-
         KeyDifference<TSubjectKey, TExpectedKey> keyDifference = CalculateKeyDifference(subject, expectation);
 
         bool hasMissingKeys = keyDifference.MissingKeys.Count > 0;
--- a/Tests/FluentAssertions.Equivalency.Specs/DataColumnSpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/DataColumnSpecs.cs
@@ -79,6 +79,24 @@ public class DataColumnSpecs : DataSpecs
     }
 
     [Fact]
+    public void Can_exclude_a_column_from_all_tables()
+    {
+        // Arrange
+        var subjectDataSet = CreateDummyDataSet<TypedDataSetSubclass>();
+        var expectationDataSet = new TypedDataSetSubclass(subjectDataSet);
+
+        var subject = subjectDataSet.TypedDataTable1.DecimalColumn;
+        var expectation = expectationDataSet.TypedDataTable1.DecimalColumn;
+
+        expectation.Unique = true;
+        expectation.Caption = "Test";
+
+        // Act & Assert
+        subject.Should().BeEquivalentTo(expectation, options => options
+            .ExcludingColumnInAllTables(expectation.ColumnName));
+    }
+
+    [Fact]
     public void When_DataColumn_has_changes_but_is_excluded_as_params_it_should_succeed()
     {
         // Arrange
--- a/Tests/FluentAssertions.Equivalency.Specs/DataTableSpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/DataTableSpecs.cs
@@ -238,6 +238,24 @@ public class DataTableSpecs : DataSpecs
     }
 
     [Fact]
+    public void When_excluding_invalid_constraint_it_should_throw()
+    {
+        // Arrange
+        var typedDataSet = CreateDummyDataSet<TypedDataSetSubclass>();
+
+        var subject = typedDataSet.ToUntypedDataSet().Tables["TypedDataTable1"];
+        var expectation = typedDataSet.ToUntypedDataSet().Tables["TypedDataTable1"];
+
+        // Act
+        Action action = () => subject.Should().BeEquivalentTo(expectation, options => options
+            .ExcludingRelated((Constraint constraint) => new object()));
+
+        // Assert
+        action.Should().Throw<ArgumentException>().WithMessage(
+            "*Expression must be a simple member access*");
+    }
+
+    [Fact]
     public void When_data_table_name_does_not_match_but_the_corresponding_property_is_excluded_equivalence_test_should_succeed()
     {
         // Arrange


Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
```diff
@@ -114,11 +114,6 @@ private static bool AssertExpectationIsNotNull(object subject, object expectatio
         // the AssertSameLength method.
         where TExpectedKey : TSubjectKey
     {
-        if (expectation.Count == subject.Count)
```
lg2de, line 117: Why do you think that this is redundant?

Tests/FluentAssertions.Equivalency.Specs/DataColumnSpecs.cs
```diff
@@ -78,6 +78,24 @@ public void When_DataColumn_has_changes_but_is_excluded_it_should_succeed()
             .ExcludingColumn(dataColumn2));
     }
 
+    [Fact]
+    public void When_DataColumn_has_changes_but_is_excluded_in_all_tables_it_should_succeed()
```
dennisdoomen, line 82: 🔧 I suggest to align with the new naming conventions `Can_exclude_a_column_from_all_tables`

Tests/FluentAssertions.Equivalency.Specs/DataColumnSpecs.cs
```diff
@@ -78,6 +78,24 @@ public void When_DataColumn_has_changes_but_is_excluded_it_should_succeed()
...
+    {
+        // Arrange
+        var dataSet1 = CreateDummyDataSet<TypedDataSetSubclass>();
```
dennisdoomen, line 85: 🔧 Please keep using terms like `subject` and `expectation`

Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
```diff
@@ -494,6 +494,20 @@ public void When_a_generic_dictionary_is_typed_as_object_it_should_respect_the_r
...
+    {
+        // Arrange
+        object object1 = new Dictionary<string, string> { ["greeting"] = "hello" };
+        object object2 = new Dictionary<string, string> { ["greeting"] = "hello" };
+
+        // Act
+        Action act = () => object1.Should().BeEquivalentTo(object2, options => options.ExcludingNestedObjects());
+
+        // Assert
+        act.Should().NotThrow();
```
dennisdoomen, line 508: 🔧 We prefer not to use `NotThrow` in the assert anymore.

Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
```diff
@@ -494,6 +494,20 @@ public void When_a_generic_dictionary_is_typed_as_object_it_should_respect_the_r
         act.Should().NotThrow();
     }
 
+    [Fact]
+    public void When_a_generic_dictionary_is_typed_as_object_without_nesting_it_should_respect_the_runtime_typed()
```
dennisdoomen, line 498: 🤔 This test also succeeds when you omit `ExcludingNestedObjects`, so I'm not sure what it is supposed to do.

Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
```diff
@@ -526,6 +540,23 @@ public void
...
+    public void
+        When_a_non_generic_dictionary_is_decided_to_be_equivalent_to_expected_without_nesting_trace_is_still_written()
+    {
+        // Arrange
+        object object1 = new NonGenericDictionary { ["greeting"] = "hello" };
+        object object2 = new NonGenericDictionary { ["greeting"] = "hello" };
+        var traceWriter = new StringBuilderTraceWriter();
+
+        // Act
+        object1.Should().BeEquivalentTo(object2, opts => opts.RespectingRuntimeTypes().ExcludingNestedObjects().WithTracing(traceWriter));
```
dennisdoomen, line 553: 🤔 Without `RespectingRuntimeTypes` it also succeeds. Did you expect that?

Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
```diff
@@ -526,6 +540,23 @@ public void
         trace.Should().Contain("Recursing into dictionary item greeting at object1");
     }
 
+    [Fact]
+    public void
+        When_a_non_generic_dictionary_is_decided_to_be_equivalent_to_expected_without_nesting_trace_is_still_written()
```
dennisdoomen, line 545: ❓ I don't quite understand what this test is trying to validate

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
jnyrup: #935 added an earlier `Count` check on line 105, but didn't remove this one.

So I'm not sure this inner one can be hit anymore.

Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
sdelarosbil: ok so all the three tests you talk about is to test the IsRecursive option which was frequently missing coverage and the way you do that is with ExcludeNestedObject

However, it seems my current tests aren't selective enough about it so I will try to improve them.

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
jnyrup: I had a closer look at this and I'm good with this change.

Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
sdelarosbil: So as I recheck my tests, I am now wondering: what cases would ExcludeNestedObjects succeed while omitting it fails?

Because I tried this:
```cs
[Fact]
    public void When_a_generic_dictionary_is_typed_as_object_without_nesting_it_should_respect_the_runtime_typed()
    {
        // Arrange
        object subject = new Dictionary<string, SomeNestedObject>
        {
            ["greeting"] = new()
            {
                MyString = "identical string",
                MyChild = new SomeNestedObject { MyString = "identical string" }
            }
        };
        object expectation = new Dictionary<string, SomeNestedObject>
        {
            ["greeting"] = new()
            {
                MyString = "identical string",
                MyChild = new SomeNestedObject { MyString = "DIFFERENT STRING" }
            }
        };

        // Act & Assert
        subject.Should().BeEquivalentTo(expectation, options => options.ExcludingNestedObjects());
    }
``` But it failed.

Even more strange is `When_nested_objects_are_excluded_from_collections_it_should_use_simple_equality_semantics` from CollectionSpecs still succeeds if you remove the ExcludingNestedObjects.

So I feel like there's some problems with this option that makes it too weird to test. Should I remove those test and only keep the other part of the pr?

jnyrup: Yeah, let's do that 👍

jnyrup:
Did you intend to remove `When_DataColumn_has_changes_but_is_excluded_in_all_tables_it_should_succeed` as well?

sdelarosbil:
> Did you intend to remove `When_DataColumn_has_changes_but_is_excluded_in_all_tables_it_should_succeed` as well?
> 
> ```cs
> [Fact]
> public void When_DataColumn_has_changes_but_is_excluded_in_all_tables_it_should_succeed()
> {
>     // Arrange
>     var dataSet1 = CreateDummyDataSet<TypedDataSetSubclass>();
>     var dataSet2 = new TypedDataSetSubclass(dataSet1);
> 
>     var dataColumn1 = dataSet1.TypedDataTable1.DecimalColumn;
>     var dataColumn2 = dataSet2.TypedDataTable1.DecimalColumn;
> 
>     dataColumn2.Unique = true;
>     dataColumn2.Caption = "Test";
> 
>     // Act & Assert
>     dataColumn1.Should().BeEquivalentTo(dataColumn2, options => options
>         .ExcludingColumnInAllTables(dataColumn2.ColumnName));
> }
> ```
That wasn't intended, I restored it.

jnyrup (approved):

dennisdoomen (approved):

PR #2184, opened 2023-04-16
Prepare for .NET 8 analyzers

jnyrup:
## IMPORTANT 

I installed the latest .NET 8 preview SDK to see what new buit-in analyzers we can expect when .NET 8 is released later this year.

CA1859 suggests using concrete types instead of interfaces for (micro) performance reasons.

### DIFF
--- a/Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingSolver.cs
+++ b/Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingSolver.cs
@@ -79,7 +79,7 @@ internal class MaximumMatchingSolver<TValue>
         return Enumerable.Empty<Match>();
     }
 
-    private IEnumerable<Element<TValue>> GetMatchingElements(Predicate<TValue> predicate)
+    private List<Element<TValue>> GetMatchingElements(Predicate<TValue> predicate)
     {
         if (!matchingElementsByPredicate.TryGetValue(predicate, out var matchingElements))
         {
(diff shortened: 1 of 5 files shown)


Src/FluentAssertions/Collections/MaximumMatching/MaximumMatchingSolver.cs
```diff
@@ -79,7 +79,7 @@ private IEnumerable<Match> FindMatchForPredicate(Predicate<TValue> predicate, Ma
         return Enumerable.Empty<Match>();
     }
 
-    private IEnumerable<Element<TValue>> GetMatchingElements(Predicate<TValue> predicate)
+    private List<Element<TValue>> GetMatchingElements(Predicate<TValue> predicate)
```
dennisdoomen, line 82: I actually dislike the changes in this PR. I often use `IEnumerable<T>` to prevent the call side from modifying the collection.

jnyrup: Did you notice that all changed methods are `private`?
If that doesn't make the situation better, then I'll just close the PR again, no worries.

dennisdoomen: That's a good point. Will that CA1859 also complain about `protected`, `internal` and `public` methods like that?

jnyrup: I'll have to check up on that.

jnyrup: It can be configured via `.editorconfig`
> CA1859: default value is private.

https://github.com/dotnet/roslyn-analyzers/blob/main/docs/Analyzer%20Configuration.md#editorconfig-format

dennisdoomen (approved):

PR #2196, opened 2023-04-25
CA1860: Avoid using 'Enumerable.Any()' extension method

Evangelink:
Fixes newly enabled .NET analyzers CA1860 rule. More information about the rule at https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1860

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -169,7 +169,7 @@ public sealed class AssertionScope : IAssertionScope
             {
                 string becauseOrEmpty = because ?? string.Empty;
 
-                return becauseArgs?.Any() == true
+                return becauseArgs?.Length != 0
                     ? string.Format(CultureInfo.InvariantCulture, becauseOrEmpty, becauseArgs)
                     : becauseOrEmpty;
             }
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -169,7 +169,7 @@ public sealed class AssertionScope : IAssertionScope
             {
                 string becauseOrEmpty = because ?? string.Empty;
 
-                return becauseArgs?.Length != 0
+                return becauseArgs?.Length > 0
                     ? string.Format(CultureInfo.InvariantCulture, becauseOrEmpty, becauseArgs)
                     : becauseOrEmpty;
             }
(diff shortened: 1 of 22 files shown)


Evangelink:
Sorry not sure what happened, I had no error locally but now I do reproduce the same errors. Will investigate and mark as ready when build is passing locally.

Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -169,7 +169,7 @@ public AssertionScope BecauseOf(string because, params object[] becauseArgs)
             {
                 string becauseOrEmpty = because ?? string.Empty;
 
-                return becauseArgs?.Any() == true
+                return becauseArgs?.Length != 0
```
jnyrup, line 172: This is not an equivalent rewrite when `becauseArgs` is null

jnyrup: but `becauseArgs?.Length > 0` is

jnyrup:
> Fixes newly enabled .NET analyzers CA1860 rule.

Since we don't have very performance sensitive code, I don't have a strong opinion on whether or not we should limit the usage of `Any`. 
But if we do, we should stricten the analyzer to enforce it.

dennisdoomen:
> Since we don't have very performance sensitive code, I don't have a strong opinion on whether or not we should limit the usage of `Any`. But if we do, we should stricten the analyzer to enforce it.

It depends on the run-time version. As of .NET Core 2.0, `Any` will be check whether you're calling it on an array or collection and directly use the `Count` property. The .NET 4.x version isn't that smart.  

But I agree. Totally fine to enable this rule.

jnyrup:
Thinking more about it, I'm pro enforcing this. 
Limiting the usage of `Any` should help spotting places where we do multiple enumerations on lazy non-materialized enumerables.

Related analyzer and issue:
https://github.com/fluentassertions/fluentassertions/blob/22a603980d977ee37ef0f3c5a9bcdffec1f2e1d5/.editorconfig#L153-L154

Evangelink:
Will fix this PR and enable the rule today. Sorry for the delay, I was held back by some other tasks.

Evangelink:
Sorry for the delay here and multiple commits.... Hope this PR will help!

Thanks for the awesome tool :)

jnyrup (approved):

dennisdoomen (approved):

PR #2203, opened 2023-05-11
Use correct precision in docs for BeApproximately

cskwrd:


docs/_pages/numerictypes.md
```diff
@@ -40,7 +40,7 @@ Notice that `Should().Be()` and `Should().NotBe()` are not available for floats
 
 ```csharp
 float value = 3.1415927F;
-value.Should().BeApproximately(3.14F, 0.01F);
+value.Should().BeApproximately(3.14F, 0.001F);
```
dennisdoomen, line 43: I actually intended to say that the value must be between `3.13F` and `3.15F`, so unless my brain is somehow short circuiting, the original example is fine.

ITaluone: I also don't get the point :)

It is also between 2 and 4
...and 1 and 5
...and -100 and 100

jnyrup:
If I run the proposed change, the assertion fails.
The current code does not fail.
https://dotnetfiddle.net/WYUjgZ

I'm closing this for now.
Please comment if we missed something.
In any cas,e thanks for taking the time to wanting to improve the docs.

cskwrd:
Maybe the confusion is coming from the wording below the example:
`This will verify that the value of the float is between 3.139 and 3.141.`

To me, that reads as `[3.139, 3.141]` not `(3.139, 3.141)`. Which would mean that the proposed change would be correct as it results in failure if the actual value deviates by more than `0.001`. The example as written allows values in the range `[3.13, 3.15]`, like you mentioned.

PR #2204, opened 2023-05-12
Add Qodana

matkoch:
Setup:

- Create an organization, team, and project on https://qodana.cloud/ (enable _Make the organization public_)
- Copy token and add [here](https://github.com/fluentassertions/fluentassertions/settings/secrets/actions) as `QODANA_TOKEN`

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -1,0 +1,24 @@
+name: Qodana
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+      - develop
+
+jobs:
+  qodana-scan:
+    name: Qodana Scan
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: 'Qodana Scan'
+        uses: JetBrains/qodana-action@main
+        with:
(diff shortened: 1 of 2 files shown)


dennisdoomen (approved):

lg2de:
Where can we see the result of this analysis?
What is the benefit for FluentAssertions?
In the past [I asked for using SonarQube](https://github.com/fluentassertions/fluentassertions/pull/1320#issuecomment-621868573), which is maybe comparable a bit. Qodana seams to be accepted without any discussion.

jnyrup:
Interesting stuff, let's give it a go and see if it's something for us.

@matkoch Have you seen that the pipeline fails to read `qodana.sarif.json`?

matkoch:
@jnyrup yes I discussed it with Dennis already. It should work once merged, this is a special case/bug for Qodana because the PR itself introduces the baseline report. If it fails anyway, just revert and I will investigate once more, but dev team said it should be fine :)

jnyrup (approved):

lg2de:
> It's working..

Unfortunately it is not visible to public. (Yes, I logged myself.)

dennisdoomen:
@matkoch is there a way to give contributors access to the report?

PR #2207, opened 2023-05-16
Extend `AssemblyAssertions` with `HavePublicKey` and `BeUnsigned`

Corniel:
To ensures that the public key of an signed assembly does not accidentally changes, having an assertion `.HasPublicKey(string)` can help. Hence this PR. For completeness/symmetry reasons I also added `.BeUnsigned()`.

## IMPORTANT 

Relevant issue: #2209


Corniel:
Note tath locally my build fails, as it claims that C#11 features are not available, and all warnings are handled as errors. Can someone enlighten me on why that could happen?

Tests/FluentAssertions.Specs/Specialized/AssemblyAssertionSpecs.cs
```diff
@@ -260,6 +260,85 @@ public void When_an_assembly_is_null_and_Should_BeNull_is_asserted_it_should_suc
...
+            // Act
+            Action act = () => assembly.Should().HavePublicKey(
+                "0024000004800000940000000602000000240000525341310004000001000100" +
+                "2d25ff515c85b13ba08f61d466cff5d80a7f28ba197bbf8796085213e7a3406f" +
+                "970d2a4874932fed35db546e89af2da88c194bf1b7f7ac70de7988c78406f762" +
+                "9c547283061282a825616eb7eb48a9514a7570942936020a9bb37dca9ff60b77" +
+                "8309900851575614491c6d25018fadb75828f4c7a17bf2d7dc86e7b6eafc5d8f");
+
+            // Assert
+            act.Should().NotThrow();
```
ITaluone, line 311: same as above

Corniel: That is fine by me, but I just followed the format applied by the other tests in this file.

IT-VBFK: This is more straight forward. If the test fails.. it fails anyway.. no need to fail another test just to recognize "oh.. the test failed". Do you know what I mean?

Corniel: @IT-VBFK Makes total sense. But I just wanted to follow the (apparent) guidelines. But I've changed them.

IT-VBFK: ah.. don't bother :) design keeps changing ;)

dennisdoomen:
Thank you for this proposal. As you can see from the checklist, the idea is that you first create an issue with an API proposal so we can discuss the consequences. But since we know each other let's see if we can take this PR further 🫢

lg2de:
When writing the API proposal, please check whether asserting signature fingerprint only is possible and - maybe - easier to use.

Corniel:
> When writing the API proposal, please check whether asserting signature fingerprint only is possible and - maybe - easier to use.

I would argue that checking that an assembly is signed is less relevant: if it an assembly you don't control you self, it out of your control, and if you do sign an assembly, the reason to have a test on it, is because you don't want to ever change the public key. But if we feel there is an desire to just test if an assembly is signed, we can add that too, obviously.

jnyrup:
It looks promising, but let's discuss the API in a [GH issue](https://github.com/fluentassertions/fluentassertions/issues/new?assignees=&labels=api-suggestion&projects=&template=02_api_proposal.yml&title=%5BAPI+Proposal%5D%3A+) first.
The issue template contains questions that help ensuring we're on the same line and reach the better design.

That people already started commenting on the implementation before the shape was agreed upon proves the need to keep these thing separate.

Corniel:
@dennisdoomen :Note that locally my build fails, as it claims that C#11 features are not available, and all warnings are handled as errors. Can someone enlighten me on why that could happen?

jnyrup:
Do you have the .NET 7 SDK installed? Our global.json file should specify the minimum required version.

Corniel:
> Do you have the .NET 7 SDK installed? Our global.json file should specify the minimum required version.

Yes I have. Both for my company and multiple open source projects I work on require that. As mentioned, also the thing that (and that only happens for this solution) all warnings are handled as errors is strange.

dennisdoomen:
> Yes I have. Both for my company and multiple open source projects I work on require that. As mentioned, also the thing that (and that only happens for this solution) all warnings are handled as errors is strange.

I recommend joining our Slack to have a more efficient discussion. You can use [this invite](https://join.slack.com/t/fluentassertions/shared_invite/zt-1w843y900-vV_vUxLbIPAhA_0ZHAKSGw).

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -177,9 +177,17 @@ public AndConstraint<AssemblyAssertions> HavePublicKey(string publicKey, string
     {
         Guard.ThrowIfArgumentIsNullOrEmpty(publicKey);
 
-        var bytes = Subject.GetName().GetPublicKey() ?? Array.Empty<byte>();
+        var bytes = Subject.GetName().GetPublicKey();
         var assemblyKey = BitConverter.ToString(bytes).Replace("-", string.Empty, StringComparison.Ordinal);
 
+        Execute.Assertion
+            .BecauseOf(because, becauseArgs)
+            .ForCondition(bytes.Length != 0)
```
IT-VBFK, line 185: NRE if bytes `null`

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -139,6 +141,64 @@ public AndConstraint<AssemblyAssertions> Reference(Assembly assembly, string bec
...
+                "but it unsigned.",
+                Subject.FullName, publicKey);
+
+        Execute.Assertion
+            .BecauseOf(because, becauseArgs)
+            .ForCondition(assemblyKey == publicKey)
+            .FailWith(
+                "Expected assembly {0} to have public key {1}{because}, " +
+                "but it has {2} instead.",
+                Subject.FullName, publicKey, assemblyKey);
```
IT-VBFK, line 197:
```suggestion
        Execute.Assertion
            .BecauseOf(because, becauseArgs)
            .WithExpectation("Expected assembly {0} to have public key {1}{because}, ", Subject.FullName , publicKey)
            .ForCondition(bytes is not null)
            .FailWith("but public key's bytes are null.")
            .Then
            .ForCondition(bytes.Length != 0)
            .FailWith("but it unsigned.")
            .Then
            .ForCondition(assemblyKey == publicKey)
            .FailWith("but it has {0} instead.", assemblyKey)
            .Then
            .ClearExpectation();
```

IT-VBFK: You can concat the assertions you avoid to repeat yourself.

FluentAssertions.sln
```diff
@@ -134,6 +136,12 @@ Global
 		{A946043D-D3F8-46A4-B485-A88412C417FE}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{A946043D-D3F8-46A4-B485-A88412C417FE}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{A946043D-D3F8-46A4-B485-A88412C417FE}.Release|Any CPU.Build.0 = Release|Any CPU
+		{6140D5E2-9792-4D48-9FA8-20AE38C6E97B}.CI|Any CPU.ActiveCfg = Debug|Any CPU
```
dennisdoomen, line 139: 🤔 Why do you need the extra configurations?

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -139,6 +141,59 @@ public AndConstraint<AssemblyAssertions> Reference(Assembly assembly, string bec
...
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
+    /// </param>
+    public AndConstraint<AssemblyAssertions> BeUnsigned(string because = "", params object[] becauseArgs)
+    {
+        Execute.Assertion
+            .BecauseOf(because, becauseArgs)
+            .ForCondition(Subject.GetName().GetPublicKey() is not { Length: > 0 })
+            .FailWith(
+                "Expected assembly {0} to be unsigned{reason}, but it is.", Subject.FullName);
```
dennisdoomen, line 158: 🔧 We prefer to use the phrase "Did not expect the assembly {0} to be signed{reason}.". See also https://github.com/fluentassertions/fluentassertions/issues/1340

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -139,6 +141,59 @@ public AndConstraint<AssemblyAssertions> Reference(Assembly assembly, string bec
...
 
+    /// <summary>Asserts that the assembly is unsigned.</summary>
+    /// <param name="because">
+    /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+    /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+    /// </param>
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
+    /// </param>
+    public AndConstraint<AssemblyAssertions> BeUnsigned(string because = "", params object[] becauseArgs)
```
dennisdoomen, line 152: 🔧 I was also expecting a `BeSigned`

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -139,6 +141,59 @@ public AndConstraint<AssemblyAssertions> Reference(Assembly assembly, string bec
...
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
+    /// </param>
+    /// <exception cref="ArgumentNullException"><paramref name="publicKey"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException"><paramref name="publicKey"/> is empty.</exception>
+    public AndConstraint<AssemblyAssertions> HavePublicKey(string publicKey, string because = "", params object[] becauseArgs)
+    {
+        Guard.ThrowIfArgumentIsNullOrEmpty(publicKey);
+
+        var bytes = Subject.GetName().GetPublicKey();
```
dennisdoomen, line 180: 🤔 What if `Subject` is `null`?

Tests/FluentAssertions.Specs/Specialized/AssemblyAssertionSpecs.cs
```diff
@@ -260,6 +260,79 @@ public void When_an_assembly_is_null_and_Should_BeNull_is_asserted_it_should_suc
...
+
+            // Act & Assert
+            assembly.Should().BeUnsigned();
+        }
+
+        [Fact]
+        public void Throws_for_signed_assembly()
+        {
+            // Arrange
+            var assembly = GetType().Assembly;
```
dennisdoomen, line 280: 🔧 I would name this as `signedAssembly` to emphasize what you're doing.

dennisdoomen (changes requested):

Tests/FluentAssertions.Specs/Specialized/AssemblyAssertionSpecs.cs
```diff
@@ -260,6 +260,79 @@ public void When_an_assembly_is_null_and_Should_BeNull_is_asserted_it_should_suc
...
+            // Assert
+            act.Should().Throw<XunitException>()
+                .WithMessage("Did not expect the assembly * to be signed, but it is.");
+        }
+    }
+
+    public class HavePublicKey
+    {
+        [Fact]
+        public void Guards_for_signed_assembly_with_expected_public_key()
```
jnyrup, line 294: Add a test to verify that having the expected key in lowercase also works

Tests/FluentAssertions.Specs/Specialized/AssemblyAssertionSpecs.cs
```diff
@@ -260,6 +260,79 @@ public void When_an_assembly_is_null_and_Should_BeNull_is_asserted_it_should_suc
...
+                "970D2A4874932FED35DB546E89AF2DA88C194BF1B7F7AC70DE7988C78406F762" +
+                "9C547283061282A825616EB7EB48A9514A7570942936020A9BB37DCA9FF60B77" +
+                "8309900851575614491C6D25018FADB75828F4C7A17BF2D7DC86E7B6EAFC5D8F");
+        }
+
+        [Fact]
+        public void Throws_for_unsigned_assembly()
+        {
+            // Arrange
+            var unsignedAssembly = typeof(UnsignedAssembly.Class).Assembly;
```
jnyrup, line 312: If @dennisdoomen is also fine with it, here's a dummy implementation of an `Assembly` to avoid adding `UnsignedAssembly`.

```cs
public class MyAssembly : Assembly
{
    private readonly AssemblyName assemblyName = new();

    public override string FullName => "MyAssembly";

    public MyAssembly(string publicKey)
    {
        assemblyName.SetPublicKey(FromHexString(publicKey));
    }

    public override AssemblyName GetName() => assemblyName;

#if NET6_0_OR_GREATER
    private static byte[] FromHexString(string chars) => Convert.FromHexString(chars);
#else
    private static byte[] FromHexString(string chars)
    {
        var bytes = new byte[chars.Length / 2];

        for (int i = 0, j = 0; j < bytes.Length; j++, i += 2)
        {
            bytes[j] = (byte)((Hex(chars[i]) << 4) | Hex(chars[i + 1]));

            static byte Hex(char c) => (byte)(c > '9' ? (c & 0x5F) - 55 : c - '0');
        }

        return bytes;
    }
#endif
}
```

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -139,6 +141,75 @@ public AndConstraint<AssemblyAssertions> Reference(Assembly assembly, string bec
...
+    /// </param>
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
+    /// </param>
+    public AndConstraint<AssemblyAssertions> BeUnsigned(string because = "", params object[] becauseArgs)
+    {
+        bool success = Execute.Assertion
+            .BecauseOf(because, becauseArgs)
+            .ForCondition(Subject is not null)
+            .FailWith("Expected assembly to be unsigned{reason}, but {context:assembly} is <null>.");
```
dennisdoomen, line 157: :wrench I think this error might be easier to understand if it was written as `Can't check for assembly signing if the assembly reference is null` or something like that.

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -139,6 +141,75 @@ public AndConstraint<AssemblyAssertions> Reference(Assembly assembly, string bec
...
+    /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
+    /// </param>
+    /// <exception cref="ArgumentNullException"><paramref name="publicKey"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException"><paramref name="publicKey"/> is empty.</exception>
+    public AndConstraint<AssemblyAssertions> BeSignedWithPublicKey(string publicKey, string because = "", params object[] becauseArgs)
+    {
+        Guard.ThrowIfArgumentIsNullOrEmpty(publicKey);
+
+        bool success = Execute.Assertion
+          .BecauseOf(because, becauseArgs)
```
dennisdoomen, line 189: ⛏️ (nitpick) The spacing is off here.

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -139,6 +141,75 @@ public AndConstraint<AssemblyAssertions> Reference(Assembly assembly, string bec
...
+          .ForCondition(Subject is not null)
+          .FailWith("Expected assembly to have public key {0}{reason}, but {context:assembly} is <null>.", publicKey);
+
+        if (success)
+        {
+            var bytes = Subject.GetName().GetPublicKey();
```
dennisdoomen, line 195: 🤔 According to the MSDN docs, `GetPublicKey` can return `null` (and my IDE agrees).

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -139,6 +141,75 @@ public AndConstraint<AssemblyAssertions> Reference(Assembly assembly, string bec
...
+            var assemblyKey = BitConverter.ToString(bytes).Replace("-", string.Empty, StringComparison.Ordinal);
+
+            Execute.Assertion
+                .BecauseOf(because, becauseArgs)
+                .WithExpectation("Expected assembly {0} to have public key {1}{because}, ", Subject.FullName, publicKey)
+                .ForCondition(bytes.Length != 0)
+                .FailWith("but it unsigned.")
```
dennisdoomen, line 202: 🔧 `but it is unsigned.`

Tests/FluentAssertions.Specs/FindAssembly.cs
```diff
@@ -1,11 +1,46 @@
...
+
+        for (int i = 0, j = 0; j < bytes.Length; j++, i += 2)
+        {
+            bytes[j] = (byte)((Hex(chars[i]) << 4) | Hex(chars[i + 1]));
+
+            static byte Hex(char c) => (byte)(c > '9' ? (c & 0x5F) - 55 : c - '0');
+        }
+
+        return bytes;
+    }
```
dennisdoomen, line 43: 🔧 I found this implementation a bit hard to read, so I asked ChatGPT to provide a suggestion:

```
        if (hexString.Length % 2 != 0)
        {
            throw new ArgumentException("Input string length must be even.");
        }

        byte[] byteArray = new byte[hexString.Length / 2];

        for (int i = 0; i < hexString.Length; i += 2)
        {
            string hexByte = hexString.Substring(i, 2);
            byte value = Convert.ToByte(hexByte, 16);
            byteArray[i / 2] = value;
        }

        return byteArray;
```

dennisdoomen (changes requested):
* Please assume comments also apply to other places in the PR. 
* Also update the `releases.md` and the `assemblies.md`

dennisdoomen (approved):

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Security.Cryptography.X509Certificates;
+using System.Xml.Linq;
```
jnyrup, line 6:
```suggestion
```

Tests/FluentAssertions.Specs/FindAssembly.cs
```diff
@@ -1,11 +1,50 @@
...
+        public override AssemblyName GetName() => assemblyName;
+
+#if NET6_0_OR_GREATER
+        private static byte[] FromHexString(string chars)
+            => string.IsNullOrEmpty(chars)
+            ? Array.Empty<byte>()
+            : Convert.FromHexString(chars);
+#else
+    private static byte[] FromHexString(string chars)
     {
```
jnyrup, line 32:
```suggestion
        private static byte[] FromHexString(string chars)
        {
```

Tests/FluentAssertions.Specs/FindAssembly.cs
jnyrup: I suggested the bit hacking implementation.
I perhaps sacrificed to much readability in my quest to avoid allocations 😅

jnyrup (commented):
Looks good 👍 

Please also add an example to the [docs](https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/assemblies.md)
and update the [release notes](https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/releases.md)

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -139,6 +139,80 @@ public AndConstraint<AssemblyAssertions> Reference(Assembly assembly, string bec
...
+    {
+        Guard.ThrowIfArgumentIsNullOrEmpty(publicKey);
+
+        bool success = Execute.Assertion
+            .ForCondition(Subject is not null)
+            .FailWith("Can't check for assembly signing if {context:assembly} reference is <null>.");
+
+        if (success)
+        {
+            var bytes = Subject.GetName().GetPublicKey() ?? Array.Empty<byte>();
```
IT-VBFK, line 191: Is it worth to add a test which checks what happens if `.GetName()` or `.GetPublicKey()` returns `null` (or fails)?

jnyrup: Nice spotted 👍 
`Throws_for_unsigned_assembly` tests the case where `GetPublicKey` would normally return `null`.
The problem is that `FindAssembly.Stub` should default to a `null` key instead of an empty key.

Tests/FluentAssertions.Specs/Specialized/AssemblyAssertionSpecs.cs
```diff
@@ -260,6 +260,105 @@ public void When_an_assembly_is_null_and_Should_BeNull_is_asserted_it_should_suc
             act.Should().NotThrow();
         }
     }
+
+    public class BeUnsigned
+    {
+        [Fact]
+        public void Guards_for_unsigned_assembly()
```
jnyrup, line 267: Let's extend this test to cover both null and empty keys to match what we check for inside `BeUnsigned`.

Tests/FluentAssertions.Specs/FindAssembly.cs
```diff
@@ -1,11 +1,50 @@
...
+            => string.IsNullOrEmpty(chars)
+            ? Array.Empty<byte>()
+            : Convert.FromHexString(chars);
+#else
+        private static byte[] FromHexString(string chars)
+        {
+            if (chars.Length % 2 != 0)
+            {
+                throw new ArgumentException("Input string length must be even.");
+            }
```
IT-VBFK, line 36: Should we add this (or a similar) check as guard inside the production code?

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -139,6 +139,80 @@ public AndConstraint<AssemblyAssertions> Reference(Assembly assembly, string bec
...
+            .ForCondition(Subject is not null)
+            .FailWith("Can't check for assembly signing if {context:assembly} reference is <null>.");
+
+        if (success)
+        {
+            Execute.Assertion
+                .BecauseOf(because, becauseArgs)
+                .ForCondition(Subject.GetName().GetPublicKey() is not { Length: > 0 })
+                .FailWith(
+                    "Did not expect the assembly {0} to be signed{reason}, but it is.", Subject.FullName);
```
IT-VBFK, line 162: Coveralls also reported a missed path here (probably the same issue with `.GetName()` as below

jnyrup: I suspect we just need to test both for null and empty key.

Tests/FluentAssertions.Specs/FindAssembly.cs
jnyrup: I don't think it's necessary. 
If one provides an expected public key of odd length the assertion will fail and include both subject and expected key in the failure message

Corniel:
I applied Qodana's suggestions. What kind of documentation should be added?

Corniel:
Documentation added.

docs/_pages/assemblies.md
```diff
@@ -7,11 +7,20 @@ sidebar:
...
+You have access to methods to assert an assembly does or does not reference another assembly.
 These are typically used to enforce layers within an application, such as for example, asserting the web layer does not reference the data layer.
 To assert the references, use the following syntax:
 
 ```csharp
 assembly.Should().Reference(otherAssembly);
 assembly.Should().NotReference(otherAssembly);
 ```
+
+Furthermore, you can assert if an assembly is assigned with a specif public key, or that the assembly is not signed at all.
```
IT-VBFK, line 19:
```suggestion
Furthermore, you can assert if an assembly is assigned with a specific public key, or that the assembly is not signed at all.
```

Corniel:
I think I addressed all concerns.

IT-VBFK:
Quick question: Are you able to run ./build.sh (or .\build.ps1 respectively) on you dev device? And is it failing there too?

Corniel:
> Quick question: Are you able to run ./build.sh (or .\build.ps1 respectively) on you dev device? And is it failing there too?

It fails with:
>   // FileNotFoundException: Could not load file or assembly 'System.Runtime, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.

I don't think that is related to my changes.

IT-VBFK:
Looks like macOS, doesn't it? Have the same problem :)

Corniel:
> Looks like macOS, doesn't it? Have the same problem :)

Might be, but this is on a good all Windows laptop. :D

FluentAssertions.sln
```diff
@@ -168,8 +167,6 @@ Global
 		{6D31FFF8-E7FD-41D2-996C-CA8DDFDAE4FD} = {963262D0-9FD5-4741-8C0E-E2F34F110EF3}
 		{08087654-2C32-4818-95E4-45362373441D} = {963262D0-9FD5-4741-8C0E-E2F34F110EF3}
 		{A946043D-D3F8-46A4-B485-A88412C417FE} = {963262D0-9FD5-4741-8C0E-E2F34F110EF3}
-		{0C0211B6-D185-4518-A15A-38AC092EDC50} = {963262D0-9FD5-4741-8C0E-E2F34F110EF3}
-		{0A69DC62-CA14-44E5-BAF9-2EB2E2E2CADF} = {963262D0-9FD5-4741-8C0E-E2F34F110EF3}
```
jnyrup, line 172: revert to include VB and F# projects again

Tests/FluentAssertions.Specs/Specialized/AssemblyAssertionSpecs.cs
jnyrup: Augment with
```cs
[Theory]
[InlineData(null)]
[InlineData("")]
```

And then we can make the `publicKey` parameter to `FindAssembly.Stub` mandatory

docs/_pages/assemblies.md
```diff
@@ -7,11 +7,20 @@ sidebar:
...
 assembly.Should().Reference(otherAssembly);
 assembly.Should().NotReference(otherAssembly);
 ```
+
+Furthermore, you can assert if an assembly is assigned with a specific public key, or that the assembly is not signed at all.
+The first can be useful, to ensure that the public key of your package is not changed unintentionally, the latter to prevent unintended siging.
+To assert this, use the following syntax:
+
+```csharp
+assembly.Should().HavePublicKey("somePublicKey");
```
jnyrup, line 24:
```suggestion
assembly.Should().HavePublicKey("e0851575614491c6d25018fadb75");
```

jnyrup (approved):

docs/_pages/assemblies.md
```diff
@@ -7,11 +7,20 @@ sidebar:
...
 These are typically used to enforce layers within an application, such as for example, asserting the web layer does not reference the data layer.
 To assert the references, use the following syntax:
 
 ```csharp
 assembly.Should().Reference(otherAssembly);
 assembly.Should().NotReference(otherAssembly);
 ```
+
+Furthermore, you can assert if an assembly is assigned with a specific public key, or that the assembly is not signed at all.
+The first can be useful, to ensure that the public key of your package is not changed unintentionally, the latter to prevent unintended siging.
```
jnyrup, line 20:
```suggestion
The first can be useful, to ensure that the public key of your package is not changed unintentionally, the latter to prevent unintended signing.
```

jnyrup (approved):

docs/_pages/assemblies.md
dennisdoomen: I'll wait for @Corniel to fix this and then we can merge.

dennisdoomen:
@Corniel well done. It took a while, but you made it ;-)

jnyrup:
Yeah, thanks for the patience. It was an API that to grow some on me.

Corniel:
Well, it was worth it. I had get custom your way of working too, obviously. I think I did.

@jnyrup Thanks for your patience too.

PR #2228, opened 2023-06-21
Prepare for .NET8 analyzers

jnyrup:
I installed the latest .NET8 preview to see what analyzers we would trigger.

These three lines trigger [CA1861: Avoid constant arrays as arguments](https://learn.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/quality-rules/ca1861)
https://github.com/fluentassertions/fluentassertions/blob/343e80ff81a0598f58d241ba7a4f1588e3fff35d/Src/FluentAssertions/Common/MemberPath.cs#L119
https://github.com/fluentassertions/fluentassertions/blob/343e80ff81a0598f58d241ba7a4f1588e3fff35d/Src/FluentAssertions/Common/StringExtensions.cs#L121
https://github.com/fluentassertions/fluentassertions/blob/343e80ff81a0598f58d241ba7a4f1588e3fff35d/Src/FluentAssertions/Equivalency/Execution/ObjectReference.cs#L42

As I see it, we have three options here:
* Follow the analyzer and extract the constant arrays to static readonly fields. 
  * Downside: Reduced locality of the values
* Suppress CA1861 locally with `#pragma`s.
  * Downside: Added preprocessor noise.
* Suppress CA1861 for the entire project like I did for tests.
  * Downside: no warnings anywhere

I'm probably fine with either, but lean mostly towards either of the first two.

## IMPORTANT 

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -154,6 +154,8 @@ dotnet_diagnostic.CA1819.severity = none
 dotnet_diagnostic.CA1851.severity = suggestion
 # CA1860: Avoid using 'Enumerable.Any()' extension method
 dotnet_diagnostic.CA1860.severity = warning
+# CA1861: Avoid constant arrays as arguments
+dotnet_diagnostic.CA1861.severity = none
 # CA2007: Do not directly await a Task
 dotnet_diagnostic.CA2007.severity = none
 # CA2225: Operator overloads have named alternates
--- a/Build/Build.cs
+++ b/Build/Build.cs
@@ -250,7 +250,7 @@ class Build : NukeBuild
         .Executes(() =>
         {
             ReportGenerator(s => s
-                .SetProcessToolPath(NuGetToolPathResolver .GetPackageExecutable("ReportGenerator", "ReportGenerator.dll", framework: "net6.0"))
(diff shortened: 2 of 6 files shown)


ITaluone:
> `.Split(new[] { '.', '[', ']' }, StringSplitOptions.RemoveEmptyEntries); `

A dump way could be: `.Split(".[]".ToCharArray(), StringSplitOptions.RemoveEmptyEntries); `

lg2de:
> A dump way could be: `.Split(".[]".ToCharArray(), StringSplitOptions.RemoveEmptyEntries); `

This creates an array on the fly, too. This is what the rule tries to prevent, isn't it?

I recommend a field. This is not that local definition, but gives the list a name.

dennisdoomen:
Will a local `const` variable also work?

dennisdoomen (approved):

jnyrup:
> Will a local `const` variable also work?

No, it gives: `CS0133: The expression being assigned must be constant`

@dennisdoomen Can I interpret this as you vote for option 2?

dennisdoomen:
> Can I interpret this as you vote for option 2?

I'm leaning a little bit towards 3, but I can live with 2.

PR #2221, opened 2023-06-13
Qodana fixes

IT-VBFK:
First few Qodana "fixes".

Please close safely if it does not suit you :)

Note that I have most likely not covered all "same errors".

## IMPORTANT 

* [ ] If the PR touches the public API, the changes have been approved in a separate issue with the "api-approved" label.
* [x] The code complies with the [Coding Guidelines for C#](https://www.csharpcodingguidelines.com/).
* [ ] The changes are covered by unit tests which follow the Arrange-Act-Assert syntax and the naming conventions such as is used [in these tests](../tree/develop/Tests/FluentAssertions.Equivalency.Specs/MemberMatchingSpecs.cs#L51-L430).
* [ ] If the PR adds a feature or fixes a bug, please update [the release notes](../tree/develop/docs/_pages/releases.md) with a functional description that explains what the change means to consumers of this library, which are published on the [website](https://fluentassertions.com/releases).
* [ ] If the PR changes the public API the changes needs to be included by running [AcceptApiChanges.ps1](../tree/develop/AcceptApiChanges.ps1) or [AcceptApiChanges.sh](../tree/develop/AcceptApiChanges.sh).
* [ ] If the PR affects [the documentation](../tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).
    * [ ] Please also run `./build.sh --target spellcheck` or `.\build.ps1 --target spellcheck` before pushing and check the good outcome

PR #2229, opened 2023-07-02
Qodana fixes

jnyrup:
Went through the Qodana report and here are fixes to some that should be non-controversial.
This is in addition to be fixed in #2221.

There are a lot of warnings about unused methods/return values of public methods.
To silence them, it seems the idiomatic way is to annotate them with [`[PublicAPI]`](https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#PublicAPIAttribute)

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/CallerIdentification/AddNonEmptySymbolParsingStrategy.cs
+++ b/Src/FluentAssertions/CallerIdentification/AddNonEmptySymbolParsingStrategy.cs
@@ -16,7 +16,7 @@ internal class AddNonEmptySymbolParsingStrategy : IParsingStrategy
         }
         else if (mode is Mode.RemoveSuperfluousWhitespace)
         {
-            if (precedingSymbol is char value && !char.IsWhiteSpace(value))
+            if (precedingSymbol is { } value && !char.IsWhiteSpace(value))
             {
                 statement.Append(symbol);
             }
--- a/Src/FluentAssertions/Common/Guard.cs
+++ b/Src/FluentAssertions/Common/Guard.cs
@@ -2,12 +2,13 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using JetBrains.Annotations;
 
 namespace FluentAssertions.Common;
 
 internal static class Guard
 {
-    public static void ThrowIfArgumentIsNull<T>([ValidatedNotNull] T obj,
+    public static void ThrowIfArgumentIsNull<T>([ValidatedNotNull][NoEnumeration] T obj,
         [CallerArgumentExpression(nameof(obj))]
         string paramName = "")
     {
@@ -17,7 +18,7 @@ internal static class Guard
         }
     }
 
-    public static void ThrowIfArgumentIsNull<T>([ValidatedNotNull] T obj, string paramName, string message)
+    public static void ThrowIfArgumentIsNull<T>([ValidatedNotNull][NoEnumeration] T obj, string paramName, string message)
     {
         if (obj is null)
         {
--- a/Src/FluentAssertions/Types/TypeSelector.cs
+++ b/Src/FluentAssertions/Types/TypeSelector.cs
@@ -293,14 +293,13 @@ public class TypeSelector : IEnumerable<Type>
     {
         types = types.ConvertAll(type =>
         {
-            if (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Task<>))
+            if (type.IsGenericType)
             {
-                return type.GetGenericArguments().Single();
-            }
-
-            if (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(ValueTask<>))
-            {
-                return type.GetGenericArguments().Single();
+                Type genericTypeDefinition = type.GetGenericTypeDefinition();
+                if (genericTypeDefinition == typeof(Task<>) || genericTypeDefinition == typeof(ValueTask<>))
+                {
+                    return type.GetGenericArguments().Single();
+                }
             }
 
             return type == typeof(Task) || type == typeof(ValueTask) ? typeof(void) : type;
--- a/qodana.yaml
+++ b/qodana.yaml
@@ -5,3 +5,7 @@ failThreshold: 0
 
 dotnet:
   solution: FluentAssertions.sln
+
+exclude:
+  - name: ConvertIfStatementToReturnStatement
+  - name: ConvertIfStatementToConditionalTernaryExpression
--- a/Src/FluentAssertions/Common/Configuration.cs
+++ b/Src/FluentAssertions/Common/Configuration.cs
@@ -40,12 +40,7 @@ public class Configuration
         {
             lock (propertiesAccessLock)
             {
-                if (!valueFormatterDetectionMode.HasValue)
-                {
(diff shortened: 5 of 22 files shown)


lg2de:
> There are a lot of warnings about unused methods/return values of public methods.
> To silence them, it seems the idiomatic way is to annotate them with [`[PublicAPI]`](https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#PublicAPIAttribute)

Public API shall be used by the unit tests.
So, it should be used, incl. return values, isn't it?

jnyrup:
> > There are a lot of warnings about unused methods/return values of public methods.
> > To silence them, it seems the idiomatic way is to annotate them with [`[PublicAPI]`](https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#PublicAPIAttribute)
> 
> Public API shall be used by the unit tests. So, it should be used, incl. return values, isn't it?

Examples:
`DateTimeAssertions.BeOneOf(IEnumerable<DateTime?> validValues, string because = "", params object[] becauseArgs)` is marked with `Member can be made private (non-private accessibility)`.
It is never called directly by unit tests, only from other overloads of `DateTimeAssertions.BeOneOf`.

The same method is also marked `Method return value is never used` because we exercise the chainability in the unit tests.

Unless we explicitly exercise all overloads including their chainability, I assume adding `[PublicAPI]` is the way to silence these warnings.
Note: I'm completely new to Qodana, so I'm open to suggestions.

lg2de:
> It is never called directly by unit tests, only from other overloads of `DateTimeAssertions.BeOneOf`.

Is there a reason to not test it directly?

If we would follow the rule and making the method private, only the API verification would fail because a method is missing.
But, there is not test validating the behavior of this method. It may get broken with internal refactoring.
I would argue for direct testing every public method directly.

I could help to provide such tests, but so far I did not find access to the Qodana report.

qodana.yaml
```diff
@@ -5,3 +5,7 @@ failThreshold: 0
 
 dotnet:
   solution: FluentAssertions.sln
+
+exclude:
```
dennisdoomen, line 9: ⛏️ Would be nice to see what it is complaining about.

Src/FluentAssertions/Common/Guard.cs
```diff
@@ -2,12 +2,13 @@
...
 using System.Linq;
 using System.Runtime.CompilerServices;
+using JetBrains.Annotations;
 
 namespace FluentAssertions.Common;
 
 internal static class Guard
 {
-    public static void ThrowIfArgumentIsNull<T>([ValidatedNotNull] T obj,
+    public static void ThrowIfArgumentIsNull<T>([ValidatedNotNull][NoEnumeration] T obj,
```
dennisdoomen, line 11: 🤔 I'm starting to wonder I like all these attributes. What do you think?

Src/FluentAssertions/Types/TypeSelector.cs
```diff
@@ -293,12 +293,8 @@ public TypeSelector UnwrapTaskTypes()
     {
         types = types.ConvertAll(type =>
         {
-            if (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Task<>))
-            {
-                return type.GetGenericArguments().Single();
-            }
-
-            if (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(ValueTask<>))
+            if (type.IsGenericType && type.GetGenericTypeDefinition() is var genericTypeDefinition
```
dennisdoomen, line 296: 🤔 I'm on the brink whether I actually like this.

Src/FluentAssertions/CallerIdentification/AddNonEmptySymbolParsingStrategy.cs
```diff
@@ -16,7 +16,7 @@ public ParsingState Parse(char symbol, StringBuilder statement)
         }
         else if (mode is Mode.RemoveSuperfluousWhitespace)
         {
-            if (precedingSymbol is char value && !char.IsWhiteSpace(value))
+            if (precedingSymbol is { } value && !char.IsWhiteSpace(value))
```
dennisdoomen, line 19: ❓ I don't get this one. Why is it better?

jnyrup: The general idea is that instead of checking if an object is an instance of a particular type, pattern matching let's you check that is not null and still assign it to a variable.
The benefit comes with inheritance where `is T` could check if the object is an instance of _another_ type, whereas `is { }` will just check if "is something", i.e. not null.

When the compiler can see that `is T` can _only_ be a null-check, i.e. the object is either null or an instance of `T`, `is T` also compiles down to a simple null check like `is {}` or `is not null`.

Src/FluentAssertions/Types/TypeSelector.cs
jnyrup: Care to share what you don't like :wink: 

The thing about the current code that annoys me a bit is that we check `type.IsGenericType` twice.
It's like the code yells at me "I JUST TOLD YOU THAT I'M A GENERIC TYPE! :rage:"

How do you feel about this less condensed version then?
```cs
if (type.IsGenericType)
{
    var genericTypeDefinition = type.GetGenericTypeDefinition();
    if (genericTypeDefinition == typeof(Task<>) || genericTypeDefinition == typeof(ValueTask<>)))
    {
        return type.GetGenericArguments().Single();
    }
}
```

Src/FluentAssertions/Common/Guard.cs
dennisdoomen: Let's see how this evolves. It if becomes too noisy, we can always decide to disable the rule and remove the attributes.

Src/FluentAssertions/Types/TypeSelector.cs
dennisdoomen: That would work well. I think my main issue was with the nested _ands_ and _ors_.

Src/FluentAssertions/CallerIdentification/AddNonEmptySymbolParsingStrategy.cs
eNeRGy164: as `is {}` is the same as `is not null`, would the last one not be the preferable notation?

dennisdoomen: OK. That makes sense. A link to the relevant post would have been enough to convince me ;-)

dennisdoomen (approved):

Src/FluentAssertions/CallerIdentification/AddNonEmptySymbolParsingStrategy.cs
jnyrup: > as `is {}` is the same as `is not null`, would the last one not be the preferable notation?

`is not null` does not support declaration patterns, i.e. assigning the result to new variable, which is used here to check that `char? precedingSymbol` has a value _and_ extract that value to `char value`.

> OK. That makes sense. A link to the relevant post would have been enough to convince me ;-)

Why do you need blog posts when you have me?

Issue #2205, opened 2023-05-15
Assertions on System.Text.Json


Corniel:
### Background and motivation

There is a separate package to assert on `NewtonSoft.Json`, but given that `System.Text.Json` is replacing the other more and more, I would expect me not being the only one who would like to see assertions that help testing JSON (de)serialization based on `System.Text.Json`.

I'm more than willing to help out to create a PR for such a feature, but before doing that (quite) some questions have to be answered.

1. Should it be a separate package, or part of the core? (I would prefer the latter, as it is part of the core of .NET since 5.0)
2. What is the object to assert on? `string`, `JsonElement`, `JsonSerializerOptions`, or ...

For me personally, the most important use case is to write specs that guard the (overall) JSON serialization contract did not change.  So something like:

``` C#
namespace Serialization_Json;

public class Serializes
{
    private readonly JsonSerializerSettings Settings = ...; // Whatever rows your boat.

    [Test]
    public void enum_values_by_name()
        =>   Settings.Serialize(NumberKind.One).Should().BeJsonString("One");
    // Or        
    [Test]
    public void enum_values_by_integer()
        =>   Settings.Serialize(NumberKind.One).Should().BeJsonNumber(1);
        
    [Test]
    public void models_skipping_null_properties()
        =>  Settings.Serialize(new Person { Name = null }).Should().BeJson("{}");
}

static class JsonSerializerOptionsExtensions
{
    public static JsonElement Serialize<T>(this JsonSerializerOptions options, T value)
    {
        var bytes = JsonSerializer.SerializeToUtf8Bytes(value, options);
        using var doc = JsonDocument.Parse(bytes );
        return doc.RootElement.Clone();
    }
}
```

And obviously also for deserialize.  But is this the kind of API that other users would also like to see? And if not, what kind of API to assert JSON (de)serialization is preferable?

dennisdoomen:
I think having this could even be a non-breaking feature, but I would prefer to move this in v7 as an incentive to have people upgrade.

Corniel:
I've created a draft PR, just to play with an actual implementation that could do the trick.

dennisdoomen:
Cool. The idea is that the API is compatible with the https://github.com/fluentassertions/fluentassertions.json.

vbreuss:
I think this issue is too broad and abstract, so I tried to split it into multiple concrete API suggestions:
- #2555 
- #2556
- #2559

lucasteles:
[...]

PR #2246, opened 2023-08-01
Support assertions on System.Text.Json.JsonSerializerOptions based (de)serialization

Corniel:
An important part of JSON (de)serialization specifications, is defining the intended behavior of it. This (draft) PR an attempt to create a readable/maintainable/extendable syntax, that fits within FluentAssertions. 

## IMPORTANT 

See:  #2205


Corniel:
Added `json.md` as a page, to describe the intent.

Src/FluentAssertions/Json/JsonElementAssertions.cs
```diff
@@ -0,0 +1,152 @@
...
+    /// </summary>
+    /// <param name="value">The value of the JSON string node.</param>
+    /// <param name="because">
+    ///     A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+    ///     is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+    /// </param>
+    /// <param name="becauseArgs">
+    ///     Zero or more objects to format using the placeholders in <see paramref="because" />.
+    /// </param>
+    public AndConstraint<JsonElementAssertions> BeString(string value, string because = "", params object[] becauseArgs)
```
ITaluone, line 76: Why not use `Be()` in every case? The type info on the method name is redundant.

Corniel: I considered that, and if we would only assert on simple nodes, that might would be sufficient.

However, if the JSON is a complex structure, you might want to assert against the JSON string it results in, what is different then the a JSON string node/element. Besides that, it also makes implicit that both the value and the element type are asserted.

In general, what the API should look like for JSON array, and JSON object results, is something I'm still thinking about. I think I personally would use this to assert on the full JSON result as a string:

``` C#
options.Should().Serialize(myComplexModel)
    And.Shoud().Be("{ 'prop': 'Value', 'other': 12, children: [ ]}");
```

but others might opt for:


``` C#
options.Should().Serialize(myComplexModel)
    And.Shoud().Be(new()
    {
        prop = "Value" ,
        other = 12,
        children = Array.Empty<object>()
     });
```

Where the anonymous object is used to represent the `JsonElement`. The latter is normally used for `BeEquivalent()`, but I'm wondering if that has meaning for this kind of assertions.

I'm also wondering if we want the opposite of `BeString()`, etc. Because, what does that say? That I except something not to be a string? Or a string, but with a different value?!

dennisdoomen:
Thanks for taking the lead to spike out the implementation. I still have doubts though. 

Consider this example from the PR.

```csharp
options.Should().Serialize("42").And.Value.Should().BeNumber(42, because: "why not?");
```

I have two issues with this:
1. It doesn't follow the same style like the rest of FA's API
1. `And` is normally used to execute another assertion on the SUT, which in this case would be `options`

To make it more idiomatic FA, it could look like this:

```csharp
"42".Should().BeSerializableToJson(options).Which.Value().Should().BeNumber(42, because: "why not?");
```

Corniel:
`.Witch`, instead of `.And` is an improvement, I'll change that.

For me, there are two types of (de)serialization tests I'd like to write: when a custom type that is decorated with custom `JsonConveter`, I would like to test that. In that case the system under test is that custom type, hence: `myCustomType.Should(),BeJsonSerializable().Witch.Value().Should().Be("{ 'someProp': 42 }")`.

The other kind of test, is about the `JsonSeralizationOptions` defined for an API. In that case, it are the options that are the system under test, and I think that `options.Should().Seralize().Which.Value()` describe that kind of scenario's better, then the other way around. Actually, in the first case, I Might not want to specify the options in most cases, but use the default settings.

Corniel:
@dennisdoomen About `.Should().JsonDeserialize()`. would you like to have that as an extension method, or just on ObjectAssertions? (And some others).

dennisdoomen:
3> The other kind of test, is about the `JsonSeralizationOptions` defined for an API. In that case, it are the options that are the system under test, and I think that `options.Should().Seralize().Which.Value()` describe that kind of scenario's better, then the other way around. Actually, in the first case, I Might not want to specify the options in most cases, but use the default settings.

I'm sorry, but I disagree. `JsonSeralizationOptions` should never be the _subject_ of a test. It's not the options that serialize or deserialize anything, it's the serializer. In other words, those options are what you pass to the serializer. 

> About `.Should().JsonDeserialize()`. would you like to have that as an extension method, or just on ObjectAssertions? (And some others).

I think so, just like `Should().BeEquivalentTo`

Corniel:
> I'm sorry, but I disagree. `JsonSeralizationOptions` should never be the _subject_ of a test. It's not the options that serialize or deserialize anything, it's the serializer. In other words, those options are what you pass to the serializer.

That is also true.  Unfortunately, the JSON Serializer of System.Text.Json is a static method. It are these options that define its behavior, and you want to test if those the setup of your serializer will result in the kind of serialization you like.  (With System.Text.Json, you could actually resolve a serializer with - in that  case settings - to have something like: `serializer.Should().Serialize()`).

in outer words: I would like to be able to say: `serializer.Should().Serializer(whatever).Which.Value().Should().Be`, but System.Text.Json will not allow me to do that. and to me: `options.Should().Serialize(whatever).Which().Value().Should().Be` is the best next thing for that scenario.

If I want to specify that because of my JsonOptions, null properties are ignored, it is not about the object that is serialized, but about that it is configured that way (in the options). So to me:

``` C#
options.Should().Serialize().Which.Value().Should().Be("{}");
```

Is then better than:

``` C#
new { Name = (string) null }.Should().BeJsonSerializable(options).Which.Value().Should().Be("{}");
```

I've thought of another name than `Serialize`, as technically, the options indeed only describe who to serialize, they do not perform it themselves, but could not come up with a name that also makes clear that the continuation with `.Which.Value()` does not come as a surprise, or are extremely vebose, like `options.Should().SpecifySeralizationOf(whatever)`.

> > About `.Should().JsonDeserialize()`. would you like to have that as an extension method, or just on ObjectAssertions? (And some others).
> 
> I think so, just like `Should().BeEquivalentTo`

Than I'll add that one.

dennisdoomen:
> new { Name = (string) null }.Should().BeJsonSerializable(options).Which.Value().Should().Be("{}");

What's wrong with this one? There's a _subject-under-test_, there are the `options` you want to pass in, it's idiomatic FA,  and you get to continue with the resulting `JObject` (or something like that.

Corniel:
@dennisdoomen About `.Should().JsonDeserialize()`. would you like to have that as an extension method, or just on ObjectAssertions? (And some others).

dennisdoomen:
> About `.Should().JsonDeserialize()`. would you like to have that as an extension method, or just on ObjectAssertions? (And some others). That is an advantage of defining it on the options: you have all assertions at one place.

If you put it on `ReferenceTypeAssertions`, it's practically available everywhere.

PR #2247, opened 2023-08-05
Group framework dependent configuration

Corniel:
To increase the readability of the csproj file, I've grouped toe framework dependent configuration. I also added some `Label`'s.

### DIFF
--- a/Src/FluentAssertions/FluentAssertions.csproj
+++ b/Src/FluentAssertions/FluentAssertions.csproj
@@ -1,4 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+
   <!-- To reduce build times, we only enable analyzers for the newest TFM -->
   <PropertyGroup>
     <TargetFrameworks>net47;net6.0;netcoreapp2.1;netcoreapp3.0;netstandard2.0;netstandard2.1</TargetFrameworks>
@@ -11,7 +12,8 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
   </PropertyGroup>
-  <PropertyGroup>
+  
+  <PropertyGroup Label="Package info">
     <Authors>Dennis Doomen;Jonas Nyrup</Authors>
     <PackageDescription>
       A very extensive set of extension methods that allow you to more naturally specify the expected outcome of a TDD or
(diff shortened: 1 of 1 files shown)


jnyrup (commented):
For the `<Compile Remove`, e.g. for `SystemExtensions.cs` I prefer the current feature-grouped approach as that'll ensure that future included TFMs don't include it.

But I have no strong opinions about this. I find both TFM and feature grouped approaches equally readable, it boils down to if you're current view is a feature or a TFM.

dennisdoomen:
> But I have no strong opinions about this. I find both TFM and feature grouped approaches equally readable, it boils down to if you're current view is a feature or a TFM

I had the same feeling about this.

dennisdoomen (approved):

Corniel:
I tried to get my head around what was disabled and when, And only be rewriting it, it started making sense.

dennisdoomen:
On second thought, I actually think the result of this PR makes it better.

Corniel:
@dennisdoomen 

If I understand correctly, the reason the Qodana steps fails is this one?

> To continue using Qodana, please ensure you have an access token and provide the token as the QODANA_TOKEN environment variable.

Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -40,58 +43,82 @@
       <_Parameter1>Benchmarks, PublicKey=00240000048000009400000006020000002400005253413100040000010001002d25ff515c85b13ba08f61d466cff5d80a7f28ba197bbf8796085213e7a3406f970d2a4874932fed35db546e89af2da88c194bf1b7f7ac70de7988c78406f7629c547283061282a825616eb7eb48a9514a7570942936020a9bb37dca9ff60b778309900851575614491c6d25018fadb75828f4c7a17bf2d7dc86e7b6eafc5d8f</_Parameter1>
     </AssemblyAttribute>
   </ItemGroup>
-  <ItemGroup>
-    <None Remove="BannedSymbols.txt" />
```
jnyrup, line 44: I can't recall why I added this.
Does removing it changes anything?
I.e. is it excluded from the nuget package and does the analyzer still flag usage of banned APIs?

dennisdoomen:
> If I understand correctly, the reason the Qodana steps fails is this one?

We do have a valid license, but Github won't pass that token to PR builds. That's something we still need to choose a workaround for.

jnyrup (approved):

PR #2258, opened 2023-08-13
Another Qodana test from the same repository

dennisdoomen:
Changed the workflow that invokes Qodana to always run the GHA workflow that is part of the target branch (using the `pull_request_target`). This should hopefully allow it to be run on PRs coming from forks as well.

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -1,7 +1,8 @@
 name: Qodana
 
 on:
-  pull_request:
+  workflow_dispatch:
+  pull_request_target:
   push:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


jnyrup (approved):
Improve the commit description from "save point"

IT-VBFK:
Should we tackle the coveralls.net NUKE problem again with this solution? 😆

dennisdoomen:
I suspect it is already solved by that.

IT-VBFK:
Hmm... I don't think so, because the `build.yaml` only runs on `push` and `pull_request`.

dennisdoomen:
Ah, then we have to update that one as well.

IT-VBFK:
Should I give this try nr. 143246536? 😂

PR #2275, opened 2023-08-23
Introduction of interface to improve extensibility

Corniel:
What about introducing an interface that can be used to write extensions like this?

``` C#
using System;
using System.Globalization;
using Xunit;

namespace FluentAssertions.Specs;

public class IAssertionsSpecs
{
    [Fact]
    public void Test()
    {
        var assertions = new FormatableAssertions<int>(42);

        assertions.BeFormatted("0.00", "42.00");
    }
}

internal sealed class FormatableAssertions<TSubject> : IAssertions<IFormattable>
    where TSubject : IFormattable
{
    public FormatableAssertions(TSubject subject)
    {
        Subject = subject;
    }

    public TSubject Subject { get; }

    IFormattable IAssertions<IFormattable>.Subject => Subject;
}

public static class FormatableExtensions
{
    public static AndConstraint<TAssertions> BeFormatted<TAssertions>(this TAssertions assertions, string format, string result)
        where TAssertions : IAssertions<IFormattable>
    {
        assertions.Subject.ToString(format, CultureInfo.InvariantCulture).Should().Be(result);

        return new AndConstraint<TAssertions>(assertions);
    }
}
```

### DIFF
--- a/Src/FluentAssertions/IAssertions.cs
+++ b/Src/FluentAssertions/IAssertions.cs
@@ -1,0 +1,8 @@
+﻿#nullable enable
+
+namespace FluentAssertions;
+
+public interface IAssertions<out TSubject>
+{
+    TSubject? Subject { get; }
+}
--- a/Tests/FluentAssertions.Specs/IAssertionsSpecs.cs
+++ b/Tests/FluentAssertions.Specs/IAssertionsSpecs.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Globalization;
+using Xunit;
+
+namespace FluentAssertions.Specs;
+
+internal class IAssertionsSpecs
+{
+    [Fact]
+    public void Test()
+    {
+        var assertions = new FormatableAssertions<int>(42);
+
+        assertions.BeFormatted("0.00", "42.00");
+    }
+}
+
+internal sealed class FormatableAssertions<TSubject> : IAssertions<IFormattable>
+    where TSubject : IFormattable
+{
+    public FormatableAssertions(TSubject subject)
+    {
+        Subject = subject;
+    }
+
+    public TSubject Subject { get; }
+
+    IFormattable IAssertions<IFormattable>.Subject => Subject;
+}
+
+public static class FormatableExtensions
+{
+    public static AndConstraint<TAssertions> BeFormatted<TAssertions>(this TAssertions assertions, string format, string result)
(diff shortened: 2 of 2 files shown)


dennisdoomen:
I don't (yet) see the advantage of this. Care to elaborate?

Corniel:
It would allow to define assertions based on contracts. 

Lets asume I want to do this assertion

``` C#
public static class ComparableExtensions
{
    public static AndConstraint<TAssertions> CompareWithNullAsOne<TAssertions>(this TAssertions assertions)
        where TAssertions : IAssertions<IComparable>
    {
        assertions.Subject.CompareTo(null).Should().Be(1);

        return new AndConstraint<TAssertions>(assertions);
    }
}
```

So every every custom assertion that has an `IComparable` subject (Guid, Date, DateOnly, etc.), would then benefit from this extra assertion. I'm definitely not sure either, but after writing some custom assertions (like [Qowaiv.Validation.TestTools](https://github.com/Qowaiv/qowaiv-validation/tree/main/src/Qowaiv.Validation.TestTools)), and sometimes I was searching for some (extra) extensibility.

The downside, obviously is, that you have define the interfaces you want to support for assertion class, and that you have to implement those interfaces with explicit declarations.

Note that the following will unfortunately not fly:
``` C#
public static class FormatableExtensions
{
    public static AndConstraint<TAssertions> BeFormatted<TAssertions, TSubject>(this TAssertions assertions, string format, string result)
        where TAssertions : IAssertions<TSubject>
        where TSubject : IFormatable
    {
        assertions.Subject.ToString(format, CultureInfo.InvariantCulture).Should().Be(result);

        return new AndConstraint<TAssertions>(assertions);
    }
} 
```

Because the compiler then has to be told which generic arguments to use. That is a pity, because it would make our life way easier.

dennisdoomen:
What does `IAssertion<T>` look like? Did you try this to see if we end up in some overloading resolution issue here and there?

Corniel:
> What does `IAssertion<T>` look like? Did you try this to see if we end up in some overloading resolution issue here and there?

Not yet. I was also wondering if you ever experimented with something like this.

dennisdoomen:
> Not yet. I was also wondering if you ever experimented with something like this.

Me neither. But it could work. Somebody should spike this.

Corniel:
I tried it, and didn't like the results.

dennisdoomen:
OK. Thanks for trying.

PR #2276, opened 2023-08-23
Fix Possible 'System.NullReferenceException' warning from Rider and Qodana

lg2de:


dennisdoomen (approved):
I still have doubts about all those `!`, but we'll see.

Corniel (approved):
LGTM.

IT-VBFK:
Do you mind adding those 7 possible NRE too?

image

lg2de:
@IT-VBFK I'll take a look tonight.

lg2de:
I focused on the issues which are fixable with `!` because the subject is already checked.
For easier review I do not handler other issues here.

jnyrup (approved):
I looked through all the changes and agree that we have more knowledge than the compiler here.

Issue #2234, opened 2023-07-18
[API Proposal]: Convert DateTime instance to DateOnly or TimeOnly


mu88:
### Background and motivation

The Fluent Assertion syntax for `DateTime` objects (e. g. `12.April(1953).At(19, 53)`) is phantastic. But IMHO it could be even better with two new extension methods `.ToDateOnly()` and `.ToTimeOnly()`. This would allow keeping a fluent syntax (instead of `new DateOnly(1953, 4, 12)`) and shorter (instead of `DateOnly.FromDateTime(12.April(1953))`) .

### API Proposal

```C#
public static class FluentDateTimeExtensions
{
    public static DateOnly ToDateOnly(this DateTime date);

    public static TimeOnly ToTimeOnly(this DateTime date);
}
```

### API Usage

```C#
var now = 12.April(1953);

var date = now.ToDateOnly();
var time = now.ToTimeOnly();
```

### Alternative Designs

/

### Risks

Since it's only a new extension method to be added, a regression or breaking change should not happen.

dennisdoomen:
I like that. If @jnyrup agrees, would you be willing to provide a PR for this?

mu88:
Yeah sure, feel free to assign the issue to me if both of you consider it worth

jnyrup:
Preamble: I'll poke and challenge the API, it's not to shoot the messenger, but how I get a better grasp and feeling of an API.

I wasn't around when the first date/time extensions were introduced in c143b577974d0edb79fc2a584559dc25146c2c55, but I assume one goal was writing time instants in a more natural language compared to using one of `DateTime`'s many constructors.
```cs
19.April(2020).At(12, 34)
// vs
new DateTime(2020, 4, 19, 12, 34, 00)
```

C# now has target typed new expressions, so one can write a `TimeOnly` quite concise
```cs
timeOfDay.Should().Be(new(13, 37))
```

If one currently wants to use chaining calls to achieve the same result I guess it would be something like this.

```cs
13.Hours(37.Minutes()).ToTimeOnly()
```

Note that the example above assumes we also add this extension method in addition to the proposed ones.
```cs
public static TimeOnly ToTimeOnly(this TimeSpan time);
```

I haven't used `DateOnly` and `TimeOnly` besides the existing tests in FA, so I would appreciate some more complete examples to design an API that fits the usage.
That is, I'd like to avoid creating some API, just "because we can".

mu88:
[...]

PR #2279, opened 2023-08-27
DateBuilder that allows DateOnly building too

Corniel:
As illustration of the discussion in #2234


ITaluone:
@Corniel I like this :)

Although I would suggest to drop the `AsOffset()` and `WithOffset()` methods inside the `DateBuilder` struct, and only use extension methods for this, like:

```csharp
public static DateTimeOffset WithOffset(this DateTime self, TimeSpan offset) => new(self, offset);
public static DateTimeOffset WithOffset(this DateBuilder self, TimeSpan offset) => new(self, offset);
public static DateTimeOffset AsOffset(this DateTime self) => new(self, 0.Hours());
public static DateTimeOffset AsOffset(this DateBuilder self) => new(self, 0.Hours());
```

This will address https://github.com/fluentassertions/fluentassertions/issues/2234#issuecomment-1694726115

Corniel:
> @Corniel I like this :)
> 
> Although I would suggest to drop the `AsOffset()` and `WithOffset()` methods inside the `DateBuilder` struct, and only use extension methods for this, like:

Why define the offset ,ethods as extensions on the date builder? It's only purpose is just to do things like that?!

IT-VBFK:
> Why define the offset methods as extensions on the date builder?

Because it shouldn't matter if you call `AsOffset()`/`WithOffset()` after the date "thing" like `10.May(2023).AsOffset()` or after the time "thing" like `10.May(2023).At(...).AsOffset()`.

Now this is not possible, because `AsOffset()` is not recognized when chaining after `At(...)`.
And if you introduce `AsOffset()` as an extension method, the `AsOffset()`in the `DateBuilder` becomes useless.

Do you see my point?

Corniel:
@IT-VBFK: I ment, that `AsOffset` should not be an extension for the `DateBuilder`. For `DateTime`  it should.

PR #1197, opened 2019-11-30
Make `ThatAreUnderNamespace` respect null namespaces and prefixes

jnyrup:
The namespace of a type not inside any namespace is `null`.
As all types are *under* the global namespace, `ThatAreUnderNamespace` should be able to gracefully handle `null` namespaces.

This also fixes that `ThatAreUnderNamespace` would incorrectly match the
namespace of a type against a partial prefix.

This fixes #1185 and #1196

PR #2292, opened 2023-09-03
Introduced a new attribute to allow global initialization of the assertion options

### LINKED ISSUE #2290: Allow extension libraries to alter options before the first assertion
Extension libraries have typically two needs that are currently not provided:

1. Being able to mark themselves as an extension library so that `CallerIdentifier` doesn't need to check for individual `CustomAssertion` attributes
2. Being able to inject additional capabilities, like custom implementations of `IEquivalencyStep` such as provided by `FluentAssertions.DataSets`.

I'm proposing to introduce an assembly-level attribute that FA will scan for before it executes any kind of assertion. This attribute will point to an initialization class/method that can be used to change FA defaults, e.g. using `AssertionOptions`. 

I'm not 100% sure yet, but I suspect this scanning could be triggered by the static constructor of `AssertionExtensions`

Is a prerequisite for finishing #2002

dennisdoomen:
This PR introduces FA's own "module initializer" through the new `[AssertionEngineInitializer]` attribute. It can be used multiple times. 

For future reference, we're not using [`ModuleInitializerAttribute`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.moduleinitializerattribute) for two reasons:

1.  .NET Framework 4.x doesn't have that attribute. 
2. Module initializers are not invoked until the assembly containing the initializer is actually needed at run-time. As a consequence of that, a package like `FluentAssertions.DataSets` may not even get a chance to register its custom implementations of `IEquivalencyStep` when somebody invokes `BeEquivalentTo` without explicitly using code from that package first. To resolve this, you could have that package add a source file as _content_ (in NuGet terms) that adds a module initializer. Then it would be invoked the first time the test project is loaded. But that would require that source file to define the `ModuleInitializerAttribute` for .NET 4.x, which is hard since you can't detect whether it is needed. [non-SDK projects do not define attributes](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives) like NETFRAMEWORK and such. If that test project already defines that attribute in a different way (e.g. using Polysharp), it would cause compilation conflicts.

Also, while testing with `FluentAssertions.DataSets`, it turns out the caller identifier was misbehaving if Fluent Assertions code preceded the data set extensions code in the stack trace. This has been fixed as well. 

Tested with the [FluentAssertions.DataSets](https://github.com/fluentassertions/fluentassertions.datasets) repository.

**Note** Review by commit

Closes #2290


dennisdoomen:
> Overall, it looks good. I think explaining what the effect is of 'using multiple times', v.s. 'called once' might be explained in more detail.

Good point. Will look a the description. Need to rebase anyway.

Src/FluentAssertions/Common/Services.cs
```diff
@@ -33,14 +38,66 @@ public static Configuration Configuration
...
+            foreach (var attribute in attributes)
+            {
+                attribute.Initialize();
+            }
+        }
+        catch
+        {
+            // Just ignore any exceptions that might happen while finding and invoking initialize
+        }
+    }
```
jnyrup, line 96: We could also try/catch each call to `Initialize` such that one bogus extension doesn't prevent the other ones from running.

Src/FluentAssertions/Common/Services.cs
```diff
@@ -33,14 +38,66 @@ public static Configuration Configuration
...
+        var currentAssembly = Assembly.GetExecutingAssembly();
+        var currentAssemblyName = currentAssembly.GetName();
+
+        try
+        {
+            var attributes = AppDomain.CurrentDomain
+                .GetAssemblies()
+                .Where(assembly => assembly != currentAssembly && !assembly.IsDynamic && !IsFramework(assembly))
+                .Where(a => a.GetReferencedAssemblies().Any(r => r.FullName == currentAssemblyName.FullName))
+                .SelectMany(a => a.GetCustomAttributes<AssertionEngineInitializerAttribute>());
```
jnyrup, line 85: We might want to be more specific on the method signature we scan for - but this could also be improved in a later PR.

Even though we catch exceptions, it's annoying if a third party library throws internally and triggers a debugger to break execution.

This example throws `System.Reflection.TargetParameterCountException: 'Parameter count mismatch.'`
```cs
public static void InitializeBeforeFirstAssertion(object _)
{
}
```

This example throws `System.Reflection.AmbiguousMatchException: 'Ambiguous match found.'`
```cs
public static void InitializeBeforeFirstAssertion()
{
    Interlocked.Increment(ref shouldBeCalledOnlyOnce);
}

public static void InitializeBeforeFirstAssertion(object _)
{
}
```

docs/_pages/extensibility.md
```diff
@@ -275,6 +275,21 @@ internal static class Initializer
 }
 ```
 
+Unfortunately, only a single module initializer is allowed. That's why Fluent Assertions supports its own "module initializer" through the `[AssertionEngineInitializer]` attribute. It can be used multiple times.
```
jnyrup, line 278: I can get several module initializers to run 🤨 
Am I missing something?

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/module-initializers
> When one or more valid methods with this attribute are found in a compilation, the compiler will emit a module initializer which calls each of the attributed methods. The calls will be emitted in a reserved, but deterministic order.

Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -48,8 +49,10 @@
     <PackageReference Include="JetBrains.Annotations" Version="2023.2.0" PrivateAssets="all" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
     <PackageReference Include="PolySharp" Version="1.13.2" PrivateAssets="all" />
+    <PackageReference Include="xunit.extensibility.core" Version="2.5.0" />
```
jnyrup, line 52: Remove 🙃

docs/_pages/extensibility.md
dennisdoomen: You're right, it can and it all seems to work. The other caveat is that .NET Framework 4.x doesn't have that attribute. So when the `FluentAssertions.DataSet` package adds a source file to the target test project with a module initializer, it would also have to define the attribute for .NET 4.x. However, [non-SDK projects do not define attributes](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives) like `NETFRAMEWORK` and such.

Src/FluentAssertions/Common/Services.cs
dennisdoomen: At this point, I'm not checking for any signatures. Am I missing something?

docs/_pages/extensibility.md
```diff
@@ -275,6 +275,21 @@ internal static class Initializer
...
 ```
 
+Unfortunately, this only works for .NET 5 and higher. That's why Fluent Assertions supports its own "module initializer" through the `[AssertionEngineInitializer]` attribute. It can be used multiple times. 
+
+```csharp
+[assembly: AssertionEngineInitializer(typeof(Initializer), nameof(Initializer.Initialize))]
+
+public static class Initializer
+{
+    public static void Initialize
```
jnyrup, line 285:
```suggestion
    public static void Initialize()
```

docs/_pages/introduction.md
```diff
@@ -63,6 +63,23 @@ xDocument.Should().HaveElement("child").Which.Should().BeOfType<XElement>().And.
...
+## Global Configurations
+
+Fluent Assertions `AssertionOptions` has several methods and properties that can be used to change the way it executes assertions or the defaults it will use for comparing object graphs. Changing those settings at the right time can be difficult, depending on the test framework. That's why Fluent Assertions offers a special assembly-level attribute that can be used to have some code executed _before_ the first assertion is executed. It will be called only once per test run, but you can use the attribute multiple times. 
+
+```csharp
+[assembly: AssertionEngineInitializer(typeof(Initializer), nameof(Initializer.Initialize))]
+
+public static class Initializer
+{
+    public static void Initialize
```
jnyrup, line 75:
```suggestion
    public static void Initialize()
```

Tests/FluentAssertions.Extensibility.Specs/FluentAssertions.Extensibility.Specs.csproj
```diff
@@ -0,0 +1,49 @@
...
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net47'">
+    <!--
+      (SRCU = System.Runtime.CompilerServices.Unsafe)
+      FluentAssertions.csproj depends on SRCU 4.5.0 (AssemblyVersion 4.0.4.0)
+      System.Collections.Immutable 1.6.0+ depends on SRCU 4.5.2+ (AssemblyVersion 4.0.4.1)
+      FluentAssertions.Specs.csproj for net47 does not generate a binding redirect from 4.0.4.0 to 4.0.4.1
+      Upgrading to 1.6.0 gives "Could not load file or assembly 'System.Runtime.CompilerServices.Unsafe, Version=4.0.4.0"
+    -->
+    <PackageReference Include="System.Collections.Immutable" Version="[1.5.0]" />
```
jnyrup, line 20: This can probably be removed.
As I recall it's only there to get access to `ImmutableArray`.

Src/FluentAssertions/Common/Services.cs
jnyrup: Right, we don't filter on signatures here, i.e. at this point we don't try to validate if the method about the be called exists.
When invoking `attribute.Initialize()` and hence `AssertionEngineInitializerAttribute.Initialize` we assume that `type` has a single static parameterless method named `methodName`.

As this hopefully is an extreme edge case and nothing we can do at this point in the code the fix it, I'm fine with the added try/catch inside the loop to prevent bogus ones interfering with working ones.

Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -50,6 +50,7 @@
     <PackageReference Include="PolySharp" Version="1.13.2" PrivateAssets="all" />
   </ItemGroup>
 
+
```
jnyrup, line 53:
```suggestion
```

Tests/FluentAssertions.Specs/FluentAssertions.Specs.csproj
```diff
@@ -47,4 +47,4 @@
     <ProjectReference Include="..\AssemblyB\AssemblyB.csproj" />
   </ItemGroup>
 
-</Project>
+  </Project>
```
jnyrup, line 50:
```suggestion
</Project>
```

jnyrup (approved):

Src/FluentAssertions/Extensibility/AssertionEngineInitializerAttribute.cs
```diff
@@ -0,0 +1,30 @@
...
+/// Can be added to an assembly so it gets a change to initialize Fluent Assertions before the first assertion happens.
+/// </summary>
+[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
+public sealed class AssertionEngineInitializerAttribute : Attribute
+{
+    private readonly string methodName;
+    private readonly Type type;
+
+    /// <summary>
+    /// Defines the static void-returned and parameterless method that should be invoked before the first assertion happens.
```
jnyrup, line 16: void-returned -> void-returning?

Tests/FluentAssertions.Extensibility.Specs/FluentAssertions.Extensibility.Specs.csproj
```diff
@@ -0,0 +1,48 @@
...
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net47'">
+    <!--
+      (SRCU = System.Runtime.CompilerServices.Unsafe)
+      FluentAssertions.csproj depends on SRCU 4.5.0 (AssemblyVersion 4.0.4.0)
+      System.Collections.Immutable 1.6.0+ depends on SRCU 4.5.2+ (AssemblyVersion 4.0.4.1)
+      FluentAssertions.Specs.csproj for net47 does not generate a binding redirect from 4.0.4.0 to 4.0.4.1
+      Upgrading to 1.6.0 gives "Could not load file or assembly 'System.Runtime.CompilerServices.Unsafe, Version=4.0.4.0"
+    -->
```
jnyrup, line 19:
```suggestion
```

jnyrup (approved):

IT-VBFK:
https://github.com/fluentassertions/fluentassertions/blob/67fa57085562d0fa290b427a88618f86ea5ba1dc/Build/Build.cs#L162C5-L168C7

```diff
    Project[] Projects => new[]
    {
        Solution.Specs.FluentAssertions_Specs,
        Solution.Specs.FluentAssertions_Equivalency_Specs,
+       Solution.Specs.FluentAssertions_Extensibility_Specs,
        Solution.Specs.FSharp_Specs,
        Solution.Specs.VB_Specs
    };
```

To actually call the tests..

dennisdoomen:
> To actually call the tests..

Good catch. Thanks.

Src/FluentAssertions/Common/Services.cs
```diff
@@ -33,14 +39,76 @@ public static Configuration Configuration
...
+            {
+                // Just ignore any exceptions that might happen while trying to find the attributes
+            }
+        }
+    }
+
+    private static bool IsFramework(Assembly assembly)
+    {
+        return assembly?.FullName?.StartsWith("Microsoft", StringComparison.OrdinalIgnoreCase) == true ||
+            assembly?.FullName?.StartsWith("System", StringComparison.OrdinalIgnoreCase) == true;
```
jnyrup, line 112: We might want to be more strict here to not allow `SystemFoo` or `MicrosoftBar`.
Both #1193 and #1197 fixed similar cases where our comparison was too loose.

jnyrup, line 112: I'm not sure how Qodana concludes that `FullName` cannot be null, but on the other hand I cannot find a realistic case where it can be null.

I guess one would have to really force it to be null.
```cs
class MyAssembly : Assembly
{
    public override string? FullName => null;
}
```

https://stackoverflow.com/questions/70569074/under-what-conditions-can-assembly-getname-name-be-null
suggests that having an assembly without a name is violating the ECMA spec.

Grepping the BCL for `'assembly.*\.FullName\!` gives 17 results.

I'm fine with either of keeping or removing the null checks.

Src/FluentAssertions/Common/Services.cs
```diff
@@ -33,14 +39,76 @@ public static Configuration Configuration
...
+            catch
+            {
+                // Just ignore any exceptions that might happen while trying to find the attributes
+            }
+        }
+    }
+
+    private static bool IsFramework(Assembly assembly)
+    {
+        return assembly?.FullName?.StartsWith("Microsoft", StringComparison.OrdinalIgnoreCase) == true ||
```
jnyrup, line 111: FYI: `assembly` cannot be null at this place as we have already called `assembly.IsDynamic` before entering this method.

IT-VBFK:
What about adding a bogous extension project, to test the failures?

IT-VBFK:
To cover the remaining four missed lines 😉

jnyrup:
> What about adding a bogous extension project, to test the failures?

Since `[AssertionEngineInitializer]` can be used multiple times, we can add a bogus initializer next to `InitializeBeforeFirstAssertion` and add an extra `[assembly:AssertionEngineInitializer` at the top.

This will exercise the second `catch`.

dennisdoomen:
I'll do that in a follow-up PR

PR #2296, opened 2023-09-10
Refactor `StringValidator` to avoid inheritance

vbreuss:
Refactor the `StringValidator` to favor composition over inheritance (see [this comment](https://github.com/fluentassertions/fluentassertions/issues/2050#issuecomment-1416710889)):
Instead of subclassing the abstract `StringValidator` class, this class now accepts an interface `IStringMismatchValidator` which provides the necessary implementation details.
If this PR is merged, it should simplify implementing #2050. 

## IMPORTANT 


Src/FluentAssertions/Primitives/IStringMismatchValidator.cs
```diff
@@ -0,0 +1,10 @@
+using FluentAssertions.Execution;
+
+namespace FluentAssertions.Primitives;
+
+internal interface IStringMismatchValidator
```
dennisdoomen, line 5: 🔧 This interface needs some documentation
🤔 What about `IStringComparisonStrategy` to emphase we're using a Strategy Pattern here?

dennisdoomen (changes requested):

Src/FluentAssertions/Primitives/IStringMismatchValidator.cs
dennisdoomen: > This interface needs some documentation

I meant both the interface _and_ its properties ;-)

vbreuss: Sorry, it seems I was too quick :-)

Src/FluentAssertions/Primitives/NegatedStringStartStrategy.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    public void ValidateAgainstMismatch(IAssertionScope assertion, string subject, string expected)
+    {
+        bool isMatch = subject.StartsWith(expected, stringComparison);
+
+        if (isMatch)
+        {
+            assertion.FailWith(ExpectationDescription + "{0}{reason}, but found {1}.",
+                expected, subject);
+        }
```
IT-VBFK, line 35: Why not just use the internal FA-API?

like

```csharp
assertion
    .ForCondition(subject.StartsWith(expected, stringComparison))
    .FailWith(ExpectationDescription + "{0}{reason}, but found {1}.",
                 expected, subject);
```

vbreuss: No reason, I just didn't refactor the existing code :-)
image

dennisdoomen: And that code pre-dates the `ForCondition` construct ;-)

Src/FluentAssertions/Primitives/StringWildcardMatchingStrategy.cs
```diff
@@ -2,39 +2,39 @@
...
 
-        if (isMatch && Negate)
+        if (Negate)
         {
-            Assertion.FailWith(ExpectationDescription + "but {1} matches.", Expected, Subject);
+            assertion.FailWith(ExpectationDescription + "but {1} matches.", expected, subject);
+        }
+        else
+        {
+            assertion.FailWith(ExpectationDescription + "but {1} does not.", expected, subject);
```
IT-VBFK, line 26: Internal FA-API:
Applies to other spots as well ;)

vbreuss: You are right. Originally I tried to change as little as possible, but will refactor more thoroughly now :-)

IT-VBFK:
Something to rethink either: I think it is odd to have a `StringStartStrategy` and a seperate `NegatedStringStartStrategy` but do not have this pattern in the other `...Strategy`s.

What about aligning them either use into `Xxx` and `NegatedXxx` files or handle it like the `StringWildcardStrategy`?

vbreuss:
> Something to rethink either: I think it is odd to have a `StringStartStrategy` and a seperate `NegatedStringStartStrategy` but do not have this pattern in the other `...Strategy`s.
> 
> What about aligning them either use into `Xxx` and `NegatedXxx` files or handle it like the `StringWildcardStrategy`?
> 
> If we agree to the latter: Should then `Negated` and `IgnoreCase` go into the contract `IStringComparisonStrategy`?

The `StringStartStrategy` and `NegatedStringStartStrategy` are implemented very differently and differ in the explicit failure messages for different scenarios.
I don't think it a good idea to also change the behavior as part of a "refactor"-PR and would rather adapt this in a separate pull request.

dennisdoomen:
I believe that enough work is done and there's no need to increase the scope.  Moreover, the involved classes are all `internal` and thus easy to change later on

dennisdoomen (approved):

Src/FluentAssertions/Primitives/StringWildcardMatchingStrategy.cs
```diff
@@ -2,39 +2,32 @@
...
-        if (isMatch && Negate)
-        {
-            Assertion.FailWith(ExpectationDescription + "but {1} matches.", Expected, Subject);
-        }
+        assertion
+            .ForCondition(!Negate || !isMatch)
+            .FailWith(ExpectationDescription + "but {1} matches.", expected, subject)
+            .Then
+            .ForCondition(Negate || isMatch)
+            .FailWith(ExpectationDescription + "but {1} does not.", expected, subject);
```
jnyrup, line 20: Let's keep this method as it was.
I haven't dug into the history, but I suspect it was constructed as it was for performance reasons.

Src/FluentAssertions/Primitives/StringStartStrategy.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        assertion
+            .ForCondition(subject.Length >= expected.Length)
+            .FailWith(ExpectationDescription + "{0}{reason}, but {1} is too short.", expected, subject)
+            .Then
+            .ForCondition(subject.StartsWith(expected, stringComparison))
+            .FailWith(ExpectationDescription + "{0}{reason}, but {1} differs near " +
+                subject.IndexedSegmentAt(subject.IndexOfFirstMismatch(expected, stringComparison)) +
+                ".",
+                expected, subject);
```
jnyrup, line 38: This will now also invoke `IndexedSegmentAt` and `IndexOfFirstMismatch` on success.

jnyrup (commented):
That's some nice untangling of this unnecessary class hierarchy 🧹

Src/FluentAssertions/Primitives/StringStartStrategy.cs
vbreuss: Changed in [b34b47b](https://github.com/fluentassertions/fluentassertions/pull/2296/commits/b34b47b567e800470868e19b80901b4ee013f0c6).

Src/FluentAssertions/Primitives/StringWildcardMatchingStrategy.cs
vbreuss: Changed in [b34b47b](https://github.com/fluentassertions/fluentassertions/pull/2296/commits/b34b47b567e800470868e19b80901b4ee013f0c6)

jnyrup (approved):

Issue #2245, opened 2023-08-01
Consider adding .NET project file analyzers


Corniel:
To improve the quality of the project files, consider adding these Roslyn analyzers:
https://dotnet-project-file-analyzers.github.io/

dennisdoomen:
Have you tried it? What does it report? It'll help us to see what value it brings.

dennisdoomen:
As discussed, we don't feel we need those analyzers.

PR #2305, opened 2023-09-17
Adding .NET project file analyzers 

Corniel:
Enabled the [.NET project file analyzers](https://dotnet-project-file-analyzers.github.io/), to test #2245.

Unfortunately, the `Directory.Build.props` mechanism is not supported (yet). This will take some time.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,8 +1,13 @@
 <Project>
+  
   <PropertyGroup>
     <LangVersion>11.0</LangVersion>
     <CheckEolTargetFramework>false</CheckEolTargetFramework>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+  </PropertyGroup>
+  
+  <PropertyGroup>
+	<IsPackable>false</IsPackable>
   </PropertyGroup>
 
   <!-- To reduce build times, we only enable analyzers for the newest TFM -->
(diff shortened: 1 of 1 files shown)


dennisdoomen:
I've checked the messages, but I don't see the rationale for them yet. Why would I add something to the project file if it has a suitable default?

Corniel:
> I've checked the messages, but I don't see the rationale for them yet. Why would I add something to the project file if it has a suitable default?

For some (like `<OutputType>` the defaults are different based on the SDK type. Others, like `<IsPackable>` and `<IsPublishable>` have defaults that are not per say preferable. And that is of course the nice thing about rules: they can be disabled if you disagree.

dennisdoomen:
Let me rephrase. I didn't see any message that I thought we should be changing. But maybe it's just me and @jnyrup has a different opinion.

Corniel:
@dennisdoomen As mentioned, the version you saw did not take into account wat is defined in the `Directory.Build.props`, so it contained some FP's.

dennisdoomen:
And it does now?

And in your opinion, what suggestion that the analyzer gave would be relevant for us?

jnyrup:
> Let me rephrase. I didn't see any message that I thought we should be changing. But maybe it's just me and @jnyrup has a different opinion.

I agree.

Corniel:
What I would at least go for:
* [Proj0005: Define package reference assets as attributes](https://dotnet-project-file-analyzers.github.io/rules/Proj0005.html)
* [Proj0015: [Order package references alphabetically](https://dotnet-project-file-analyzers.github.io/rules/Proj0015.html)
* [Proj0209: Define the project readme file explicitly](https://dotnet-project-file-analyzers.github.io/rules/Proj0209.html)

There are more warnings/suggestions in the 0200 range (related to packaging) that might be beneficial. They should help with creating NuGet packages according to MS' guidelines.

jnyrup:
I agree with Proj0015 👍 
I wouldn't blindly apply Proj0005, i.e. I don't care if one uses XML attributes or XML elements, but I don't like long lines.
I don't find our readme.md suitable for display on nuget.org so I would ignore that.

Directory.Build.props
```diff
@@ -1,42 +1,41 @@
...
-      <PrivateAssets>all</PrivateAssets>
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-    </PackageReference>
-    <PackageReference Include="Roslynator.Analyzers" Version="4.3.0">
-      <PrivateAssets>all</PrivateAssets>
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-    </PackageReference>
-    <PackageReference Include="Meziantou.Analyzer" Version="2.0.70">
-      <PrivateAssets>all</PrivateAssets>
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
```
jnyrup, line 39: Is this the default behavior or is this disabling transitive inclusion which should be fine since we include the packages in all projects anyways.

dennisdoomen: This happens automatically with these development-only packages.

Corniel: @jnyrup Is is about including a README that is useful for NuGet, not including the root README.

jnyrup: I'm all open to augmenting https://github.com/fluentassertions/fluentassertions/blob/269ac689e7ef414050d3b37c8ed365cf97f836c3/Src/FluentAssertions/FluentAssertions.csproj#L19-L25
with another readme.md for display on nuget.org

jnyrup: > This happens automatically with these development-only packages.

_What_ happens automatically?

jnyrup: Hadn't read up on the analyzers listed in https://github.com/dotnet-project-file-analyzers/dotnet-project-file-analyzers/issues/26, but happy to know we're following the essential ones.

There's nothing in here I'm eager to changing right now.

dennisdoomen: Me neither

PR #2311, opened 2023-09-20
Fix various Qodana issues and update the threshold

dennisdoomen:
* Removed interface that we forgot while removing DataSet support
* Disabled Qodana rule that complains about similar code in tests.
* Fixed a namespace issue reported by Qodana
* Set the threshold at the current level so that builds succeed as long as it doesn't get worse.


jnyrup (approved):
I also noticed `ICollectionWrapper` the other day but hadn't got around to look at it.
I guess we need to add it to https://github.com/fluentassertions/fluentassertions.datasets

PR #2313, opened 2023-09-20
Run the Qodana on Windows so .NET 4.7 scanning works.

dennisdoomen:
According to JetBrains, in order for Qodana to properly detect .NET 4.7 usage, we need to run the build on Windows.

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   qodana-scan:
     name: Qodana Scan
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jnyrup (approved):

IT-VBFK:
image


I think we cannot get around using linux for now:
https://youtrack.jetbrains.com/issue/QD-4758/Azure-Devops-QodanaScan-fails-on-windows-2019-hosted-agent#focus=Comments-27-6752589.0-0

dennisdoomen:
We've fixed this with the help of JetBrains. Should be fine now.

PR #2318, opened 2023-09-24
Pass down the reason to inner `AssertionScope`

### LINKED ISSUE #2293: Equivalency assertion for dictionary does not include because message
### Description

When comparing the values of dictionaries for equivalency, the `because` message is not included in the failure message.

### Reproduction Steps

```csharp
var subject = new Dictionary<string, object>
{
    ["a"] = new List<int>()
};

var expected = new Dictionary<string, object>
{
    ["a"] = new List<int> { 42 }
};

subject.Should().BeEquivalentTo(expected, "FOO");
```

### Expected behavior

Exception message should include `FOO`

### Actual behavior

```
FluentAssertions.Execution.AssertionFailedException: 'Expected subject[a] to be a collection with 1 item(s), but found an empty collection.

With configuration:
- Compare enums by value
'
```

### Regression?

no

### Known Workarounds

_No response_

### Configuration

Fluent Assertions 6.12
.NET 8.0

### Other information

_No response_

### Could you help with a pull-request?

Yes

IT-VBFK:
This fixes #2293 

## IMPORTANT 

### DIFF
--- a/Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
@@ -1184,8 +1184,8 @@ public class DictionarySpecs
             ["a"] = new List<int> { 42 }
         };
 
-        Action act = () => subject.Should().BeEquivalentTo(expected, "FOO");
+        Action act = () => subject.Should().BeEquivalentTo(expected, "FOO {0}", "BAR");
 
-        act.Should().Throw<XunitException>().WithMessage("*FOO*");
+        act.Should().Throw<XunitException>().WithMessage("*FOO BAR*");
     }
 }
--- a/Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
@@ -1170,4 +1170,22 @@ public class DictionarySpecs
             .WhenTypeIs<double>()
         );
     }
+
+    [Fact]
+    public void Passing_the_reason_to_the_inner_equivalency_assertion_works()
+    {
+        var subject = new Dictionary<string, object>
+        {
+            ["a"] = new List<int>()
+        };
+
+        var expected = new Dictionary<string, object>
+        {
+            ["a"] = new List<int> { 42 }
+        };
+
+        Action act = () => subject.Should().BeEquivalentTo(expected, "FOO");
+
+        act.Should().Throw<XunitException>().WithMessage("*FOO*");
+    }
 }
--- a/Src/FluentAssertions/Execution/AssertionScope.cs
+++ b/Src/FluentAssertions/Execution/AssertionScope.cs
@@ -148,10 +148,7 @@ public sealed class AssertionScope : IAssertionScope
     /// </summary>
     public FormattingOptions FormattingOptions { get; } = AssertionOptions.FormattingOptions.Clone();
 
-    internal bool Succeeded
-    {
-        get => succeeded == true;
-    }
+    internal bool Succeeded => succeeded == true;
 
     /// <summary>
     /// Adds an explanation of why the assertion is supposed to succeed to the scope.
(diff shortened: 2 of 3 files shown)


IT-VBFK:
Qodana thinks we aren't in pr_mode? 🤔

jnyrup:
> Qodana thinks we aren't in pr_mode? 🤔

We fiddled with the [Qodana pipeline](https://github.com/fluentassertions/fluentassertions/blob/develop/.github/workflows/code_quality.yml) today. Apparently we didn't nail it 😞 
I have pushed a change to `develop` 🤞 
Try rebasing on `develop`.

jnyrup:
Try rebasing one more time.

image

IT-VBFK:
Can I close #2319 then?

jnyrup:
> Can I close #2319 then?

Yes.

Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
```diff
@@ -34,6 +34,8 @@ public EnumerableEquivalencyValidator(IEquivalencyValidator parent, IEquivalency
 
     public void Execute<T>(object[] subject, T[] expectation)
     {
+        AssertionScope.Current.BecauseOf(context.Reason);
```
dennisdoomen, line 37: 🤔 I think you need to solve this problem at a different place. When a new nested `AssertionScope` is created, such as [here](https://github.com/fluentassertions/fluentassertions/blob/a832de3ed8987188d13cf4ae33bf4413e6137f9a/Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs#L157), it should automatically pick up the _reason_ from the parent. This should be fixed [here](https://github.com/fluentassertions/fluentassertions/blob/de9fa4d63ffad9444657558ae690ccbcc789dfc2/Src/FluentAssertions/Execution/AssertionScope.cs#L106).

dennisdoomen (changes requested):

Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
```diff
@@ -1170,4 +1170,22 @@ public void When_a_custom_rule_is_applied_on_a_dictionary_it_should_apply_it_on_
...
+        };
+
+        var expected = new Dictionary<string, object>
+        {
+            ["a"] = new List<int> { 42 }
+        };
+
+        Action act = () => subject.Should().BeEquivalentTo(expected, "FOO");
+
+        act.Should().Throw<XunitException>().WithMessage("*FOO*");
```
jnyrup, line 1189: To ensure we pass both `because` and `becauseArgs` all the way.
```suggestion
        Action act = () => subject.Should().BeEquivalentTo(expected, "FOO {1}", "BAR");

        act.Should().Throw<XunitException>().WithMessage("*FOO BAR*");
```

Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
IT-VBFK: Thank you at pointing me at the correct location ♥️

Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
IT-VBFK: I think this should be `{0}`, isn't it? 😉

jnyrup: 🤐

jnyrup:
Something's up with our Qodana pipeline.
It seems to be scanning the wrong commit.

@dennisdoomen 
Should this line https://github.com/fluentassertions/fluentassertions/blob/73ff4e012398a64d30244ff8993199706c6a7cb7/.github/workflows/code_quality.yml#L18 also use `pull_request_target` instead of `pull_request`?

dennisdoomen:
> also use `pull_request_target` instead of `pull_request`?

No, there's no such thing. Instead, I think we should use 

`ref: "${{ github.event.pull_request.merge_commit_sha }}"`

See also https://github.com/actions/checkout/issues/518#issuecomment-890401887

IT-VBFK:
> Something's up with our Qodana pipeline.
> It seems to be scanning the wrong commit.

Rebased on current `develop`... let's see whether this helps..

PR #2349, opened 2023-09-29
Checkout the merge commit for Qodana to generate correct results

ITaluone:
## IMPORTANT 

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ github.event.pull_request.head.sha }}  # to check out the actual pull request commit, not the merge commit
+          ref: ${{ github.event.pull_request.merge_commit_sha }}
           fetch-depth: 0
       - name: 'Qodana Scan'
         uses: JetBrains/qodana-action@v2023.2


ITaluone:
Using `fetch_depth=0` is also unnecessary for Qodana, right?

dennisdoomen:
> Using `fetch_depth=0` is also unnecessary for Qodana, right?

According to the [docs](https://www.jetbrains.com/help/qodana/github.html#Usage), we do...

IT-VBFK:
Interestingly it uses `fetch_depth=0` in every case (see the build logs), even if it is set to 1

jnyrup:
> Interestingly it uses `fetch_depth=0` in every case (see the build logs), even if it is set to 1

I don't see it, can you point at it?


> > Using `fetch_depth=0` is also unnecessary for Qodana, right?
> 
> According to the [docs](https://www.jetbrains.com/help/qodana/github.html#Usage), we do...

As I read it `fetch-depth=0` is only strictly needed for pr-mode, so we could use `fetch-depth=1` for non-pr mode.
But I have no idea of what difference it would make, but I estimate it won't be that much since the checkout step with `fetch-depth=0` takes 11s 🤷

dennisdoomen:
> so we could use `fetch-depth=1` for non-pr mode.

I'm pretty sure it'll break GitVersion, since it needs the completely history.

IT-VBFK:
> I'm pretty sure it'll break GitVersion, since it needs the completely history.

As I understand this: The Qodana pipeline is a different one than the usual `build`. So this shouldn't matter, I believe.

dennisdoomen (approved):

jnyrup (approved):

IT-VBFK:
>  I don't see it, can you point at it?

Take this as an example:
https://github.com/fluentassertions/fluentassertions/actions/runs/6347891564/job/17243569462

This was a Qodana scan for this commit: https://github.com/fluentassertions/fluentassertions/commit/12df7d0b7672dbecbc94b3a3f95889073487a601
(with `fetch_depth=1`)

and the output shows:
image

PR #2354, opened 2023-10-01
Add a note to XML-docs that explains the `(Not)ContainEquivalentOf` problem

### LINKED ISSUE #2352: [Docs]: Improve documentation for `ContainEquivalentOf`
### Background and motivation

#1795, #2260, #2283 and #2328 all thought that `subjectList.ContainEquivalentOf(expectedList)` tries to find a subset in `subjectList` that is equivalent to `expectedList`.
What `ContainEquivalentOf` does is trying to find an element within `subjectList` that is equivalent to `expectedList`.

Is there some xml-documentation or website docs/examples that are missing, incorrect or could be improved to avoid these confusions?

### Alternative Concerns

_No response_

### Are you willing help with a pull-request?

No

IT-VBFK:
This ref. and hopefully closes #2352 

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -810,13 +810,18 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> : Referenc
     }
 
     /// <summary>
-    /// Asserts that a collection of objects contains at least one object equivalent to another object.
+    /// Asserts that at least one element in the collection is equivalent to <paramref name="expectation"/>.
     /// </summary>
     /// <remarks>
-    /// Objects within the collection are equivalent to the expected object when both object graphs have equally named properties with the same
-    /// value, irrespective of the type of those objects. Two properties are also equal if one type can be converted to another
-    /// and the result is equal.
+    /// <para>
+    /// <b>Important:</b> You cannot use this method to assert whether a subset of the collection is equivalent to the <paramref name="expectation"/>.
+    /// This usually means that the expectation is meant to be a <i>single</i> item.
+    /// </para>
+    /// <para>
+    /// By default, objects within the collection are seen as equivalent to the expected object when both object graphs have equally named properties with the same
+    /// value, irrespective of the type of those objects.
     /// Notice that actual behavior is determined by the global defaults managed by <see cref="AssertionOptions"/>.
+    /// </para>
     /// </remarks>
     /// <param name="expectation">The expected element.</param>
     /// <param name="because">
@@ -833,13 +838,18 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> : Referenc
     }
 
     /// <summary>
-    /// Asserts that a collection of objects contains at least one object equivalent to another object.
+    /// Asserts that at least one element in the collection is equivalent to <paramref name="expectation"/>.
     /// </summary>
     /// <remarks>
-    /// Objects within the collection are equivalent to the expected object when both object graphs have equally named properties with the same
-    /// value, irrespective of the type of those objects. Two properties are also equal if one type can be converted to another
-    /// and the result is equal.
+    /// <para>
+    /// <b>Important:</b> You cannot use this method to assert whether a subset of the collection is equivalent to the <paramref name="expectation"/>.
+    /// This usually means that the expectation is meant to be a <i>single</i> item.
+    /// </para>
+    /// <para>
+    /// By default, objects within the collection are seen as equivalent to the expected object when both object graphs have equally named properties with the same
+    /// value, irrespective of the type of those objects.
     /// Notice that actual behavior is determined by the global defaults managed by <see cref="AssertionOptions"/>.
+    /// </para>
     /// </remarks>
     /// <param name="expectation">The expected element.</param>
     /// <param name="config">
@@ -2283,13 +2293,18 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> : Referenc
     }
 
     /// <summary>
-    /// Asserts that a collection of objects does not contain any object equivalent to another object.
+    /// Asserts that no element in the collection is equivalent to <paramref name="unexpected"/>.
     /// </summary>
     /// <remarks>
-    /// Objects within the collection are equivalent to the expected object when both object graphs have equally named properties with the same
-    /// value, irrespective of the type of those objects. Two properties are also equal if one type can be converted to another
-    /// and the result is equal.
+    /// <para>
+    /// <b>Important:</b> You cannot use this method to assert whether a subset of the collection is not equivalent to the <paramref name="unexpected"/>.
+    /// This usually means that the expectation is meant to be a <i>single</i> item.
+    /// </para>
+    /// <para>
+    /// By default, objects within the collection are seen as not equivalent to the expected object when both object graphs have unequally named properties with the same
+    /// value, irrespective of the type of those objects.
     /// Notice that actual behavior is determined by the global defaults managed by <see cref="AssertionOptions"/>.
+    /// </para>
     /// </remarks>
     /// <param name="unexpected">The unexpected element.</param>
     /// <param name="because">
@@ -2306,13 +2321,18 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> : Referenc
     }
 
     /// <summary>
-    /// Asserts that a collection of objects does not contain any object equivalent to another object.
+    /// Asserts that no element in the collection is equivalent to <paramref name="unexpected"/>.
     /// </summary>
     /// <remarks>
-    /// Objects within the collection are equivalent to the expected object when both object graphs have equally named properties with the same
-    /// value, irrespective of the type of those objects. Two properties are also equal if one type can be converted to another
-    /// and the result is equal.
+    /// <para>
+    /// <b>Important:</b> You cannot use this method to assert whether a subset of the collection is not equivalent to the <paramref name="unexpected"/>.
+    /// This usually means that the expectation is meant to be a <i>single</i> item.
+    /// </para>
+    /// <para>
+    /// By default, objects within the collection are seen as not equivalent to the expected object when both object graphs have unequally named properties with the same
+    /// value, irrespective of the type of those objects.
     /// Notice that actual behavior is determined by the global defaults managed by <see cref="AssertionOptions"/>.
+    /// </para>
     /// </remarks>
     /// <param name="unexpected">The unexpected element.</param>
     /// <param name="config">


Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -813,6 +813,8 @@ public AndConstraint<TAssertions> Contain(IEnumerable<T> expected, string becaus
     /// Asserts that a collection of objects contains at least one object equivalent to another object.
     /// </summary>
     /// <remarks>
+    /// <b>Note:</b> You cannot use this method to assert whether a collection contains a subset of another collection.
+    /// Use: e.g. <see cref="BeSubsetOf"/> with the complete collection as expectation.
```
jnyrup, line 817: The remarks are not shown when scrolling in intellisense suggestions in VS.
So I think we should improve the summary instead.
And reading the summary now I can see why consumers can get confused by it.

image

dennisdoomen: That's a Visual Studio limitation. This is what I see in Rider:

image

jnyrup: I see that as well in VS when hovering over the method definition.

dennisdoomen: Here's an improved version:

```csharp
/// <summary>
    /// Asserts that at least one object in the collection is equivalent to <paramref name="expectation"/>.
    /// </summary>
    /// <remarks>
    /// <para>
    /// By default, objects within the collection are seen as equivalent to the expected object when both object graphs have equally named properties with the same
    /// value, irrespective of the type of those objects.
    /// Notice that actual behavior is determined by the global defaults managed by <see cref="AssertionOptions"/>.
    /// </para>
    /// <para>
    /// <b>Important:</b> You cannot use this method to assert whether a collection contains a subset of another collection.
    /// Use e.g. <see cref="BeSubsetOf"/> instead.
    /// </para>
    /// </remarks>
```

image

eNeRGy164: Can we use `<em>` instead of `<b>`?

jnyrup: > That's a Visual Studio limitation. This is what I see in Rider:

VS also renders remarks etc, when hovering the method definition.

image

IT-VBFK: > Can we use `<em>` instead of `<b>`?

Sure... but why?

eNeRGy164: Well, I guess it should be `<strong>`, but because you make it bold for a reason, so your probably want to convey a message. In this case the urgency that it is important?

https://www.w3docs.com/snippets/html/what-is-the-difference-between-b-and-strong-i-and-em-tags.html

IT-VBFK: Thanks for sharing this link 👍🏻

See clear now 🙃

jnyrup:
How does these two views look in Rider?
I imagine the vast majority of our users only look at the text presented when IntelliSense view.
image
image

ITaluone:
Rider:
image
image

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -813,10 +813,15 @@ public AndConstraint<TAssertions> Contain(IEnumerable<T> expected, string becaus
...
-    /// value, irrespective of the type of those objects. Two properties are also equal if one type can be converted to another
-    /// and the result is equal.
+    /// <para>
+    /// By default, objects within the collection are seen as equivalent to the expected object when both object graphs have equally named properties with the same
+    /// value, irrespective of the type of those objects.
     /// Notice that actual behavior is determined by the global defaults managed by <see cref="AssertionOptions"/>.
+    /// </para>
+    /// <para>
+    /// <strong>Important:</strong> You cannot use this method to assert whether a collection contains a subset of another collection.
+    /// Use e.g. <see cref="BeSubsetOf"/> instead.
```
jnyrup, line 823: I read "collection contains a subset of another collection" if `subject` and `expectation` has a non-empty intersection.

To ask if `subject` contains at least the items in `expectation` is to ask whether `subject` is a superset of `expectation`.
In FA that is `subject.Should().Contain(expectation)`.
But `Contain` compares items by equality and not equivalence.

jnyrup:
> Are there some tags that shows in VS but not in Rider?

Not to my knowledge.
But this is why I think we should improve the text in `<summary>` which will benefit all FA users, not matter if they use VS or Rider.

ITaluone:
What about this? I moved the "Important" `para` up to summary

```csharp
/// <summary>
/// Asserts that a collection of objects contains at least one object equivalent to another object.
/// <br />
/// <para>
/// <strong><b>Important:</b></strong> You cannot use this method to assert whether a collection contains a subset of another collection.
/// Use e.g. <see cref="BeSubsetOf"/> instead.
/// </para>
/// </summary>
/// <remarks>
/// <para>
/// By default, objects within the collection are seen as equivalent to the expected object when both object graphs have equally named properties with the same
/// value, irrespective of the type of those objects.
/// Notice that actual behavior is determined by the global defaults managed by <see cref="AssertionOptions"/>.
/// </para>
/// </remarks>
/// <param name="expectation">The expected element.</param>
/// <param name="because">
/// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
/// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
/// </param>
/// Zero or more objects to format using the placeholders in <paramref name="because"/>.
/// </param>
```

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
ITaluone: What do you mean? Should I change `BeSubsetOf` to `Contain`?

I previously wrote something like: Use `BeSubsetOf` instead using the complete collections as expectation.

dennisdoomen:
> ```cs
> /// Asserts that a collection of objects contains at least one object equivalent to another object.
> ```

I think what I had was bettter, since it explicitly referred to the parameter.

dennisdoomen:
Check out what it looks like in Rider (notice the `<string>` XML element)

image

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
IT-VBFK: Hmm I don't like how this looks in Rider.. I will stick with `<b>` only

IT-VBFK:
Yeah.. I had the same feelings. I will stick with `<b>` tags only

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -810,13 +810,18 @@ public AndConstraint<TAssertions> Contain(IEnumerable<T> expected, string becaus
     }
 
     /// <summary>
-    /// Asserts that a collection of objects contains at least one object equivalent to another object.
+    /// Asserts that at least one object in the collection is equivalent to <paramref name="expectation"/>.
+    /// <para>
```
dennisdoomen, line 814: `<para>` doesn't work in the summary (and that is expected). Two `<br/>`s will work, but I think moving it to the remarks section is better.

image

dennisdoomen (changes requested):

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
IT-VBFK: What does @jnyrup think? 

As I recall, `<remarks>` were not shown in VisualStudio.

jnyrup: That's a ~Visual Studio~ Rider limitation. This is what I see in ~Rider~ Visual Studio: 😉 
image

> The `<para>` tag is for use inside a tag, such as `<summary>`, `<remarks>`, or `<returns>` [...]

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#format-documentation-output

On a more serious note, the change of the summary to `Asserts that at least one object in the collection is equivalent to <paramref name="expectation"/>` is enough for me and improves the current situation for both Rider and VS users.
I would only change it from using "object" to "element".

I'm fine having the "you cannot..." part in `<remarks>` - that also seems to be the semantically more correct place to put it.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
jnyrup: What people seems to be thinking `subject.Should().ContainEquivalentOf(expectation)` does is checking if `expectation` is a subset of `subject`, where elements are compared by equivalency.
image
(I conclude this also by a first-hand experience with a colleague)

This is how I read "a collection contains a subset of another collection"
image


`subject.Should().BeSubsetOf(expectation)` is the opposite of what people seems to seek.
image


Currently FA does not contain a method to assert that an expectation of elements is equivalent to a subset of the subject.
`subject.Should().Contain(expectation)` is the closest we offer, but this compares elements by equality and not equivalency.


I think either of these would be a bit more precise.
(And I'm fine with either)
```cs
/// <b>Important:</b> You cannot use this method to assert whether a subset of the collection is equivalent to the <paramref name="expectation"/>.
/// <b>Important:</b> You cannot use this method to assert whether the <paramref name="expectation"/> is equivalent to a subset of the collection.
```

Btw, I really like your note about a "single item" 👍 
It's much more applicable to most programmers than some formal definition about a [surjective function](https://en.wikipedia.org/wiki/Surjective_function) between the subject and the expectation.

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
dennisdoomen: Ah, if it works in VS, then keep it there. I'll report an issue for Rider.

jnyrup: To be fair, this is what VS shows in the IntelliSense, but I guess (to their favor) that's a conscious choice to clutter the UI less.

image

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -2283,13 +2293,18 @@ public AndConstraint<TAssertions> NotContain(IEnumerable<T> unexpected, string b
     }
 
     /// <summary>
-    /// Asserts that a collection of objects does not contain any object equivalent to another object.
+    /// Asserts that at least one element in the collection is equivalent to <paramref name="unexpected"/>.
```
jnyrup, line 2296: This should be negated. E.g. "Asserts that no element in the collection..."

jnyrup (approved):

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -2283,13 +2293,18 @@ public AndConstraint<TAssertions> NotContain(IEnumerable<T> unexpected, string b
     }
 
     /// <summary>
-    /// Asserts that a collection of objects does not contain any object equivalent to another object.
+    /// Asserts that at least no element in the collection is equivalent to <paramref name="unexpected"/>.
```
jnyrup, line 2296: Remove "at least"

IT-VBFK: ups :)

jnyrup (approved):

dennisdoomen (approved):

Issue #2308, opened 2023-09-18
`BeEquivalentTo` cannot compare generic and non-generic dictionary


jnyrup:
### Description

In #2298 we discovered a regression with 6.0.0 when comparing a non-generic and a generic dictionary.

### Reproduction Steps

These works
```csharp
[Fact]
public void NonGeneric_NonGeneric()
{
    var subject = new ListDictionary
    {
        ["id"] = 22,
        ["CustomerId"] = 33
    };

    var expected = new ListDictionary
    {
        ["id"] = 22,
        ["CustomerId"] = 33
    };

    subject.Should().BeEquivalentTo(expected);
}

[Fact]
public void NonGeneric_Generic_cast_to_non_generic()
{
    var subject = new ListDictionary
    {
        ["id"] = 22,
        ["CustomerId"] = 33
    };

    var expected = (IDictionary)new Dictionary<object, object>
    {
        ["id"] = 22,
        ["CustomerId"] = 33
    };

    subject.Should().BeEquivalentTo(expected);
}

[Fact]
public void Generic_NonGeneric()
{
    var subject = new Dictionary<object, object>
    {
        ["id"] = 22,
        ["CustomerId"] = 33
    };

    var expected = new ListDictionary
    {
        ["id"] = 22,
        ["CustomerId"] = 33
    };

    subject.Should().BeEquivalentTo(expected);
}
```

This one fails:
```cs

[Fact]
public void NonGeneric_Generic()
{
    var subject = new ListDictionary
    {
        ["id"] = 22,
        ["CustomerId"] = 33
    };

    var expected = new Dictionary<object, object>
    {
        ["id"] = 22,
        ["CustomerId"] = 33
    };

    subject.Should().BeEquivalentTo(expected);
}
```

### Expected behavior

All tests should pass

### Actual behavior

`NonGeneric_Generic` fails.

### Regression?

Yes.
All examples above worked in 5.10.3 and with 6.0.0 `NonGeneric_Generic` started failing.

### Known Workarounds

Cast `Dictionary<,>` to `IDictionary` or use a non-generic dictionary type like `HashTable`

### Configuration

FA 6.12.0
.NET 8

### Are you willing to help with a pull-request?

Yes, please assign this issue to me.

ITaluone:
As a side note: this is why the error message looks strange: `It implements .`, because `actualDictionary is null`.

```csharp
if (actualDictionary is null)
{
    AssertionScope.Current.FailWith(
[...]

PR #2358, opened 2023-10-03
Handle comparing an `IDictionary` subject with an `IDictionary<,>` expectation

ITaluone:
Fixes: #2308 

~~**Note**: The very first two commits are to fix the actual problem. Commit 3-6 are for code quality (and therefore can moved to a separate PR if needed).~~

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
@@ -1,5 +1,7 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
 using FluentAssertions.Execution;
 
@@ -17,32 +19,53 @@ public class GenericDictionaryEquivalencyStep : IEquivalencyStep
     public EquivalencyResult Handle(Comparands comparands, IEquivalencyValidationContext context,
         IEquivalencyValidator nestedValidator)
     {
-        if (comparands.Expectation != null)
+        if (comparands.Expectation is null)
         {
-            Type expectationType = comparands.GetExpectedType(context.Options);
-            if (DictionaryInterfaceInfo.FindFrom(expectationType, "expectation") is { } expectedDictionary)
-            {
-                if (AssertSubjectIsNotNull(comparands.Subject)
-                    && EnsureSubjectIsDictionary(comparands, expectedDictionary) is { } actualDictionary)
-                {
-                    AssertDictionaryEquivalence(comparands, context, nestedValidator, actualDictionary, expectedDictionary);
-                }
-
-                return EquivalencyResult.AssertionCompleted;
-            }
+            return EquivalencyResult.ContinueWithNext;
         }
 
-        return EquivalencyResult.ContinueWithNext;
+        Type expectationType = comparands.GetExpectedType(context.Options);
+
+        if (DictionaryInterfaceInfo.FindFrom(expectationType, "expectation") is not { } expectedDictionary)
+        {
+            return EquivalencyResult.ContinueWithNext;
+        }
+
+        if (IsNonGenericDictionary(comparands.Subject))
+        {
+            // Because we handle non-generic dictionaries later
+            return EquivalencyResult.ContinueWithNext;
+        }
+
+        if (IsNotNull(comparands.Subject)
+            && EnsureSubjectIsOfTheExpectedDictionaryType(comparands, expectedDictionary) is { } actualDictionary)
+        {
+            AssertDictionaryEquivalence(comparands, context, nestedValidator, actualDictionary, expectedDictionary);
+        }
+
+        return EquivalencyResult.AssertionCompleted;
     }
 
-    private static bool AssertSubjectIsNotNull(object subject)
+    private static bool IsNonGenericDictionary(object subject)
+    {
+        if (subject is not IDictionary)
+        {
+            return false;
+        }
+
+        return !subject.GetType().GetInterfaces()
+            .Any(@interface => @interface.IsGenericType
+                && @interface.GetGenericTypeDefinition() == typeof(IDictionary<,>));
+    }
+
+    private static bool IsNotNull(object subject)
     {
         return AssertionScope.Current
             .ForCondition(subject is not null)
             .FailWith("Expected {context:Subject} not to be {0}{reason}.", new object[] { null });
     }
 
-    private static DictionaryInterfaceInfo EnsureSubjectIsDictionary(Comparands comparands,
+    private static DictionaryInterfaceInfo EnsureSubjectIsOfTheExpectedDictionaryType(Comparands comparands,
         DictionaryInterfaceInfo expectedDictionary)
     {
         var actualDictionary = DictionaryInterfaceInfo.FindFromWithKey(comparands.Subject.GetType(), "subject",
@@ -57,8 +80,8 @@ public class GenericDictionaryEquivalencyStep : IEquivalencyStep
         if (actualDictionary is null)
         {
             AssertionScope.Current.FailWith(
-                $"Expected {{context:subject}} to be a dictionary or collection of key-value pairs that is keyed to type {expectedDictionary.Key}. " +
-                $"It implements {actualDictionary}.");
+                "Expected {context:subject} to be a dictionary or collection of key-value pairs that is keyed to " +
+                $"type {expectedDictionary.Key}.");
         }
 
         return actualDictionary;
--- a/Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Collections.Specialized;
 using System.Globalization;
 using System.Linq;
 using FluentAssertions.Equivalency.Tracing;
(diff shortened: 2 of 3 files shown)


ITaluone:
Out of curiosity:

Should this test fail or pass?

```csharp
[Fact]
public void A_enumerable_of_key_value_pairs_and_generic_dict()
{
    var subject = new List<KeyValuePair<string, object>>
    {
        new("id", 22),
        new("CustomerId", 33)
    };

    var expected = new Dictionary<object, object>
    {
        ["id"] = 22,
        ["CustomerId"] = 33
    };

    subject.Should().BeEquivalentTo(expected);
}
```

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
```diff
@@ -35,14 +43,25 @@ public class GenericDictionaryEquivalencyStep : IEquivalencyStep
         return EquivalencyResult.ContinueWithNext;
     }
 
+    private static bool SubjectIsNonGenericDictionary(object subject)
```
ITaluone, line 46: Here I followed the naming like `AssertSubjectIsNotNull` below, but I think this is a bit redundant.

So I would probably rename this to `IsNonGenericDictionary()`.

And I propose to change the method name below also to something like: `AssertIsNotNull()` or simply `IsNotNull()`.

(Or alternatively pass in the comparands object rather than `comparands.Subject`  🤔 )

What do you think?

dennisdoomen:
> The very first two commits are to fix the actual problem. Commit 3-6 are for code quality (and therefore can moved to a separate PR if needed).

It's totally fine to combine the fix and the refactorings in a single PR. But that requires you to ensure that every commit has a clear purpose so we can rebase-merge them.

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
dennisdoomen: > So I would probably rename this to IsNonGenericDictionary().

Yeah, that is a better name

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
```diff
@@ -20,10 +21,17 @@ public class GenericDictionaryEquivalencyStep : IEquivalencyStep
...
             {
+                if (SubjectIsNonGenericDictionary(comparands.Subject))
+                {
+                    // Because we handle non-generic dictionaries later
+                    return EquivalencyResult.ContinueWithNext;
+                }
+
                 if (AssertSubjectIsNotNull(comparands.Subject)
-                    && EnsureSubjectIsDictionary(comparands, expectedDictionary) is { } actualDictionary)
+                    && EnsureSubjectIsGenericDictionary(comparands, expectedDictionary) is { } actualDictionary)
```
dennisdoomen: 🤔 I think renaming this to `EnsureSubjectIsOfTheExpectedDictionaryType` makes it clearer why there is also a `SubjectIsNonGenericDictionary`

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
```diff
@@ -20,10 +21,17 @@ public class GenericDictionaryEquivalencyStep : IEquivalencyStep
         if (comparands.Expectation != null)
         {
             Type expectationType = comparands.GetExpectedType(context.Options);
+
             if (DictionaryInterfaceInfo.FindFrom(expectationType, "expectation") is { } expectedDictionary)
             {
+                if (SubjectIsNonGenericDictionary(comparands.Subject))
```
dennisdoomen: 🤔 Can't you move this before everything else as an early exit?

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
```diff
@@ -76,8 +76,9 @@ private static bool AssertSubjectIsNotNull(object subject)
         if (actualDictionary is null)
         {
             AssertionScope.Current.FailWith(
-                $"Expected {{context:subject}} to be a dictionary or collection of key-value pairs that is keyed to type {expectedDictionary.Key}. " +
-                $"It implements {actualDictionary}.");
+                "Expected {context:subject} to be a dictionary or collection of key-value pairs that is keyed to " +
+                $"type {expectedDictionary.Key}. " +
+                "It implements <null>.");
```
dennisdoomen, line 83: 🔧 The commit message is wrong. This code will never be executed if the subject is non-generic.

dennisdoomen (changes requested):

IT-VBFK:
> But that requires you to ensure that every commit has a clear purpose so that we can review them one by one and we can rebase-merge them. But in your case, it looks like some commits are rework on earlier commits. Same for the "release notes" commit.

How can a refactoring commit be a "non-rework" of previous commits? 🤔

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
ITaluone: No, unfortunately not, because `DictinaryInterfaceInfo.FindFrom` guard for dictionaries which implements multiple `IDictionary` and if these are non-generic we will move to the `DictionaryEquivalencyStep` immediately and therefore we do not fail in such a case.

ITaluone: Of course we could combine the `if` with `||` but why? 😄

ITaluone:
@dennisdoomen I have combined all commits to one, because I had no clue, how I could add refactoring commits, but not interfere (or non-rework) with changes in previous commits.

Sorry :/ 

But I am happy if you could point me into the correct direction (for PRs in future)

dennisdoomen:
Using fixup commits and interactive squashing. See https://www.continuousimprover.com/2020/03/keep-source-control-history-clean.html

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
```diff
@@ -57,8 +78,9 @@ private static bool AssertSubjectIsNotNull(object subject)
...
```
jnyrup, line 83: How does something implement null? :upside_down_face: 
Let's either remove this line or e.g. print the all `IDictionary<,>` interfaces implemented by `subject` like it did before #1603.
I'm fine with either.

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
```diff
@@ -17,32 +18,52 @@ public class GenericDictionaryEquivalencyStep : IEquivalencyStep
...
+
+        return EquivalencyResult.AssertionCompleted;
+    }
+
+    private static bool IsNonGenericDictionary(object subject)
+    {
+        if (subject is null)
+        {
+            return false;
         }
```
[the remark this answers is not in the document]
ITaluone: True :)

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
ITaluone: I will remove it then.

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
```diff
@@ -17,32 +18,52 @@ public class GenericDictionaryEquivalencyStep : IEquivalencyStep
...
 
-        return EquivalencyResult.ContinueWithNext;
+        return subject is IDictionary &&
+            !subject.GetType().IsGenericType;
```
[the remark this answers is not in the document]
ITaluone: Hmm.. returns true.. 

But until now it does the same: it fails only if the expected implements multiple dictionaries.

Should I add a check here?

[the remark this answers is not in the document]
ITaluone: And: what should be the desired behavior?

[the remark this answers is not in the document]
ITaluone: The check for `IsNonGenericDictionary` is now more precise and also takes custom dictionary implementations into account.

Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
```diff
@@ -1188,4 +1189,170 @@ public void Passing_the_reason_to_the_inner_equivalency_assertion_works()
...
+
+internal class MyDictionary1 : Dictionary<string, int>
+{
+    public new void Add(string key, int value)
+    {
+        base.Add(key, value);
+    }
+}
+
+internal class MyDictionary2 : IDictionary<string, int>
```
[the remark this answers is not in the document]
ITaluone: I just tried with your example above :)

I will change them.

Are your suggestions meant as "either one" or `MyDictionary1 `-> `NonGenericChildDictionary`; `MyDictionary2 `-> `NonGenericDictionary`

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
```diff
@@ -17,32 +19,60 @@ public class GenericDictionaryEquivalencyStep : IEquivalencyStep
...
+        if (IsNotNull(comparands.Subject)
+            && EnsureSubjectIsOfTheExpectedDictionaryType(comparands, expectedDictionary) is { } actualDictionary)
+        {
+            AssertDictionaryEquivalence(comparands, context, nestedValidator, actualDictionary, expectedDictionary);
+        }
+
+        return EquivalencyResult.AssertionCompleted;
+    }
+
+    private static bool IsNonGenericDictionary(object subject)
```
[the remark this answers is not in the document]
ITaluone: Hmm.. Could work indeed :)

Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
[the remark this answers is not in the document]
ITaluone: As a side-note: Your example above was `class MyDictionary2: IDictionary<string, int>, IDictionary`, but this does not work, because either one interface requires a property `Keys` and a property `Values`, which cannot live together in one class.

The same with a few other methods

jnyrup (approved):
Great work!

Dennis is conferencing these days, so he's pretty occupied.
https://twitter.com/ddoomen/status/1712069826494050455

Src/FluentAssertions/Equivalency/Steps/GenericDictionaryEquivalencyStep.cs
```diff
@@ -17,32 +19,53 @@ public class GenericDictionaryEquivalencyStep : IEquivalencyStep
...
+    {
+        if (subject is not IDictionary)
+        {
+            return false;
         }
 
-        return EquivalencyResult.ContinueWithNext;
+        return !subject.GetType().GetInterfaces()
```
dennisdoomen, line 56: ⛏️ (nitpick), I prefer

```
        return !subject.GetType().GetInterfaces().Any(@interface => 
            @interface.IsGenericType && @interface.GetGenericTypeDefinition() == typeof(IDictionary<,>));
```

dennisdoomen (approved):
Nice work. It's always hard to keep a clean PR when moving around code _and_ fixing bugs. Becoming proficient with fixup commits and interactive rebases is a skill you wonder why you never used it before.

PR #2314, opened 2023-09-21
Add missing `DateTimeOffset` assertions

ITaluone:
This also ref. qodana complaining about 'Method return value is never used'

## IMPORTANT 

* [ ] If the PR touches the public API, the changes have been approved in a separate issue with the "api-approved" label.
* [x] The code complies with the [Coding Guidelines for C#](https://www.csharpcodingguidelines.com/).
* [x] The changes are covered by unit tests which follow the Arrange-Act-Assert syntax and the naming conventions such as is used [in these tests](../tree/develop/Tests/FluentAssertions.Equivalency.Specs/MemberMatchingSpecs.cs#L51-L430).
* [ ] If the PR adds a feature or fixes a bug, please update [the release notes](../tree/develop/docs/_pages/releases.md) with a functional description that explains what the change means to consumers of this library, which are published on the [website](https://fluentassertions.com/releases).
* [ ] If the PR changes the public API the changes needs to be included by running [AcceptApiChanges.ps1](../tree/develop/AcceptApiChanges.ps1) or [AcceptApiChanges.sh](../tree/develop/AcceptApiChanges.sh).
* [ ] If the PR affects [the documentation](../tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).
    * [ ] Please also run `./build.sh --target spellcheck` or `.\build.ps1 --target spellcheck` before pushing and check the good outcome

PR #2378, opened 2023-10-12
Ignore ` UnusedMember.Global` Qodana scan rule and add missing tests

ITaluone:
Qodana is failing due to "The number of problems exceeds the fail threshold".

This was a surprise to me, because no PR was merged which had other than "Qodana alright!"-messages.

As I dug into the logs (and also checked the uploaded results on qodana.cloud) I realized, that sometimes `develop` (with the correct fail threshold) and sometimes `refs/heads/develop` (which has a higher fail rate) is chosen from qodana. I couldn't figure out when this happens.. seems a little random to me.

As a possible solution we could try the hint from YouTrack: https://youtrack.jetbrains.com/issue/QD-7084/Getting-refs-heads-branch-instead-of-just-branch-Qodana-GitHub-Actions#focus=Comments-27-8090586.0-0

Note: 
 - I think the real impact will be seen only after this is merged.
 - The diffenences between this two branch reports are
image
However I also don't get why qodana reports this 🤔 

## IMPORTANT 

### DIFF
--- a/Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.HaveCountGreaterThanOrEqualTo.cs
+++ b/Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.HaveCountGreaterThanOrEqualTo.cs
@@ -69,5 +69,15 @@ public partial class CollectionAssertionSpecs
             act.Should().Throw<XunitException>()
                 .WithMessage("*at least*1*we want to test the behaviour with a null subject*found <null>*");
         }
+
+        [Fact]
+        public void Chaining_after_one_assertion()
+        {
+            // Arrange
+            var collection = new[] { 1, 2, 3 };
+
+            // Act / Assert
+            collection.Should().HaveCountGreaterThanOrEqualTo(3).And.Contain(1);
+        }
     }
 }
--- a/Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.HaveCountLessThanOrEqualTo.cs
+++ b/Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.HaveCountLessThanOrEqualTo.cs
@@ -69,5 +69,15 @@ public partial class CollectionAssertionSpecs
             act.Should().Throw<XunitException>()
                 .WithMessage("*at most*1*we want to test the behaviour with a null subject*found <null>*");
         }
+
+        [Fact]
+        public void Chaining_after_one_assertion()
+        {
+            // Arrange
+            var collection = new[] { 1, 2, 3 };
+
+            // Act / Assert
+            collection.Should().HaveCountLessThanOrEqualTo(3).And.Contain(1);
+        }
     }
 }
--- a/Tests/FluentAssertions.Specs/Numeric/ComparableSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Numeric/ComparableSpecs.cs
@@ -556,6 +556,17 @@ public class ComparableSpecs
             // Assert
             act.Should().NotThrow();
         }
+
+        [Fact]
+        public void Chaining_after_one_assertion()
+        {
+            // Arrange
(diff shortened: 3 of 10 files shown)


ITaluone:
Another possibility could be just to ignore these type of errors.

Maybe this is a more reliable solution, because usually we don't have such code, that is never used, do we?

dennisdoomen:
> As I dug into the logs (and also checked the uploaded results on qodana.cloud) I realized, that sometimes `develop` (with the correct fail threshold) and sometimes `refs/heads/develop` (which has a higher fail rate) is chosen from qodana. I couldn't figure out when this happens.. seems a little random to me.

But those are the same. The full name of `develop` _is_ `refs/heads/develop`. Although I do see that as of October 8th, the reports to Qodana Cloud switched from `develop` to `refs/heads/develop`

IT-VBFK:
But why do they have different failures?

dennisdoomen:
> But why do they have different failures?

Because the `develop` report is from October 8th. I just ran Qodana locally using `qodana scan --baseline qodana.sarif.json --ide QDNET` and it reports 61 new issues as well. Apparently we have new issues. I'm about to report the false positive for `FluentAssertions.Specialized.ActionAssertions.NotThrow`.

dennisdoomen:
But it doesn't hurt fixing the branch name so we are not hurt by the weird checkout behavior. I'm not sure about ignoring those issues.

dennisdoomen:
I remember now. We're targeting .NET Standard 2.0 and 2.1, but we don't have tests for these. Qodana is correctly treating those as missing. I've [reported a change request](https://youtrack.jetbrains.com/issue/QD-7401/Allow-excluding-or-including-certain-.NET-framework-targets) so we can exclude/include certain targets.

jnyrup:
I'm fine disabling `UnusedMember.Global` until Qodana hopefully get's an option to ignore certain TFMs, our Code Coverage should cover us here.

The [reports](https://qodana.cloud/projects/AgRgk/reports/KRYRm?genericFilters=N4IgzgpgbhBOCWAXAniAXAbRAWQPYBM4BDRCEAXQBoQBjEiAc11lUxACUJ8BXAO3yK8a8CGAAE8XmIDKyALYAjXABsxAEQg1lRWCXi5eYCtRQAHMm2wREACwJjY17rClQiy7hAnje0OGO5IfDEACl4DAFpTBDdSMSIaGlEweAV4ZSRkAEoKAF8gA#ded3a4bfc999e4ba21c010d7854bec83132b946693e1fdf33d51a60787e89ee2) on `UnusedMethodReturnValue.Global` look valid. They are all for net47/net6
image

dennisdoomen:
> The [reports](https://qodana.cloud/projects/AgRgk/reports/KRYRm?genericFilters=N4IgzgpgbhBOCWAXAniAXAbRAWQPYBM4BDRCEAXQBoQBjEiAc11lUxACUJ8BXAO3yK8a8CGAAE8XmIDKyALYAjXABsxAEQg1lRWCXi5eYCtRQAHMm2wREACwJjY17rClQiy7hAnje0OGO5IfDEACl4DAFpTBDdSMSIaGlEweAV4ZSRkAEoKAF8gA#ded3a4bfc999e4ba21c010d7854bec83132b946693e1fdf33d51a60787e89ee2) on `UnusedMethodReturnValue.Global` look valid. They are all for net47/net6

OK, so we'll keep checking for `UnusedMethodReturnValue.Global`

IT-VBFK:
I think the `UnusedMethodReturnValue.Global` error/warning is only gone if you chain at least once assertions with `.And`. This is why tests #2314 have the pretty unnecessary `.And.Be(xxx)`. This was more or less to silence this qodana complaint.

ITaluone:
Thanks for clarfication.

I addressed all comments :)

jnyrup (approved):

dennisdoomen (approved):

Issue #1677, opened 2021-09-01
Assorted v8 improvements


lg2de:
After release of V6 there are issues reported which cannot be (completely) realized before V7 because of breaking changes.
This issue lists all pending changes including link to related issue(s). It will be updated permanently.

- [ ] Make `ComparableAssertions` more generic - [draft](https://github.com/jnyrup/fluentassertions/tree/GenericIComparableAssertions)
- [x] Remove `GreaterOrEqualTo` (obsolete) methods replaced by `GreaterThanOrEqualTo` (#1673)
- [ ] In `EquivalentValidator` replace the last `IEquivalencyStep` with a hard-coded assertion to replace the `NotImplementedException`. See https://github.com/fluentassertions/fluentassertions/blob/master/Src/FluentAssertions/Equivalency/EquivalencyValidator.cs#L76
- [x] Rename `RespectingRuntimeTypes` to `RespectingRuntimeTypeProperties` or something like that. See #1830. Also, it can be confusing because we also have `IncludingAllRuntimeProperties` and `IncludingAllDeclaredProperties`
- [x] Check whether assertions on `AndWhichConstraint` should be executed within `AssertionScope` after base assertion has failed. See discussion on [Slack](https://fluentassertions.slack.com/archives/C01055CMSD9/p1660070451083999).
- [x] Refactor `AsyncFunctionAssertions` as "real" base class (preparation in #1972).
- [ ] Investigate if we can replace/aid `CallerIdentifier` using `[CallerArgumentExpression]`.
  - sometimes it could also be useful to extract the variable name of the `expected` parameter.
  - this might also let us replace `Expression<>` with `Func<>`
  - investigate whether this works for .NET 4.8 as well
- [ ] Consider renaming `AssertionScope.Current` to `AssertionScope.ThreadInstance` 
- [ ] Consider renaming `FailWith` since it does not always throw
- [x] Consider renaming `ExcludingNestedObjects` to something with _recursive_ in it or _deep_ vs _shallow_
- [ ] Align terminology between "object graph comparison" in the [docs](https://fluentassertions.com/objectgraphs/) vs "structural comparison" in the code.
- [ ] Consider renaming "expected" to "expectation" and "unexpected" to ?, to use nouns instead of adjectives
- [ ] Consider returning `AndWhichConstraint` from `GenericCollectionAssertions.ContainItemsAssignableTo<T>`
- [x] Consider making `EquivalencyValidator` and some other "internal" classes `internal`

[...]

PR #2380, opened 2023-10-13
Add [NotNull] attribute on the Should() method for object assertions

### LINKED ISSUE #1115: Support c# 8 nullable reference

0xced:
Compelling example:

```csharp
[Theory]
[InlineData(true)]
[InlineData(false)]
public void Test1(bool currentUser)
{
    IPrincipal principal = currentUser ? new WindowsPrincipal(WindowsIdentity.GetCurrent()) : new ClaimsPrincipal();
    IIdentity? identity = principal.Identity;
    
    identity.Should().NotBeOfType<GenericIdentity>();
    identity.IsAuthenticated.Should().BeFalse();
}
```

Fixes #1115

### DIFF
--- a/Src/FluentAssertions/AssertionExtensions.cs
+++ b/Src/FluentAssertions/AssertionExtensions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -18,6 +18,7 @@ using FluentAssertions.Streams;
 using FluentAssertions.Types;
 using FluentAssertions.Xml;
 using JetBrains.Annotations;
+using NotNullAttribute = System.Diagnostics.CodeAnalysis.NotNullAttribute;
 #if !NETSTANDARD2_0
 using FluentAssertions.Events;
 #endif
@@ -174,7 +175,7 @@ public static class AssertionExtensions
     /// current <see cref="Assembly"/>.
     /// </summary>
     [Pure]
-    public static AssemblyAssertions Should(this Assembly assembly)
+    public static AssemblyAssertions Should([NotNull] this Assembly assembly)
     {
         return new AssemblyAssertions(assembly);
     }
@@ -184,7 +185,7 @@ public static class AssertionExtensions
     /// current <see cref="XElement"/>.
     /// </summary>
     [Pure]
-    public static XDocumentAssertions Should(this XDocument actualValue)
+    public static XDocumentAssertions Should([NotNull] this XDocument actualValue)
     {
         return new XDocumentAssertions(actualValue);
     }
@@ -194,7 +195,7 @@ public static class AssertionExtensions
     /// current <see cref="XElement"/>.
     /// </summary>
     [Pure]
-    public static XElementAssertions Should(this XElement actualValue)
+    public static XElementAssertions Should([NotNull] this XElement actualValue)
     {
         return new XElementAssertions(actualValue);
     }
@@ -204,7 +205,7 @@ public static class AssertionExtensions
     /// current <see cref="XAttribute"/>.
     /// </summary>
     [Pure]
-    public static XAttributeAssertions Should(this XAttribute actualValue)
+    public static XAttributeAssertions Should([NotNull] this XAttribute actualValue)
     {
         return new XAttributeAssertions(actualValue);
     }
@@ -214,7 +215,7 @@ public static class AssertionExtensions
     /// current <see cref="Stream"/>.
     /// </summary>
     [Pure]
-    public static StreamAssertions Should(this Stream actualValue)
+    public static StreamAssertions Should([NotNull] this Stream actualValue)
     {
         return new StreamAssertions(actualValue);
     }
@@ -224,7 +225,7 @@ public static class AssertionExtensions
     /// current <see cref="BufferedStream"/>.
     /// </summary>
     [Pure]
-    public static BufferedStreamAssertions Should(this BufferedStream actualValue)
+    public static BufferedStreamAssertions Should([NotNull] this BufferedStream actualValue)
     {
         return new BufferedStreamAssertions(actualValue);
     }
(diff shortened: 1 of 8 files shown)


Src/FluentAssertions/AssertionExtensions.cs
```diff
@@ -277,7 +277,7 @@ private static void ForceEnumeration<T>(T subject, Func<T, IEnumerable> enumerab
     /// current <see cref="object"/>.
     /// </summary>
     [Pure]
-    public static ObjectAssertions Should(this object actualValue)
+    public static ObjectAssertions Should([System.Diagnostics.CodeAnalysis.NotNull] this object actualValue)
```
dennisdoomen, line 280: Why not add the attribute to _all_ `Should()` methods?

IT-VBFK:
Isn't this something that pretends what isn't there?

In such a case we can get rid of all `null` checks inside in the assertions?

dennisdoomen:
> Isn't this something that pretends what isn't there?
> 
> In such a case we can get rid of all `null` checks inside in the assertions?

I'm not entirely sure what you are saying? If you mean that we're kind of pretending that the subject is never `null`, that's true. We're relying on the assertion method itself to take care of that.

IT-VBFK:
Ok, then this is just to silence the IDE, right?

lg2de:
> Isn't this something that pretends what isn't there?
> 
> In such a case we can get rid of all `null` checks inside in the assertions?

Newer compiler version use such attributes to create warnings. Older version does not.
And: It is a warning which the user can ignore or even disable.
So, the `Should` method cannot **ensure** non-nullability with the attribute.

IT-VBFK:
> So, the Should method cannot ensure non-nullability with the attribute.

Ahh.. thanks for clarification :)

Src/FluentAssertions/AssertionExtensions.cs
0xced: That's a very good question! I think we should add the [NotNull] attribute on all `Should` methods that return a `ReferenceTypeAssertions<,>`. So I started to write a test and noticed that I could not assert `parameterInfo.IsDecoratedWith<NotNullAttribute>`, hence #2385. 😄

0xced:
I have rebased this pull request on top of #2385 because the new specs (`Should_methods_returning_reference_type_assertions_are_annotated_with_not_null_attribute` and `Should_methods_not_returning_reference_type_assertions_are_not_annotated_with_not_null_attribute`) require the new assertions on `ParameterInfo`.

Src/FluentAssertions/AssertionExtensions.cs
dennisdoomen: > So I started to write a test and noticed that I could not assert parameterInfo.IsDecoratedWith<NotNullAttribute>, hence https://github.com/fluentassertions/fluentassertions/pull/2385. 😄

I don't think we need to have tests for that.

0xced: Well, at least it was very interesting to dive into all the assertion types, I learnt a lot. And the test is now written anyway, alongside other tests on the `Should` methods. But this pull request could also be merged without the tests added in `AssertionExtensionsSpecs` and thus without requiring #2385 to also be merged.

0xced:
I just rebased this pull request on the `develop` branch and make it **not** depend on #2385 at all.

The assertion is now written like this:

```csharp
parameter.GetCustomAttribute<NotNullAttribute>().Should().NotBeNull();
```

instead of this (which would require #2385 to be merged first):
```csharp
parameter.Should().BeDecoratedWith<NotNullAttribute>();
```

zhangpengchen:
Hi @dennisdoomen ,

Would you consider releasing this feature soon so we can finally avoid workarounds by mixing the native assertions and fluent assertions in the tests.

dennisdoomen (approved):

dennisdoomen (changes requested):
Please also update the `releases.md`

0xced:
I just added an entry in the release notes under improvements in 6ac05e84f9e7aeb8df561b3e997db04f4e488fd5.

dennisdoomen (approved):
Awesome

jnyrup (approved):
Noticed that `[NotNull]` is not added to these overloads of `Should`, but guess it doesn't really matter that much.

```
AssertionExtensions.cs:168:    public static ExecutionTimeAssertions Should(this ExecutionTime executionTime)
AssertionExtensions.cs:753:    public static TypeSelectorAssertions Should(this TypeSelector typeSelector)
AssertionExtensions.cs:788:    public static MethodInfoSelectorAssertions Should(this MethodInfoSelector methodSelector)
AssertionExtensions.cs:813:    public static PropertyInfoSelectorAssertions Should(this PropertyInfoSelector propertyInfoSelector)
AssertionExtensions.cs:865:    public static TaskCompletionSourceAssertions<T> Should<T>(this TaskCompletionSource<T> tcs)
AssertionExtensions.cs:894:    public static TaskCompletionSourceAssertions Should(this TaskCompletionSource tcs)
```

thj-dk:
When is this planned to be released?

ITaluone:
I think there is no specified date for now :)

I assume, that at least this ticket have to be completely addressed: https://github.com/fluentassertions/fluentassertions/issues/1677

dennisdoomen:
@ITaluone is correct

dggsax:
Is it possible to release it in an alpha release at the very least? Or are there items in #1677 that you need assistance with that could help speed this along. I very much look forward to taking advantage of the changes in this PR, thanks!

dennisdoomen:
> Is it possible to release it in an alpha release at the very leas

I just tagged alpha 4. Since it's a pre-release, all bets are off.

0xced:
Oopsie. 😖

https://github.com/fluentassertions/fluentassertions/actions/runs/9291245214/job/25569332022#step:5:551

> [ERR] Push: error: Response status code does not indicate success: 403 (The specified API key is invalid, has expired, or does not have permission to access the specified package.).

IT-VBFK:
Why not just compiling the lates develop yourself?

- Clone this repo
- run `build.ps1` or `build.sh` and you have your "alpha version" 🤷‍♂️

dennisdoomen:
Fixed the API key and pushed again.

PR #2381, opened 2023-10-13
Simplify the ApiApproval test

0xced:
* Simplify how paths relative to the solution directory are accessed.
* Actually use the Verify.DiffPlex integration instead of using DiffPlex directly.

### DIFF
--- a/Tests/Approval.Tests/ApiApproval.cs
+++ b/Tests/Approval.Tests/ApiApproval.cs
@@ -20,10 +20,10 @@ public class ApiApproval
 
     [Theory]
     [ClassData(typeof(TargetFrameworksTheoryData))]
-    public Task ApproveApi(string frameworkVersion)
+    public Task ApproveApi(string framework)
     {
         var configuration = typeof(ApiApproval).Assembly.GetCustomAttribute<AssemblyConfigurationAttribute>()!.Configuration;
-        var assemblyFile = GetPath("Src", "FluentAssertions", "bin", configuration, frameworkVersion, "FluentAssertions.dll");
+        var assemblyFile = CombinedPaths("Src", "FluentAssertions", "bin", configuration, framework, "FluentAssertions.dll");
         var assembly = Assembly.LoadFile(assemblyFile);
         var publicApi = assembly.GeneratePublicApi(options: null);
 
@@ -31,19 +31,15 @@ public class ApiApproval
             .Verify(publicApi)
             .ScrubLinesContaining("FrameworkDisplayName")
             .UseDirectory(Path.Combine("ApprovedApi", "FluentAssertions"))
-            .UseFileName(frameworkVersion)
+            .UseFileName(framework)
             .DisableDiff();
     }
-
-    private static string GetSolutionDirectory([CallerFilePath] string path = "") => Path.Combine(Path.GetDirectoryName(path)!, "..", "..");
-
-    private static string GetPath(params string[] paths) => Path.GetFullPath(Path.Combine(paths.Prepend(GetSolutionDirectory()).ToArray()));
 
     private class TargetFrameworksTheoryData : TheoryData<string>
     {
         public TargetFrameworksTheoryData()
         {
-            var csproj = GetPath("Src", "FluentAssertions", "FluentAssertions.csproj");
+            var csproj = CombinedPaths("Src", "FluentAssertions", "FluentAssertions.csproj");
             var project = XDocument.Load(csproj);
             var targetFrameworks = project.XPathSelectElement("/Project/PropertyGroup/TargetFrameworks");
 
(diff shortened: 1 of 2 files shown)


Tests/Approval.Tests/ApiApproval.cs
```diff
@@ -19,72 +16,34 @@ namespace Approval.Tests;
...
-            }
+    private static string GetSolutionDirectory([CallerFilePath] string path = "") => Path.Combine(Path.GetDirectoryName(path)!, "..", "..");
 
-            builder.AppendLine(line.Text);
-        }
-
-        var compareResult = CompareResult.NotEqual(builder.ToString());
-        return Task.FromResult(compareResult);
-    }
+    private static string GetPath(params string[] paths) => Path.GetFullPath(Path.Combine(paths.Prepend(GetSolutionDirectory()).ToArray()));
```
dennisdoomen, line 40: 🔧 We like to order the methods in calling order, so this method should be positioned at the bottom. 
🤔 I think naming this `CombinedPaths` would better cover its purpose, right?

Tests/Approval.Tests/ApiApproval.cs
```diff
@@ -19,72 +16,34 @@ namespace Approval.Tests;
...
+        var configuration = typeof(ApiApproval).Assembly.GetCustomAttribute<AssemblyConfigurationAttribute>()!.Configuration;
+        var assemblyFile = GetPath("Src", "FluentAssertions", "bin", configuration, frameworkVersion, "FluentAssertions.dll");
+        var assembly = Assembly.LoadFile(assemblyFile);
         var publicApi = assembly.GeneratePublicApi(options: null);
 
         return Verifier
             .Verify(publicApi)
             .ScrubLinesContaining("FrameworkDisplayName")
             .UseDirectory(Path.Combine("ApprovedApi", "FluentAssertions"))
-            .UseStringComparer(OnlyIncludeChanges)
```
dennisdoomen, line 45: ❓ Why is this no longer needed?

dennisdoomen (changes requested):
👍🏻 Nice improvement. Thanks.

🔧 You need to format the file to apply the `.editorconfig` settings. Some lines are too long.

Tests/Approval.Tests/ApiApproval.cs
0xced: Because `VerifyDiffPlex.Initialize(OutputType.Compact);` is used instead. This is how the [Verify.DiffPlex](https://github.com/VerifyTests/Verify.DiffPlex) documentation says it should be integrated. Before this pull request, the `Verify.DiffPlex` package was referenced but not actually used at all!

Tests/Approval.Tests/ApiApproval.cs
0xced: Addressed in 58fbc6de6255e71bf806f708bc4d6bf46d1c388a

0xced:
I think I addressed everything in 58fbc6de6255e71bf806f708bc4d6bf46d1c388a.

dennisdoomen (approved):

jnyrup:
Sorry for the late reply, I had written a reply but apparently not pressed the comment button...

A historical note: I [added](9414c7c85409d408614c3bbf082a832527a4fd73) this custom code before the `OutputType` mode was [added](https://github.com/VerifyTests/Verify.DiffPlex/pull/71).

Here's a comparison of my custom comparer and `OutputType.Compact` and having looked at this a couple of times now, I still prefer my minimal output.
I've opened https://github.com/VerifyTests/Verify.DiffPlex/issues/184 proposing a new Minimal mode.

```
Compare Result:
-         public FluentAssertions.AndConstraint<FluentAssertions.Types.MethodInfoAssertions> BeVirtual(string because = "", params object[] becauseArgs) { }
+         public FluentAssertions.AndConstraint<FluentAssertions.Types.MethodInfoAssertions> BeVirtualFoo(string because = "", params object[] becauseArgs) { }
```

```
Compare Result:
2419         public FluentAssertions.AndConstraint<FluentAssertions.Types.MethodInfoAssertions> BeAsync(string because = "", params object[] becauseArgs) { }
   -         public FluentAssertions.AndConstraint<FluentAssertions.Types.MethodInfoAssertions> BeVirtual(string because = "", params object[] becauseArgs) { }
   +         public FluentAssertions.AndConstraint<FluentAssertions.Types.MethodInfoAssertions> BeVirtualFoo(string because = "", params object[] becauseArgs) { }
2421         public FluentAssertions.AndConstraint<FluentAssertions.Types.MethodInfoAssertions> NotBeAsync(string because = "", params object[] becauseArgs) { }
```

0xced:
Thanks for the explanation. I must admit I had not noticed the slight difference between the custom comparer and the `Compact` output type.

If Simon is fine with a new `Minimal` output type then it will be trivial to update this pull request to use it instead of `Compact`. Let's wait for Simon's input before moving forward.

0xced:
I just rebased onto develop and applied the new minial diff style introduced in https://github.com/VerifyTests/Verify.DiffPlex/pull/185 (released in [Verify.DiffPlex 2.3.0](https://www.nuget.org/packages/Verify.DiffPlex/2.3.0)).

jnyrup (approved):

PR #2385, opened 2023-10-14
Introduce new assertions on ParameterInfo

0xced:
All the (Not)BeDecoratedWith methods were implemented for `MemberInfo` but implementing them for `ICustomAttributeProvider` instead allows for a _free_ `ParameterInfoAssertions` implementation.

## IMPORTANT 


Src/FluentAssertions/Types/CustomAttributeProviderAssertions.cs
```diff
@@ -0,0 +1,151 @@
...
+using FluentAssertions.Execution;
+using FluentAssertions.Primitives;
+
+namespace FluentAssertions.Types;
+
+/// <summary>
+/// Contains a number of methods to assert that a <see cref="ICustomAttributeProvider"/> is in the expected state.
+/// </summary>
+[DebuggerNonUserCode]
+public abstract class CustomAttributeProviderAssertions<TSubject, TAssertions> : ReferenceTypeAssertions<TSubject, TAssertions>
```
dennisdoomen, line 17: 🤔 Given that this type may contain other assertions that are reusable across the reflection assertions, I think we should name this something like `ReflectionAssertions`
🌱 I hope using inheritance isn't going to cause issues later on.

Tests/FluentAssertions.Specs/Types/ParameterInfoAssertionSpecs.cs
```diff
@@ -0,0 +1,213 @@
...
+using System.Linq.Expressions;
+using System.Reflection;
+using Xunit;
+using Xunit.Sdk;
+
+namespace FluentAssertions.Specs.Types;
+
+public class ParameterInfoAssertionSpecs
+{
+    public class BeDecortatedWithOfT
```
dennisdoomen, line 12: ⛏️ Typo

Tests/FluentAssertions.Specs/Types/ParameterInfoAssertionSpecs.cs
```diff
@@ -0,0 +1,213 @@
...
+    {
+        [Fact]
+        public void When_asserting_a_parameter_is_decorated_with_attribute_and_it_is_it_succeeds()
```
dennisdoomen, line 15: 🔧 We have been adopting a more fact-based naming convention for new tests, e.g. `A_parameter_that_is_decorated_with_the_expected_attribute_is_valid` or `Succeeds_for_a_parameter_with_the_expected_attribute`. For example, It's completely clear from the parent class that every test is about checking whether a parameter is decorated with an attribute, so you can leave that out. Same for the other tests.

Tests/FluentAssertions.Specs/Types/ParameterInfoAssertionSpecs.cs
```diff
@@ -0,0 +1,213 @@
...
+
+            // Act
+            Action act = () => parameterInfo.Should().BeDecoratedWith<DummyParameterAttribute>(d => d.Value == "Value");
+
+            // Assert
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_subject_is_null_be_decorated_withOfT_should_fail()
```
dennisdoomen, line 113: 🔧 Stay away from technical terms like `OfT`

Tests/FluentAssertions.Specs/Types/ParameterInfoAssertionSpecs.cs
```diff
@@ -0,0 +1,213 @@
...
+            Action act = () =>
+                parameterInfo.Should().NotBeDecoratedWith((Expression<Func<DummyParameterAttribute, bool>>)null);
+
+            // Assert
+            act.Should().ThrowExactly<ArgumentNullException>()
+                .WithParameterName("isMatchingAttributePredicate");
+        }
+    }
+
+    #region Internal classes used in unit tests
```
dennisdoomen, line 180: 🔧 We prefer not to use regions anymore.

dennisdoomen (changes requested):
🤔 We now miss a `Parameters()` extension method like we have in our [Reflection API](https://fluentassertions.com/typesandmethods/)

Src/FluentAssertions/Types/CustomAttributeProviderAssertions.cs
jnyrup: Currently our inheritance hierarchy of type-related assertions follows the inheritance hierarchy of the corresponding BCL types.
I notice that e.g. `Assembly` also implements `ICustomAttributeProvider`.

How difficult would it be to implement this feature without changing our inheritance hierarchy?

It's not show stopper for me, it just feels a little odd to me to mix these concepts.

Src/FluentAssertions/Common/TypeExtensions.cs
```diff
@@ -31,13 +31,10 @@ public static bool IsDecoratedWith<TAttribute>(this Type type)
...
 
-    public static bool IsDecoratedWith<TAttribute>(this MemberInfo type)
+    public static bool IsDecoratedWith<TAttribute>(this ICustomAttributeProvider type)
         where TAttribute : Attribute
     {
-        // Do not use MemberInfo.IsDefined
-        // There is an issue with PropertyInfo and EventInfo preventing the inherit option to work.
-        // https://github.com/dotnet/runtime/issues/30219
-        return Attribute.IsDefined(type, typeof(TAttribute), inherit: false);
+        return type.IsDefined(typeof(TAttribute), inherit: false);
```
jnyrup, line 37: It seems this will be calling the `MemberInfo.IsDefined` that the comment said _not_ to use.
I dug into this and the related PRs are #748 and #1095.

I checked out develop at #748 and used the prohibited method.
I don't have net45 or netcoreapp1.2 installed, but for the remaining TFMs no tests failed.
Since we no longer support these TFMs directly or indirectly, I think it's safe to use the `ICustomAttributeProvider.IsDefined`.

0xced:
I won't be able to address all the feedback in the coming weeks so please don't hesitate to adapt everything that needs to be adapted if have the time to.

Also, note that all the specs were mostly copy/pasted from `PropertyInfoAssertionSpecs`.

0xced:
Wow, I did not realize that this pull request was opening Pandora's box! 😅

Indeed a new `Parameters()` extension method would be warranted. This also implies designing all the properties that would make sense on a `ParameterInfoSelector` (e.g. `ThatAreInput`, `ThatHaveDefaultValue` etc.) Would you prefer me to add it as part of this pull request or in a follow-up pull request?

And yes, `Assembly` also implements `ICustomAttributeProvider`, so does `Type`! 😁

I'm currently reading [Multiple Inheritance in C#](https://stackoverflow.com/questions/178333/multiple-inheritance-in-c-sharp) on Stack Overflow to see if I can come up with an elegant solution that does not involve changing the inheritance hierarchy.

In the end, that's quite some work ahead for something that was supposed to _just_ be some support for #2380. 😂

dennisdoomen:
> Wow, I did not realize that this pull request was opening Pandora's box!

Sorry about that. This tends to happen with an open-source library that has close to 300 million downloads ;-).

> In the end, that's quite some work ahead for something that was supposed to just be some support for https://github.com/fluentassertions/fluentassertions/pull/2380. 😂

You could still decouple #2380 from this one.

IT-VBFK:
Think about coverage and qodana scan, if you can :)

dennisdoomen:
@0xced did you give up on this?

0xced:
I did not give up on this but since that was just a side effect of #2380 and not something I actually need I did not gave it a high priority.

dennisdoomen:
Closed because of a lack of progress

Issue #2002, opened 2022-09-19
Move DataSet/DataTable/DataRow assertions to separate library


ericnewton76:
Should DataSet/DataTable/DataRow assertions get lumped into a separate library?  These are types that are over 20 years old, introduced with NetFramework 1.0 beta 1... and I know this because yes I was porting things from VB6 back then as fast as I could... ;-D

Of note, starting a fresh testing project with nunit 3.12 and fluentassertions 6.7.0, I get reference errors about missing reference to System.Data v4.0.0.  Otherwise FluentAssertions is pretty clean.

- FluentAssertions 6.7.0

* [x] Copy all data-related code to https://github.com/fluentassertions/fluentassertions.datasets
* [x] How can the data project inject additional equivalency steps?
* [x] Delete all data-related code from the main repo
* [x] Move the data-related docs to the read-me of the data repo
* [x] Ensure that the data repo has its own discussions, issues and releases
* [x] Update the breaking changes of the main repo
* [x] Push an alpha version of FA 7.0 and update dependencies in the data set project to make it build again
* [x] Introduce a mechanism for an extension library like `FluentAssertions.DataSets` to register custom equivalency steps
* [x] Review all namespace imports and make sure everything works as expected
* [x] Ensure the build works correctly
* [x] Identify any code analysis issues
* [x] Check package contents
* [x] Update the readme
* [x] Setup branch protection rules

jnyrup:
It's definitely something we talked about for vNext (7.0.0), we just never got around to create an issue for it.

Any good ideas for what name it should have?

The first thing that comes to my mind is `FluentAssertions.System.Data`.

ITaluone:
Do you really inteded to close this? 
I think this was a false positive (PR description: .... resolves #2002)

dennisdoomen:
No, I didn't. Thanks for heads up.

PR #2386, opened 2023-10-15
Allow specifying the build configuration.

dennisdoomen:
1. Made the solution build configuration a build parameter
2. Use the `Debug` configuration for a local build
3. Use the `CI` configuration on a server

Needed this to get a local package for testing #2002

### DIFF
--- a/.nuke/build.schema.json
+++ b/.nuke/build.schema.json
@@ -6,6 +6,14 @@
     "build": {
       "type": "object",
       "properties": {
+        "Configuration": {
+          "type": "string",
+          "description": "The solution configuration to build. Default is 'Debug' (local) or 'CI' (server)",
+          "enum": [
+            "CI",
+            "Debug"
+          ]
+        },
         "Continue": {
           "type": "boolean",
           "description": "Indicates to continue a previously failed build attempt"
--- a/Build/Build.cs
+++ b/Build/Build.cs
@@ -40,6 +40,9 @@ class Build : NukeBuild
 
     string PullRequestBase => GitHubActions?.BaseRef;
 
+    [Parameter("The solution configuration to build. Default is 'Debug' (local) or 'CI' (server).")]
+    readonly Configuration Configuration = IsLocalBuild ? Configuration.Debug : Configuration.CI;
+
     [Parameter("Use this parameter if you encounter build problems in any way, " +
         "to generate a .binlog file which holds some useful information.")]
     readonly bool? GenerateBinLog;
@@ -126,7 +129,7 @@ class Build : NukeBuild
 
             DotNetBuild(s => s
                 .SetProjectFile(Solution)
-                .SetConfiguration(Configuration.CI)
+                .SetConfiguration(Configuration)
                 .When(GenerateBinLog is true, _ => _
                     .SetBinaryLog(ArtifactsDirectory / $"{Solution.Core.FluentAssertions.Name}.binlog")
                 )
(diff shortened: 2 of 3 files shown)


Build/Build.cs
```diff
@@ -40,6 +40,9 @@ class Build : NukeBuild
 
     string PullRequestBase => GitHubActions?.BaseRef;
 
+    [Parameter]
+    readonly Configuration Configuration = IsLocalBuild ? Configuration.Debug : Configuration.CI;
```
jnyrup, line 44: I think we would need to introduce targeted parameters to make this work.
E.g. Something along the lines of this.

```cs
[Parameter]
readonly Configuration Configuration = Configuration.Debug;

readonly Configuration BuildConfiguration = IsLocalBuild ? Configuration : Configuration.CI;
readonly Configuration TestConfiguration = IsLocalBuild ? Configuration : Configuration.Debug;
readonly Configuration PackConfiguration = IsLocalBuild ? Configuration : Configuration.Release;

```

dennisdoomen: Why? The solution configurations `Debug` and `CI` already use the right project configurations.

jnyrup: The solution configuration `CI` is (to the best of my knowledge) for building a subset of the projects in debug or release mode.
At least I'm not sure what `dotnet pack --configuration CI` would output.

dennisdoomen: You're right. The "configuration" means different things depending on which `dotnet` command we pass. On the build-level, it's the solution config.

Build/Configuration.cs
```diff
@@ -4,11 +4,9 @@
 [TypeConverter(typeof(TypeConverter<Configuration>))]
 public class Configuration : Enumeration
 {
-    public static Configuration Debug { get; } = new() { Value = nameof(Debug) };
+    public static Configuration Debug = new() { Value = nameof(Debug) };
 
-    public static Configuration Release { get; } = new() { Value = nameof(Release) };
-
-    public static Configuration CI { get; } = new() { Value = nameof(CI) };
+    public static Configuration CI = new() { Value = nameof(CI) };
```
jnyrup, line 9: This change triggers several analyzers for me :suspect: 
[SA1401](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1401.md)
[MA0069](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0069.md)

dennisdoomen: Weird. I'm not getting those. I can ignore those, but I'm wondering why you are getting them and I don't.

Nuke needs those to be fields, otherwise you can't pass them as CLI configuration parameters or environment variables.

jnyrup: > Weird. I'm not getting those. I can ignore those, but I'm wondering why you are getting them and I don't.

I didn't get them on a regular build of `_build` but only when manually triggering code analysis.

> Nuke needs those to be fields, otherwise you can't pass them as CLI configuration parameters or environment variables.

Aha... 
It's [`[TypeConverter(typeof(TypeConverter<Configuration>))]`](https://github.com/nuke-build/nuke/blob/6b080532916f16742454c41e8534cc64e3608c63/source/Nuke.Tooling/Enumeration.cs#L101) that looks for fields but not properties.

Let's suppress the analyzers: SA1401, S1104, S2223, MA0069 and move on.

jnyrup (approved):

.nuke/build.schema.json
```diff
@@ -6,6 +6,15 @@
     "build": {
       "type": "object",
       "properties": {
+        "Configuration": {
+          "type": "string",
+          "description": "The solution configuration to build. Default is 'Debug' (local) or 'CI' (server)",
+          "enum": [
+            "CI",
+            "Debug",
+            "Release"
```
jnyrup, line 15: There is no corresponding "Configuration.Release" 🤔

Build/Configuration.cs
jnyrup: Discussed 1:1 - Since the analyzers aren't triggered automatically I'll take care of this later separately from this PR.

.nuke/build.schema.json
dennisdoomen: Fixed that. Forgot to include the updated `json

PR #2398, opened 2023-10-20
Do not continue asserting on the concrete exception type when the exception is `null`

### LINKED ISSUE #2395: Extra message when in AssertionScope using ThrowExactly
### Description

This is pretty minor. It might also be by design.

Basically, if you use `ThrowExactly` in an `AssertionScope`, it generates a second message that has the "wrong" context name. (This is apparent for me since I'm trying to work around strange context names in custom assertions, but actually, it's possible you consider the regular way this works to be intended not sure.)

### Reproduction Steps

```csharp
// Arrange
var a = () => { };

// Act

// Assert
using AssertionScope scope = new();
a.Should().ThrowExactly<InvalidOperationException>();
```

### Expected behavior

Should (and does) output this:

```
Expected System.InvalidOperationException, but no exception was thrown.
```

### Actual behavior

Outputs this instead:

```
Expected System.InvalidOperationException, but no exception was thrown.
Expected a to be System.InvalidOperationException, but found <null>.
```

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

You probably already know but it seems to be from this line in `ThrowExactly`:

```csharp
exception.Should().BeOfType(expectedType, because, becauseArgs);
```

I think because it is in an `AssertionScope` this check adds the additional message when the previous one already failed. This is maybe ok, except that because of the way `AssertionScope`s work, the context name is wrong (sort of).

In a different example:

```csharp
using AssertionScope scope = new();
value.Invoking(x => x.DoSomething()).Should().ThrowExactly<InvalidOperationException>();
```

you get the message:

```
Expected System.InvalidOperationException, but no exception was thrown.
Expected value.Invoking(x => x.DoSomething()) to be System.InvalidOperationException, but found <null>.
```

I suppose this could be by design, although the wording is not really correct,

No

IT-VBFK:
Fixes #2395 

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Specialized/DelegateAssertions.cs
+++ b/Src/FluentAssertions/Specialized/DelegateAssertions.cs
@@ -113,12 +113,15 @@ public abstract class DelegateAssertions<TDelegate, TAssertions> : DelegateAsser
 
             Type expectedType = typeof(TException);
 
-            Execute.Assertion
+            success = Execute.Assertion
                 .ForCondition(exception is not null)
                 .BecauseOf(because, becauseArgs)
                 .FailWith("Expected {0}{reason}, but no exception was thrown.", expectedType);
 
-            exception.Should().BeOfType(expectedType, because, becauseArgs);
+            if (success)
+            {
+                exception.Should().BeOfType(expectedType, because, becauseArgs);
+            }
 
             return new ExceptionAssertions<TException>(new[] { exception as TException });
         }
--- a/Tests/FluentAssertions.Specs/Specialized/DelegateAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Specialized/DelegateAssertionSpecs.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Threading.Tasks;
+using FluentAssertions.Execution;
 using FluentAssertions.Specialized;
 using Xunit;
 
@@ -33,5 +35,23 @@ public class DelegateAssertionSpecs
         // Act
         act.Should().ThrowExactly<ArgumentNullException>()
             .WithParameterName("clock");
+    }
+
+    public class ThrowExactly
+    {
+        [Fact]
+        public void Does_not_continue_assertion_on_exact_exception_type()
+        {
+            // Arrange
+            var a = () => { };
+
+            // Act
+            using var scope = new AssertionScope();
+            a.Should().ThrowExactly<InvalidOperationException>();
+
+            // Assert
+            scope.Discard().Should().ContainSingle()
+                .Which.Should().Match("*InvalidOperationException*no exception*");
+        }
     }
 }
--- a/docs/_pages/releases.md
+++ b/docs/_pages/releases.md
@@ -25,6 +25,8 @@ sidebar:
 * Capitalize `true` and `false` in failure messages and make them formattable to a custom `BooleanFormatter` - [#2390](https://github.com/fluentassertions/fluentassertions/pull/2390), [#2393](https://github.com/fluentassertions/fluentassertions/pull/2393)
 * Improved the failure message for `NotBeOfType` when wrapped in an `AssertionScope` and the subject is null  - [#2399](https://github.com/fluentassertions/fluentassertions/pull/2399)
 * Improved the failure message for `BeWritable`/`BeReadable` when wrapped in an `AssertionScope` and the subject is read-only/write-only - [#2399](https://github.com/fluentassertions/fluentassertions/pull/2399)
+* Improved the failure message for `ThrowExactly[Async]` when wrapped in an `AssertionScope` and no exception is thrown - [#2398](https://github.com/fluentassertions/fluentassertions/pull/2398)
+
 
 ### Breaking Changes (for users)
 * Moved support for `DataSet`, `DataTable`, `DataRow` and `DataColumn` into a new package `FluentAssertions.DataSet` - [#2267](https://github.com/fluentassertions/fluentassertions/pull/2267)
--- a/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
+++ b/Src/FluentAssertions/Specialized/AsyncFunctionAssertions.cs
@@ -91,12 +91,15 @@ public class AsyncFunctionAssertions<TTask, TAssertions> : DelegateAssertionsBas
         {
             Exception exception = await InvokeWithInterceptionAsync(Subject);
 
-            Execute.Assertion
+            success = Execute.Assertion
                 .ForCondition(exception is not null)
                 .BecauseOf(because, becauseArgs)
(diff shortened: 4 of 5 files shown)


Src/FluentAssertions/Specialized/DelegateAssertions.cs
```diff
@@ -113,12 +113,15 @@ public AndConstraint<TAssertions> NotThrow<TException>(string because = "", para
...
             Type expectedType = typeof(TException);
 
-            Execute.Assertion
+            success = Execute.Assertion
                 .ForCondition(exception is not null)
                 .BecauseOf(because, becauseArgs)
                 .FailWith("Expected {0}{reason}, but no exception was thrown.", expectedType);
 
-            exception.Should().BeOfType(expectedType, because, becauseArgs);
+            if (success)
```
jnyrup, line 121: We seem to have the same bug in `ThrowExactly<TException>`.
Wanna fix that as well?

docs/_pages/releases.md
```diff
@@ -23,6 +23,7 @@ sidebar:
 * `BeEquivalentTo` can again compare a non-generic `IDictionary` with a generic one - [#2358](https://github.com/fluentassertions/fluentassertions/pull/2358)
 * Fixed that the `FormattingOptions` were not respected in inner `AssertionScope` - [#2329](https://github.com/fluentassertions/fluentassertions/pull/2329)
 * Capitalize `true` and `false` in failure messages and make them formattable to a custom `BooleanFormatter` - [#2390](https://github.com/fluentassertions/fluentassertions/pull/2390), [#2393](https://github.com/fluentassertions/fluentassertions/pull/2393)
+* Fixed that an assertion on an exception type doesn't continue asserting the concrete type when wrapped into an assertion scope and the exception is `null` - [#2395](https://github.com/fluentassertions/fluentassertions/pull/2390), [#2395](https://github.com/fluentassertions/fluentassertions/pull/2395)
```
jnyrup, line 26: These release notes are meant to be at functional level, so I try formulate the sentences such that they are understandable to an FA user that has no knowledge of the FA internals.
The exception to this guideline is the "for extensions" section where it again makes sense to mention the the deep level APIs.

```suggestion
* Improved the failure message for `ThrowExactly[Async]` when wrapped in an `AssertionScope` and no exception is thrown - [#2398](https://github.com/fluentassertions/fluentassertions/pull/2398)

File: Src/FluentAssertions/Specialized/DelegateAssertions.cs (hunk 1 above)
Comment (IT-VBFK) [reply Src/FluentAssertions/Specialized/DelegateAssertions.cs:121]: Sure :)

**Review by jnyrup — APPROVED** [review]

**Comment by IT-VBFK** [comment]
Out of curiosity: Why is the assertion class for synchronous functions called `DelegateAssertions` and for asynchronous functions `AsyncFunctionAssertions`?
And why do the `FunctionAssertions` exist? This class (on first glance) does nothing special compared to `DelegateAssertions`.

What about aligning them with naming (and inheritance hierarchy respectively)?

**Comment by jnyrup** [comment]
> Out of curiosity: Why is the assertion class for synchronous functions called `DelegateAssertions` and for asynchronous functions `AsyncFunctionAssertions`? And why do the `FunctionAssertions` exist? This class (on first glance) does nothing special compared to `DelegateAssertions`.

I think it boils down to compatibility and avoiding/minimizing breaking changes.

> What about aligning them with naming (and inheritance hierarchy respectively)?

What would the benefits be contra what breaking changes it would bring?

**Comment by IT-VBFK** [comment]
> What would the benefits be contra what breaking changes it would bring?

I don't know :) Was just curious about it.

File: docs/_pages/releases.md (hunk 2 above)
Comment (IT-VBFK) [reply docs/_pages/releases.md]: You are completely right :)

And.. if someone want details about what exactly was the problem... there is a link beside :D

**Comment by dennisdoomen** [comment]
> Out of curiosity: Why is the assertion class for synchronous functions called `DelegateAssertions` and for asynchronous functions `AsyncFunctionAssertions`? 

Because _delegate_ is a good name to represent both `Action` and `Func`

> And why do the `FunctionAssertions` exist? This class (on first glance) does nothing special compared to `DelegateAssertions`.

It probably could be moved to the base-class, but it doesn't hurt.

File: Tests/FluentAssertions.Specs/Specialized/DelegateAssertionSpecs.cs [hunk 3]
```diff
@@ -34,4 +36,36 @@ public void When_injecting_a_null_clock_it_should_throw()
...
+        a.Should().ThrowExactly<InvalidOperationException>();
+        var failures = scope.Discard();
+
+        // Assert
+        failures.Should().ContainSingle()
+            .Which.Should().Match("*InvalidOperationException*no exception*");
+    }
+
+    [Fact]
+    public async Task If_an_async_function_is_supposed_to_throw_an_exact_type_of_an_exception_while_wrapped_into_an_assertion_scope_it_do_not_proceed_when_the_exception_is_null()
```
dennisdoomen, line 57: 🔧 This test belongs in `TaskAssertionSpecs`. This class uses nested classes, which makes it possible to keep the test names shorter

Tests/FluentAssertions.Specs/Specialized/DelegateAssertionSpecs.cs
```diff
@@ -34,4 +36,36 @@ public void When_injecting_a_null_clock_it_should_throw()
...
+    {
+        // Arrange
+        var a = async () => await Task.Delay(1);
```
dennisdoomen, line 60: 🙃 `async` is not needed here.

Tests/FluentAssertions.Specs/Specialized/DelegateAssertionSpecs.cs
```diff
@@ -34,4 +36,36 @@ public void When_injecting_a_null_clock_it_should_throw()
...
+    [Fact]
+    public void If_a_function_is_supposed_to_throw_an_exact_type_of_an_exception_while_wrapped_into_an_assertion_scope_it_do_not_proceed_when_the_exception_is_null()
+    {
+        // Arrange
+        var a = () => { };
+
+        // Act
+        using var scope = new AssertionScope();
+        a.Should().ThrowExactly<InvalidOperationException>();
+        var failures = scope.Discard();
```
dennisdoomen, line 49: 🤔 I would move the call to `Discard` in the _assert_ part

dennisdoomen (changes requested):

Tests/FluentAssertions.Specs/Specialized/DelegateAssertionSpecs.cs
IT-VBFK: Are you sure?

The build fails if omitted?

dennisdoomen (approved):

PR #2402, opened 2023-10-22
Remove unecessary call to `Subject.Should()`

jnyrup:
Unless `Subject.Should()` resolves to a another overload of `Should` than the initial `sut.Should()`, calling `Subject.Should().Foo()` instead of `Foo()` directly should not add any value, but on the negative side lengthen the stack trace in case the assertion fails.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -1845,7 +1845,7 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> : Referenc
 
         using (var scope = new AssertionScope())
         {
-            Subject.Should().BeEquivalentTo(unexpected, config);
+            BeEquivalentTo(unexpected, config);
 
             failures = scope.Discard();
         }
--- a/Src/FluentAssertions/Primitives/ObjectAssertions.cs
+++ b/Src/FluentAssertions/Primitives/ObjectAssertions.cs
@@ -361,7 +361,7 @@ public class ObjectAssertions<TSubject, TAssertions> : ReferenceTypeAssertions<T
 
         using (var scope = new AssertionScope())
         {
-            Subject.Should().BeEquivalentTo(unexpected, config);
+            BeEquivalentTo(unexpected, config);
             hasMismatches = scope.Discard().Length > 0;
         }
 
--- a/Src/FluentAssertions/Primitives/StringAssertions.cs
+++ b/Src/FluentAssertions/Primitives/StringAssertions.cs
@@ -143,7 +143,7 @@ public class StringAssertions<TAssertions> : ReferenceTypeAssertions<string, TAs
 
         using (var scope = new AssertionScope())
         {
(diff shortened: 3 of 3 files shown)


Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -1845,7 +1845,7 @@ public AndConstraint<TAssertions> NotBeEmpty(string because = "", params object[
 
         using (var scope = new AssertionScope())
         {
-            Subject.Should().BeEquivalentTo(unexpected, config);
+            BeEquivalentTo(unexpected, config);
```
IT-VBFK, line 1848: What about adding methods called `SubjectShouldXXXX(...)` which just calls `XXX(...)`? It's just for the fluent readability 😉

jnyrup: Or... 😉 
```cs
Subject.Should().Subject.Should().Subject.Should().Subject.Should().Foo();
```

IT-VBFK: Ok.. got it ;)

IT-VBFK: Tounge-Twister ;)

dennisdoomen (commented):
I see what you are trying to do, but it doesn't read very fluent to me.

jnyrup:
> I see what you are trying to do, but it doesn't read very fluent to me.

One can hardly disagree that `Subject.Should().Foo()` reads closer to prose than `Foo()`.

To get some numbers on the table, this pattern gives 134 results
```sh
grep -rInE --include=*.cs '^ *(return|=>)? *[A-Z][A-Za-z]+\(.*because, becauseArgs' | wc -l
```

The vast majority are simple one-liner forwards to another overload.
When filtering them out, left is a handful of places like these two examples.

https://github.com/fluentassertions/fluentassertions/blob/52622a034062e65dbf981970c97c6ab773ca4fe3/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs#L127-L136

https://github.com/fluentassertions/fluentassertions/blob/52622a034062e65dbf981970c97c6ab773ca4fe3/Src/FluentAssertions/NumericAssertionsExtensions.cs#L848-L867

Another pattern we use around a dozen places is to have non-public methods having `Assert` or `Validate` in their names.

Combined with #2403 the four ones calling `Implement(interfaceType, because, becauseArgs)` could be named `AssertSubjectImplements(interfaceType, because, becauseArgs)`

IT-VBFK:
> ...could be named AssertSubjectImplements(interfaceType, because, becauseArgs)

Is this a third proposal (for #2403)? :)

dennisdoomen:
> Combined with #2403 the four ones calling `Implement(interfaceType, because, becauseArgs)` could be named `AssertSubjectImplements(interfaceType, because, becauseArgs)`

Funny that this is exactly what I just suggested for #2403.

dennisdoomen (approved):

PR #2403, opened 2023-10-22
Guard methods against assertion scope `[Not]HaveExplicit(Property|Method)`

### LINKED ISSUE #2400: Methods are not guarded against being wrapped in an `AssertionScope`
### Description

When doing #2399 I found that these four methods in `TypeAssertions` are not guarded against being wrapped in an `AssertionScope`.

* `HaveExplicitProperty`
* `NotHaveExplicitProperty`
* `HaveExplicitMethod`
* `NotHaveExplicitMethod`

See how we call `Should().Implement()` and the make the next assertion without any guards.
https://github.com/fluentassertions/fluentassertions/blob/f4c3d56d5524996f6dc6bd02024fce88f55941d9/Src/FluentAssertions/Types/TypeAssertions.cs#L976-L984

### Reproduction Steps

```cs
var act = () =>
{
    typeof(int).Should().HaveExplicitProperty(typeof(IExplicitInterface), "Foo");
};

    .WithMessage("Expected type System.Int32 to*implement IExplicitInterface.Foo, but it does not.");
```

### Expected behavior

pass

### Actual behavior

```
Expected exception message to match the equivalent of "Expected type System.Int32 to*implement IExplicitInterface.Foo, but it does not.", but "Expected type System.Int32 to implement interface FluentAssertions.Specs.Types.IExplicitInterface, but it does not.
Expected System.Int32 to explicitly implement FluentAssertions.Specs.Types.IExplicitInterface.Foo, but it does not.
" does not.
```

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

Yes, please assign this issue to me.

IT-VBFK:
Fixes #2400 

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Types/TypeAssertions.cs
+++ b/Src/FluentAssertions/Types/TypeAssertions.cs
@@ -1124,14 +1124,17 @@ public class TypeAssertions : ReferenceTypeAssertions<Type, TypeAssertions>
         {
             success = AssertSubjectImplements(interfaceType, because, becauseArgs);
 
-            var explicitlyImplementsMethod = Subject.HasMethod($"{interfaceType}.{name}", parameterTypes);
+            if (success)
+            {
+                var explicitlyImplementsMethod = Subject.HasMethod($"{interfaceType}.{name}", parameterTypes);
 
-            Execute.Assertion
-                .BecauseOf(because, becauseArgs)
-                .ForCondition(explicitlyImplementsMethod)
-                .FailWith(
-                    $"Expected {Subject} to explicitly implement {interfaceType}.{name}" +
-                    $"({GetParameterString(parameterTypes)}){{reason}}, but it does not.");
+                Execute.Assertion
+                    .BecauseOf(because, becauseArgs)
+                    .ForCondition(explicitlyImplementsMethod)
+                    .FailWith(
+                        $"Expected {Subject} to explicitly implement {interfaceType}.{name}" +
+                        $"({GetParameterString(parameterTypes)}){{reason}}, but it does not.");
+            }
         }
 
         return new AndConstraint<TypeAssertions>(this);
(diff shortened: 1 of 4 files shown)


Src/FluentAssertions/Types/TypeAssertions.cs
```diff
@@ -462,18 +462,7 @@ public AndConstraint<TypeAssertions> Implement(Type interfaceType, string becaus
...
-            .BecauseOf(because, becauseArgs)
-            .WithExpectation("Expected type {0} to implement interface {1}{reason}", Subject, interfaceType)
-            .ForCondition(interfaceType.IsInterface)
-            .FailWith(", but {0} is not an interface.", interfaceType)
-            .Then
-            .ForCondition(containsInterface)
-            .FailWith(", but it does not.")
-            .Then
-            .ClearExpectation();
+        ImplementInternal(interfaceType, because, becauseArgs);
```
IT-VBFK, line 465: This is just a suggestion.. could most definitely be a better name...

`SubjectImplenents(...)` ?
`DoesImplement(...)` ?

jnyrup (approved):

Src/FluentAssertions/Types/TypeAssertions.cs
IT-VBFK: I have called it `SubjectImplenents()`

Src/FluentAssertions/Types/TypeAssertions.cs
```diff
@@ -495,6 +484,22 @@ public AndConstraint<TypeAssertions> Implement<TInterface>(string because = "",
         return Implement(typeof(TInterface), because, becauseArgs);
     }
 
+    private bool SubjectImplements(Type interfaceType, string because = "", params object[] becauseArgs)
```
dennisdoomen, line 487: 🔧 We prefer to have a method always follow its invocations (like a book). So it should be moved below the last member that uses it.
🤔 Since this method actually throws under normal circumstances, I think naming it some like `AssertSubjectImplements`

Src/FluentAssertions/Types/TypeAssertions.cs
```diff
@@ -495,6 +484,22 @@ public AndConstraint<TypeAssertions> Implement<TInterface>(string because = "",
...
+    {
+        bool containsInterface = interfaceType.IsAssignableFrom(Subject) && interfaceType != Subject;
```
dennisdoomen, line 489: 🤔 Why not use that condition inline, inside the `ForCondition`?

dennisdoomen (changes requested):

dennisdoomen (approved):

IT-VBFK:
Label `bug`? 😉

Issue #2364, opened 2023-10-05
New string-specific options for string equivalency assertions


vbreuss:
#### Initial discussion in #2339 and #2064

I would like to add some overloads to most string equivalency assertions to tackle various use cases (e.g. ignore white space, ignore casing, ignore newline, etc.) with the goal to tackle most existing use cases and to allow for extensibility of the provided mechanism. In order to achieve this, I would add an overload with a `Func<EquivalencyAssertionOptions<string>, EquivalencyAssertionOptions<string>>` to specify additional options on the string comparison like ignoring leading or trailing white space or newlines to the following string assertions:
- BeEquivalentTo
- ContainEquivalentOf
- EndWithEquivalentOf
- MatchEquivalentOf
- NotBeEquivalentTo
- NotContainEquivalentOf
- NotEndWithEquivalentOf
- NotMatchEquivalentOf
- NotStartWithEquivalentOf
- StartWithEquivalentOf

This also allows specifying an `IEqualityComparer<string>` as requested in #2339 via the [`.Using` method on the assertion options](https://github.com/fluentassertions/fluentassertions/blob/75aa13e9e813b677a93dd950a93a41d32526b4a6/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs#L526).

### API Proposal

Example for `BeEquivalentTo`:
```C#
    public AndConstraint<TAssertions> BeEquivalentTo(string expected,
        Func<EquivalencyAssertionOptions<string>, EquivalencyAssertionOptions<string>> config,
        string because = "", params object[] becauseArgs)
    {
      // ...
    }

    public AndConstraint<TAssertions> NotBeEquivalentTo(string unexpected,
        Func<EquivalencyAssertionOptions<string>, EquivalencyAssertionOptions<string>> config,
        string because = "", params object[] becauseArgs)
    {
      // ...
    }
```

and add the following methods to the `EquivalencyAssertionOptions<T>` class:

```C#

    public EquivalencyAssertionOptions<T> IgnoringLeadingWhitespace()
    {
        // ...
        return this;
    }

    public EquivalencyAssertionOptions<T> IgnoringTrailingWhitespace()
    {
        // ...
        return this;
    }

    public EquivalencyAssertionOptions<T> IgnoringCase()
    {
        // ...
        return this;
    }
```

### API Usage

```C#
subject.Should().BeEquivalentTo(expected, o => o
    .IgnoringLeadingWhitespace()
[...]

PR #2372, opened 2023-10-09
Allow `IEqualityComparer` in string `(Not)BeEquivalentTo` assertion

vbreuss:
Add an overload for `StringAssertions.BeEquivalentTo` and `StringAssertions.NotBeEquivalentTo` that accepts a `IEqualityComparer<string>`.

Details are discussed in #2364.
Fixes #2339.

## IMPORTANT 

* [ ] If the PR touches the public API, the changes have been approved in a separate issue with the "api-approved" label.
* [x] The code complies with the [Coding Guidelines for C#](https://www.csharpcodingguidelines.com/).
* [x] The changes are covered by unit tests which follow the Arrange-Act-Assert syntax and the naming conventions such as is used [in these tests](../tree/develop/Tests/FluentAssertions.Equivalency.Specs/MemberMatchingSpecs.cs#L51-L430).
* [x] If the PR adds a feature or fixes a bug, please update [the release notes](../tree/develop/docs/_pages/releases.md) with a functional description that explains what the change means to consumers of this library, which are published on the [website](https://fluentassertions.com/releases).
* [x] If the PR changes the public API the changes needs to be included by running [AcceptApiChanges.ps1](../tree/develop/AcceptApiChanges.ps1) or [AcceptApiChanges.sh](../tree/develop/AcceptApiChanges.sh).
* [ ] If the PR affects [the documentation](../tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).
    * [ ] Please also run `./build.sh --target spellcheck` or `.\build.ps1 --target spellcheck` before pushing and check the good outcome

Issue #2339, opened 2023-09-28
Allow IEqualityComparer in string equality assertion


bart-vmware:
### Background and motivation

I'm trying to assert two strings are equal using a custom comparer. The code below works, though it would be nice if the cast to `object` wouldn't be necessary and `Be()` could be used instead of `BeEquivalentTo()`.

```c#
using System.Text.RegularExpressions;
using FluentAssertions;

string logLines = @"one\ntwo\nthree";
string expected = @"one\r\ntwo\r\nthree";

((object)logLines).Should().BeEquivalentTo(expected, options => options.Using(IgnoreLineEndingsComparer.Instance));

public sealed class IgnoreLineEndingsComparer : IEqualityComparer<string>
{
    public static readonly IgnoreLineEndingsComparer Instance = new();

    public bool Equals(string x, string y)
    {
        if (x == y)
        {
            return true;
        }

        if (x == null || y == null)
        {
            return false;
        }

        var xLines = Regex.Split(x, "\r\n|\r|\n");
        var yLines = Regex.Split(x, "\r\n|\r|\n");

        return xLines.SequenceEqual(yLines);
    }

    public int GetHashCode(string obj)
    {
        return obj.GetHashCode();
    }
}
```

### API Proposal

```C#
public class StringAssertions<TAssertions> : ReferenceTypeAssertions<string, TAssertions>
    where TAssertions : StringAssertions<TAssertions>
{
    public AndConstraint<TAssertions> Be(string expected, IEqualityComparer<string> comparer,
        string because = "", params object[] becauseArgs);
}
```

### API Usage

```C#
using System.Text.RegularExpressions;
using FluentAssertions;

string logLines = @"one\ntwo\nthree";
string expected = @"one\r\ntwo\r\nthree";

logLines.Should().Be(expected, IgnoreLineEndingsComparer.Instance);
```

### Are you willing to help with a proof-of-concept (as PR in that or a separate repo) first and as pull-request later on?

No

jnyrup:
For the provided example you should be able to use the existing:
```cs
logLines.Should().MatchEquivalentOf(expected);
```

Note: the `BeEquivalentTo` assertion method for `object` and `string` works differently as `StringAssertions.BeEquivalentTo` is case insensitive, while `ObjectAssertions.BeEquivalentTo` boils down to `StringAssertions.Be`.

With v6 you can e.g. write these extension method.
```cs
public static class StringAssertionsExtensions
{
[...]

PR #1312, opened 2020-04-18
Optimize IndexOfFirstMismatch

jnyrup:
`StringAssertions.Be[EquivalentTo]` uses an `StringEqualityValidator`
and in there `IndexOfFirstMismatch` to compare if two `string`s are
equal/equivalent.

It had annoyed me for some time that we compared strings characterwise by
slicing of `char`s and converting them to single character `string`s
that we then compare.

The benchmark is calling `IndexOfFirstMismatch` two identical strings using
`Ordinal`/`OrdinalIgnoreCase` and for different lengths.

```
|                  Method |    N |        Comparison | Character |         Mean |      Error |     StdDev | Ratio | RatioSD |   Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------ |----- |------------------ |---------- |-------------:|-----------:|-----------:|------:|--------:|--------:|------:|------:|----------:|
|    IndexOfFirstMismatch |   10 |           Ordinal |         A |    155.13 ns |   3.061 ns |   2.863 ns |  1.00 |    0.00 |  0.1528 |     - |     - |     480 B |
| IndexOfFirstMismatchNew |   10 |           Ordinal |         A |     25.48 ns |   0.439 ns |   0.367 ns |  0.16 |    0.00 |       - |     - |     - |         - |
|                         |      |                   |           |              |            |            |       |         |         |       |       |           |
|    IndexOfFirstMismatch |   10 |           Ordinal |         a |    153.99 ns |   2.850 ns |   2.527 ns |  1.00 |    0.00 |  0.1528 |     - |     - |     480 B |
| IndexOfFirstMismatchNew |   10 |           Ordinal |         a |     25.12 ns |   0.509 ns |   0.476 ns |  0.16 |    0.01 |       - |     - |     - |         - |
|                         |      |                   |           |              |            |            |       |         |         |       |       |           |
|    IndexOfFirstMismatch |   10 | OrdinalIgnoreCase |         A |    156.03 ns |   2.405 ns |   1.878 ns |  1.00 |    0.00 |  0.1528 |     - |     - |     480 B |
| IndexOfFirstMismatchNew |   10 | OrdinalIgnoreCase |         A |    123.76 ns |   2.462 ns |   2.303 ns |  0.79 |    0.02 |       - |     - |     - |         - |
|                         |      |                   |           |              |            |            |       |         |         |       |       |           |
[...]

PR #2413, opened 2023-10-29
Allow specifying `EquivalencyAssertionOptions` in string assertions

### LINKED ISSUE #1247: Feature request: make a version from string BeEquivalentTo which ignores \n <> \r\n
### Description
The current`BeEquivalentTo()` method asserts that a string is exactly the same as another string, including any leading or trailing whitespace, with the exception of the casing.

It would be very useful to add a new method or extend this method to support also `\n` and `\r\n`.

```csharp
// Arrange
string expected= "foo\r\n";

// Act
string result = "foo\n";

// Assert
result.Should().BeEquivalentTo(expected);
```

### Expected behavior:
Assertion is OK

### Actual behavior:
Assertion FAILS

### Versions
* Latest version

vbreuss:
Enhance the `EquivalencyAssertionOptions` to support the following methods:
- **IgnoringLeadingWhitespace**
   Instructs the comparison to ignore leading whitespace when comparing strings.
- **IgnoringTrailingWhitespace**
   Instructs the comparison to ignore trailing whitespace when comparing strings.
- **IgnoringNewlines**
   Instructs the comparison to ignore newlines when comparing strings.
- **IgnoringCase**
   Instructs the comparison to compare strings case-insensitive.
- 

Add an overload for the following string assertions that accept a `Func<EquivalencyAssertionOptions<string>, EquivalencyAssertionOptions<string>> config`.
- Be
- BeEquivalentTo
- ContainEquivalentOf
- EndWithEquivalentOf
- MatchEquivalentOf
- NotBe
- NotBeEquivalentTo
- NotContainEquivalentOf
- NotEndWithEquivalentOf
- NotMatchEquivalentOf
- NotStartWithEquivalentOf
- StartWithEquivalentOf

Details are discussed in #2364.
Fixes #1247.

## IMPORTANT 


vbreuss:
If this pull request is accepted, it replaces #2372

Src/FluentAssertions/Common/StringExtensions.cs
```diff
@@ -9,15 +10,13 @@ internal static class StringExtensions
...
     /// </summary>
-    public static int IndexOfFirstMismatch(this string value, string expected, StringComparison stringComparison)
+    public static int IndexOfFirstMismatch(this string value, string expected, IEqualityComparer<string> comparer)
     {
-        Func<char, char, bool> comparer = GetCharComparer(stringComparison);
-
         for (int index = 0; index < value.Length; index++)
         {
-            if (index >= expected.Length || !comparer(value[index], expected[index]))
+            if (index >= expected.Length || !comparer.Equals(value[index..(index + 1)], expected[index..(index + 1)]))
```
jnyrup, line 19: Just for reference:
I made this optimization in #1312 but since this is only called on the failure path I'm fine with loosing it again.

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -110,18 +144,53 @@ public AndConstraint<TAssertions> BeOneOf(IEnumerable<string> validValues, strin
...
+        Guard.ThrowIfArgumentIsNull(config);
+
+        EquivalencyAssertionOptions<string> options = config(AssertionOptions.CloneDefaults<string>());
+
+        var expectation = new StringValidator(
+            new StringEqualityStrategy(options.GetStringComparerOrDefault()),
+            because, becauseArgs);
+
+        var subject = options.ApplyStringSettings(Subject);
+        expected = options.ApplyStringSettings(expected);
```
jnyrup, line 188: Won't be indices in the failure message be wrong now?

Src/FluentAssertions/Common/StringExtensions.cs
vbreuss: Unfortunately I see no other way, when using `IEqualityComparer<string>`, as it has no option to compare single characters.

Src/FluentAssertions/Primitives/StringAssertions.cs
vbreuss: That's a very good question. I am unsure, what the expected behaviour should be, when I say "IgnoringLeadingWhitespace".
Do I expect the index to ignore it too or not?

I think you can argue both ways, so I would settle for the easier solution and leave it as it is.

But it would probably a good idea to explicitely mention it in the docs of `IgnoringLeadingWhitespace`...

vbreuss:
@dennisdoomen :
I don't understand why the Qodana Scan fails. Is it a problem with the pipeline or with the code changes? Can I see somewhare, what the actual failures are (the [raw build log](https://github.com/fluentassertions/fluentassertions/actions/runs/6692083023/job/18180551853?pr=2413) is not really helpful for me)?

dennisdoomen:
> I don't understand why the Qodana Scan fails. Is it a problem with the pipeline or with the code changes? Can I see somewhare, what the actual failures are (the [raw build log](https://github.com/fluentassertions/fluentassertions/actions/runs/6692083023/job/18180551853?pr=2413) is not really helpful for me)?

Is that because you're branch is behind and needs to rebase?

vbreuss:
> Is that because you're branch is behind and needs to rebase?

I doubt it, as the build ran and the failure was before the merge of #2416 to `develop`.
Nonetheless I rebased the branch to get rid of the conflicting files.

Src/FluentAssertions/Primitives/StringContainsStrategy.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    private readonly IEqualityComparer<string> comparer;
+    private readonly OccurrenceConstraint occurrenceConstraint;
+    private readonly bool ignoringCase;
+
+    public StringContainsStrategy(IEqualityComparer<string> comparer, OccurrenceConstraint occurrenceConstraint)
+    {
+        this.comparer = comparer;
+        this.occurrenceConstraint = occurrenceConstraint;
+        ignoringCase = comparer.Equals("A", "a");
```
dennisdoomen, line 17: 🤔 I get a strong feeling we should not take an `IEqualityComparer<string>`, but just use a `StringComparer`.

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.BeEquivalentTo.cs
```diff
@@ -158,6 +158,50 @@ public void Default_immutable_lists_should_be_equivalent()
             // Act / Assert
             collection.Should().BeEquivalentTo(collection1);
         }
+
+        [Fact]
+        public void Succeed_for_collection_with_case_different_strings_when_IgnoringCase()
```
dennisdoomen, line 163:
```suggestion
        public void Can_ignore_casing_while_comparing_collections_of_strings()
```

We definitely _do not_ want to include names of code elements in test names. We also prefer to avoid terms like _when_ and _should_. Same for the other tests.

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.BeEquivalentTo.cs
```diff
@@ -158,6 +158,50 @@ public void Default_immutable_lists_should_be_equivalent()
...
+        {
+            // Arrange
+            var actual = new[] { "first", "test", "last" };
+            var expect = new[] { "first", "TEST", "last" };
```
dennisdoomen, line 167: 🙃 We prefer to name this the `expectation`

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.cs
```diff
@@ -23,4 +24,30 @@ public void When_chaining_multiple_assertions_it_should_assert_all_conditions()
             .Contain(substring).And
             .HaveLength(length);
     }
+
+    private sealed class MatchingEqualityComparer : IEqualityComparer<string>
```
dennisdoomen, line 28:
```suggestion
    private sealed class AlwaysMatchingEqualityComparer : IEqualityComparer<string>
```

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.cs
```diff
@@ -23,4 +24,30 @@ public void When_chaining_multiple_assertions_it_should_assert_all_conditions()
...
+            return true;
+        }
+
+        public int GetHashCode(string obj)
+        {
+            return obj.GetHashCode();
+        }
+    }
+
+    private sealed class NotMatchingEqualityComparer : IEqualityComparer<string>
```
dennisdoomen, line 41:
```suggestion
    private sealed class NeverMatchingEqualityComparer : IEqualityComparer<string>
```

dennisdoomen (changes requested):
That's a pretty impressive PR. It just needs a little cleaning up here and there.

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -54,14 +55,47 @@ public StringAssertions(string value)
...
+    /// The equivalency options.
+    /// </param>
+    /// <param name="because">
+    /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+    /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+    /// </param>
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+    /// </param>
+    public AndConstraint<TAssertions> Be(string expected,
```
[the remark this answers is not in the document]
vbreuss: No, that's no longer the case. As @jnyrup commented [here](https://github.com/fluentassertions/fluentassertions/issues/2364#issuecomment-1793423504), I removed the overload for `Be` in [d126fae](https://github.com/fluentassertions/fluentassertions/pull/2413/commits/d126faece902149dade45d70f489c306019fdaab).

Src/FluentAssertions/Primitives/StringContainsStrategy.cs
vbreuss: What would be the advantage, when I don't use any other property or method than the one from the interface?

dennisdoomen: 1. It would avoid the type check in `Using`
2. It would avoid the weird logic to determine whether compare is case insensitive.

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.BeEquivalentTo.cs
vbreuss: Done!

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.cs
vbreuss: Done!

Tests/FluentAssertions.Specs/Collections/CollectionAssertionSpecs.BeEquivalentTo.cs
vbreuss: Done!

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.cs
vbreuss: Done!

Src/FluentAssertions/Primitives/StringContainsStrategy.cs
vbreuss: > 1. It would avoid the type check in `Using`

Yes, I could do this, but I could do this as well with an explicit additional overload for `IEqualityComparer<string>`. Would this really be better?

> 2. It would avoid the weird logic to determine whether compare is case insensitive.
How so? Also a `StringComparer` does not tell me, if it is case-sensitive or not? I could always create a custom `StringComparer` and would then need to check in the strategy how it handles case-sensitivity.

docs/_pages/releases.md
```diff
@@ -15,6 +15,7 @@ sidebar:
 
 ### Improvements
 * Improve failure message for string assertions when checking for equality - [#2307](https://github.com/fluentassertions/fluentassertions/pull/2307)
+* Allow specifying `EquivalencyAssertionOptions` in string assertions - [#2413](https://github.com/fluentassertions/fluentassertions/pull/2413)
```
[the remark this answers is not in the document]
vbreuss: I added both changes to the releases.md.

Regarding the update of the docs, you mean adding a paragraph to [strings.md](https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/strings.md)?
Should this also be part of this pull request or a separate one?

Src/FluentAssertions/Primitives/StringContainsStrategy.cs
dennisdoomen: > How so? Also a `StringComparer` does not tell me, if it is case-sensitive or not? I could always create a custom `StringComparer` and would then need to check in the strategy how it handles case-sensitivity.

Than you can check which `StringComparer` it is, e.g. `StringComparer.StringContainsStrategy`

vbreuss: I can still write custom `StringComparer` and would have the same issue with it. Additionally it would no longer solve the issue in #2339 (or would require the user to write a custom `StringComparer` with two additional methods).

I only need this information (if it is case sensitive or not) for the `ExpectationDescription`, e.g. in `StringContainsStrategy`:
```csharp
    string predicateDescription = ignoringCase ? "contain the equivalent of" : "contain";
    return "Expected {context:string} to " + predicateDescription + " ";
```

What if I always print out "contain the equivalent of" when calling `Should().ContainEquivalentOf`?

docs/_pages/releases.md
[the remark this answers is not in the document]
vbreuss: I added a paragraph to `strings.md` in [5b84065](https://github.com/fluentassertions/fluentassertions/pull/2413/commits/5b840659e03c6f928d36d7519afb0d4301b236b5).

Src/FluentAssertions/Primitives/StringContainsStrategy.cs
dennisdoomen: Ah yes, you're right. Ignore this comment. However, we do need to add an overload of `SelfReferenceEquivalencyOptions<TSelf>.Using`

vbreuss: I got rid of the strange `comparer.Equals("A", "a")` in [42bb1d2c](https://github.com/fluentassertions/fluentassertions/pull/2413/commits/42bb1d2c766d024c81d46341dea9036c1cfffc2a).

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyAssertionOptions.cs
```diff
@@ -527,6 +544,11 @@ public TSelf Using<T>(IEqualityComparer<T> comparer)
     {
         userEquivalencySteps.Insert(0, new EqualityComparerEquivalencyStep<T>(comparer));
 
+        if (comparer is IEqualityComparer<string> c)
```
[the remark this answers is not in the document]
vbreuss: I added an explicit overload for `IEqualityComparer<string>` in [9575859](https://github.com/fluentassertions/fluentassertions/pull/2413/commits/957585952b26bb7da683d5bb0c07d94ea330357a).

Src/FluentAssertions/Primitives/StringContainsStrategy.cs
dennisdoomen: We try to avoid boolean parameters. And since it feels a bit weird how that boolean is used inside the strategies, why not replace the boolean with the actual string that it results in?

vbreuss: Great idea. I changed it accordingly!

Tests/FluentAssertions.Equivalency.Specs/CollectionSpecs.cs
```diff
@@ -1177,8 +1177,8 @@ public void When_some_string_in_the_collection_is_not_equal_to_the_expected_stri
 
         // Assert
         action.Should().Throw<XunitException>().WithMessage(
-            "Expected subject[1]*to be \"one\", but \"two\" differs near \"two\" (index 0).*" +
-            "Expected subject[2]*to be \"one\", but \"six\" differs near \"six\" (index 0).*");
+            "Expected subject[1]*to be equivalent to \"one\", but \"two\" differs near \"two\" (index 0).*" +
```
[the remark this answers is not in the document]
vbreuss: Fixed it. The reason was, that the `StringEqualityEquivalencyStep` used `BeEquivalentTo`.

Src/FluentAssertions/Equivalency/Execution/CollectionMemberOptionsDecorator.cs
```diff
@@ -20,26 +20,14 @@ public CollectionMemberOptionsDecorator(IEquivalencyOptions inner)
     }
 
     public IEnumerable<IMemberSelectionRule> SelectionRules
-    {
-        get
-        {
-            return inner.SelectionRules.Select(rule => new CollectionMemberSelectionRuleDecorator(rule)).ToArray();
-        }
-    }
+        => inner.SelectionRules.Select(rule => new CollectionMemberSelectionRuleDecorator(rule)).ToArray();
```
dennisdoomen, line 23: 🤔 Was it Qodana? If so, we should disable that rule. I don't think it's always better to use expression-bodied members.

dennisdoomen (approved):
Nice work

IT-VBFK:
@vbreuss Can you try `behavior` instead of `behaviour` in string.md and check if that silence the spell check?
And if so, probably you could add the british version to the custom dict. ;)

vbreuss:
> @vbreuss Can you try `behavior` instead of `behaviour` in string.md and check if that silence the spell check? And if so, probably you could add the british version to the custom dict. ;)

Yes, changing it to `behavior` does indeed fix the spell check error.
image

Sorry, I was not aware, that I need to run it locally! 😃 

@IT-VBFK : Where can I find the custom dictionary?

Src/FluentAssertions/Equivalency/Execution/CollectionMemberOptionsDecorator.cs
vbreuss: Yes, I get this notice from Qodana:
image

IT-VBFK:
in the repository root `cSpell.json` -> words :)

dennisdoomen:
> And if so, probably you could add the british version to the custom dict. ;)

It's not completely obvious, but we're using US English as our standard.

Src/FluentAssertions/Equivalency/Execution/CollectionMemberOptionsDecorator.cs
dennisdoomen: Let's disable it then. I'm fine if you do that in this PR.

vbreuss: > Let's disable it then. I'm fine if you do that in this PR.

What do you mean, @dennisdoomen?
1. Keep the refactorings regarding expression body syntax in the PR or
2. Disable the Qodana rule as part of this PR?

If you mean the second point, I am not sure where the rule would be disabled. Can you point me in the direction?

dennisdoomen: I prefer 2. This needs to happen in the `qodana.yml`

vbreuss: I disabled the rule [`ArrangeAccessorOwnerBody`](https://www.jetbrains.com/help/rider/ArrangeAccessorOwnerBody.html) and undid the refactoring in [CollectionMemberOptionsDecorator.cs](https://github.com/fluentassertions/fluentassertions/pull/2413/files/aff14810149d3017523d0a9faf6677f9be747133#diff-32d01e9ed629abd7e7461b4bd2283efe9267baaca071a6a519d1fb358c3991cb).

ITaluone:
I think the PR title is litte outdated? Should this go to "Allow specifying `EquivalencyOptions` in string assertions" instead?

Src/FluentAssertions/Primitives/StringAssertions.cs
vbreuss: I mentioned the behaviour in the XML-Doc for `IgnoringLeadingWhitespace`.

dennisdoomen:
What's the status of this PR?

vbreuss:
> What's the status of this PR?

From my side it is complete, I am waiting for reviews. I am not sure, what the issue is with the Qodana scan. Should I rebase and try again?

dennisdoomen:
> I am not sure, what the issue is with the Qodana scan. Should I rebase and try again?

Some changes were made to the Qodana scanner since the last time your PR got build. Although it's weird that your PR didn't get rebuild.

vbreuss:
> Some changes were made to the Qodana scanner since the last time your PR got build. Although it's weird that your PR didn't get rebuilt.

I merged the latest `develop` branch and now all seems fine 😄

vbreuss:
> I also requeued the build so that wasn't necessary. Our builds build the (virtual) result of merging your PR into the target branch.

@dennisdoomen :
I saw that you manually triggered the Qodana build, but it failed again, so it seems the automatic GitHub merge was incorrect or incomplete. After the manual merge, the build was fine...

> Either way, we never merge `develop` or `master` in our feature branches. We _always_ use rebases.

I was not aware of this guideline. The build is now also successful after a rebase 😄 
Maybe we could add this information to the [DOs and DON'Ts](https://github.com/fluentassertions/fluentassertions/blob/develop/CONTRIBUTING.md#dos-and-donts)?

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -821,19 +1101,19 @@ public AndConstraint<TAssertions> NotEndWith(string unexpected, string because =
...
-                .BecauseOf(because, becauseArgs)
-                .FailWith("Expected {context:string} {0} not to end with {1}{reason}.", Subject, unexpected);
+            Subject.Should().EndWith(unexpected);
+            notEquivalent = scope.Discard().Length > 0;
         }
 
+        Execute.Assertion
+            .ForCondition(Subject != null && notEquivalent)
+            .BecauseOf(because, becauseArgs)
+            .FailWith("Expected {context:string} not to end with {0}{reason}, but found {1}.", unexpected, Subject);
```
jnyrup, line 1115: I agree that this phrasing is better than the "that does not" 👍 
Will you also adjust the four remaining occurrences of "that does"?

jnyrup (commented):
Just two small things.
very impressive work ❤️

Src/FluentAssertions/Primitives/StringAssertions.cs
vbreuss: You are right, that would be more consistent. I replaced the five remaining occurences of "that does not" with "not to".

jnyrup (approved):

bart-vmware:
I don't understand how the linked PR that closes this issue solves my problem. It doesn't fix the next bug, which I pointed out earlier:
```c#
public static string RemoveNewLines(this string @this)
{
    return @this.Replace("\n", string.Empty, StringComparison.Ordinal)
        .Replace("\r", string.Empty, StringComparison.Ordinal)
        .Replace(@"\r\n", string.Empty, StringComparison.Ordinal);
}
```
Note the `@` verbatim string modifier, which is equivalent to matching `"\\r\\n"` instead of `"\r\n"`.

Aside from that, how do these changes affect my use case? In other words, what FA extension method should I be using now to pass my custom string comparer (without also ignoring leading/trailing whitespace, line breaks, casing, etc)?

vbreuss:
@bart-vmware :
You can now specify the `IEqualityComparer<string>` that should be used, e.g. with the mentioned `IgnoreLineEndingsComparer` in #2339:
```csharp
subject.Should().BeEquivalentTo(expected, o => o
    .Using(IgnoreLineEndingsComparer.Instance));
```

> I don't understand how the linked PR that closes this issue solves my problem. It doesn't fix the next bug, which I pointed out earlier
Or do you mean another issue?

bart-vmware:
@vbreuss Thanks, but that doesn't actually work. Tracked at #2566.

PR #2415, opened 2023-10-29
Add `releases.md` to solution items

vbreuss:
With each change it is necessary to also update the `docs/_pages/releases.md` file.
I added it to the solution items, so that it can be updated directly from within the IDE.

## IMPORTANT 


IT-VBFK:
At least in Rider you can switch to "folder view". I think this should also be available in VS?

vbreuss:
> At least in Rider you can switch to "folder view". I think this should also be available in VS?

Yes, it is possible, but it is an unnecessary manual step (and unloads the solution) and I always have to look/find the correct directory, as the `releases.md` is not in the root. Especially for beginners, I think this is easier, if it can be found directly in the solution.

jnyrup:
My gut reaction is that we shouldn't put a subset of the docs in the solution items 🤔 

But it could be nice to have the readable changelog in the root folder, I often look for a `changelog.md` or `releases.md` when surfing GitHub repositories.

Perhaps we could put the pure markdown in `/` and include it from `/docs/_pages/releases.md`?
Note: the fix in #2181 would when have to done otherwise to avoid having mixing liquid template and markdown.

Having such a `/releases.md` included in "solutions items" (somehow) seems more reasonable to me.

dennisdoomen:
> Perhaps we could put the pure markdown in `/` and include it from `/docs/_pages/releases.md`?
> Note: the fix in #2181 would when have to done otherwise to avoid having mixing liquid template and markdown.

I like that. Let's do that.

vbreuss:
@dennisdoomen 
I tried working on this, and it seems to work locally, but I am not sure, how it will work on the actual homepage.

However if the approach is not how you want to do it, I can also abandon the PR...

IT-VBFK:
Did you miss a check in `_build/Build.cs` (`HasDocumentationChanges`)

dennisdoomen:
> However if the approach is not how you want to do it, I can also abandon the PR...

I do, but I don't think it'll work. Github's default [build workflow](https://github.com/fluentassertions/fluentassertions/actions/runs/6252023597) will only publish the contents under the `docs` folder. To change that, we would need to create [a custom GHA workflow](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow), but I have no experience with that.

jnyrup:
As I recall GHA also doesn't support extra ruby gems.
[`jekyll-remote-include`](https://github.com/netrics/jekyll-remote-include) has also been archived.

I was thinking if we could hack this by adding a symlink from somewhere inside `docs` to `/RELEASES.md`, but that doesn't seem easy on Windows.

vbreuss:
What about adding a build step, that copies the file to the target location before deploying it?

jnyrup:
> What about adding a build step, that copies the file to the target location before deploying it?

Sounds reasonable to me.
Curious how it would be done.

vbreuss:
> Curious how it would be done.

I pushed a solution that works locally, when you first copy the `RELEASE.md` from the root directory to the `docs/_pages` directory.
This step would have to be automated by creating a [custom github pages workflow](https://docs.github.com/en/pages/getting-started-with-github-pages/using-custom-workflows-with-github-pages) in the [GitHub Pages settings](https://github.com/fluentassertions/fluentassertions/settings/pages).

The default "GitHub Pages Jekyll" workflow looks like the following:
```yaml
# Sample workflow for building and deploying a Jekyll site to GitHub Pages
name: Deploy Jekyll with GitHub Pages dependencies preinstalled

on:
  # Runs on pushes targeting the default branch
  push:
    branches: ["main"]

  # Allows you to run this workflow manually from the Actions tab
  workflow_dispatch:

# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
permissions:
  contents: read
  pages: write
  id-token: write

# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
concurrency:
  group: "pages"
  cancel-in-progress: false

jobs:
  # Build job
  build:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3
      - name: Setup Pages
        uses: actions/configure-pages@v3
      - name: Build with Jekyll
        uses: actions/jekyll-build-pages@v1
        with:
          source: ./
          destination: ./_site
      - name: Upload artifact
        uses: actions/upload-pages-artifact@v2

  # Deployment job
  deploy:
    environment:
      name: github-pages
      url: ${{ steps.deployment.outputs.page_url }}
    runs-on: ubuntu-latest
    needs: build
    steps:
      - name: Deploy to GitHub Pages
        id: deployment
        uses: actions/deploy-pages@v2

```

In the default workflow we would then have to add a manual copy step of the file, e.g. 
```yaml
      - name: Copy the RELEASE.md
        run: |
          cp "./RELEASE.md" "./docs/_pages/RELEASE.md"
```

**As I am no maintainer, someone with the corresponding rights would have to create the custom github pages workflow first.**

dennisdoomen:
> What about adding a build step, that copies the file to the target location before deploying it?

That's exactly what I had in mind.

dennisdoomen:
> As I am no maintainer, someone with the corresponding rights would have to create the custom github pages workflow first.

You can do that yourself, but you'll need to target `master` since that's where the build is being triggered from.

RELEASE.md
```diff
@@ -0,0 +1,562 @@
...
+* Added native support for `XDocument`, `XElement` and `XAttribute` properties and fields to `BeEquivalentTo` - [#1572](https://github.com/fluentassertions/fluentassertions/pull/1572)
+* The equivalency failure message will include information on how tuples, anonymous types, records and other types are compared - [#1571](https://github.com/fluentassertions/fluentassertions/pull/1571)
+* Improved formatting a dictionary when key or value is a complex type - [#1577](https://github.com/fluentassertions/fluentassertions/pull/1577).
+* Added `NotBe(string)` for symmetry with `Be(string)` for GuidAssertions - [#1597](https://github.com/fluentassertions/fluentassertions/pull/1597).
+* Added `BeOneOf` for enum assertions - [#1637](https://github.com/fluentassertions/fluentassertions/pull/1637)
+* Homogenized assertion message formatting of predicate expressions - [#1619](https://github.com/fluentassertions/fluentassertions/pull/1619).
+
+### Fixes
+
+* Reported actual value when it contained {% raw %}`{{{{` or `}}}}`{% endraw %} - [#1234](https://github.com/fluentassertions/fluentassertions/pull/1234).
```
jnyrup, line 299: To make it valid markdown and not interfere with liquid templates
```suggestion
* Reported actual value when it contained four consecutive `{` or `}` - [#1234](https://github.com/fluentassertions/fluentassertions/pull/1234).
```

vbreuss:
> You can do that yourself, but you'll need to target `master` since that's where the build is being triggered from.

I don't understand how you think that could work. Currently there is no explicit workflow file, that I could adapt, even in the `master` branch. For this to work, you would first have to change the settings to use the "GitHub Actions" source and create an initial workflow file.
Only afterwards would I be able to update this workflow file.

dennisdoomen:
> I don't understand how you think that could work. Currently there is no explicit workflow file, that I could adapt, even in the `master` branch. For this to work, you would first have to change the settings to use the "GitHub Actions" source and create an initial workflow file.
> Only afterwards would I be able to update this workflow file.

Correct. That's exactly what would be needed to make that work.

ITaluone:
> In the default workflow we would then have to add a manual copy step of the file, e.g.
> ```yml
> - name: Copy the RELEASE.md
>       run: |
>         cp "./RELEASE.md" "./docs/_pages/RELEASE.md"
> ```
     
I think we should keep the file name casing.. On windows it makes no difference, but unix is case-sensitive on file/directory names.

dennisdoomen:
Unless you're willing to recreate the workflow that Github Pages is using, I don't think there's any point continuing this PR.

ITaluone:
@vbreuss I think you can add the custom pages workflow file whether this project has it enabled or not.

vbreuss:
> I think you can add the custom pages workflow file whether this project has it enabled or not.

I would not know how and would not know how to test it (I never worked with Github Pages before).

@ITaluone
Do you know how to create a custom pages workflow file?
Would you be willing to participate in this pull request?

*Otherwise I would close the PR...*

vbreuss:
I don't think that I am the right person to continue pursuing this PR. I will leave the branch active for some time, if someone else want's to continue with this approach, but close the PR.

dennisdoomen:
> Yes, it is possible in VS, but it is an unnecessary manual step (and unloads the solution) and I always have to look/find the correct directory, as the `releases.md` is not in the root. Especially for beginners, I think this is easier, if it can be found directly in the solution.

Why not add that file as a link to the solution items? Then you can easily edit it.

vbreuss:
> Why not add that file as a link to the solution items? Then you can easily edit it.

That was my very first commit in this pull request, but then it got out of hand.

As we had quite some discussion on an alternative solution here, I created a new pull request (#2469) which only adds the releases to the solution items.

PR #2437, opened 2023-11-04
Correct Qodana `failThreshold`

IT-VBFK:
Actually I don't know if `protected` would satisfy the qodana rule..

Let's try...

## IMPORTANT 

### DIFF
--- a/qodana.yaml
+++ b/qodana.yaml
@@ -1,7 +1,7 @@
 version: "1.0"
 
 linter: jetbrains/qodana-dotnet:latest
-failThreshold: 13
+failThreshold: 19
 
 dotnet:
   solution: FluentAssertions.sln


qodana.yaml
```diff
@@ -1,7 +1,7 @@
 version: "1.0"
 
 linter: jetbrains/qodana-dotnet:latest
-failThreshold: 13
+failThreshold: 15
```
IT-VBFK, line 4: This is going up, because we have a few errors for `.netstandard2.0` and `.netstandard2.1`.

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyOptions.cs
```diff
@@ -245,7 +245,7 @@ public TSelf IncludingInternalFields()
     /// <remarks>
     /// This does not preclude use of `Including`.
     /// </remarks>
-    public TSelf ExcludingFields()
+    protected TSelf ExcludingFields()
```
dennisdoomen, line 248: This will make the functionality unavailable and thus isn't something we want.

IT-VBFK: Unavailable for what? No test failed and class is `abstract`?

jnyrup: Just because we haven't invoked it directly from our own unit tests doesn't imply that no-one is using it.

To silence analyzers we could perhaps instead write a test that directly invokes this method.

IT-VBFK: I have meant that, because `SelReferenceEquivalencyOptions` is abstract you have to inherit from it and therefore you can access this. Am I missing something? 🤔

jnyrup: The case where the derived class is cast to the abstract class.
[SharpLab](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK4wD4CEIGcCmABBIQLyEz4DuhAIvslAG74AmAFAJQDcAsAFAB9MsQB0ANQgAbVPj78BEYLgTIIAYwSEAAgCZCOAgIDeAwucIAHZAHsE+TW0KwtkmUWOEA5vgTdCAL4CQQr8enQMzE4gBnj4JgFAA=)

```cs
Base a = new Derived();
_ = a.Value; // error CS0122: 'Base.Value' is inaccessible due to its protection level

abstract class Base
{
    protected int Value { get; }
}

class Derived : Base
{}
```

IT-VBFK: oh.. I see 👍 

What's also interesting: qodana does not complain about the other `public` methods in this class.
Can this rule be disabled only for this file?

dennisdoomen: We always use the `EquivalencyOptions` in `BeEquivalentTo` and we just forgot to add tests that use this method.

Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyOptions.cs
```diff
@@ -384,7 +384,7 @@ public Restriction<TProperty> Using<TProperty>(Action<IAssertionContext<TPropert
     /// <remarks>
     /// This is the default behavior. You can override this using <see cref="ExcludingNestedObjects"/>.
     /// </remarks>
-    public TSelf IncludingNestedObjects()
+    protected TSelf IncludingNestedObjects()
```
dennisdoomen, line 387: 🔧 Same comment here. It's a public API.

dennisdoomen (changes requested):

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -1231,6 +1311,30 @@ public void When_excluding_virtual_or_abstract_property_exclusion_works_properly
...
+
+            object class2 = new ClassWithSomeFieldsAndProperties
+            {
+                Property1 = "sit",
+                Property2 = "amet",
+                Property3 = "consectetur"
+            };
+
+            // Act / Assert
+            class1.Should().BeEquivalentTo(class2, opts => opts.ExcludingFields().RespectingRuntimeTypes());
```
dennisdoomen, line 1336: 🔧 Even if you would not exclude the fields, this test would still succeed. 
🙃 We prefer to use terms like `subject` and `expectation`

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -602,6 +602,86 @@ public void When_both_field_and_properties_are_configured_for_inclusion_both_sho
...
+                },
+                B = "D"
+            };
+
+            // Act / Assert
+            subject.Should().BeEquivalentTo(expectation, opts => opts.IncludingNestedObjects());
+        }
+
+        [Fact]
+        public void Taking_nested_objects_into_account_throws()
```
dennisdoomen, line 658: 🤔 It's not that it throws because nested objects are included, but because the nested objects differ. By the way, this the default.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -602,6 +602,86 @@ public void When_both_field_and_properties_are_configured_for_inclusion_both_sho
...
+
+            // Act
+            Action act = () => subject.Should().BeEquivalentTo(expected, opts => opts.IncludingInternalFields());
+
+            // Assert
+            act.Should().Throw<XunitException>();
+        }
+
+        [Fact]
+        public void Taking_nested_objects_into_account()
```
dennisdoomen, line 632: 🤔 Does what?

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
```diff
@@ -602,6 +602,86 @@ public void When_both_field_and_properties_are_configured_for_inclusion_both_sho
             // Assert
             act.Should().Throw<XunitException>().Which.Message.Should().Contain("Field1").And.Contain("Property1");
         }
+
+        [Fact]
+        public void Taking_internal_fields_into_consideration()
+        {
+            // Arrage
+            var subject = new ClassWithAllAccessModifiersForMembers("Lorem", "protectedValue", "ipsum", "protectedInternalValue", "privateValue", "privateProtectedValue");
```
dennisdoomen, line 610: 🔧 I find this example too hard to read. I'd prefer that you introduce a dedicated test class directly under the tests.

dennisdoomen (changes requested):

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
IT-VBFK: I don't get it. If I remove the `ExcludeProperties()` the test fails.
Am I missing something?

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
IT-VBFK: Hmm... true

If this is the default.. how can I construct a test case then?

What comes to my mind first: 
Run two assertions one with excluded and one with included and check if they are the same. 🤔

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
dennisdoomen: Hmm, I was convinced fields were excluded by default, but it turns out I was wrong. Sorry about that.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
dennisdoomen: Yeah, something like that. Although I agree it looks a bit weird. Maybe it's enough to change the title.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.cs
IT-VBFK: No problem :)

IT-VBFK:
@jnyrup @dennisdoomen What do you think? Are the test from this PR obsolete? 

IMHO they are, but it's up to you keeping or removing them.

dennisdoomen:
> What do you think? Are the tests from this PR obsolete?

According to the coverage report, they don't seem to add value.

eNeRGy164:
Not only branch coverage can be a reason for a test, mutation coverage can also be a factor.

ITaluone:
> Not only branch coverage can be a reason for a test, mutation coverage can also be a factor.

Note, that the whole reason for this PR was to silence Qodana.

IT-VBFK:
> > What do you think? Are the tests from this PR obsolete?
> 
> 
> 
> According to the coverage report, they don't seem to add value.

Ok, I will remove them

dennisdoomen (approved):

PR #1416, opened 2020-10-27
Order by multiple properties sequentially

Serg046:
## IMPORTANT 

* [x] The code complies with the [Coding Guidelines for C#](https://www.csharpcodingguidelines.com/).
* [x] The changes are covered by a new or existing set of unit tests which follow the Arrange-Act-Assert syntax such as is used [in this example](https://github.com/fluentassertions/fluentassertions/blob/daaf35b9b59b622c96d0c034e8972a020b2bee55/Tests/FluentAssertions.Shared.Specs/BasicEquivalencySpecs.cs#L33).
* [x] If the contribution adds a feature or fixes a bug, please update [**the release notes**](https://github.com/fluentassertions/fluentassertions/tree/master/docs), which are published on the [website](https://fluentassertions.com/releases).
* [x] If the contribution changes the public API the changes needs to be included by running `AcceptApiChanges.ps1`/`AcceptApiChanges.sh`.
* [ ] If the contribution affects [the documentation](https://github.com/fluentassertions/fluentassertions/tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).

Issue #1160

PR #2439, opened 2023-11-05
Simplify inheritance of `SubsequentOrderingAssertions`

vbreuss:
The `SubsequentOrderingGenericCollectionAssertions<TCollection, T>` is only used as a base class for `SubsequentOrderingAssertions`. By moving the functionality into the (previously empty) `SubsequentOrderingAssertions` the inheritance tree can be simplified.

This should also increase the code coverage according to #1823.

## IMPORTANT 


Src/FluentAssertions/Collections/SubsequentOrderingAssertions.cs
```diff
@@ -6,7 +6,7 @@ namespace FluentAssertions.Collections;
 
 [DebuggerNonUserCode]
 public class SubsequentOrderingAssertions<T>
-    : SubsequentOrderingGenericCollectionAssertions<IEnumerable<T>, T, SubsequentOrderingAssertions<T>>
+    : SubsequentOrderingGenericCollectionAssertions<IEnumerable<T>, T>
```
dennisdoomen, line 9: 🤔 I wonder why we even need `SubsequentOrderingAssertions` _and_ a `SubsequentOrderingGenericCollectionAssertions`. Maybe it's a weird usage of generic types that won't be apparent until you remove it.
🏕️ There's another `SubsequentOrderingGenericCollectionAssertions` that is also not used. Maybe we can remove that one as well.

vbreuss: You are right. I moved the code from `SubsequentOrderingGenericCollectionAssertions` to `SubsequentOrderingAssertions` and could completely delete the `SubsequentOrderingGenericCollectionAssertions` class without breaking any tests.

dennisdoomen (approved):

jnyrup:
I don't know or recall why #1416 used `SubsequentOrderingAssertions` in `GenericCollectionAssertions` instead of `SubsequentOrderingGenericCollectionAssertions`.
The latter retains `TCollection` whereas the former uses `IEnumerable<T>` which `TCollection` is constrained to.

Could we use `SubsequentOrderingGenericCollectionAssertions` instead of `SubsequentOrderingAssertions`?

vbreuss:
> Could we use `SubsequentOrderingGenericCollectionAssertions` instead of `SubsequentOrderingAssertions`?

I don't have any preference regarding the naming. I was simply wondering why it needed so many levels of inheritance and could find no reason for it. Should I rename `SubsequentOrderingAssertions` to `SubsequentOrderingGenericCollectionAssertions`?

jnyrup:
> > Could we use `SubsequentOrderingGenericCollectionAssertions` instead of `SubsequentOrderingAssertions`?
> 
> I don't have any preference regarding the naming. I was simply wondering why it needed so many levels of inheritance and could find no reason for it. Should I rename `SubsequentOrderingAssertions` to `SubsequentOrderingGenericCollectionAssertions`?

Not sure I get you.
I didn't mean rename `SubsequentOrderingAssertions` to `SubsequentOrderingGenericCollectionAssertions`, but let the relevant assertions in `GenericCollectionAssertions` return `SubsequentOrderingGenericCollectionAssertions` instead of `SubsequentOrderingAssertions`.

vbreuss:
> Not sure I get you.
> I didn't mean rename `SubsequentOrderingAssertions` to `SubsequentOrderingGenericCollectionAssertions`, but let the relevant assertions in `GenericCollectionAssertions` return `SubsequentOrderingGenericCollectionAssertions` instead of `SubsequentOrderingAssertions`.

That would practically be the same. All methods were in one class and all other classes were empty. I chose `SubsequentOrderingAssertions` to keep (move all logic inside and remove the inheritance). Alternatively I could keep the logic in `SubsequentOrderingGenericCollectionAssertions` and replace all references of `SubsequentOrderingAssertions` with it.

I expected the first change to have less chance of breaking things, as I don't change the return type of existing methods. That was my reason for choosing this way...

dennisdoomen:
@jnyrup I think this is waiting for you, isn't it?

ITaluone:
What is the state of this one?

vbreuss:
I'm not sure. It's more a refactoring change missing a review from @jnyrup.

I can rebase the branch, or abandon the PR.

dennisdoomen:
Please consider rebasing on `develop`

vbreuss:
> Please consider rebasing on `develop`

Oh, wow, you finished your big PR 😊

vbreuss:
@dennisdoomen : I am not sure, why the Qodana build failed, but it seems not to be related with my changes...

dennisdoomen:
> I am not sure, why the Qodana build failed, but it seems not to be related with my changes...

Something is wrong with the commit SHA we pass. Somebody at JetBrains is getting some confirmation for me.

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
```diff
@@ -553,20 +553,9 @@ namespace FluentAssertions.Collections
         public FluentAssertions.AndConstraint<TAssertions> Equal(params string[] expected) { }
         public FluentAssertions.AndConstraint<TAssertions> NotContainMatch(string wildcardPattern, string because = "", params object[] becauseArgs) { }
     }
-    public class SubsequentOrderingAssertions<T> : FluentAssertions.Collections.SubsequentOrderingGenericCollectionAssertions<System.Collections.Generic.IEnumerable<T>, T, FluentAssertions.Collections.SubsequentOrderingAssertions<T>>
+    public class SubsequentOrderingAssertions<T> : FluentAssertions.Collections.GenericCollectionAssertions<System.Collections.Generic.IEnumerable<T>, T>
```
dennisdoomen, line 556: 🔧Since you've also changed the public contract, please also update the `releases.md` file.

dennisdoomen (changes requested):

PR #2448, opened 2023-11-09
Avoid possible 'null' assignment in `CallerIdentifier`

vbreuss:
Fix issue detected by ReSharper:
- Possible 'null' assignment to non-nullable entity

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/CallerIdentifier.cs
+++ b/Src/FluentAssertions/CallerIdentifier.cs
@@ -27,9 +27,7 @@ public static class CallerIdentifier
         {
             var stack = new StackTrace(fNeedFileInfo: true);
 
-            var allStackFrames = stack.GetFrames()
-                .Where(frame => frame is not null && !IsCompilerServices(frame))
-                .ToArray();
+            var allStackFrames = GetFrames(stack);
 
             int searchStart = allStackFrames.Length - 1;
 
@@ -83,9 +81,7 @@ public static class CallerIdentifier
         {
             var stack = new StackTrace();
 
-            var allStackFrames = stack.GetFrames()
-                .Where(frame => frame is not null && !IsCompilerServices(frame))
-                .ToArray();
+            var allStackFrames = GetFrames(stack);
 
             int firstUserCodeFrameIndex = 0;
 
@@ -116,9 +112,7 @@ public static class CallerIdentifier
 
     internal static bool OnlyOneFluentAssertionScopeOnCallStack()
     {
-        var allStackFrames = new StackTrace().GetFrames()
-            .Where(frame => frame is not null && !IsCompilerServices(frame))
-            .ToArray();
+        var allStackFrames = GetFrames(new StackTrace());
 
         int firstNonFluentAssertionsStackFrameIndex = Array.FindIndex(
             allStackFrames,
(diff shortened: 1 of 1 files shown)


Src/FluentAssertions/CallerIdentifier.cs
```diff
@@ -267,4 +261,11 @@ private static bool IsBooleanLiteral(string candidate)
     {
         return candidate is "true" or "false";
     }
+
+    private static StackFrame[] GetStackFrameArray(StackTrace stack)
```
dennisdoomen, line 265: 🔧 Let's keep the type out of the name

dennisdoomen (approved):
🙃 (nitpick) I'd prefer more functional commit titles.

vbreuss:
> 🙃 (nitpick) I'd prefer more functional commit titles.

Do you mean the pull request title (that will be visible in the commit history)?
If so, I could split this PR in three different pull requests, with a better title...

dennisdoomen:
> I split the pull request in two. This one now only updates the `CallerIdentifier` and the consolidation of namespace and file location is moved to #2449.

It could have been two separate commits. But all commits and PR titles need to be as functional as possible.

Src/FluentAssertions/CallerIdentifier.cs
```diff
@@ -267,4 +261,18 @@ private static bool IsBooleanLiteral(string candidate)
     {
         return candidate is "true" or "false";
     }
+
+    private static StackFrame[] GetFrames(StackTrace stack)
+    {
+        var frames = stack.GetFrames();
+#if !NET6_0_OR_GREATER
```
jnyrup, line 268: I had to look up why this might be necessary for .NET Framework.

[net48](https://referencesource.microsoft.com/#mscorlib/system/diagnostics/stacktrace.cs,601)
[net core 3.0+](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Diagnostics/StackTrace.cs,165)
[change](https://github.com/dotnet/coreclr/pull/25303/files#diff-cf1bd43d732ab10d08dc2126f73384119656794951312c072eff4b21c34c3bebL165-R165)

jnyrup (approved):

Src/FluentAssertions/CallerIdentifier.cs
ITaluone: Why not just `#if NET47`? I find this easier to understand 🤷‍♂️

vbreuss: Because `NETSTANDARD` also is affected...

ITaluone: Oh.. I see :)

dennisdoomen:
What about the Qodana issues?

vbreuss:
> What about the Qodana issues?

The two issues in the comment are from an older run before [commit 2caaf39](https://github.com/fluentassertions/fluentassertions/pull/2448/commits/2caaf394459263e7367b88a3ee34a62f4179871e).
That's why I introduced the `#if !NET6_0_OR_GREATER` condition.

I don't know, why the comment was not updated 😕

PR #2449, opened 2023-11-09
Consolidate namespace and file location

vbreuss:
Fix ReSharper issue: "[Namespace does not correspond to file location](https://www.jetbrains.com/help/resharper/CheckNamespace.html)"

## IMPORTANT 


dennisdoomen (approved):

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
```diff
@@ -6,7 +6,7 @@
 using FluentAssertions.Execution;
 using FluentAssertions.Primitives;
 
-namespace FluentAssertions.Reflection;
+namespace FluentAssertions.Specialized;
```
jnyrup, line 9: <rant>
Breaking changes are not justified by an analyzer following the "namespace must equal folder" dogma.
Folder structure and namespaces are related but orthogonal matters.
</rant>

With that said, all other `[ReflectionFoo]Assertions` types are located in `Types/` so let's move `AssemblyAssertions` over there and adjust its namespace to `FluentAssertions.Types`.

jnyrup (changes requested):

Src/FluentAssertions/Polyfill/StringBuilderExtensions.cs
```diff
@@ -1,5 +1,8 @@
-﻿using System.Collections.Generic;
+﻿#if NET47 || NETSTANDARD2_0 || NETSTANDARD2_1
```
jnyrup, line 1: FYI: these files are excluded from net6 and netstandard21 in 
https://github.com/fluentassertions/fluentassertions/blob/9d0daeea0cb24ef1c48c9c8f6386d5fd53549893/Src/FluentAssertions/FluentAssertions.csproj#L58C17-L59

https://github.com/fluentassertions/fluentassertions/blob/9d0daeea0cb24ef1c48c9c8f6386d5fd53549893/Src/FluentAssertions/FluentAssertions.csproj#L69C10-L69C10

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
IT-VBFK: V7 will be full ob breaking changes? I think this few make not much difference, do they? 😉

vbreuss:
@jnyrup :
I agree with your sentiment, that analyzers should be taken with a grain of salt, but they normally indicate a good spot to have a better look. If the decision is that the code is correct, I try to disable the analyzer in this case, so that it is clear that this was intentional.

I now moved the `AssemblyAssertions` class to the `Types/` directory and adjusted the namespace accordingly.
Additionally I kept the namespace in the two polyfill classes `SystemExtensions` and `StringBuilderExtensions` and moved them to a separate directory. This way it is clear, that these are intended as polyfill and they can be excluded from analyzers.

Src/FluentAssertions/Polyfill/StringBuilderExtensions.cs
vbreuss: Ah, I was looking, how this worked before without excluding .NET6 😃

Src/FluentAssertions/Specialized/AssemblyAssertions.cs
jnyrup: I think every breaking change should be justified.
In any case it's a good practice justify why something users rely on should suddenly change.
I found it useful to watch some of .NET's API [reviews](https://www.youtube.com/@NETFoundation/search?query=review) to see the amount of detail they put into designing the .NET APIs.

The Linux kernel has the famous "Don't break userspace" mantra.
For .NET 6, the BCL had 25 documented [breaking changes](https://learn.microsoft.com/en-us/dotnet/core/compatibility/6.0#core-net-libraries) and these includes behavioral changes.
The [BinaryFormatter Obsoletion Strategy](https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md#timeline-overview) is a 4 year progress
We're not the Linux Kernel or .NET BCL, but take e.g. `Newtonsoft.Json` which is used virtually everywhere.
Especially with `packages.config` I recall having problems with needing to manually bump `Newtonsoft.Json` to resolve diamond dependency problems.
Luckily James Newton-King seems to have taken breaking changes very seriously, so for my ~8 years of .NET development I don't recall having problems just picking the newest version.
We're lucky that we're (hopefully) only referenced from test libraries, which limits the amount of dependency entanglement we're part of.

If one sees "don't break things" as the default promise, Semantic Versioning is a "relaxation" of that promise (whatever that means) to a model where breaking changes are permitted, but can only be introduced during major versions.
That doesn't imply that a major version should seek to break as much as possible.

I imagine to myself a major bump as a sort of apology for breaking stuff.
"I promised not tell anyone your deep secret, but because I adhere to Semantic Versioning, I bumped the major to relax that promise and told everyone your secret anyways."

See e.g. this great [talk](https://www.youtube.com/watch?v=oyLBGkS5ICk) by Rich Hickey.
(I don't agree with everything, but he has some really great perspectives)

The funny ones:
https://www.hyrumslaw.com/
https://xkcd.com/1172/

jnyrup (approved):

Src/FluentAssertions/Polyfill/StringBuilderExtensions.cs
IT-VBFK: Should this then be:

```suggestion
#if NET47 || NETSTANDARD2_0
```

vbreuss: No, the `StringBuilderExtensions` were only excluded from .NET 6.

dennisdoomen (approved):

Issue #2253, opened 2023-08-13
Revisit AssertionScope and how it travels over chained calls


dennisdoomen:
Although I'm still thinking about the design, I propose the following changes:

* `AssertionScope` is only used by consumers as they do now _and_ by certain APIs such as `BeEquivalentTo`
* Every `Should` method will create a new object called `Assertion` (but I'm open to a better name) that contains the context of the assertion and a link to the scope. When no ambient `AssertionScope` is available, it will be a new one that is created at that point. This is different from v6 where we create a new scope per call to `Execute.Assertion`. 
* Every `Should` method will also take an extra parameter annotated with `[CallerArgumentExpression]`. If this returns a non-`null` value, we can use that instead of using the `CallerIdentifier`.
* This `Assertion` instance will be explicitly passed to all the assertion classes (like `StringAssertions`). 
* No public APIs will depend on `AssertionScope.Current` anymore
* Most of the APIs that are now part of `AssertionScope` will need to move to `Assertion`.
* This should allow methods like `ContainSingle` to update the `Assertion` instance with extra information to solve #1502 
* This should also allow us to drop `ClearExpectations`
* Also provides a nice split between the responsibility of `AssertionScope` as a way to group related assertions and its responsibilities as the internal assertion API

Corniel:
Do you have some (pseudo) code in mind on how the body of a certain assert method should look like?

dennisdoomen:
I'm going to do a POC soon to proof this idea to myself ;-)

jnyrup:
Perhaps related to this idea: 
Use `[CallerArgumentExpression]` to retrieve the name of the subject at compile-time and populate this `Assertion` object with that.
The idea is to remove the need for `CallerIdentifier`, which is both slow and fragile.

dennisdoomen:
Yeah, that could work indeed. And in case it doesn't work (.NET 6 SDK not installed and using .NET Framework 4.7), we can fall back to the `CallerIdentifier`.

dennisdoomen:
The [first attempt](https://github.com/dennisdoomen/fluentassertions/tree/Spike/RevisitAssertionScope) where `Which` would return a `WhichResult` that includes both the result as well as the current assertion failed for a couple of reasons. But sometimes we directly call methods or access properties on the result of `Which`, which means the type of what `Which` returns must be `TElement`. So trying to pass a compound object from one `Should` to another through a special `AndWhichConstraint` was more difficult than I thought.

Issue #1502, opened 2021-03-02
Misleading caller identity when `Which` fails


jnyrup:
### Description

When chaining into an assertion using `Which` the failure message can be confusing, when the inner assertion fails, as the determined caller identity is from the outer assertion.

### Complete minimal example reproducing the issue

```csharp
var languages = new[] { "C#" };

languages.Should().ContainSingle()
    .Which.Should().EndWith("script");
```

### Expected behavior:

Some nice message about "C#" not ending with "script".

### Actual behavior:

```
Expected languages to end with "script", but "C#" is too short.
```

### Versions

* latest version of develop: ee0aa439c542cbab2d83cdb41ea65849585114d0

dennisdoomen:
Yeah, also noticed this while preparing the demo video.

sdelarosbil:
I tried to look into this and it seems complex.

You would have to somehow have a way for the call identifier's parser have the ability to look ahead and see if we don't have a Which.Should call down the line which seems difficult to detect right on top of the identifier being dependant on the current assertion context.

The later seems doable, but I am really not sure how you'd tackle the former. Unless I am missing something, this doesn't seem easy to address.

dennisdoomen:
Or `Which` should leave some clue to the caller identification code that it should not try to get the caller.

dennisdoomen:
Assuming the following example:

> ```cs
> var languages = new[] { "C#" };
> 
> languages.Should().ContainSingle()
>     .Which.Should().EndWith("script");
> ```

I would expect a failing `EndWith` to throw the appropriate exception with

`Expected languages[0] to end with "script", but "C#" is too short.` 

or

`Expected languages.Single() to end with "script", but "C#" is too short.`

Either way, we need some way to understand what `ContainSingle().Which` will do. I can think of two solutions:

1. Make the `CallerIdentifier` smart enough so that it knows what the various methods like `ContainSingle` do and how that affects the identifier. 
2. Give the `ContainSingle` the means to pass along information on how to combine the variable `languages` extracted by `CallerIdentifier` with something specific like `[0]` or `Single()`. 

I don't like option 1, but option 2 most definitely will require a breaking change. I'm thinking of a way where each `Should` call will create an `AssertionScope` (either a new one, or one connected to the ambient scope) and pass that on to the assertion class and its assertion methods. In that way, a method like `ContainSingle` can update the scope with the relevant context, which is then passed on to the chained `Should`.

comecme:
[...]

PR #2452, opened 2023-11-10
Spike to revisit AssertionScope

dennisdoomen:
* `AssertionScope` is only used by consumers as they do now _and_ by certain APIs such as `BeEquivalentTo`
* Every `Should` method will create a new object called `Assertion` (but I'm open to a better name) that contains the context of the assertion and a link to the scope. When no ambient `AssertionScope` is available, it will be a new one that is created at that point. This is different from v6 where we create a new scope per call to `Execute.Assertion`. 
* Every `Should` method will also take an extra parameter annotated with `[CallerArgumentExpression]`. If this returns a non-`null` value, we can use that instead of using the `CallerIdentifier`.
* This `Assertion` instance will be explicitly passed to all the assertion classes (like `StringAssertions`). 
* No public APIs will depend on `AssertionScope.Current` anymore
* Most of the APIs that are now part of `AssertionScope` will need to move to `Assertion`.
* This should allow methods like `ContainSingle` to update the `Assertion` instance with extra information to solve #1502 
* This should also allow us to drop `ClearExpectations`
* Also provides a nice split between the responsibility of `AssertionScope` as a way to group related assertions and its responsibilities as the internal assertion API
#2253 

## IMPORTANT 


dennisdoomen:
This first attempt where Which would return a WhichResult that includes both the result as well as the current assertion failed for a couple of reasons. But sometimes we directly call methods or access properties on the result of Which, which means the type of what Which returns must be TElement. So trying to pass a compound object from one Should to another through a special AndWhichConstraint was more difficult than I thought.

Issue #2450, opened 2023-11-09
Fix Qodana issues


vbreuss:
### Background and motivation

The Qodana baseline scan has [currently](https://github.com/fluentassertions/fluentassertions/actions/runs/6801258831/job/18491687872) 1,049 problems

```
Qodana - Baseline summary
Analysis results: 1,049 problems detected
Grouping problems according to baseline: UNCHANGED: 1,030, NEW: 19
---------------------------------------------------------------------------------------
Name                                                 Baseline  Severity  Problems count
---------------------------------------------------------------------------------------
Possible 'System.NullReferenceException'                  NEW      High               1
Member can be made private (non-private accessibil        NEW  Moderate               9
ity)                                                                                   
Method return value is never used (non-private acc        NEW  Moderate               6
essibility)                                                                            
Member can be made protected (non-private accessib        NEW  Moderate               2
ility)                                                                                 
Class is never instantiated (non-private accessibi        NEW  Moderate               1
lity)                                                                                  
Possible 'System.InvalidOperationException'         UNCHANGED      High              87
Possible multiple enumeration                       UNCHANGED      High              83
Possible 'System.NullReferenceException'            UNCHANGED      High              36
Redundant using directive                           UNCHANGED      High              23
Possible 'null' assignment to non-nullable entity   UNCHANGED      High              10
Auto-property accessor is never used (non-private   UNCHANGED      High               4
accessibility)                                                                         
Expression is always 'true' or always 'false'       UNCHANGED      High               4
Namespace does not correspond to file location      UNCHANGED      High               4
Local function hides method                         UNCHANGED      High               3
Parameter hides member                              UNCHANGED      High               3
Co-variant array conversion                         UNCHANGED      High               2
Access to modified captured variable                UNCHANGED      High               1
Inconsistent synchronization on field               UNCHANGED      High               1
Non-readonly type member referenced in 'GetHashCod  UNCHANGED      High               1
e()'                                                                                   
Possible overflow in unchecked context              UNCHANGED      High               1
Type member is never used (private accessibility)   UNCHANGED      High               1
Use null check pattern instead of a type check suc  UNCHANGED      High               1
ceeding on any not-null value                                                          
Variable in local function hides variable from out  UNCHANGED      High               1
er scope                                                                               
Method return value is never used (non-private acc  UNCHANGED  Moderate             426
essibility)                                                                            
Member can be made private (non-private accessibil  UNCHANGED  Moderate             162
ity)                                                                                   
[...]

PR #2453, opened 2023-11-12
Avoid possible 'System.NullReferenceException' in `PredicateLambdaExpressionValueFormatter`

vbreuss:
The `ExpressionVisitor.Visit` method can accept `null` as node expression (see [here (.NET core)](https://source.dot.net/#System.Linq.Expressions/System/Linq/Expressions/ExpressionVisitor.cs,f0615a823a9efe36) or [here (.NET Framework 4.8)](https://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/ExpressionVisitor.cs,f0615a823a9efe36)).

This fixes a `Possible 'System.NullReferenceException'` in #2450 

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Formatting/PredicateLambdaExpressionValueFormatter.cs
+++ b/Src/FluentAssertions/Formatting/PredicateLambdaExpressionValueFormatter.cs
@@ -134,7 +134,7 @@ public class PredicateLambdaExpressionValueFormatter : IValueFormatter
 
         public override Expression Visit(Expression node)
         {
-            if (node?.NodeType == ExpressionType.AndAlso)
+            if (node!.NodeType == ExpressionType.AndAlso)
             {
                 var binaryExpression = (BinaryExpression)node;
                 Visit(binaryExpression.Left);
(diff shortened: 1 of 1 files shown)


dennisdoomen (approved):

dennisdoomen:
Although we would love to get test coverage, I don't care enough to bother. But maybe Jonas has a different opinion.

Src/FluentAssertions/Formatting/PredicateLambdaExpressionValueFormatter.cs
```diff
@@ -134,7 +134,7 @@ private sealed class AndOperatorChainExtractor : ExpressionVisitor
 
         public override Expression Visit(Expression node)
         {
-            if (node.NodeType == ExpressionType.AndAlso)
+            if (node?.NodeType == ExpressionType.AndAlso)
```
jnyrup, line 137: We had some discussion about this line in #2264 (be aware that the discussions are spread across multiple comment sections).

As I recall when diving into `Expressions` back then my conclusion was that one cannot create a `BinaryExpression` with a null `Left` or `Right` child. That should prevent the two recursive calls on line 140 and 141 from calling `Visit` on a null `node`.
The initial call to `Visit()` on line 62 is guarded by the conditional on line 22.

With the proposed change and if my analysis is wrong, instead of throwing an NRE here, we would add a null `node` to `AndChain` which would then lead to an NRE on line 25 when trying to invoke `ToString()` on the `null`.

I would be fine with `node!.NodeType` to silence Qodana.

vbreuss: @jnyrup : I really appreciate how thorougly you review pull requests!

I changed it as you suggested.

jnyrup (approved):

PR #2454, opened 2023-11-12
Add missing tests for ActionAssertions and DelegateAssertions

vbreuss:
Fixes `Method return value is never used (non-private accessibility)` in #2450 

## IMPORTANT 

### DIFF
--- a/Tests/FluentAssertions.Specs/Specialized/DelegateAssertionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Specialized/DelegateAssertionSpecs.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using FluentAssertions.Execution;
-using FluentAssertions.Specialized;
 using Xunit;
 
 namespace FluentAssertions.Specs.Specialized;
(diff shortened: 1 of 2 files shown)


Tests/FluentAssertions.Specs/Specialized/DelegateAssertionSpecs.cs
```diff
@@ -7,37 +7,52 @@ namespace FluentAssertions.Specs.Specialized;
...
+        [Fact]
+        public void When_injecting_a_null_extractor_it_should_throw()
+        {
+            // Arrange
+            Action subject = () => { };
 
-        // Act
-        Func<ActionAssertions> act = () => new ActionAssertions(subject, extractor, clock: null);
+            // Act
+            Func<ActionAssertions> act = () => new ActionAssertions(subject, extractor: null);
```
dennisdoomen, line 34: 🔧 `ActionAssertions` is not supposed to be used directly and thus should not be directly tests.

dennisdoomen (changes requested):

Tests/FluentAssertions.Specs/Specialized/DelegateAssertionSpecs.cs
vbreuss: I am not sure, I understand your issue. The two tests that use the `ActionAssertions` class existed previously. I only moved them to the `ThrowExactly` nested class.
Do you want me to remove them? Don't they test a valid scenario?

dennisdoomen: Those where never supposed to be tested directly. So, if we move forward with this PR, I believe we should fix that.

vbreuss: I removed the two tests.

dennisdoomen: I think we need to take a different route. Instead of deleting those tests (and reducing the coverage), we should see if we can achieve the following:
1. Hide the constructors by making them `internal`
2. Rewrite the tests so they always use `Should` 
3. Maybe mark the `AssertionExtensions` as `sealed`

What do you think @jnyrup ?

vbreuss: Please note, that the constructors of all "*Assertions" classes I checked were public, so I think it would be inconsistent to change these to internal.

Regarding rewriting the tests:
- The test checking for a `null` clock is simple to rewrite
- The second test checking for `null` on the `extractor` is more difficult. All internally used constructor calls use the static [`Extractor`](https://github.com/fluentassertions/fluentassertions/blob/develop/Src/FluentAssertions/AssertionExtensions.cs#L33) field in `AssertionExtensions` which can not be `null`.

dennisdoomen: My rationale was that `AssertionExtensions` is not supposed to be used outside the library itself. Oftentimes we keep the constructors `public` or `protected` to allow consumer to build extensions. But I can't think of an example where that would be necessary.

dennisdoomen (approved):

jnyrup (approved):

PR #2488, opened 2023-11-27
Allow anonymous object for selecting fields/properties at `Exclude` and `Include`

### LINKED ISSUE #2479: [API Proposal]: Support for selecting multiple properties in single expression.
### Background and motivation

If I write assertion this way:
```
 actualEnv.Should().BeEquivalentTo(env, _ => _.Excluding(_ => new { _.Id, _.Creator, _.LastEditor }));
```
the exception is thrown:
> System.ArgumentException : Expression <new <>f__AnonymousType0`3(Id = _.Id, Creator = _.Creator, LastEditor = _.LastEditor)> cannot be used to select a member. (Parameter 'expression')

So I have to write it slightly more verbose.
```
 actualEnv.Should().BeEquivalentTo(env, _ => _.Excluding(_ => _.Id).Excluding(_ => _.Creator).Excluding(_ => _.LastEditor));
```

### API Proposal

So, my proposal is to treat expression for anonymous object creation as an enumeration of properties to exclude.
For example EF Core uses this approach for defining composite key.

### API Usage

```
a.Shoul().BeEquivalentTo(b, _ => _.Excluding(b => new { b.Prop1, ...., b.PropN }));
```

### Alternative Designs

_No response_

### Risks

_No response_

### Are you willing to help with a proof-of-concept (as PR in that or a separate repo) first and as pull-request later on?

No

IT-VBFK:
This refers to `Excluding()` / `.For().Exclude()` and `Including()`.

This closes #2479 

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Common/ExpressionExtensions.cs
+++ b/Src/FluentAssertions/Common/ExpressionExtensions.cs
@@ -31,21 +31,22 @@ internal static class ExpressionExtensions
         (((expression.Body as UnaryExpression)?.Operand ?? expression.Body) as MemberExpression)?.Member;
 
     /// <summary>
-    /// Gets a dotted path of property names representing the property expression, including the declaring type.
+    /// Gets one or more dotted paths of property names representing the property expression, including the declaring type.
     /// </summary>
     /// <example>
-    /// E.g. Parent.Child.Sibling.Name.
+    /// E.g. ["Parent.Child.Sibling.Name"] or ["A.Dotted.Path1", "A.Dotted.Path2"].
     /// </example>
     /// <exception cref="ArgumentNullException"><paramref name="expression"/> is <see langword="null"/>.</exception>
 #pragma warning disable MA0051
-    public static MemberPath GetMemberPath<TDeclaringType, TPropertyType>(
+    public static IEnumerable<MemberPath> GetMemberPaths<TDeclaringType, TPropertyType>(
         this Expression<Func<TDeclaringType, TPropertyType>> expression)
 #pragma warning restore MA0051
     {
         Guard.ThrowIfArgumentIsNull(expression, nameof(expression), "Expected an expression, but found <null>.");
 
-        var segments = new List<string>();
-        var declaringTypes = new List<Type>();
+        string singlePath = null;
+        List<string> selectors = [];
+        List<Type> declaringTypes = [];
         Expression node = expression;
 
         while (node is not null)
@@ -68,7 +69,7 @@ internal static class ExpressionExtensions
                     var memberExpression = (MemberExpression)node;
                     node = memberExpression.Expression;
 
-                    segments.Add(memberExpression.Member.Name);
+                    singlePath = $"{memberExpression.Member.Name}.{singlePath}";
                     declaringTypes.Add(memberExpression.Member.DeclaringType);
                     break;
 
@@ -77,7 +78,7 @@ internal static class ExpressionExtensions
                     var indexExpression = (ConstantExpression)binaryExpression.Right;
                     node = binaryExpression.Left;
 
-                    segments.Add($"[{indexExpression.Value}]");
+                    singlePath = $"[{indexExpression.Value}].{singlePath}";
                     break;
 
                 case ExpressionType.Parameter:
@@ -87,13 +88,26 @@ internal static class ExpressionExtensions
                 case ExpressionType.Call:
                     var methodCallExpression = (MethodCallExpression)node;
 
-                    if (methodCallExpression is not { Method.Name: "get_Item", Arguments: [ConstantExpression argumentExpression] })
+                    if (methodCallExpression is not
+                        { Method.Name: "get_Item", Arguments: [ConstantExpression argumentExpression] })
                     {
                         throw new ArgumentException(GetUnsupportedExpressionMessage(expression.Body), nameof(expression));
                     }
 
                     node = methodCallExpression.Object;
-                    segments.Add($"[{argumentExpression.Value}]");
+                    singlePath = $"[{argumentExpression.Value}].{singlePath}";
+                    break;
+                case ExpressionType.New:
+                    var newExpression = (NewExpression)node;
+
+                    foreach (Expression member in newExpression.Arguments)
+                    {
+                        var expr = member.ToString();
+                        selectors.Add(expr[expr.IndexOf('.', StringComparison.Ordinal)..]);
+                        declaringTypes.Add(((MemberExpression)member).Member.DeclaringType);
+                    }
+
+                    node = null;
                     break;
 
                 default:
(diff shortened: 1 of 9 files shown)


IT-VBFK:
Hmm.. discovered a bad bug...

Will fix it in the evening 🫣

Src/FluentAssertions/Common/ExpressionExtensions.cs
```diff
@@ -95,21 +96,41 @@ public static MemberPath GetMemberPath<TDeclaringType, TPropertyType>(
...
                     throw new ArgumentException(GetUnsupportedExpressionMessage(expression.Body), nameof(expression));
             }
         }
 
-        // If any members were accessed in the expression, the first one found is the last member.
         Type declaringType = declaringTypes.FirstOrDefault() ?? typeof(TDeclaringType);
 
-        IEnumerable<string> reversedSegments = segments.AsEnumerable().Reverse();
-        string segmentPath = string.Join(".", reversedSegments);
+        if (selectors.Count == 0)
```
IT-VBFK: It is necessary to use a new `List` which stores multiple complete paths instead of n tokens (which results in one path).

An alternative would be to use e.g. a `Dictionary<int, List<string>>`, but I don't think this makes much better here...

Src/FluentAssertions/Common/ExpressionExtensions.cs
```diff
@@ -87,29 +88,51 @@ public static MemberPath GetMemberPath<TDeclaringType, TPropertyType>(
...
+                        { Method.Name: "get_Item", Arguments: [ConstantExpression argumentExpression] })
                     {
                         throw new ArgumentException(GetUnsupportedExpressionMessage(expression.Body), nameof(expression));
                     }
 
                     node = methodCallExpression.Object;
-                    segments.Add($"[{argumentExpression.Value}]");
+                    singlePath = $"[{argumentExpression.Value}].{singlePath}";
+                    break;
+                case ExpressionType.New:
```
dennisdoomen, line 100: ♻️ This entire method is become too bloated and needs some refactoring (but maybe not now). An implementation of the Strategy Pattern might be a better fit.

Src/FluentAssertions/Common/ExpressionExtensions.cs
```diff
@@ -87,29 +88,51 @@ public static MemberPath GetMemberPath<TDeclaringType, TPropertyType>(
...
                     throw new ArgumentException(GetUnsupportedExpressionMessage(expression.Body), nameof(expression));
             }
         }
 
-        // If any members were accessed in the expression, the first one found is the last member.
         Type declaringType = declaringTypes.FirstOrDefault() ?? typeof(TDeclaringType);
 
-        IEnumerable<string> reversedSegments = segments.AsEnumerable().Reverse();
-        string segmentPath = string.Join(".", reversedSegments);
+        if (singlePath is null)
```
dennisdoomen, line 120: ❓ What happens if `singlePath` is non-null and `selectors` isn't empty?

Src/FluentAssertions/Common/ExpressionExtensions.cs
```diff
@@ -87,29 +88,51 @@ public static MemberPath GetMemberPath<TDeclaringType, TPropertyType>(
...
+        {
+            return selectors.Select(selector =>
+                GetNewInstance<TDeclaringType>(declaringType, selector));
```
dennisdoomen, line 123: 🔧 This is dangerous as you're returning a lazy-evulated expression.

Src/FluentAssertions/Common/ExpressionExtensions.cs
```diff
@@ -87,29 +88,51 @@ public static MemberPath GetMemberPath<TDeclaringType, TPropertyType>(
...
+        }
 
-        return new MemberPath(typeof(TDeclaringType), declaringType, segmentPath.Replace(".[", "[", StringComparison.Ordinal));
+        return [GetNewInstance<TDeclaringType>(declaringType, singlePath)];
     }
 
+    private static MemberPath GetNewInstance<TReflectedType>(Type declaringType, string dottedPath) =>
+        new(typeof(TReflectedType), declaringType, dottedPath.Trim('.').Replace(".[", "[", StringComparison.Ordinal));
+
+    public static MemberPath GetMemberPath<TDeclaringType, TPropertyType>(
```
dennisdoomen, line 132: 🔧 Please document this functionallity.

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.Excluding.cs
```diff
@@ -687,5 +687,416 @@ public void Abstract_properties_cannot_be_excluded()
 
             act.Should().Throw<ArgumentException>().WithMessage("*(o.AbstractProperty + \"B\")*cannot be used to select a member*");
         }
+
+        [Fact]
+        public void An_anonymous_object_with_multiple_fields()
```
dennisdoomen, line 692: 🔧 This does not state a fact

docs/_pages/objectgraphs.md
```diff
@@ -191,6 +207,8 @@ orderDto.Should().BeEquivalentTo(order, options => options
 
 This configuration affects the initial inclusion of members and happens before any `Exclude`s or other `IMemberSelectionRule`s. This configuration also affects matching. For example, that if properties are excluded, properties will not be inspected when looking for a match on the expected object.
 
+The behavior with anonymous objects for selection also applies to `Include` how it does to `Exclude`.
```
dennisdoomen, line 210:
```suggestion
The behavior with anonymous objects for selection also applies to `Include` as it does to `Exclude`.
```

docs/_pages/releases.md
```diff
@@ -39,6 +39,7 @@ sidebar:
 * All `Should()` methods on reference types are now annotated with the `[NotNull]` attribute for a better Fluent Assertions experience when nullable reference types are enabled - [#2380](https://github.com/fluentassertions/fluentassertions/pull/2380)
 * All assertions that support chaining using the `.Which` construct will now amend the caller identifier - [2539](https://github.com/fluentassertions/pull/2539)
 * Introduced a `MethodInfoFormatter` and improved the `PropertyInfoFormatter` - [2539](https://github.com/fluentassertions/pull/2539)
+* `Excluding()` / `For().Exclude()` and `Including()` on `BeEquivalentTo()` now also accepts an anonymous object - [#2488](https://github.com/fluentassertions/fluentassertions/pull/2488)
```
dennisdoomen, line 42:
```suggestion
* `Excluding()` / `For().Exclude()` and `Including()` on `BeEquivalentTo()` now also accepts an anonymous object to include/exclude multiple members at once - [#2488](https://github.com/fluentassertions/fluentassertions/pull/2488)
```

dennisdoomen (changes requested):
👍🏻 Very nice feature
🔧 We only use `var` if the type is obvious.

Src/FluentAssertions/Common/ExpressionExtensions.cs
IT-VBFK: As I can see, this cannot happen, because you either have: `.Excluding(p => p.NestedField)`(which results in `singlePath` to be non-`null` and `segments` to be empty) or `.Excluding(p => new { p.NestedField } )` (which results in `singlePath` null + `segments` not empty).

Src/FluentAssertions/Common/ExpressionExtensions.cs
IT-VBFK: The method itself returning an `IEnumerable<T>`? Should I change that to `List<T>` too?

Src/FluentAssertions/Common/ExpressionExtensions.cs
IT-VBFK: I would rather make this internal? WDYT?

IT-VBFK: Ahh.. mis-understood.. sorry :)

Src/FluentAssertions/Common/ExpressionExtensions.cs
dennisdoomen: My point was that the code doesn't make that clear.

Src/FluentAssertions/Common/ExpressionExtensions.cs
dennisdoomen: Either that. Or call `ToList`

dennisdoomen: And please don't resolve comments if you have not addressed the comment yet.

Src/FluentAssertions/Common/ExpressionExtensions.cs
IT-VBFK: Should I add a comment for this?

Src/FluentAssertions/Common/ExpressionExtensions.cs
IT-VBFK: I have resolved it? (See the `ConvertAll()` instead if `Select`) 

When I write `Select().ToList()` Rider suggested to use `ConvertAll()`

Src/FluentAssertions/Common/ExpressionExtensions.cs
IT-VBFK: Actually, we have do do it exactly like this, because in case of `.Excluding(p => new { })` `singlePath` stays `null` and the `segments` stays empty.

dennisdoomen: In that case, I would make it an `if`, `else if` and `else`. Even though the huge `switch` is difficult to process, you at least can see what will happen at the end.

Src/FluentAssertions/Common/ExpressionExtensions.cs
dennisdoomen: Ah yes, I didn't see the change when I looked at your comment. Maybe I've missed it or maybe you pushed your change _after_ I re-reviewed the code.

IT-VBFK: No, this `ConvertAll` thing I pushed before your second comment and after the first comment 😊

Issue #2480, opened 2023-11-27
BeEquivalent failed with nullably struct property and EqualityComparer


MartinDemberger:
### Description

I have a complex object structure. A few properties are of a nullable struct which must be compared by my own EqualityComparer.

The BeEquivalentTo-call always fails if the struct value is `null`.

The EqualityComparer is necessary because not all properties of the struct should be compared and the compared values must be decrypted to be compared.

### Reproduction Steps

```csharp
 [Fact]
 public void FluentAssertionsTest()
 {
     var art1 = new MyTestClass();
     art1.Should().BeEquivalentTo(new MyTestClass(), o => o
     //.Using<MyProperty?, MyPropertyComparer>()
     .Using<MyProperty, MyPropertyComparer>()
     );
 }

 public class MyTestClass
 {
     public MyProperty? ArticlePriceWithOtherParts { get; set; }
 }

 public struct MyProperty
 {
     public int Value1 { get; set; }
     public int Value2 { get; set; }
 }

 public class MyPropertyComparer : IEqualityComparer<MyProperty?>, IEqualityComparer<MyProperty>
 {
     public bool Equals(MyProperty? x, MyProperty? y)
     {
         var result = Equals(x?.Value1, y?.Value1);
         return result;
     }

     public bool Equals(MyProperty x, MyProperty y)
     {
         var result = Equals(x.Value1, y.Value1);
         return result;
     }

     public int GetHashCode([DisallowNull] MyProperty? obj)
     {
         if (obj is null)
             return 0;
         return obj.Value.Value1;
     }

     public int GetHashCode([DisallowNull] MyProperty obj)
     {
         return obj.Value1;
     }
 }
```

### Expected behavior

The test should succeed.

### Actual behavior

the test fail with the message

Expected property art1.ArticlePriceWithOtherParts to be of type RootNine.ProjectNet.Tests.Calculations.TestOutgoingOfferCalculationElPro+MyProperty, but found <null>

### Configuration

.NET 7.0
FluentAssertions: 6.12.0

### Are you willing to help with a pull-request?

Yes

dennisdoomen:
Is your problem that the `MyPropertyComparer` isn't even used when your `ArticlePriceWithOtherParts` property is `null`? 

Did you try to use `Using<MyProperty?, MyPropertyComparer>()` or `Using<Nullable<MyProperty>, MyPropertyComparer>()`?

PR #2489, opened 2023-11-27
Correct null handling when using a custom comparer

MartinDemberger:
Currently a custom comparer surpresses the `BeEquivalent` compare with a nullable struct property against a expected `null` value.
This change doesn't use the custom comparer for a struct with a expected `null`.

#2480

## IMPORTANT 


docs/_pages/releases.md
```diff
@@ -32,7 +32,7 @@ sidebar:
 * Improved the failure message for `[Not]HaveExplicitProperty` when wrapped in an `AssertionScope` and not implementing the interface - [#2403](https://github.com/fluentassertions/fluentassertions/pull/2403)
 * Improved the failure message for `[Not]HaveExplicitMethod` when wrapped in an `AssertionScope` and not implementing the interface - [#2403](https://github.com/fluentassertions/fluentassertions/pull/2403)
 * Changed `BeEquivalentTo` to exclude `private protected` members from the comparison - [#2417](https://github.com/fluentassertions/fluentassertions/pull/2417)
-
+* `BeEquivalentTo` compares a nullable struct property against a expected `null` value. [#2480](https://github.com/fluentassertions/fluentassertions/issues/2480)
```
IT-VBFK, line 35: Typo: a expected -> an expected

MartinDemberger: Thank you, is corrected

Src/FluentAssertions/Equivalency/Steps/EqualityComparerEquivalencyStep.cs
```diff
@@ -21,6 +21,11 @@ public EqualityComparerEquivalencyStep(IEqualityComparer<T> comparer)
             return EquivalencyResult.ContinueWithNext;
         }
 
+        if (comparands.Subject is null && typeof(T).IsValueType)
```
dennisdoomen, line 24: 🤔 I don't understand why this solves the problem. Care to elaborate?

MartinDemberger: After I wrote this I tried the same issue with a class instead of a struct. It's the same error. So I will add another test for this case and remove the `typeof(T).IsValueType)`.

Tests/FluentAssertions.Equivalency.Specs/ExtensibilitySpecs.cs
```diff
@@ -863,14 +863,129 @@ public DoEquivalencyStep(Action doAction)
     }
 
     [Fact]
-    public void Can_compare_null_with_nullable_struct_property_with_custom_comparer()
+    public void Can_compare_null_again_null_with_custom_comparer_for_nullable_property()
```
IT-VBFK: again -> against

Tests/FluentAssertions.Equivalency.Specs/ExtensibilitySpecs.cs
```diff
@@ -863,14 +863,129 @@ public DoEquivalencyStep(Action doAction)
...
-        var art1 = new ClassWithNullableStructProperty();
-        art1.Should().BeEquivalentTo(new ClassWithNullableStructProperty(), o => o
+        var comperand = new ClassWithNullableStructProperty();
+        comperand.Should().BeEquivalentTo(new ClassWithNullableStructProperty(), o => o
             .Using<StructWithProperties, StructWithPropertiesComparer>()
         );
     }
 
+    [Fact]
+    public void Can_compare_null_again_not_null_with_custom_comparer_for_nullable_property()
```
IT-VBFK: again -> against

and following

Tests/FluentAssertions.Equivalency.Specs/ExtensibilitySpecs.cs
IT-VBFK: oh.. you were quicker :D

Src/FluentAssertions/Equivalency/Steps/EqualityComparerEquivalencyStep.cs
MartinDemberger: I have added the additional tests and also a extended solution for other null cases. I also added a comment in code to explain the reason of the condition.

Please feel free to provide feedback if any part of my explanation is unclear.

Tests/FluentAssertions.Equivalency.Specs/ExtensibilitySpecs.cs
dennisdoomen, line 1: 🔧 Please use the AAA comment blocks we always use

Src/FluentAssertions/Equivalency/Steps/EqualityComparerEquivalencyStep.cs
```diff
@@ -21,6 +21,14 @@ public EqualityComparerEquivalencyStep(IEqualityComparer<T> comparer)
             return EquivalencyResult.ContinueWithNext;
         }
 
+        if (comparands.Subject is null || comparands.Expectation is null)
+        {
+            // The later check for `comparands.Subject is T` leads to an failure even if the expectation is null
+            // IEqualityComparer isn't null aware. To preserve NullReferenceExceptions a compare with null is
+            // not handled by the custom comparer.
```
dennisdoomen, line 28: ❓ I don't get the remark about NREs. Isn't this change just going to make sure that the entire step (and the custom implementation of `IEqualityComparer<T>` is not going to be used) when either the subject or expectation is `null`?

MartinDemberger: You are right. I removed this part of the comment. I'm not sure if the commet at all is usefull. Should I remove it?

dennisdoomen (approved):

Src/FluentAssertions/Equivalency/Steps/EqualityComparerEquivalencyStep.cs
```diff
@@ -21,6 +21,12 @@ public EqualityComparerEquivalencyStep(IEqualityComparer<T> comparer)
             return EquivalencyResult.ContinueWithNext;
         }
 
+        if (comparands.Subject is null || comparands.Expectation is null)
+        {
+            // The later check for `comparands.Subject is T` leads to an failure even if the expectation is null.
```
ITaluone, line 26:
```suggestion
            // The later check for `comparands.Subject is T` leads to a failure even if the expectation is null.
```

little typo

Tests/FluentAssertions.Equivalency.Specs/ExtensibilitySpecs.cs
MartinDemberger: What do you mean by AAA comment blocks? Is this the method comment with three dashes?

I think the comment is only relevant for the if statement and not for the method. If you think this comment is for the method I can move it to the method comment.

ITaluone:
A few qodana issues are left :)

jnyrup (approved):

PR #2504, opened 2023-12-05
Turn off RoslynAnalyzers in Qodana WF

hybloid:
This PR turns off RoslynAnalyzers in Qodana by introducing new profile that is based on qodana.starter. This profile creates group InspectionsToExclude and excludes them from the default sets.


hybloid:
@dennisdoomen please approve the switch to 2023.3.0. The expected version of the IDE is `233.12259.111.73`

dennisdoomen (approved):

PR #2514, opened 2023-12-14
Update code_quality.yml to use newer Qodana linter

architectbat:
Migrate to the new version of Qodana GitHub Action


ITaluone:
dup #2513 ?

PR #2515, opened 2023-12-14
Update qodana.yaml to limit target frameworks

hybloid:
Using `frameworks:` property one could limit analysed TFMs.

### DIFF
--- a/qodana.yaml
+++ b/qodana.yaml
@@ -8,6 +8,7 @@ profile:
 
 dotnet:
   solution: FluentAssertions.sln
+  frameworks: "!netstandard2.0;!netstandard2.1"
 
 exclude:
   - name: ConvertIfStatementToReturnStatement


hybloid:
I will turn off pr-analysis to verify the report

qodana.yaml
```diff
@@ -8,6 +8,7 @@ profile:
 
 dotnet:
   solution: FluentAssertions.sln
+  frameworks: net47;net6.0;netstandard2.0;netstandard2.1
```
IT-VBFK, line 11: I think we don't want .netstandard2.x to be reported

dennisdoomen: Correct

IT-VBFK: I guess we are close to a 0 error baseline report then :D

dennisdoomen:
I'm not sure it works. There are now _no_ problems, which raises my suspicions...

qodana.yaml
hybloid: Folks, need a little bit more time. We've implemented some kind of validation after the limitation feature was implemented. Gonna find the way how to supply the same config in a valid way.

hybloid: image
image

hybloid: image

hybloid: @dennisdoomen the last configuration works (with plus list). We will think how we could supply the minus list then for 241.

hybloid:
@dennisdoomen having

`Using 'local-changes' script as qodana run scenario` 

in the log means that we are running in local changes mode, where we analyze only changed files.

dennisdoomen:
So does this work now?

hybloid:
> So does this work now?

@dennisdoomen it works right now, checked once again and squashed to one commit. You won't see any issues in that PR, since you have pr-mode on. I've double tested it on https://github.com/hybloid/fluentassertions/actions/runs/7460599955

dennisdoomen:
I thought you said that limiting by framework wasn't finished yet.

dennisdoomen:
It works!

Before:

image

After

image

jnyrup:
Thanks for the help @hybloid

PR #2530, opened 2023-12-31
BeEmpty() materializes IEnumerable<T> only once, even on failure

### LINKED ISSUE #2490: IEnumerable gets evaluated multiple times when using BeEmpty()
### Description

There seems to be an issue with `BeEmpty()` when used with `IEnumerable`. The `IEnumerable` gets evaluated multiple times, leading to unexpected behavior.

### Reproduction Steps

```csharp
private bool a = false;

public IEnumerable<string> Get()
{
    if (a)
    {
        yield break;
    }

    a = true;
    yield return "123";
}

[TestMethod]
public void Execute()
{
    var result = Get();

    result.Should().BeEmpty();
}
```

### Expected behavior

I expect the test to fail, with the message: "Expected result to be empty, but found {"123"}."

### Actual behavior

Test fails, but with the message: "Expected result to be empty, but found {empty}."

### Regression?

_No response_

### Known Workarounds

Converting the `IEnumerable` to a list before asserting solves the issue:
```csharp
[TestMethod]
public void Execute()
{
    var result = Get().ToList();

    result.Should().BeEmpty();
}
```

### Configuration

FluentAssertions version: 6.12.0
.NET version: 7.0

### Other information

The issue seems to arise from how Fluent Assertions handles the evaluation of `IEnumerable`. Ensuring the collection is evaluated only once before the emptiness check might resolve this issue.

No

louis-z:
- Make BeEmpty() materialize only the first item, as no further processing is needed for validation
- The error message now mentions "but found at least one item" and outputs the said item

Fixes #2490

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -639,13 +639,14 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> : Referenc
     /// </param>
     public AndConstraint<TAssertions> BeNullOrEmpty(string because = "", params object[] becauseArgs)
     {
-        var nullOrEmpty = Subject is null || !Subject.Any();
+        var singleItemArray = Subject?.Take(1).ToArray();
+        var nullOrEmpty = singleItemArray is null || singleItemArray.Length == 0;
 
         Execute.Assertion.ForCondition(nullOrEmpty)
             .BecauseOf(because, becauseArgs)
             .FailWith(
-                "Expected {context:collection} to be null or empty{reason}, but found {0}.",
-                Subject);
+                "Expected {context:collection} to be null or empty{reason}, but found at least one item {0}.",
+                singleItemArray);
 
         return new AndConstraint<TAssertions>((TAssertions)this);
     }
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -294,7 +294,7 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> : Referenc
             .FailWith("but found <null>.")
             .Then
             .ForCondition(subject => subject.Length == 0)
-            .FailWith("but found at least 1 item {0}.", singleItemArray)
+            .FailWith("but found at least one item {0}.", singleItemArray)
             .Then
             .ClearExpectation();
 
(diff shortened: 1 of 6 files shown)


Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -285,15 +285,16 @@ public AndConstraint<TAssertions> AllBeOfType(Type expectedType, string because
...
             .WithExpectation("Expected {context:collection} to be empty{reason}, ")
-            .Given(() => Subject)
+            .Given(() => singleItemArray)
             .ForCondition(subject => subject is not null)
             .FailWith("but found <null>.")
             .Then
-            .ForCondition(subject => !subject.Any())
-            .FailWith("but found {0}.", Subject)
+            .ForCondition(subject => subject.Length == 0)
+            .FailWith("but found at least 1 item {0}.", singleItemArray)
```
dennisdoomen, line 297: 🤔 I've been trained that numbers up to 10 should be spelled out as "one", "two", etc. Not sure if that applies here.

dennisdoomen (approved):

jnyrup (approved):

jnyrup:
A note: This partly reverts #227, but I'm fine with the change here.
A similar change should be applied to `BeNullOrEmpty` to keep them aligned.

louis-z:
> [...] A similar change should be applied to `BeNullOrEmpty` to keep them aligned.

Will do.

jnyrup:
Aren't we missing a test for `BeNullOrEmpty` identical to the added `When_asserting_non_empty_collection_is_empty_it_should_enumerate_only_once` test to verify the change?

dennisdoomen:
Fair point.

bogataj:
@dennisdoomen did this change also unintentionally affect WithMessage on ExceptionAssertions? Release notes only mentions BeEmpty() and BeNullOrEmpty(), but I have tests failing now after upgrading 6.12.0 -> 6.12.1 where the iterator seem to stop after the first item at that method also. 

Specifically my use case is a generic convention test that finds and prints all missing subscription registrations on a service bus, stopping the iteration on the first item while evaluating WithMessage would be a somewhat unfortunate feature, while I appreciate the efforts for performance gains 👍

jnyrup:
@bogataj Please share a code example of what no longer works for you.

In the meantime I would suggest that if your test requires complete enumeration, that you enforce it on your side instead of relying on that we (or any other library) does that.

bogataj:
@jnyrup My apologize, I usually start with a sample but for some reason this time I didnt. While deconstructing my tests to prepare a sample, I found that we were actually chaining the results between tests, basing the original message off an BeEmpty(), so this is my bad and the update to the library does exactly as is says, thank you.

PR #2535, opened 2024-01-02
Implement support for IAsyncEnumerable the same way how Nunit4 implemented it

zachsaw:
Draft PR to implement support for IAsyncEnumerable - option 3 as discussed in Issue 1213? The first commit does not take dependency on `System.Linq.Async` but does not offer strongly typed `BeEquiavlentOf`. Taking dependency on `System.Linq.Async` should be no worse than `System.Threading.Tasks` though.

This is also the [implementation NUnit4](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/AsyncEnumerableAdapter.NetCore.cs) has gone with. Yes, it does violate sync-over-async practice, but for all intents and purposes, FluentAssertion is a unit test framework. Blocking enumeration of AsyncEnumerable isn't as big a sin here vs high throughput web servers.

Note: This is a draft PR to demo a concept, not prod ready in any shape or form.

## IMPORTANT 


Src/FluentAssertions/Collections/AsyncEnumerableAssertions.cs
```diff
@@ -0,0 +1,3651 @@
...
+using FluentAssertions.Common;
+using FluentAssertions.Equivalency;
+using FluentAssertions.Execution;
+using FluentAssertions.Formatting;
+using FluentAssertions.Primitives;
+
+namespace FluentAssertions.Collections;
+
+[DebuggerNonUserCode]
+public class AsyncEnumerableAssertions<T> : AsyncEnumerableAssertions<IAsyncEnumerable<T>, T, AsyncEnumerableAssertions<T>>
```
dennisdoomen, line 18: Wouldn't it be much simpler to just have the new `Should` overload enumerate the async collection and then return the existing `GenericCollectionAssertions`?

zachsaw-lmg: lol yeah you're right. That also allows us to avoid taking dependency on `System.Linq.Async`!

zachsaw-lmg: ok that actually failed when we have a class that implements multiple `IAsyncEnumerable` generic types. With `RespectingRuntimeTypes` option, it no longer sees the type has having multiple generic types for the `IAsyncEnumerable` interface.

i.e. this fails.

```
    [Fact]
    public void
        When_a_object_implements_multiple_IAsyncEnumerable_interfaces_but_the_declared_type_is_assignable_to_only_one_and_runtime_checking_is_configured_it_should_fail()
    {
        // Arrange
        IAsyncEnumerable<string> collection1 = new AsyncEnumerableOfStringAndObject();
        IAsyncEnumerable<string> collection2 = new AsyncEnumerableOfStringAndObject();

        // Act
        Action act =
            () => collection1.Should().BeEquivalentTo(collection2, opts => opts.RespectingRuntimeTypes());

        // Assert
        act.Should().Throw<XunitException>("the runtime type is assignable to two IAsyncEnumerable interfaces")
            .WithMessage("*cannot determine which one*");
    }
```

zachsaw-lmg: `AsyncEnumerableAssertions` is gross as it violates DRY having lots of shared code with `GenericCollectionAssertions`. However, I can't see any other way due to the use of inheritance (violating composition over inheritance). :(

zachsaw-lmg: Ok I've updated `AsyncEnumerableAssertions` to reuse `GenericCollectionAssertions` whereever possible.

zachsaw-lmg: I've had to revert the above. We lose `ReferenceEquals` checks when we do that as the references are no longer the original one so they are no longer "equal".

PR #2537, opened 2024-01-05
Bump cspell to 8.3.2

IT-VBFK:
**Reason:**

Before it had a strange error when running `yarn install` (GHA build _and_ locally)
image

After googling this error I found this issue: https://github.com/yarnpkg/yarn/issues/4812

And one of the tips was: https://github.com/yarnpkg/yarn/issues/4812#issuecomment-1193692415

So I deleted the lock file and run `yarn install` again, but this time it installs the latest 8 (8.3.2 at this moment) and the error was gone afterwards

## IMPORTANT 

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,10 +2,10 @@
 # yarn lockfile v1
 
 
-"@cspell/cspell-bundled-dicts@8.3.1":
-  version "8.3.1"
-  resolved "https://registry.yarnpkg.com/@cspell/cspell-bundled-dicts/-/cspell-bundled-dicts-8.3.1.tgz#342f205d79ba4c074023d67f30531dece66e37b6"
-  integrity sha512-pITQe2B9CGm7WFe/BsKyzOlBtuTRtP4uSC6baijCOJX5xduTCfJsS/l4iX3oXNx2ewYny+K3yUE9KRI0QPKtQA==
+"@cspell/cspell-bundled-dicts@8.3.2":
+  version "8.3.2"
+  resolved "https://registry.yarnpkg.com/@cspell/cspell-bundled-dicts/-/cspell-bundled-dicts-8.3.2.tgz#649ed168a72cb49a7d83f3840ab6933a8beba68d"
+  integrity sha512-3ubOgz1/MDixJbq//0rQ2omB3cSdhVJDviERZeiREGz4HOq84aaK1Fqbw5SjNZHvhpoq+AYXm6kJbIAH8YhKgg==
   dependencies:
     "@cspell/dict-ada" "^4.0.2"
     "@cspell/dict-aws" "^4.0.1"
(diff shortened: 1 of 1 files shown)


jnyrup:
Thanks for looking into this.

I was mislead by seemingly passing pipelines in #2531
The problematic changes are
https://github.com/fluentassertions/fluentassertions/pull/2531/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR986-R1008

which, as far as I understand the lock-file, maps both `string-width-cjs` and `string-width` to `string-width-cjs`, which makes the disk caching mechanism trying to write both to same destination.

https://github.com/fluentassertions/fluentassertions/actions/runs/7378193454/job/20073037546
image

We need a way to distinguish between failing due to spelling mistakes and failing due to anything else...

IT-VBFK:
Hmm.. I don't get why it keeps failing on develop? Locally it works (for me)

IT-VBFK:
Ok.. 
Because this package is no longer maintained.. Should we just ignore this errors?

e.g. something like:

```csharp
Node($"{YarnCli} --silent install", workingDirectory: RootDirectory,
    logger: (_, msg) =>
    {
        if (msg.Contains("warning Pattern")) Information(msg);
        else Error(msg);
    });
```

or even:

```csharp
Node($"{YarnCli} --silent install", workingDirectory: RootDirectory,
    logger: (_, msg) =>
    {
        if (!msg.Contains("warning Pattern")) Error(msg);
    });
```

IT-VBFK:
#2538

jnyrup:
Running `git clean -fdx; .\build.ps1 SpellCheck` consistently provokes the error for me locally.
That explains why it behaves different locally than in GH pipelines.

> Because this package is no longer maintained.. Should we just ignore these sort of errors?

Normally that implies to me, that I should start looking for an alternative to `Yarn.MSBuild`.
I have little experience with npm/yarn/bun/whatever and frankly what to spent as little time as possible myself with that ecosystem.

Brainstorming:
* is there an easy way to use npm - spending five minutes Googling it seems that the speed benefits that Yarn gave has also been incorporated into npm.
* Instead of downloading yarn via a nuget package, could we just download it from "somewhere" more official?

IT-VBFK:
What about ignoring for now and hope for https://github.com/nuke-build/nuke/pull/1151 ?

jnyrup:
I'm fine with ignoring for now, I'm just thinking about the future.
e.g. when cspell or one it's dependencies start using a feature that is completely unsupported by yarn 1.22.19. 
Then we're stuck with the previous version of cspell.

I'm also subscribing to that PR, but as I understand it, it doesn't _include_ yarn, but expects it to be present on the system.

dennisdoomen:
Yes, we can switch to NPM as well, but we have to investiage whether we can do that without having to install NPM on the build agent.

IT-VBFK:
We probably can mimic what this script does: https://www.npmjs.com/package/npm/v/10.2.5#direct-download

IT-VBFK:
What's quite cool is: we don't have to install npm manually..
Digging into the package `node.js.redist` I noticed following:
image

So.. actually... this works, and also could be injected to a NUKE tool, like Node itself.

PR #2538, opened 2024-01-05
Ignore `yarn install` errors

IT-VBFK:
## IMPORTANT 

### DIFF
--- a/Build/Build.cs
+++ b/Build/Build.cs
@@ -354,11 +354,33 @@ class Build : NukeBuild
         .ProceedAfterFailure()
         .Executes(() =>
         {
-            Node($"{YarnCli} --silent install", workingDirectory: RootDirectory);
+            Node($"{YarnCli} --silent install", workingDirectory: RootDirectory,
+                logger: YarnInstallLogger);
 
             Node($"{YarnCli} --silent run cspell", workingDirectory: RootDirectory,
                 logger: (_, msg) => Error(msg));
         });
+
+    Action<OutputType, string> YarnInstallLogger = (_, msg) =>
+    {
+        // See this PR here: https://github.com/fluentassertions/fluentassertions/pull/2537
+        // and several comments and references.
+        // This seems to be a bug in yarn and the package 'Yarn.MSBuild' is no longer maintained;
+        // So: ignore 'yarn install' errors when the same cache directory is defined in yarn.lock.
+        // This errors have the pattern like: 'Error: warning Pattern [packagename@version]...'
+
+        if (!msg.Contains("is trying to unpack in the same destination"))
+        {
+            if (msg.Contains("warning Pattern"))
+            {
+                Warning(msg);
+            }
+            else
+            {
+                Error(msg);
+            }
+        }
+    };
 
     bool HasDocumentationChanges => Changes.Any(x => IsDocumentation(x));
 
--- a/Build/Build.cs
+++ b/Build/Build.cs
@@ -371,7 +371,7 @@ class Build : NukeBuild
 
         if (!msg.Contains("is trying to unpack in the same destination"))
         {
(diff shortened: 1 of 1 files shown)


Build/Build.cs
```diff
@@ -354,7 +354,11 @@ void ReportTestOutcome(params string[] globFilters)
         .ProceedAfterFailure()
         .Executes(() =>
         {
-            Node($"{YarnCli} --silent install", workingDirectory: RootDirectory);
+            Node($"{YarnCli} --silent install", workingDirectory: RootDirectory,
+                logger: (_, msg) =>
+                {
+                    if (!msg.Contains("warning Pattern")) Error(msg);
```
dennisdoomen, line 360: 🔧 We need some comments to explain what we're doing here. 
🙃 (nitpick) Please use curly braces here

dennisdoomen (changes requested):

Build/Build.cs
```diff
@@ -354,7 +354,20 @@ void ReportTestOutcome(params string[] globFilters)
...
+                    // See this PR here: https://github.com/fluentassertions/fluentassertions/pull/2537
+                    // and several comments and references.
+                    // This seems to be a bug in yarn and the package 'Yarn.MSBuild' is no longer maintained;
+                    // So: ignore 'yarn install' errors when the same cache directory is defined in yarn.lock.
+                    // This errors have the pattern like: 'Error: warning Pattern [packagename@version]...'
+
+                    if (!msg.Contains("warning Pattern"))
+                    {
+                        Error(msg);
+                    }
```
jnyrup, line 369: I assume `warning Pattern` covers all kinds of warnings.
So how about we only completely ignore `"is trying to unpack in the same destination"` to avoid silencing too much?
```cs
if (!msg.Contains("is trying to unpack in the same destination"))
{
	if (msg.Contains("warning Pattern"))
	{
		Warning(msg);
	}
	else
	{
		Error(msg);
	}
}
```

IT-VBFK: Is this really the case? If I googling "warning pattern", it only comes up with this sort of error.

jnyrup: `Pattern` comes from
https://github.com/yarnpkg/yarn/blob/158d96dce95313d9a00218302631cd263877d164/src/reporters/lang/en.js#L127-L128

`Warning` comes from `config.reporter.warn(`
https://github.com/yarnpkg/yarn/blob/master/src/package-fetcher.js#L118-L123

grepping for `reporter.warn(` in the yarn src gives 60 hits.

IT-VBFK: Hmm true, but the casing (with pattern in uppercase) is only used for this error.

But I will change it according to you suggestion :)

Build/Build.cs
```diff
@@ -354,12 +354,34 @@ void ReportTestOutcome(params string[] globFilters)
...
+    {
+        // See this PR here: https://github.com/fluentassertions/fluentassertions/pull/2537
+        // and several comments and references.
+        // This seems to be a bug in yarn and the package 'Yarn.MSBuild' is no longer maintained;
+        // So: ignore 'yarn install' errors when the same cache directory is defined in yarn.lock.
+        // This errors have the pattern like: 'Error: warning Pattern [packagename@version]...'
+
+        if (!msg.Contains("is trying to unpack in the same destination"))
+        {
+            if (msg.Contains("warning Pattern"))
```
jnyrup, line 374: Forgot to adjust this one to be all warnings.

```suggestion
            if (msg.StartsWith("warning"))
```

IT-VBFK: OK

dennisdoomen (approved):

jnyrup (approved):

Issue #2407, opened 2023-10-25
Using which after BeOfType shows wrong information if assertions fails


comecme:
### Description

I'd expect a failure on an assertion after using `.Which` to report the property I am asserting if the assertions fails.

I must admit that the documentation doesn't tell me what to expect if the assertions fails, it just lists all kinds of assertions you can make, including `Which`. If it is a known limitation of FluentAssertions that is unable to tell me the name of the property on which the assertion fails, than please make that clear in the documentation.

### Reproduction Steps

```csharp
// Arrange
var ex = new Exception("Message");

// Assert
ex.Should().BeOfType<Exception>()
    .Which.Message.Should().Be("Other Message");
```

I know that for Exceptions, I'd better use `Should().Throw<Exception>().WithMessage`. But I am not actualy trying this on an Exception but on a custom type. The above code however shows the problem, using a type that is generally available.

### Expected behavior

Print out an error stating that `Message` was expected to be "Other Message"
```
Expected ex.Message to be "Other Message" with a length of 13, but "Message" has a length of 7, differs near "Mes" (index 0).
```

### Actual behavior

FluentAssertions prints out the following error, stating that `ex` itself was expected to be "Other Message":
```
Expected ex to be "Other Message" with a length of 13, but "Message" has a length of 7, differs near "Mes" (index 0).
```

### Known Workarounds

Split the assertion in to two assertions, althoug that just means you don't use the `Which` method after all.
```csharp
ex.Should().BeOfType<Exception>();
ex.Message.Should().Be("Other Message");
```

### Configuration

Fluent Assertions 6.12.0 on .NET Framework 4.8

### Other information

This example is one of the examples [on the website](https://fluentassertions.com/basicassertions/#:~:text=You%20can%20do%20that%20by%20chaining%20those%20assertions%20onto%20the%20Which%20property%20like%20this.).

### Are you willing to help with a pull-request?

No

IT-VBFK:
#1502 ?

comecme:
As @IT-VBFK suggested, this seems to be a duplicate of #1502

PR #2539, opened 2024-01-06
Spike: Revisit AssertionScope and how it travels over chained calls

dennisdoomen:
## IMPORTANT 


jnyrup:
Had a rather quick look at this and it looks very promising 🤩 
Nothing caught my attention as being bad, regressing the current behavior or having pitfalls.

dennisdoomen:
> Had a rather quick look at this and it looks very promising 🤩
> Nothing caught my attention as being bad, regressing the current behavior or having pitfalls.

Shall I continue with the real implementation? Or do you want to take a deeper look first?

dennisdoomen:
I'm wondering whether the [new Assertion class](https://github.com/fluentassertions/fluentassertions/pull/2539/files#diff-4aee91175f15aef5a100c0add702ad89e66dd07d00b0be41966daed5ccbb9c54R9) in this redesign is a good name, given that we also have all the classes postfixed with `Assertions`. ChatGPT and Gemini suggested `FluentAssertion`, `FluentAssertionBuilder`, `AssertChain`, `Verifier` and `FluentVerifier`. Any thoughts? For reference, it is used in

```csharp
Assertion.GetOrCreate()
    .ForCondition(Subject.CompareTo(unexpected) != Equal)
    .FailWith("Expected {context:object} {0} not to be ranked as equal to {1}{reason}.", Subject, unexpected);
```

I do like `FluentAssertion` myself. I also like `FluentAssertionBuilder`, but it's not like we're really building something. The conditions are evaluated right away. `AssertionChain` is a better fit from a functional perspective, but sounds very much like `AssertionScope`, which is something completely different.

IT-VBFK:
`FluentAssertionProvider` 🤔

jnyrup:
`AssertionEvaluator`?
In any case I would avoid the `Fluent` prefix as that is about _how_ it does something compared to _what_ it does.

codymullins:
Assertion makes sense to me. AssertionBuilder does seem to map more closely to what it’s doing, though yes it isn’t strictly speaking a builder. 

How would this be used by external consumers, if at all? I have only really used Fluent Assertions via .Should(…)

obiwanjacobi:
If the conditions are evaluated right away I would think `Assert` and then maybe rename the `GetOrCreate()` function(s) ...?
(I have no knowledge of the internals and the constructs used - so maybe this is bogus)

jscarle:
`AssertionContext`?

dennisdoomen:
> How would this be used by external consumers, if at all? I have only really used Fluent Assertions via .Should(…)

It's used both as the internal engine as well as by anybody that builds extensions. See also https://fluentassertions.com/extensibility/#building-your-own-extensions

> If the conditions are evaluated right away I would think Assert and then maybe rename the GetOrCreate() function(s) ...?
(I have no knowledge of the internals and the constructs used - so maybe this is bogus

`Assert` is a verb and not a noun, so that wouldn't really work

> AssertionEvaluator?

It doesn't evaluate an assertion. You're building and executing a, potentially, chained assertion. 

> AssertionContext?

It is also not the context of an assertion. It's _doing_ the assertion work. 

PS. For reference, I've updated the original question with a more elaborate example.

jscarle:
Considering you already have `Verifier` and `FluentVerifier`, wouldn't a natural fit be `Asserter` and `FluentAsserter`?

vbreuss:
What about `AssertThat`?

This way, your chain would read almost like an English sentence
```csharp
assertThat
.WithExpectation("Expected {context:collection} to be empty{reason}, ")
.Given(() => singleItemArray)
.ForCondition(subject => subject is not null)
.Then
.ForCondition(subject => subject.Length == 0)
.FailWith("but found at least one item {0}.", singleItemArray);
```

victorvogelpoel:
Something with 'Pipeline' or 'Predicate'?

dennisdoomen:
Hmm, I kinda like `AssertionChain`

dennisdoomen:
I'm running into a design challenge while finishing up the work for introducing the new `AssertionChain`. In the new design, an instance of an `AssertionChain` is now passed by the `Should` extension method to the assertion class and then used in code like this.

    assertionChain
          .WithExpectation("Did not expect the minute part of {context:the time} to be {0}{reason}, ", unexpected)
          .ForCondition(Subject.HasValue)
          .FailWith("but found a <null> DateTimeOffset.")
          .Then
          .ForCondition(Subject.Value.Minute != unexpected)
          .FailWith("but it was.");

The problem is that I'm trying to avoid the need to match `WithExpectation` with a terminating `ClearExpectation`. However, the `assertionChain` instance is reused when chaining calls using `And` and causes the expectation to "leak" into the next assertion chain. 

Options I'm considering:
1. Reintroducing `ClearExpectation`
2. Changing `WithExpectation` so that it returns an `IDisposable` that defines the lifetime, and accepts something like a `Action<AssertionChain>` parameter that should be used to put multiple calls like `ForCondition` and `FailWith` explicitly within the lifetime.

I don't particularly like 2, but I hate 1. I'm looking for alternative solutions.

Corniel:
I have no preference for a name. I have another question though: wouldn't it be preferable if the chain was immutable?

dennisdoomen:
> wouldn't it be preferable if the chain was immutable?

I'm not sure what you mean, but the chain is supposed to be passed from the first `Should` all the way to the end. That's how we will fix #1502.

dennisdoomen:
Since there's no feedback yet, I'm going to implement option 2.

dennisdoomen:
It seems option 3 can result in some pretty ugly code. See `GenericCollectionAssertions.NotEqual`, which resulted in 

```csharp
using (var _ = assertionChain
           .WithExpectation("Expected collections not to be equal{reason}, "))
{
    assertionChain
        .Given(() => Subject)
        .ForCondition(subject => subject is not null)
        .Then
        .ForCondition(subject => !ReferenceEquals(subject, unexpected))
        .FailWith("but they both reference the same object.");
}

if (assertionChain.Succeeded)
{
    assertionChain
        .Given(() => Subject.ConvertOrCastToCollection())
        .ForCondition(actualItems => !actualItems.SequenceEqual(unexpected))
        .FailWith("Did not expect collections {0} and {1} to be equal{reason}.", _ => unexpected,
            actualItems => actualItems);
}
```

With option 2, this would become a lot nicer:

```csharp
assertionChain
    .WithExpectation("Expected collections not to be equal{reason}, chain => chain
        .Given(() => Subject)
        .ForCondition(subject => subject is not null)
        .Then
        .ForCondition(subject => !ReferenceEquals(subject, unexpected))
        .FailWith("but they both reference the same object."))
    .Then    
    .Given(() => Subject.ConvertOrCastToCollection())
    .ForCondition(actualItems => !actualItems.SequenceEqual(unexpected))
    .FailWith("Did not expect collections {0} and {1} to be equal{reason}.", _ => unexpected, actualItems => actualItems);
}
```

jnyrup:
While option 2 has increased nesting, it very explicitly ensures that the state of `WithExpecation` shouldn't leak as it ends with the lifetime of the inner scope 👍

Could this also help preventing having to call `assertionChain.Succeeded` a gazillion places to guard the assertion being wrapped in an `AssertionScope`?

IT-VBFK:
I really like how option 2 looks like, because it's so explicit about lifetime and such 👍

dennisdoomen:
> Could this also help preventing having to call `assertionChain.Succeeded` a gazillion places to guard the assertion being wrapped in an `AssertionScope`?

Yes, although a lot of these have also become less necessary because of the way `AssertionChain` handles chained assertions

ITaluone:
Is it just me, who wonders whether `.WithExpectation` (for option 2) is still the right term, because it now does more than simply "cache" the expectation.

I asked chatgpt about this specific scenario and one suggestion was `.WithPersistentExpection`, which is almost right from the perspective of what it is supposed to do.

After that also `WithScopedExpectation` or `WithScopedExpectation().For()` comes to my mind.

e.g.
```csharp
assertionChain
    .WithScopedExpectation("Expected collections not to be equal{reason}", chain => chain
        .Given(() => Subject)
        .ForCondition(subject => subject is not null)
        .Then
        .ForCondition(subject => !ReferenceEquals(subject, unexpected))
        .FailWith("but they both reference the same object."))
    .Then    
    .Given(() => Subject.ConvertOrCastToCollection())
    .ForCondition(actualItems => !actualItems.SequenceEqual(unexpected))
    .FailWith("Did not expect collections {0} and {1} to be equal{reason}.", _ => unexpected, actualItems => actualItems);
```

or

```csharp
assertionChain
    .WithScopedExpectation("Expected collections not to be equal{reason}")
    .For(chain => chain
        .Given(() => Subject)
        .ForCondition(subject => subject is not null)
        .Then
        .ForCondition(subject => !ReferenceEquals(subject, unexpected))
        .FailWith("but they both reference the same object."))
    .Then    
    .Given(() => Subject.ConvertOrCastToCollection())
    .ForCondition(actualItems => !actualItems.SequenceEqual(unexpected))
    .FailWith("Did not expect collections {0} and {1} to be equal{reason}.", _ => unexpected, actualItems => actualItems);
```

eNeRGy164:
Not looking at implementation complexity, I like both suggestions, with the last one being the most descriptive.

Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -14,60 +10,40 @@ namespace FluentAssertions.Execution;
 /// Represents an implicit or explicit scope within which multiple assertions can be collected.
 /// </summary>
 /// <remarks>
-/// This class is supposed to have a very short lifetime and is not safe to be used in assertions that cross thread-boundaries
+/// This class is supposed to have a very short life time and is not safe to be used in assertion that cross thread-boundaries
```
IT-VBFK, line 13: I think this should be reverted, because "lifetime" != "life time" and I think former is correct here...
same with "assertion"

IT-VBFK:
BTW.. weird version here
image

Build/Build.cs
```diff
@@ -86,6 +82,11 @@ class Build : NukeBuild
         {
             SemVer = GitVersion.SemVer;
 
+            foreach (var pi in GitVersion.GetType().GetProperties())
+            {
+                Information("{Name} = {Value}", pi.Name, pi.GetValue(GitVersion));
+            }
+
```
IT-VBFK, line 89: @dennisdoomen Do you mind if I take this "SemVer"-thing over?

dennisdoomen: I bumped GitVersion to 6.0.2 to see if this fixes the recurring issue of GitVersion not being able to calculate the version. But indeed, it also changed something else.

dennisdoomen: But sure. If you can create a new PR that bumps the GitVersion version to 6.0.2 _and_ maintain the correct version, that would be great.

IT-VBFK: Cool :)

IT-VBFK: For now, I think we cannot rely on GitVersion to give us the correct one, so I think I have to go another way (for now)

dennisdoomen: Have you tried a clean project?

IT-VBFK: No

I actually think that this could be related to NUKE, because of breaking changes in the command response

Corniel:
Too bad NUnit already uses it, otherwise, I would prefer:

``` C#
Assert.That(() => Subject.Length != 0)
    // ...
    .Execute();
```

Build/Build.cs
dennisdoomen: I can't think of anything. Nuke's GitVersion support just exposes all the properties that GitVersion returns.

ITaluone:
Should we also add a note about #2407?

ITaluone:
I _really_ like how the look and feel of the new chainings looks like 👍 

But I am thinking about finding a more general naming for `AndConstraint` and `AndWhich`, because this seams a little odd in perspective of consistency.
Thinking aloud: Why `AndWhichConstraint` can be modified to additionally fit the new behavior?

Src/FluentAssertions/Xml/XDocumentAssertions.cs
```diff
@@ -171,14 +174,14 @@ public AndWhichConstraint<XDocumentAssertions, XElement> HaveRoot(XName expected
...
-        Execute.Assertion
+        assertionChain
             .ForCondition(root is not null && root.Name == expected)
             .BecauseOf(because, becauseArgs)
             .FailWith(
                 "Expected {context:subject} to have root element {0}{reason}, but found {1}.",
                 expected.ToString(), Subject);
 
-        return new AndWhichConstraint<XDocumentAssertions, XElement>(this, root);
+        return new AndWhich<XDocumentAssertions, XElement>(this, root, assertionChain, $"/{expected}");
```
IT-VBFK, line 184: What about concatenating this with `.` like we access members in C#:

```csharp
var doc = ...
doc.Should().HaveRoot("member").Which...
```

which would then result in `Expected doc.member to have...`

Src/FluentAssertions/Formatting/PropertyInfoFormatter.cs
```diff
@@ -18,6 +18,15 @@ public bool CanHandle(object value)
 
     public void Format(object value, FormattedObjectGraph formattedGraph, FormattingContext context, FormatChild formatChild)
     {
-        formattedGraph.AddFragment(((PropertyInfo)value).Name);
+        var property = (PropertyInfo)value;
```
Corniel, line 21:
``` suggestion
        if (value is not PropertyInfo property)
```

Src/FluentAssertions/Xml/XDocumentAssertions.cs
dennisdoomen: Because we're talking about XML paths, and they are usually displayed using `/` as a separator.

Src/FluentAssertions/Formatting/MethodInfoFormatter.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// </returns>
+    public bool CanHandle(object value)
+    {
+        return value is MethodInfo;
+    }
+
+    public void Format(object value, FormattedObjectGraph formattedGraph, FormattingContext context, FormatChild formatChild)
+    {
+        var method = (MethodInfo)value;
+        if (method is null)
```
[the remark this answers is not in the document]
dennisdoomen: That doesn't help anymore with the latest version of `MethodInfoFormatter`

Src/FluentAssertions/Xml/XDocumentAssertions.cs
IT-VBFK: Hmm.. I think we need something different than what nested `AssertionScope`s use when chaining their contexts together.

If I get an exception that says something like this: `Expected Foo/Bar/Baz/doc/member to have...` it wouldn't be entirely sure what maps to what. Do you see my concern?

docs/_pages/releases.md
```diff
@@ -53,6 +55,7 @@ sidebar:
 * Fixed `RaisePropertyChangeFor` to return a filtered list of events - [#2677](https://github.com/fluentassertions/fluentassertions/pull/2677)
 
 ### Breaking Changes (for users)
+* Replaced `Execute.Assertion` with `assertionChain
```
IT-VBFK, line 58:
```suggestion
* Replaced `Execute.Assertion` with `assertionChain`
```

eNeRGy164:
No more savepoint 🎉

eNeRGy164:
Only 214 files to review for @jnyrup 🤣 
Luckily, the law of code reviews will apply: the more changes, the less review 😉

jnyrup:
> Only 214 files to review for @jnyrup 🤣 
> 
> Luckily, the law of code reviews will apply: the more changes, the less review 😉 

Are you trying to nerd snipe me? 🤣

I shall squint my eyes so hard when reviewing this, that any bug will come out crying.

dennisdoomen:
> Luckily, the law of code reviews will apply: the more changes, the less review

It's clear you haven't been reviewed by @jnyrup yet 😅

docs/_pages/extensibility.md
```diff
@@ -59,13 +66,46 @@ This is quite an elaborate example which shows some of the more advanced extensi
 
 * The `Subject` property is used to give the base-class extensions access to the current `DirectoryInfo` object.
 * `[CustomAssertion]` attribute enables correct subject identification, allowing Fluent Assertions to render more meaningful test fail messages
-* `Execute.Assertion` is the point of entrance into the internal fluent assertion API.
+* The variable `chain` of type `AssertionChain` is the point of entrance into the internal fluent I.
```
ITaluone, line 69:
```suggestion
* The variable `chain` of type `AssertionChain` is the point of entrance into the internal fluent API.
```

docs/_pages/introduction.md
```diff
@@ -162,13 +170,39 @@ The above will batch the two failures, and throw an exception at the point of di
 E.g. Exception thrown at point of dispose contains:
 
 ```text
-Expected value to be 10, but found 5.
+Expected value to be 10, but found 5 (difference of -5).
```
ITaluone, line 173: I think if the difference is less than or equal to ABS(10) no difference is shown (at least for non-floating point types).

Src/FluentAssertions/AndWhichConstraint.cs
```diff
@@ -2,45 +2,84 @@
...
+        : base(parent)
     {
-        TMatchedElement[] matchedElements = matchedConstraint.ToArray();
+        getSubject = new Lazy<TSubject>(() => SingleOrDefault(subjects));
+
+        this.assertionChain = assertionChain;
+        this.pathPostfix = pathPostfix;
+    }
+
+    private static TSubject SingleOrDefault(IEnumerable<TSubject> subjects)
```
ITaluone, line 75: 🤔 I wonder if this should be named `SingleOrThrow`, becasue there is no `Default` returned like in LINQs `SingleOrDefault`
Or return the first item if >1 are found?

dennisdoomen: I agree.

ITaluone: > I agree.

For `SingleOrThrow` or `SingleOrFirst`?

docs/_pages/introduction.md
dennisdoomen: Yeah, you're right. Thanks.

Src/FluentAssertions/AndWhichConstraint.cs
dennisdoomen: Actually, I now named it `Single`, since `Single` on collections will also throw if the collection contains more than one item.

docs/_pages/introduction.md
IT-VBFK: Am I missing something? 100 - 95 is also difference < ABS(10) 🤔

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
```diff
@@ -67,7 +69,6 @@ namespace FluentAssertions
         public static FluentAssertions.Specialized.NonGenericAsyncFunctionAssertions Should([System.Diagnostics.CodeAnalysis.NotNull] this System.Func<System.Threading.Tasks.Task> action) { }
         public static FluentAssertions.Primitives.GuidAssertions Should(this System.Guid actualValue) { }
         public static FluentAssertions.Primitives.NullableGuidAssertions Should(this System.Guid? actualValue) { }
-        public static FluentAssertions.Streams.BufferedStreamAssertions Should([System.Diagnostics.CodeAnalysis.NotNull] this System.IO.BufferedStream actualValue) { }
```
lg2de, line 70: I'm wondering whether is change is related to the AssertionScope feature.
This also applies to the removal of `BufferedStreamAssertions`.
I did not find this in your PR description.

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
```diff
@@ -461,7 +462,7 @@ namespace FluentAssertions.Collections
         public FluentAssertions.AndConstraint<TAssertions> NotBeSubsetOf(System.Collections.Generic.IEnumerable<T> unexpectedSuperset, string because = "", params object[] becauseArgs) { }
         public FluentAssertions.AndConstraint<TAssertions> NotContain(System.Collections.Generic.IEnumerable<T> unexpected, string because = "", params object[] becauseArgs) { }
         public FluentAssertions.AndConstraint<TAssertions> NotContain(System.Linq.Expressions.Expression<System.Func<T, bool>> predicate, string because = "", params object[] becauseArgs) { }
-        public FluentAssertions.AndWhichConstraint<TAssertions, T> NotContain(T unexpected, string because = "", params object[] becauseArgs) { }
+        public FluentAssertions.AndConstraint<TAssertions> NotContain(T unexpected, string because = "", params object[] becauseArgs) { }
```
lg2de, line 465: I'm wondering whether this (breaking) change is related to the AssertionScope feature.
This applies to `GenericDictionaryAssertions` too.
In your PR description you wrote: "Replace all usages of AndWhichConstraint with AndWhich". But this not applies here and I did not find an explanation why this is required or useful in this context.

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
```diff
@@ -1119,76 +1120,63 @@ namespace FluentAssertions.Events
...
-        public FluentAssertions.Execution.GivenSelector<T> Given<T>(System.Func<T> selector) { }
         public bool HasFailures() { }
-        public FluentAssertions.Execution.AssertionScope WithDefaultIdentifier(string identifier) { }
-        public FluentAssertions.Execution.AssertionScope WithExpectation(string message, params object[] args) { }
     }
     public class Continuation
     {
-        public FluentAssertions.Execution.IAssertionScope Then { get; }
-        public static bool op_Implicit(FluentAssertions.Execution.Continuation continuation) { }
+        public FluentAssertions.Execution.AssertionChain Then { get; }
```
lg2de, line 1174: Did you check whether creating an interface `IAssertionChain` would be helpful for the FA unit tests itself (now and in future)?

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
```diff
@@ -1119,76 +1120,63 @@ namespace FluentAssertions.Events
 }
 namespace FluentAssertions.Execution
 {
+    public sealed class AssertionChain
```
lg2de, line 1123: In your PR description you wrote: "Every Should method will create a new object called Assertion".
I only found this new `Assertion*` class. What I'm missing?

Src/FluentAssertions/Execution/AssertionChain.cs
```diff
@@ -0,0 +1,355 @@
+﻿using System;
+using System.Globalization;
+using System.Linq;
+using System.Threading;
+using FluentAssertions.Common;
+
+namespace FluentAssertions.Execution;
+
+public sealed class AssertionChain
```
lg2de, line 9: Would be great to have some more API documentation on this new core class.
There are also some public members left without documentation.

Src/FluentAssertions/Execution/AssertionChain.cs
```diff
@@ -0,0 +1,355 @@
...
+            AssertionChain assertionChain = Instance.Value;
+            Instance.Value = null;
+            return assertionChain;
+        }
+
+        return new AssertionChain(() => AssertionScope.Current,
+            () => FluentAssertions.CallerIdentifier.DetermineCallerIdentity());
+    }
+
+    private AssertionChain(Func<AssertionScope> getCurrentScope, Func<string> getCallerIdentifier)
```
lg2de, line 67: The one and only constructor is private?
This makes writing some tests very complicated, at least for extensions of FA.

lg2de (commented):
I was wondering a bit that only few unit tests were changed.
This is good because it indicates that the public API is quite stable.
But wondering what effect these changes should have. For me it is quite unclear from description of this PR as well as the referenced issue.

Src/FluentAssertions/Execution/AssertionChain.cs
dennisdoomen: You're supposed to use `GetOrCreate`. Can you explain why you would need to access this constructor directly?

Src/FluentAssertions/Execution/AssertionChain.cs
dennisdoomen: I documented the class itself, but copied most of the other members directly from the previous version of `AssertionScope`.

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
dennisdoomen: I've updated the PR description with how I ultimately implemented it. The main reason of introducing `AssertionChain` was to allow amending the caller identifier when you chain another assertion using `Which`. See also `upgradingtov7.md`

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
dennisdoomen: Why would we need it?

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
dennisdoomen: I've updated the PR description. I started calling this new thing `Assertion`, but later switched to `AssertionChain`.

Src/FluentAssertions/Execution/AssertionChain.cs
lg2de: Currently I do not have an explicit use case.
But for writing tests explicit with or for `AssertionChain` I must not use a static instance.
This is an experience of writing public libraries. Everything which I build initially private or internal will get some day public...

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
lg2de: For the same reason I added to `AssertionChain`.
Interfaces allows mocking in unit test, I'm sure you know this.
So, according to my experience nearly everything provided to public should be an interface instead of a class to be mockable.
The old version of this property returned an interface too. Why did you decide in the past to provide an interface, but not today?

Src/FluentAssertions/Execution/AssertionChain.cs
dennisdoomen: This is absolutely true. However, `AssertionChain.GetOrCreate` always returns a new instance, unless you use `ReuseOnce`, but that only works once, and travels over the async control flow.

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
dennisdoomen: Why on earth would you ever want to mock an assertion chain? Can you give me an example?

lg2de: I do not have a real example. I analyzed just with my theoretical eye.
I think, public API should use interfaces as much as possible.

When I'm not able to create my very own, non-static instance of `AssertionChain`  then I would like to have an instance where `AssertionChain` is used.

We can postpone this to the time when such change is required for a specific test. (I do not think of a user test but on an FA internal test.)

dennisdoomen: > I think, public API should use interfaces as much as possible.

I do not share that opinion unfortunately.

ITaluone: > I think, public API should use interfaces as much as possible.

IMHO this applies to projects like https://github.com/quartznet/quartznet where the only task of the "library" is to provide a "shape" and let the user do the whole implementation work.
This doesn't fully apply to FA (apart from the whole custom `Formatters` and `IEquivalencyStep`s).

docs/_pages/upgradingtov7.md
```diff
@@ -10,3 +10,142 @@ sidebar:
...
+```
+
+Prior to version 7, if the `HaveElement` assertion succeeded, but the `NotBeNull` failed, you would get the following exception:
+
+    Expected element to contain 1 item(s), but found 3: {<child />, <child />, <child />}.
+
+Now, in v7, it'll will return the following:
+
+    Expected element/child to contain 1 item(s), but found 3: {<child />, <child />, <child />}.
+
```
ITaluone, line 125: Is this example correct?
1) `NotBeNull` ?
2) Is this `HaveCount(1)` checking anything? I think `.Which` throws, because there isn't a single returned matched element => actually returns two
3) Does this failure message really fits (the new one)? I think (apart from throwing when not `Single`) `Expected element/child to contain 1 item(s)...` isn't quite right? 🤔
Because when I read the failure, I assume following xml structure:

```xml
<parent>
    <child>
        <this_one_containing_item />
    </child>
    <child>
        <this_one_containing_item />
    </child>
</parent>
```

dennisdoomen: No, because in this case, `HaveElement` returns an `AndWhichConstraint` that takes an `IEnumerable<T>`.

Src/FluentAssertions/Types/MethodBaseAssertions.cs
```diff
@@ -41,20 +44,23 @@ public AndConstraint<TAssertions> HaveAccessModifier(
...
 
-        if (success)
+        if (assertionChain.Succeeded)
         {
             CSharpAccessModifier subjectAccessModifier = Subject.GetCSharpAccessModifier();
 
-            Execute.Assertion
+            var subject = assertionChain.HasOverriddenCallerIdentifier
+                ? assertionChain.CallerIdentifier
+                : "method " + Subject.ToFormattedString();
```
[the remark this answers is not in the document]
dennisdoomen: You're right. I missed that and will need to lazy-evaluate that everywhere.

Tests/FluentAssertions.Specs/Execution/AssertionScopeSpecs.cs
```diff
@@ -326,12 +352,12 @@ public void HandleFailure(string message)
...
         internal class FailWithStupidMessageAssertionStrategy : IAssertionStrategy
         {
-            public IEnumerable<string> FailureMessages => new string[0];
+            public IEnumerable<string> FailureMessages => Array.Empty<string>();
 
             public void HandleFailure(string message) =>
                 Services.ThrowException("Good luck with understanding what's going on!");
 
-            public IEnumerable<string> DiscardFailures() => new string[0];
+            public IEnumerable<string> DiscardFailures() => Array.Empty<string>();
```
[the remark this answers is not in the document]
Corniel: That can be.  But be aware you do not get an empty array back, but an other empty collection. (using `[]` here is fine by me though).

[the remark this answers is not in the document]
jnyrup: Have you tried looking at the lowered code?

[SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABATARgLABQGADAAQY4oDchhFAzADwCWAdgC4B8pAYhBAAoAlKQC83AIJQoAQwCeAOgCiAWwAO7OSw6dhNIjiZsupAEIyowsdwDaAXSpA=)

Src/FluentAssertions/Primitives/NullableGuidAssertions.cs
```diff
@@ -24,9 +24,12 @@ public NullableGuidAssertions(Guid? value)
...
 {
-    public NullableGuidAssertions(Guid? value)
-        : base(value)
+    private readonly AssertionChain assertionChain;
+
+    public NullableGuidAssertions(Guid? value, AssertionChain assertionChain)
+        : base(value, assertionChain)
     {
+        this.assertionChain = assertionChain;
     }
```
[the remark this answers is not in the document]
jnyrup: We stored `assertionChain` both on `NullableGuidAssertions` and it's base class `GuidAssertions`.
So my question is about whether `NullableGuidAssertions` should use the `assertionChain` from it's base class.
E.g. `ReferenceTypeAssertions` exposes `CurrentAssertionChain`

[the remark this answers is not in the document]
dennisdoomen: I prefer not to expose base-class properties, unless I have to. In this case, I don't feel I have to.

Tests/FluentAssertions.Specs/Execution/AssertionScopeSpecs.cs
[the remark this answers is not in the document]
dennisdoomen: The lowered code? Is that a new thing? 😜

Tests/FluentAssertions.Specs/Collections/GenericDictionaryAssertionSpecs.ContainValues.cs
```diff
@@ -65,6 +65,23 @@ public void
...
+        {
+            [1] = "One",
+            [2] = "Two"
+        };
+
+        // Act
+        Action act = () => dictionary.Should().ContainValues("Two", "One").Which.Should().Contain("Three");
+
+        // Assert
+        act.Should().Throw<XunitException>().WithMessage("Expected dictionary[1 and 2]*to contain*Three*");
```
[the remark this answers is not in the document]
dennisdoomen: Yeah, thought hard about it, but couldn't find a decent alternative. You have a better suggestion?

Tests/FluentAssertions.Specs/Execution/AssertionScopeSpecs.cs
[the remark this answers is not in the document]
jnyrup: Lowering is when the compiler rewrites higher level syntax into lower level syntax. 
You can sometimes also hear the term desugaring when turning syntax sugar into plain syntax.

Src/FluentAssertions/Types/MethodBaseAssertions.cs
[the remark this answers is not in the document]
dennisdoomen: 🤔Trying to come up with a new idiom for evaluating elements in the `FailWith` statement in a deferred way.

Tests/FluentAssertions.Specs/Execution/AssertionScopeSpecs.cs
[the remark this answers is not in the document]
dennisdoomen: And I thought your auto-correct messed up ;-)

Tests/FluentAssertions.Specs/Formatting/FormatterSpecs.cs
```diff
@@ -254,13 +254,13 @@ public void When_the_object_is_a_user_defined_type_it_should_show_the_name_on_th
...
+            .Which.Message.Should().Match(
             """
             Expected stuff to be <null>, but found FluentAssertions.Specs.Formatting.FormatterSpecs+StuffRecord
             {
-                RecordChildren = {10, 20, 30, 40}, 
-                RecordDescription = "description", 
-                RecordId = 42, 
+                RecordChildren = {10, 20, 30, 40},*
+                RecordDescription = "description",*
+                RecordId = 42,*
```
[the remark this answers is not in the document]
dennisdoomen: I forgot to undo that after I fixed this in the formatter.

Src/FluentAssertions/Equivalency/Steps/DictionaryEquivalencyStep.cs
```diff
@@ -23,15 +27,15 @@ protected override EquivalencyResult OnHandle(Comparands comparands, IEquivalenc
...
-                        context.AsDictionaryItem<object, IDictionary>(key));
+                    nestedValidator.AssertEquivalencyOf(new Comparands(subject[key], expectation[key], typeof(object)), context.AsDictionaryItem<object, IDictionary>(key));
                 }
                 else
                 {
                     context.Tracer.WriteLine(member =>
                         Invariant(
                             $"Comparing dictionary item {key} at {member.Description} between subject and expectation"));
 
+                    assertionChain.WithCallerPostfix($"[{key.ToFormattedString()}]").ReuseOnce();
```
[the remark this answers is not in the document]
dennisdoomen: True. But how can we prevent that? Maybe by always using `ToString` and accepting that you might end up with the type description?

Tests/Approval.Tests/ApprovedApi/FluentAssertions/net47.verified.txt
```diff
@@ -1610,13 +1587,14 @@ namespace FluentAssertions.Numeric
...
-        public NumericAssertions(T value) { }
+        public NumericAssertions(T value, FluentAssertions.Execution.AssertionChain assertionChain) { }
     }
     public class NumericAssertions<T, TAssertions>
         where T :  struct, System.IComparable<T>
         where TAssertions : FluentAssertions.Numeric.NumericAssertions<T, TAssertions>
     {
-        public NumericAssertions(T value) { }
+        public NumericAssertions(T value, FluentAssertions.Execution.AssertionChain assertionChain) { }
+        public FluentAssertions.Execution.AssertionChain CurrentAssertionChain { get; }
```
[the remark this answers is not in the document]
dennisdoomen: Yes, since `NumericAssertions` doesn't rely on `ReferenceTypeAssertions`.

Src/FluentAssertions/Types/MethodBaseAssertions.cs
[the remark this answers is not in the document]
dennisdoomen: Turns out that we already solved this problem. 

```
            assertionChain
                .ForCondition(!Subject!.CanRead)
                .FailWith(() =>
                {
                    var subjectDescription = assertionChain.HasOverriddenCallerIdentifier
                        ? assertionChain.CallerIdentifier
                        : "property " + Subject.ToFormattedString();

                    return new FailReason($"Did not expect {subjectDescription} to have a getter{{reason}}.");
                });
```

Issue #2556, opened 2024-01-14
Add `BeValidJson` extensions for strings 


vbreuss:
### Background and motivation

As discussed in #2205 we want support for `System.Text.Json`. I would suggest to keep it as similar as possible to the existing extension package [`FluentAssertions.Json`](https://github.com/fluentassertions/fluentassertions.json).

### API Proposal

Add the following methods to the `StringAssertions` (similar to [`StringAssertionsExtensions` in `FluentAssertions.Json`](https://github.com/fluentassertions/fluentassertions.json/blob/master/Src/FluentAssertions.Json/StringAssertionsExtensions.cs)):
```C#
  public AndWhichConstraint<TAssertions, JsonElement> BeValidJson(string because = "", params object[] becauseArgs);
  public AndWhichConstraint<TAssertions, JsonElement> BeValidJson(JsonDocumentOptions options, string because = "", params object[] becauseArgs);
```

These methods should assert that the string is a valid JSON document.

### API Usage

```C#
string subject = """{ "id": 42, "admin": true }""";
subject.Should().BeValidJson();
```

### Are you willing to help with a proof-of-concept (as PR in that or a separate repo) first and as pull-request later on?

Yes, please assign this issue to me.

dennisdoomen:
I like that too. What about you @jnyrup ?

dennisdoomen:
@jnyrup ?

vbreuss:
I updated the proposal to use `JsonElement` instead of `JsonDocument`, as latter is `IDisposable` and should not be returned directly.
See [this comment](https://github.com/fluentassertions/fluentassertions/pull/2558#discussion_r1478863042) for more information.

vbreuss:
I won't work on this any more [due to the license change](https://github.com/fluentassertions/fluentassertions/pull/2943#issuecomment-2591040698)!

PR #2558, opened 2024-01-14
Add `BeValidJson` extensions for strings

vbreuss:
Draft for #2556 

Add the following methods to the `StringAssertions` (similar to [`StringAssertionsExtensions` in `FluentAssertions.Json`](https://github.com/fluentassertions/fluentassertions.json/blob/master/Src/FluentAssertions.Json/StringAssertionsExtensions.cs)):
```csharp
public AndWhichConstraint<TAssertions, JsonDocument> BeValidJson(string because = "", params object[] becauseArgs);
public AndWhichConstraint<TAssertions, JsonDocument> BeValidJson(JsonDocumentOptions options, string because = "", params object[] becauseArgs);
```

These methods should assert that the string is a valid JSON document.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Primitives/StringAssertions.cs
+++ b/Src/FluentAssertions/Primitives/StringAssertions.cs
@@ -3,6 +3,9 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+#if NET6_0_OR_GREATER
+using System.Text.Json;
+#endif
 using System.Text.RegularExpressions;
 using FluentAssertions.Common;
 using FluentAssertions.Equivalency;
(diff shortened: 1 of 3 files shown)


Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -1973,6 +1976,62 @@ public AndConstraint<TAssertions> NotBeLowerCased(string because = "", params ob
...
+            json = JsonDocument.Parse(Subject, options);
+        }
+        catch (JsonException jsonException)
+        {
+            Execute.Assertion
+                .BecauseOf(because, becauseArgs)
+                .FailWith("Expected {context:string} to be valid JSON{reason}, but parsing failed with {0}.", jsonException.Message);
+        }
+
+        return new AndWhichConstraint<TAssertions, JsonDocument>((TAssertions)this, json);
```
ivarne, line 2031: This way of using JsonDocument seems to be against the documentation that states that it is important to dispose (or clone) JsonDocument instances. I'm not sure I fully understand the consequences they warn about here, but my suggestion would be to use the raw `Utf8JsonReader` for `BeValidJson` and not provide the document for further usage.

https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/use-dom#jsondocument-is-idisposable

> JsonDocument builds an in-memory view of the data into a pooled buffer. Therefore the JsonDocument type implements IDisposable and needs to be used inside a using block.

vbreuss:
I won't work on this any more [due to the license change](https://github.com/fluentassertions/fluentassertions/pull/2943#issuecomment-2591040698)!

Issue #2566, opened 2024-01-16
Incorrect line breaks handling with custom string comparer


bart-vmware:
### Description

Follow-up for https://github.com/fluentassertions/fluentassertions/pull/2413#issuecomment-1891809390. #2413 was supposed to fix this, but it still doesn't work correctly.

### Reproduction Steps

Checkout the latest develop branch (600d3603) and run the below tests. Both tests should succeed, but they fail with:
```
Xunit.Sdk.XunitException
Expected logLines to be equivalent to "one\r\ntwo\r\nthree", but it misses some extra whitespace at the end.
   at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message) in C:\Source\Repos\fluentassertions\Src\FluentAssertions\Execution\XUnit2TestFramework.cs:line 38
   at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message) in C:\Source\Repos\fluentassertions\Src\FluentAssertions\Execution\TestFrameworkProvider.cs:line 39
   at FluentAssertions.Execution.DefaultAssertionStrategy.HandleFailure(String message) in C:\Source\Repos\fluentassertions\Src\FluentAssertions\Execution\DefaultAssertionStrategy.cs:line 18
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc) in C:\Source\Repos\fluentassertions\Src\FluentAssertions\Execution\AssertionScope.cs:line 274
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc) in C:\Source\Repos\fluentassertions\Src\FluentAssertions\Execution\AssertionScope.cs:line 242
   at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args) in C:\Source\Repos\fluentassertions\Src\FluentAssertions\Execution\AssertionScope.cs:line 296
   at FluentAssertions.Primitives.StringEqualityStrategy.ValidateAgainstSuperfluousWhitespace(IAssertionScope assertion, String subject, String expected) in C:\Source\Repos\fluentassertions\Src\FluentAssertions\Primitives\StringEqualityStrategy.cs:line 69
   at FluentAssertions.Primitives.StringEqualityStrategy.ValidateAgainstMismatch(IAssertionScope assertion, String subject, String expected) in C:\Source\Repos\fluentassertions\Src\FluentAssertions\Primitives\StringEqualityStrategy.cs:line 23
[...]

PR #2569, opened 2024-01-20
Avoid treating `"\\r\\n"` as new line.

vbreuss:
As mentioned in #2566 treating the verbatim string `@"\r\n"` as a newline seems suspicius, as it is equivalent to `"\\r\\n"`.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Common/StringExtensions.cs
+++ b/Src/FluentAssertions/Common/StringExtensions.cs
@@ -118,8 +118,7 @@ internal static class StringExtensions
     public static string RemoveNewLines(this string @this)
     {
         return @this.Replace("\n", string.Empty, StringComparison.Ordinal)
-            .Replace("\r", string.Empty, StringComparison.Ordinal)
-            .Replace(@"\r\n", string.Empty, StringComparison.Ordinal);
+            .Replace("\r", string.Empty, StringComparison.Ordinal);
     }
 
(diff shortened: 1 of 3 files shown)


vbreuss:
This code was added a long time ago in #84 ([see here](https://github.com/fluentassertions/fluentassertions/pull/84/files#diff-afb7da184f4c4a85d327db5f03007ae8ce9b5f2c7ae04a179b93c692232db626)).

dennisdoomen (approved):

dennisdoomen:
I think the title is misleading. You're merely removing some code that doesn't do anything.

vbreuss:
> I think the title is misleading. You're merely removing some code that doesn't do anything.

Not really. The following test failed previously, but now no longer:
```csharp
        [Fact]
        public void Does_not_treat_escaped_newlines_as_newlines()
        {
            // Arrange
            string actual = "te\r\nst";
            string expect = "te\\r\\nst";

            // Act / Assert
            actual.Should().NotMatchEquivalentOf(expect);
        }
```

image

Do you think it makes sense, to add this as a unit test?

dennisdoomen:
> Not really. The following test failed previously, but now no longer

Ah, in that case, you're right and this is actually fixing a bug. Probably also needs a release notes entry then.

jnyrup (approved):

Issue #2374, opened 2023-10-10
BeEquivalentTo incorrectly compare JsonElements?


Smurfa:
### Description

I am trying to compare if two json strings are equivalent (don't care about whitespaces or order). Using suggestion from previous issue: https://github.com/fluentassertions/fluentassertions/issues/1212 yields incorrect results.

### Reproduction Steps

```csharp
[Fact]
public void Test()
{
    var foo = JsonDocument.Parse("""{"test":"something"}""").RootElement;
    var bar = JsonDocument.Parse("""{"test":"something else"}""").RootElement;

    foo.Should().BeEquivalentTo(bar, opt => opt.ComparingByMembers<JsonElement>());
}
```

### Expected behavior

Would expect this to fail, "something" != "something else".

### Actual behavior

Test passes.

### Configuration

.NET 7.0
xunit 2.5.0
FluentAssertions 6.12.0

### Are you willing to help with a pull-request?

No

jnyrup:
My snippet in #1212 only compares the `ValueKind` property of a `JsonElement`.
The design of `JsonElement` doesn't let us use the built-in structural equivalency which reflects over the properties on the type.
To make it work, you'll have to write a custom `IEquivalencyStep`.

Here's some code I hacked together in five minutes.
It works for your example, but it most likely doesn't work in the general case, so no guarantees are given.

```cs
[Fact]
public void Test()
{
    var foo = JsonDocument.Parse("""{"test":"something"}""").RootElement;
    var bar = JsonDocument.Parse("""{"test":"something else"}""").RootElement;

    foo.Should().BeEquivalentTo(bar, opt => opt.Using(new JsonObjectEquivalencyStep()));
}

public class JsonObjectEquivalencyStep : IEquivalencyStep
{
    public EquivalencyResult Handle(Comparands comparands, IEquivalencyValidationContext context, IEquivalencyValidator nestedValidator)
    {
        if (comparands.Subject is not JsonElement subject || comparands.Expectation is not JsonElement expectation)
        {
            return EquivalencyResult.ContinueWithNext;
        }

        if (subject.ValueKind is JsonValueKind.Object && expectation.ValueKind is JsonValueKind.Object)
        {
            var newComparands = new Comparands(subject.EnumerateObject(), expectation.EnumerateObject(), typeof(IEnumerable<JsonProperty>));
            nestedValidator.RecursivelyAssertEquality(newComparands, context);
            return EquivalencyResult.AssertionCompleted;
[...]

PR #2576, opened 2024-02-04
Add JsonElementEquivalencyStep

ivarne:
Taking some inspiration from https://github.com/fluentassertions/fluentassertions/issues/2374, I improved the implementation of `IEquivalencyStep` to actually keep the path when recursing into a json structure and added some tests.

I'm a little puzzled about how to compare json numbers. In JS they are 64 bit floating point, so I ended up using `GetDouble()` so that `100` and `1e2` is equivalent, but used the string for comparing `NaN`, `Inf` and for generating the error message.

As this is part of adding proper support for `System.Text.Json`, I'm not sure how the release notes should be written.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Equivalency/Steps/JsonElementEquivalencyStep.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/JsonElementEquivalencyStep.cs
@@ -72,9 +72,11 @@ public class JsonElementEquivalencyStep : IEquivalencyStep
     {
         // Json does not specify a range for numbers, but allows scientific syntax
         // (1E2 for 100), so we can't just compare strings.
-        // The solution here is to first check if the numbers are equal in 64bit floating point
-        // then compare the actual raw text to get a proper error message.
-        if (!NumbersEqualsIgnoreInfAndNan(subject.GetDouble(), expectation.GetDouble()))
+        // The solution here is to run multiple checks
+        // 1. Check if they are both representable as 64 bit integers and compare them
+        // 2. If not check if they are equal in 64bit floating point
+        // 3. Then compare the actual raw text to get a proper error message.
+        if (!NumbersEqualsIgnoreInfAndNan(subject, expectation))
         {
             var subjectString = subject.GetRawText();
             var expectationString = expectation.GetRawText();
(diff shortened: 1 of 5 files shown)


dennisdoomen:
> I'm a little puzzled about how to compare json numbers. In JS they are 64 bit floating point, so I ended up using `GetDouble()` so that `100` and `1e2` is equivalent, but used the string for comparing `NaN`, `Inf` and for generating the error message.

According to ChatGPT:

Here's a strategy you can use to parse a JSON number without knowing in advance if it's a double or an integer:

Parse the JSON string to get a JsonDocument or directly work with a JsonElement if you have one.
Use the JsonElement.TryGetInt32, TryGetInt64, TryGetDouble, etc., methods to attempt to get the number in the desired format.
Based on the success of the TryGet methods, you can decide how to handle the value.

ITaluone:
> The build failures seem to be something about GitVersion. Anything I need to do?

Rebase to current develop and force push afterwards

dennisdoomen:
> As this is part of adding proper support for `System.Text.Json`, I'm not sure how the release notes should be written.

You were a bit ahead of the process. Normally we start with an API proposal before people start to implement a PR. So let's wait until the issue is approved.

ivarne:
Before participating in discussions on APIs I had to make sure I understood the possibilities and limitations of the library, and when I had something that seemed useful I decided to share.

dennisdoomen:
I've approved the issue, so this PR can move forward. Sorry for the delay. We were too occupied with some large-scale breaking changes.

ivarne:
That's great news! I'm really looking forward to better json support in fluent assertions.

After a quick reread of the PR, I don't see anything that stands out as something I should fix because of the approval. If you have any suggestions for extra tests or anything else that makes sense for me to do, I'll look into it.

dennisdoomen:
You'll have to rebase on `develop` and adopt to the breaking changes. And we need some documentation updates and release notes as mentioned in the PR checklist.

ivarne:
I think I managed to do the update. I have a few decisions left
* Where does this fit in the docs? (I could not find docs on other equivalency steps (eg xml))
* I implemented strict ordering for json arrays, and loose ordering for json objects. Should that be configurable somehow?
* Loose comparison on numbers (they sometimes end up not strictly equal)?
* Case sensitivity? I forwarded string comparisons for values, but should we respect that for property names too?
* The error message for number comparison could possibly not talk about the position of the failure.
* Could there be a way to specify that object properties missing from expected should be ignored? How would that configuration look?

dennisdoomen:
> Where does this fit in the docs? (I could not find docs on other equivalency steps (eg xml))

https://github.com/fluentassertions/fluentassertions/blob/develop/docs/_pages/objectgraphs.md

> * I implemented strict ordering for json arrays, and loose ordering for json objects. Should that be configurable somehow?

By default, collections are compared in a loosely ordered fashion, but this is configurable through the `WithStrictOrdering` overloads

> Loose comparison on numbers (they sometimes end up not strictly equal)?

I'll check the implementation when it's ready to review.

> Case sensitivity? I forwarded string comparisons for values, but should we respect that for property names too?

No, property names must match exactly. For real properties, we can change that using `WithMapping`

> The error message for number comparison could possibly not talk about the position of the failure.

What do you mean with that?

> Could there be a way to specify that object properties missing from expected should be ignored? How would that configuration look?

No, we don't support this right now.

PR #1660, opened 2021-08-22
Resolved a significant performance degradation in `BeEquivalentTo`

dennisdoomen:
This was caused by the increased dependency on the CallerIdentifier, and the additional work caller identification takes in v6.

#1657

Issue #1657, opened 2021-08-20
Performance issue in IEnumerable .BeEquivalentTo on enumerables with different order when going from v5 to v6.0.0


cjvanwyk3:
### Description

The BeEquivalentTo method for IEnumerable, when not using strict ordering or any options, now performs very poorly on version 6.0.0.  For example To compare 100 objects in version 5 it is a reasonable amount of time, ~1 second, but in version 6 it is > 10 seconds. The amount of time it takes depends on the number of properties as well.

### Complete minimal example reproducing the issue

```cs
ExampleObject GetObject(int i)
{
    return new ExampleObject
    {
        Id = i.ToString(),
        Value1 = i.ToString(),
        Value2 = i.ToString(),
        Value3 = i.ToString(),
        Value4 = i.ToString(),
        Value5 = i.ToString(),
        Value6 = i.ToString(),
        Value7 = i.ToString(),
        Value8 = i.ToString(),
        Value9 = i.ToString(),
        Value10 = i.ToString(),
        Value11 = i.ToString(),
        Value12 = i.ToString(),
    };
}

var list1 = new List<ExampleObject>();
var list2 = new List<ExampleObject>();
var maxAmount = 100;
for (var i = 0; i < maxAmount; i++)
{
    list1.Add(GetObject(i));
    list2.Add(GetObject(maxAmount - 1 - i));
}

var timer = Stopwatch.StartNew();
list1.Should().BeEquivalentTo(list2);
timer.Stop();
Debug.WriteLine(timer.ElapsedMilliseconds); // 18,539 on version 6.0.0 1,512 on version 5.10.3

public class ExampleObject
{
    [Key]
    public string Id { get; set; }

    public string Value1 { get; set; }

    public string Value2 { get; set; }

    public string Value3 { get; set; }

    public string Value4 { get; set; }

    public string Value5 { get; set; }

    public string Value6 { get; set; }

    public string Value7 { get; set; }

    public string Value8 { get; set; }

    public string Value9 { get; set; }

    public string Value10 { get; set; }

    public string Value11 { get; set; }

    public string Value12 { get; set; }
}
```

### Expected behavior:

Expected better performance

### Actual behavior:

Poor performance

### Versions

Version 6.0.0

### Additional Information

One workaround is to perform the ordering and use the performant strict ordering. 
So instead of this which performs poorly:
`list1.Should().BeEquivalentTo(list2);`
Something like this would be used which performs well:
[...]

PR #2597, opened 2024-02-29
Add support for `Satisfy` on `ReferenceTypeAssertions`

### LINKED ISSUE #2516: Allow `Should().Satisfy()` on objects and not only collections
### Background and motivation

When I'm asserting many properties on multiple instances, I often find it hard to keep track of the context of what I'm asserting.
Like with collections, we have a `collection.Should().AllSatisfy(item => { /* do FluentAssertions on item here */ }`, but I feel like this is missing for single instances.

The closest thing I've found is `instance.Should().Match(/* something that returns true or false */)`, which is not really fluent as I see it.
The alternative is to do object comparison, where I need to construct the expected object before and then do `instance.Should().BeEquivalentTo(expected)`.

What I'm suggesting is to allow scoped assertions on a single instance, like this:
```cs
var person = new Person { Name = "Peter", Age = 30 };

person.Should().Satisfy(p => 
{
    p.Name.Should().StartWith("P");
    p.Age.Should().BeGreaterThan(25);
});
```

Providing this, also allows me to extract generic assertions as methods, which can be reused.

I am aware that the above is equivalent to this:
```cs
var person = new Person { Name = "Peter", Age = 30 };

person.Name.Should().StartWith("P");
person.Age.Should().BeGreaterThan(25);
```

However, if I have many assertions where the variable names are long, it becomes quite a lot of noise in the tests.
For instance:
```cs
Person someInstanceOfPersonComingFormACallbackFromNSubstitute;

someInstanceOfPersonComingFormACallbackFromNSubstitute.Name.Should().StartWith("P");
someInstanceOfPersonComingFormACallbackFromNSubstitute.Age.Should().BeGreaterThan(25);
```

This would be, IMHO, more readable like this:
```cs
Person someInstanceOfPersonComingFormACallbackFromNSubstitute;

someInstanceOfPersonComingFormACallbackFromNSubstitute.Should().Satisfy(p => 
{
    p.Name.Should().StartWith("P");
    p.Age.Should().BeGreaterThan(25);
});
```

I am well aware that there might be a good reason for this to not exist and that it's possible to create a simple extension method like this:
```cs
public static void ShouldSatisfy<T>(this  T actual, Action<T> assert) where T : class
{
    assert(actual);
}
```

I also understand that this approach could lead to confusion since the `Satisfy()` method would have to exist on all types being asserted.

It might also be the case that my tests are too complex, however, it's an integration test, so I have a lot to assert, which makes the need clear to me.

### Alternative Concerns

_No response_

### Are you willing help with a pull-request?

Yes, please assign this issue to me.

siewers:
Added support for `Satisfy` on `ReferenceTypeAssertions` allowing to use element inspectors as an alternative to using predicates on Match.

Example:
```cs
var myDto = new Dto { Name = "Some Name" };

myDto.Should().Satisfy<Dto>(dto => dto.Name.Should().Be("Some Name"));
```

Fixed #2516

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
+++ b/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
@@ -423,15 +423,18 @@ public abstract class ReferenceTypeAssertions<TSubject, TAssertions>
     }
 
     /// <summary>
-    /// Asserts that the criteria provided by the element inspector is satisfied.
+    /// Allows combining one or more assertions using the other assertion methods that this library offers on an instance of <typeparamref name="T"/>.
     /// </summary>
-    /// <param name="expected">The element inspector which must be satisfied by the <typeparamref name="TSubject" />.</param>
+    /// <remarks>
+    /// If multiple assertions executed by the <paramref name="assertion"/> fail, they will be raised as a single failure.
+    /// </remarks>
+    /// <param name="assertion">The element inspector which must be satisfied by the <typeparamref name="TSubject" />.</param>
     /// <returns>An <see cref="AndConstraint{T}" /> which can be used to chain assertions.</returns>
-    /// <exception cref="ArgumentNullException"><paramref name="expected"/> is <see langword="null"/>.</exception>
-    public AndConstraint<TAssertions> Satisfy<T>(Action<T> expected)
+    /// <exception cref="ArgumentNullException"><paramref name="assertion"/> is <see langword="null"/>.</exception>
+    public AndConstraint<TAssertions> Satisfy<T>(Action<T> assertion)
         where T : TSubject
     {
-        Guard.ThrowIfArgumentIsNull(expected, nameof(expected), "Cannot verify an object against a <null> inspector.");
+        Guard.ThrowIfArgumentIsNull(assertion, nameof(assertion), "Cannot verify an object against a <null> inspector.");
 
         Execute.Assertion
             .ForCondition(Subject is T)
@@ -442,7 +445,7 @@ public abstract class ReferenceTypeAssertions<TSubject, TAssertions>
 
         using (var assertionScope = new AssertionScope())
         {
-            expected((T)Subject);
+            assertion((T)Subject);
             failuresFromInspector = assertionScope.Discard();
         }
 
--- a/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
+++ b/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
@@ -443,7 +443,7 @@ public abstract class ReferenceTypeAssertions<TSubject, TAssertions>
             .Then
             .ForCondition(Subject is T)
             .WithDefaultIdentifier(Identifier)
-            .FailWith("Expected {context:object} to be assignable to {0}{reason}, but {1} is not.", typeof(T), Subject!.GetType());
+            .FailWith("Expected {context:object} to be assignable to {0}{reason}, but {1} is not.", typeof(T), Subject?.GetType());
 
         if (success)
         {
--- a/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
+++ b/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using System.Linq.Expressions;
 using FluentAssertions.Common;
 using FluentAssertions.Execution;
@@ -417,6 +418,44 @@ public abstract class ReferenceTypeAssertions<TSubject, TAssertions>
             .BecauseOf(because, becauseArgs)
             .WithDefaultIdentifier(Identifier)
             .FailWith("Expected {context:object} to match {1}{reason}, but found {0}.", Subject, predicate);
+
+        return new AndConstraint<TAssertions>((TAssertions)this);
+    }
+
+    /// <summary>
+    /// Asserts that the criteria provided by the element inspector is satisfied.
+    /// </summary>
+    /// <param name="expected">The element inspector which must be satisfied by the <typeparamref name="TSubject" />.</param>
+    /// <returns>An <see cref="AndConstraint{T}" /> which can be used to chain assertions.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="expected"/> is <see langword="null"/>.</exception>
+    public AndConstraint<TAssertions> Satisfy<T>(Action<T> expected)
+        where T : TSubject
+    {
+        Guard.ThrowIfArgumentIsNull(expected, nameof(expected), "Cannot verify an object against a <null> inspector.");
+
+        Execute.Assertion
+            .ForCondition(Subject is T)
+            .WithDefaultIdentifier(Identifier)
+            .FailWith("Expected {context:object} to be assignable to {0}{reason}, but {1} is not.", typeof(T), Subject.GetType());
+
+        string[] failuresFromInspector;
+
+        using (var assertionScope = new AssertionScope())
+        {
+            expected((T)Subject);
+            failuresFromInspector = assertionScope.Discard();
+        }
+
+        if (failuresFromInspector.Length > 0)
+        {
+            string failureMessage = Environment.NewLine
+                + string.Join(Environment.NewLine, failuresFromInspector.Select(x => x.IndentLines()));
+
+            Execute.Assertion
+                .WithDefaultIdentifier(Identifier)
+                .WithExpectation("Expected {context:object} to match inspector, but the inspector was not satisfied:", Subject)
+                .FailWithPreFormatted(failureMessage);
+        }
 
         return new AndConstraint<TAssertions>((TAssertions)this);
     }
--- a/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
+++ b/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
@@ -436,7 +436,7 @@ public abstract class ReferenceTypeAssertions<TSubject, TAssertions>
     {
         Guard.ThrowIfArgumentIsNull(assertion, nameof(assertion), "Cannot verify an object against a <null> inspector.");
 
-        Execute.Assertion
+        var success = Execute.Assertion
             .ForCondition(Subject is not null)
             .WithDefaultIdentifier(Identifier)
             .FailWith("Expected {context:object} to be assignable to {0}{reason}, but found <null>.", typeof(T))
@@ -445,23 +445,26 @@ public abstract class ReferenceTypeAssertions<TSubject, TAssertions>
             .WithDefaultIdentifier(Identifier)
             .FailWith("Expected {context:object} to be assignable to {0}{reason}, but {1} is not.", typeof(T), Subject!.GetType());
 
-        string[] failuresFromInspector;
-
-        using (var assertionScope = new AssertionScope())
+        if (success)
         {
-            assertion((T)Subject);
-            failuresFromInspector = assertionScope.Discard();
-        }
+            string[] failuresFromInspector;
 
-        if (failuresFromInspector.Length > 0)
-        {
-            string failureMessage = Environment.NewLine
-                + string.Join(Environment.NewLine, failuresFromInspector.Select(x => x.IndentLines()));
+            using (var assertionScope = new AssertionScope())
+            {
+                assertion((T)Subject);
+                failuresFromInspector = assertionScope.Discard();
+            }
 
-            Execute.Assertion
-                .WithDefaultIdentifier(Identifier)
-                .WithExpectation("Expected {context:object} to match inspector, but the inspector was not satisfied:", Subject)
-                .FailWithPreFormatted(failureMessage);
+            if (failuresFromInspector.Length > 0)
+            {
+                string failureMessage = Environment.NewLine
+                    + string.Join(Environment.NewLine, failuresFromInspector.Select(x => x.IndentLines()));
+
+                Execute.Assertion
+                    .WithDefaultIdentifier(Identifier)
+                    .WithExpectation("Expected {context:object} to match inspector, but the inspector was not satisfied:", Subject)
+                    .FailWithPreFormatted(failureMessage);
+            }
         }
 
         return new AndConstraint<TAssertions>((TAssertions)this);
--- a/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
+++ b/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
@@ -437,9 +437,13 @@ public abstract class ReferenceTypeAssertions<TSubject, TAssertions>
         Guard.ThrowIfArgumentIsNull(assertion, nameof(assertion), "Cannot verify an object against a <null> inspector.");
 
         Execute.Assertion
+            .ForCondition(Subject is not null)
+            .WithDefaultIdentifier(Identifier)
+            .FailWith("Expected {context:object} to be assignable to {0}{reason}, but found <null>.", typeof(T))
+            .Then
             .ForCondition(Subject is T)
             .WithDefaultIdentifier(Identifier)
-            .FailWith("Expected {context:object} to be assignable to {0}{reason}, but {1} is not.", typeof(T), Subject.GetType());
+            .FailWith("Expected {context:object} to be assignable to {0}{reason}, but {1} is not.", typeof(T), Subject!.GetType());
 
         string[] failuresFromInspector;
 
--- a/Tests/FluentAssertions.Specs/Primitives/ReferenceTypeAssertionsSpecs.Satisfy.cs
+++ b/Tests/FluentAssertions.Specs/Primitives/ReferenceTypeAssertionsSpecs.Satisfy.cs
@@ -1,0 +1,220 @@
+﻿using System;
+using Xunit;
+using Xunit.Sdk;
+
+namespace FluentAssertions.Specs.Primitives;
+
+public partial class ReferenceTypeAssertionsSpecs
+{
+    public class Satisfy
+    {
+        [Fact]
+        public void When_object_satisfies_inspector_it_should_not_throw()
+        {
+            // Arrange
+            var someObject = new object();
+
+            // Act
+            Action act = () => someObject.Should().Satisfy<object>(x => x.Should().NotBeNull());
+
+            // Assert
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_typed_object_satisfies_inspector_it_should_not_throw()
+        {
+            // Arrange
+            var someObject = new PersonDto
+            {
+                Name = "Name Nameson",
+                Birthdate = new DateTime(2000, 1, 1),
+            };
+
+            // Act
+            Action act = () => someObject.Should().Satisfy<PersonDto>(o => o.Age.Should().BeGreaterThan(0));
+
+            // Assert
+            act.Should().NotThrow();
+        }
+
+        [Fact]
+        public void When_object_does_not_satisfy_the_inspector_it_should_throw()
+        {
+            // Arrange
+            var someObject = new object();
+
+            // Act
+            Action act = () => someObject.Should().Satisfy<object>(o => o.Should().BeNull("it is not initialized yet"));
+
+            // Assert
+            act.Should().Throw<XunitException>().WithMessage(
+                $"""
+                 Expected {nameof(someObject)} to match inspector, but the inspector was not satisfied:
+                 *Expected o to be <null> because it is not initialized yet, but found System.Object*
+                 """);
+        }
+
+        [Fact]
+        public void When_a_typed_object_does_not_satisfy_the_inspector_it_should_throw()
+        {
+            // Arrange
+            const string personName = "Name Nameson";
+
+            var somePerson = new PersonDto
+            {
+                Name = personName,
+                Birthdate = new DateTime(1973, 9, 20),
+            };
+
+            // Act
+            Action act = () =>
+                somePerson.Should().Satisfy<PersonDto>(d => d.Name.Should().HaveLength(0, "it is not initialized yet"));
+
+            // Assert
+            act.Should().Throw<XunitException>().WithMessage(
+                $"""
+                 Expected {nameof(somePerson)} to match inspector, but the inspector was not satisfied:
+                 *Expected d.Name with length 0 because it is not initialized yet, but found string "{personName}" with length {personName.Length}.
+                 """);
+        }
+
+        [Fact]
+        public void When_a_complex_typed_object_does_not_satisfy_inspector_it_should_throw()
+        {
+            // Arrange
+            var someComplexDto = new PersonAndAddressDto
+            {
+                Person = new PersonDto
+                {
+                    Name = "Buford Howard Tannen",
+                    Birthdate = new DateTime(1937, 3, 26),
+                },
+                Address = new AddressDto
+                {
+                    Street = "Mason Street",
+                    Number = "1809",
+                    City = "Hill Valley",
+                    Country = "United States",
+                    PostalCode = "CA 91905",
+                },
+            };
+
+            // Act
+            Action act = () => someComplexDto.Should().Satisfy<PersonAndAddressDto>(dto =>
+            {
+                dto.Person.Should().Satisfy<PersonDto>(person =>
+                {
+                    person.Name.Should().Be("Biff Tannen");
+                    person.Age.Should().Be(48);
+                    person.Birthdate.Should().Be(new DateTime(1937, 3, 26));
+                });
+
+                dto.Address.Should().Satisfy<AddressDto>(address =>
+                {
+                    address.Street.Should().Be("Mason Street");
+                    address.Number.Should().Be("1809");
+                    address.City.Should().Be("Hill Valley, San Diego County, California");
+                    address.Country.Should().Be("United States");
+                    address.PostalCode.Should().Be("CA 91905");
+                });
+            });
+
+            // Assert
+            act.Should().Throw<XunitException>().WithMessage(
+                $"""
+                 Expected {nameof(someComplexDto)} to match inspector, but the inspector was not satisfied:
+                 *Expected dto.Person to match inspector*
+                 *Expected person.Name*
+                 *Expected dto.Address to match inspector*
+                 *Expected address.City*
+                 """);
+        }
+
+        [Fact]
+        public void When_a_typed_object_is_satisfied_against_an_incorrect_type_it_should_throw()
+        {
+            // Arrange
+            var personDto = new PersonDto();
+
+            // Act
+            Action act = () => personDto.Should().Satisfy<AddressDto>(dto => dto.Should().NotBeNull());
+
+            // Assert
+            act.Should().Throw<XunitException>()
+                .WithMessage(
+                    $"Expected {nameof(personDto)} to be assignable to {typeof(AddressDto)}, but {typeof(PersonDto)} is not.");
+        }
+
+        [Fact]
+        public void When_a_typed_object_does_not_match_multiple_inspectors_it_should_throw()
+        {
+            // Arrange
+            var somePerson = new PersonDto
+            {
+                Name = "Name Nameson",
+                Birthdate = new DateTime(2000, 1, 1),
+            };
+
+            // Act
+            Action act = () => somePerson.Should().Satisfy<PersonDto>(d =>
+            {
+                d.Name.Should().Be("Someone Else");
+                d.Age.Should().BeLessThan(20);
+                d.Birthdate.Should().BeAfter(new DateTime(2001, 1, 1));
+            });
+
+            // Assert
+            act.Should().Throw<XunitException>().WithMessage(
+                $"""
+                 Expected {nameof(somePerson)} to match inspector, but the inspector was not satisfied:
+                 *Expected d.Name*
+                 *Expected d.Age*
+                 *Expected d.Birthdate*
+                 """);
+        }
+
+        [Fact]
+        public void When_object_is_satisfied_against_a_null_inspector_it_should_throw()
+        {
+            // Arrange
+            var someObject = new object();
+
+            // Act
+            Action act = () => someObject.Should().Satisfy<object>(null);
+
+            // Assert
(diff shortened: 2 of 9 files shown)


Tests/FluentAssertions.Specs/Primitives/ReferenceTypeAssertionsSpecs.Satisfy.cs
```diff
@@ -0,0 +1,213 @@
...
+public partial class ReferenceTypeAssertionsSpecs
+{
+    [Fact]
+    public void When_object_satisfies_inspector_it_should_not_throw()
+    {
+        // Arrange
+        var someObject = new object();
+
+        // Act
+        Action act = () => someObject.Should().Satisfy<object>(x => x.Should().NotBeNull());
```
ITaluone, line 16: We usually do not write this `Action x....; x.Should.NotThrow()` anymore, because if it fails, it fails anyway

ITaluone:
Missing release notes?

Tests/FluentAssertions.Specs/Primitives/ReferenceTypeAssertionsSpecs.Satisfy.cs
siewers: Okay?
It just combines the act and assert, which I find a bit awkward. The reason it fails is also unclear from the test since it might fail for any number of reasons.
However, I'm fine doing whatever you find best and there is no consistency in how the tests are written, so it's difficult to get right. I tried to follow the pattern of the tests for the class I am modifying, but even in that class, there is no consistency.

@dennisdoomen Is there a preferred way of writing tests? I can't find any guidelines, so I'd like some "official" answer to this.
The same goes for the usage of nested classes, although I can that pattern being used in several places, I just missed that.

siewers:
I didn't add it to the release notes yet as I'm unsure what version this will be in.

Tests/FluentAssertions.Specs/Primitives/ReferenceTypeAssertionsSpecs.Satisfy.cs
dennisdoomen: For the test that are not supposed to throw, we do not use the `NotThrow` anymore.

siewers: I've merged the Act / Assert for non-throwing assertions

ITaluone:
> I didn't add it to the release notes yet as I'm unsure what version this will be in.

I think while on develop for the next version it doesn't really matter

siewers:
@ITaluone Alright, I added it the the most recent section. I guess it will be aligned when the final release goes out, however, it might be a bit confusing for people using the Alpha versions to see the release notes not reflecting the latest version.

siewers:
@dennisdoomen @ITaluone I believe I've addressed all comments now.

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
```diff
@@ -416,6 +417,44 @@ public AndConstraint<TAssertions> NotBeAssignableTo(Type type, string because =
         return new AndConstraint<TAssertions>((TAssertions)this);
     }
 
+    /// <summary>
+    /// Asserts that the criteria provided by the element inspector is satisfied.
```
dennisdoomen, line 421: 🤔 I think that this can be phrased better. We're providing a method that is supposed to be used to execute one or more assertions on an object. And if you run multiple assertions that fail, they will be reported as one failure. We need to capture that in the name.

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
```diff
@@ -416,6 +417,44 @@ public AndConstraint<TAssertions> NotBeAssignableTo(Type type, string because =
...
+    /// </summary>
+    /// <param name="expected">The element inspector which must be satisfied by the <typeparamref name="TSubject" />.</param>
```
dennisdoomen, line 423: 🤔 Same here. We're not satisfying an inspector but executing one or more assertions.

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
```diff
@@ -416,6 +417,44 @@ public AndConstraint<TAssertions> NotBeAssignableTo(Type type, string because =
...
+
+        string[] failuresFromInspector;
+
+        using (var assertionScope = new AssertionScope())
+        {
+            expected((T)Subject);
+            failuresFromInspector = assertionScope.Discard();
+        }
+
+        if (failuresFromInspector.Length > 0)
```
dennisdoomen, line 444: 🤔 Do we even need to do this? What if you just relied on the assertion scope to throw a combined message? What kind of information would you be missing?

docs/_pages/basicassertions.md
```diff
@@ -70,6 +70,35 @@ dummy.Should().Match<string>(d => (d == "System.Object"));
...
+        Name = "Some product name",
+        Price = 19.95,
+        SKU = "ABC12345",
+        Store = new Store
+            {
+                Country = "Germany",
+                Quantity = 42
+            }
+    };
+productDto.Should().Satisfy<ProductDto>(dto =>
```
dennisdoomen, line 87: ⛏️ The layout is a bit messed up here.

docs/_pages/releases.md
```diff
@@ -15,6 +15,7 @@ sidebar:
 * Ensure compatibility with .NET 8 - [#2466](https://github.com/fluentassertions/fluentassertions/pull/2466)
 * Add support for NUnit 4 - [#2483](https://github.com/fluentassertions/fluentassertions/pull/2483)
 * Added `NotBeIn` to check if a `DateTime` is not in a given `DateTimeKind` - [#2536](https://github.com/fluentassertions/fluentassertions/pull/2536)
+* Introduced a new `Satisfy` method to `ReferenceTypeAssertions` to allow for custom assertions - [#2597](https://github.com/fluentassertions/fluentassertions/pull/2597)
```
dennisdoomen, line 18:
```suggestion
* Introduced a new `Satisfy` method available to all reference types to allow for nested assertions - [#2597](https://github.com/fluentassertions/fluentassertions/pull/2597)
```

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
siewers: I don't know, I'll test it to see if it's necessary 👍

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
siewers: How about:
>Asserts that all element inspectors for the specific instance are satisfied
I'll see if it can be aligned more closely to what is written elsewhere (this one is similar to what is written for `Match`).

dennisdoomen: I doubt whether the term "element inspectors" will resonate well with people (even though we've used it before). 

What about 

```csharp
    /// <summary>
    /// Allows combining one or more assertions using the other assertion methods that this library offers on an instance of <typeparamref name="T"/>.
    /// </summary>
    /// <remarks>
    /// If multiple assertions executed by the <paramref name="assertion"/> fail, they will be raised as a single failure.
    /// </remarks>
    /// <param name="assertion">The element inspector which must be satisfied by the <typeparamref name="TSubject" />.</param>
    /// <returns>An <see cref="AndConstraint{T}" /> which can be used to chain assertions.</returns>
    /// <exception cref="ArgumentNullException"><paramref name="assertion"/> is <see langword="null"/>.</exception>
    public AndConstraint<TAssertions> Satisfy<T>(Action<T> assertion)
```

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
dennisdoomen: And?

dennisdoomen:
> I believe I've addressed all comments now.

Apparently not.

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
siewers: Sorry, I haven't had time to do it yet (I'm working on this in my company's "personal time", which is only every second week).
I will look into it this evening.

siewers: I tested the changes you suggested, and it looks like we will be losing the outer object context. There is also no indentation on the nested assertions, which I think is useful.

I don't know what is preferred though, however, this is how it's done in `Match`, and I believe the result should be the same or similar, no?

Using a single assertion scope:
```
Expected person.Name to be the same string, but they differ at index 1:
    ↓ (actual)
  "Buford Howard Tannen"
  "Biff Tannen"
    ↑ (expected).
Expected person.Age to be 48, but found 87 (difference of 39).
Expected address.City to be "Hill Valley, San Diego County, California" with a length of 41, but "Hill Valley" has a length of 11, differs near "y" (index 10).
```

Using `Discard`:
```
Expected complexDto to match inspector, but the inspector was not satisfied:
	Expected dto.Person to match inspector, but the inspector was not satisfied:
		Expected person.Name to be the same string, but they differ at index 1:
		    ↓ (actual)
		  "Buford Howard Tannen"
		  "Biff Tannen"
		    ↑ (expected).
		Expected person.Age to be 48, but found 87 (difference of 39).
	Expected dto.Address to match inspector, but the inspector was not satisfied:
		Expected address.City to be "Hill Valley, San Diego County, California" with a length of 41, but "Hill Valley" has a length of 11, differs near "y" (index 10).
```

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
siewers: Sounds good to me :+1:

docs/_pages/basicassertions.md
siewers: Is this better?

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
dennisdoomen: You have not addressed this one either?

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
dennisdoomen: When I rely solely on the assertion scope, I get this

```
Expected dto.Person to be the same string, but they differ at index 1:
    ↓ (actual)
  "Buford Howard Tannen"
  "Biff Tannen"
    ↑ (expected).
Expected dto.Person to be 48, but found 87 (difference of 39).
Expected dto.Address to be "Hill Valley, San Diego County, California" with a length of 41, but "Hill Valley" has a length of 11, differs near "y" (index 10).
```

I personally feel like this is much more readable without all the noisy mentioning of the inspector (which is the wrong term anyway), don't you agree?

siewers: Something is different in your case. It's not the Person or Address properties of the dto that failed, it's the person.Name and address.City. I think the assertion is presenting the wrong thing.
However, if it's at least showing the correct property in context of the parent, I believe either case is fine. Again, this is based on the implementation Match, so I'm not really trying to make any decisions, I'm just attempting to following an existing pattern 🙂

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
siewers: I must be confused. I thought both comments were addressed by your suggestion? Honestly, I'm not used to working with GitHub code reviews, so I'm sorry if I'm a bit slow (I'm using Azure DevOps daily, so it's a very different way of working 🙂)

ITaluone:
@siewers Could you rebase to current develop and force push again?

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
dennisdoomen: Just so we talk about the same thing. Which test did you use to reproduce this scenario?

siewers: The `Complex_typed_object_not_satisfying_inspector_throws` test.

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
```diff
@@ -418,15 +418,18 @@ public AndConstraint<TAssertions> NotBeAssignableTo(Type type, string because =
...
-    /// <param name="expected">The element inspector which must be satisfied by the <typeparamref name="TSubject" />.</param>
+    /// <remarks>
+    /// If multiple assertions executed by the <paramref name="assertion"/> fail, they will be raised as a single failure.
+    /// </remarks>
+    /// <param name="assertion">The element inspector which must be satisfied by the <typeparamref name="TSubject" />.</param>
     /// <returns>An <see cref="AndConstraint{T}" /> which can be used to chain assertions.</returns>
-    /// <exception cref="ArgumentNullException"><paramref name="expected"/> is <see langword="null"/>.</exception>
-    public AndConstraint<TAssertions> Satisfy<T>(Action<T> expected)
+    /// <exception cref="ArgumentNullException"><paramref name="assertion"/> is <see langword="null"/>.</exception>
+    public AndConstraint<TAssertions> Satisfy<T>(Action<T> assertion)
```
IT-VBFK, line 429: I think `assertion` is a bit misleading here 🤔

dennisdoomen: You're supposed to use that action to build a composite assertion, so I think it's a suitable term.

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
siewers: What's the conclusion on this one?

dennisdoomen: I need to investigate this one a bit better, but haven't had any free time this week. Will try to take a look this weekend.

dennisdoomen: I tried again and that made me realize that I don't like the parts about `Expected complexDto to match inspector, but the inspector was not satisfied:`. My suggestion is to change the implementation to 

```csharp
        using (new AssertionScope(CallerIdentifier.DetermineCallerIdentity()))
        {
            assertion((T)Subject);
        }

```

This will make your test result in 

```
Expected dto.Person to be the same string, but they differ at index 1:
    ↓ (actual)
  "Buford Howard Tannen"
  "Biff Tannen"
    ↑ (expected).
Expected dto.Person to be 48, but found 87 (difference of 39).
Expected dto.Address to be "Hill Valley, San Diego County, California" with a length of 41, but "Hill Valley" has a length of 11, differs near "y" (index 10).
```

I'll create a separate PR to make sure that nested calls to `AssertionScope(string)` will chain the context you provide so that you'll get something like 

```
Expected complexDto/dto.Person to be the same string, but they differ at index 1:
    ↓ (actual)
  "Buford Howard Tannen"
  "Biff Tannen"
    ↑ (expected).
Expected complexDto/dto.Person to be 48, but found 87 (difference of 39).
Expected complexDto/dto.Address to be "Hill Valley, San Diego County, California" with a length of 41, but "Hill Valley" has a length of 11, differs near "y" (index 10).
```

dennisdoomen:
Did you close it on purpose?

siewers:
Wow, no, I didn't... I don't know what happened.

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
dennisdoomen: Created #2607

ITaluone:
```
Expected complexDto/dto.Person to be the same string, but they differ at index 1:
    ↓ (actual)
  "Buford Howard Tannen"
  "Biff Tannen"
    ↑ (expected).
Expected complexDto/dto.Person to be 48, but found 87 (difference of 39).
```

I think what @siewers meant was, that this message presenting the wrong "objects" on which the assertion fails.
Both failing on `Person` but the first mentioning a `string` which doesn't match, and the latter suddenly mentioning (for the same property) an `int` which is failing.

dennisdoomen: When you rebase this PR on `develop` (and thereby include #2607), you'll get this:

```
Expected complexDto to match inspector, but the inspector was not satisfied:
	Expected dto.Person to match inspector, but the inspector was not satisfied:
		Expected person.Name to be the same string, but they differ at index 1:
		    ↓ (actual)
		  "Buford Howard Tannen"
		  "Biff Tannen"
		    ↑ (expected).
		Expected person.Age to be 48, but found 87 (difference of 39).
	Expected dto.Address to match inspector, but the inspector was not satisfied:
		Expected address.City to be "Hill Valley, San Diego County, California" with a length of 41, but "Hill Valley" has a length of 11, differs near "y" (index 10).
```

It is still a bit noisy because of the superfluous "Expected complexDto to match inspector, but the inspector was not satisfied", but I'm pretty happy with the result.

siewers: I've rebased now, but am I supposed to then only the following in `Satisfy`?
```cs
using (new AssertionScope(CallerIdentifier.DetermineCallerIdentity()))
{
    assertion((T)Subject);
}
```
If I do that, I'm getting this assertion message, which isn't the same as you wrote (and doesn't really make sense) :)
```
Expected complexDto/dto.Person to be the same string, but they differ at index 1:
    ↓ (actual)
  "Buford Howard Tannen"
  "Biff Tannen"
    ↑ (expected).
Expected complexDto/dto.Person to be 48, but found 87 (difference of 39).
Expected complexDto/dto.Address to be "Hill Valley, San Diego County, California" with a length of 41, but "Hill Valley" has a length of 11, differs near "y" (index 10).
```

siewers: The problem with the message is that it doesn't mention the failed nested property names:
- `complexDto.Person.Name`
- `complexDto.Person.Age`
- `complexDto.Address.City`
If those exact property names aren't mentioned in the message, it's not helpful.

As it looks with the new changes, both the `Name` and `Age` properties gets merged into a single nonsensical error on the same `Person` property.
It's wrong to say that `Person` was expected to be the same string _and_ it was expected to be a certain number :)
Also, `Address` is another "complex" object, but the error message doesn't present the exact property that failed the assertion.

I hope that makes sense :)

dennisdoomen: I need to look into that (again). Later today.

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
```diff
@@ -416,6 +417,47 @@ public AndConstraint<TAssertions> NotBeAssignableTo(Type type, string because =
...
+        Execute.Assertion
+            .ForCondition(Subject is T)
+            .WithDefaultIdentifier(Identifier)
+            .FailWith("Expected {context:object} to be assignable to {0}{reason}, but {1} is not.", typeof(T), Subject.GetType());
+
+        string[] failuresFromInspector;
+
+        using (var assertionScope = new AssertionScope())
+        {
+            assertion((T)Subject);
```
jnyrup, line 443: If `Satisfy` is used inside an `AssertionScope`, then the execution won't stop if `Subject is T` above fails, meaning we would send in a `null` to `action`.

```cs
object subject = null;

Action act = () =>
{
    subject.Should().Satisfy<object>(x => x.Should().NotBeNull());
};
```

What we usually do to guard against this.
```cs
    .ForCondition(...)
    .FailWith(...);

if (success)
{
    Execute.Assertion
        .ForCondition(...)
        .FailWith(...);
}
```

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
dennisdoomen: Crap, you're right. The call to `new AssertionScope(CallerIdentifier.DetermineCallerIdentity())` doesn't actually help at all. Sorry about that. I thought I could fix that by changing some of the behavior `AssertionScope` uses. But it turns out, it doesn't. So maybe we can rewrite `Expected dto.Person to match inspector, but the inspector was not satisfied:` to be less verbose.

jnyrup: Please also note that `CallerIdentifier.DetermineCallerIdentity()` is very expensive to call and is only needed when the test fails.
See #1660 and #1657

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
```diff
@@ -416,6 +417,47 @@ public AndConstraint<TAssertions> NotBeAssignableTo(Type type, string because =
...
+    /// <exception cref="ArgumentNullException"><paramref name="assertion"/> is <see langword="null"/>.</exception>
+    public AndConstraint<TAssertions> Satisfy<T>(Action<T> assertion)
+        where T : TSubject
+    {
+        Guard.ThrowIfArgumentIsNull(assertion, nameof(assertion), "Cannot verify an object against a <null> inspector.");
+
+        Execute.Assertion
+            .ForCondition(Subject is T)
+            .WithDefaultIdentifier(Identifier)
+            .FailWith("Expected {context:object} to be assignable to {0}{reason}, but {1} is not.", typeof(T), Subject.GetType());
```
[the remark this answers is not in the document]
siewers: I've added a check for null, although the message is almost duplicated.
I also used `.Then` to chain the calls, but I'm not sure that does the same thing as you've suggested using the result of the first assertion to execute the next.

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
siewers: I don't think I understand exactly what you mean.
Is it about using an assertion scope within `Satisfy` like this:
```cs
personAndAddressDto.Should().Satisfy<PersonAndAddressDto>(dto => {
    dto.Person.Name.Should().NotBeNullOrEmpty();
    using(new AssertionScope())
    {
        dto.Address.Should().Satisfy<AddressDto>(address => address.City.Should().NotBeNullOrEmpty());
    }
}
```

After I added the null-check, it seems to work as I would expect.

I'm not sure why anyone would add a nested assertion scope though, but of course it needs to be handled correctly.

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
dennisdoomen: > Please also note that `CallerIdentifier.DetermineCallerIdentity()` is very expensive to call and is only needed when the test fails.

True, but it's no longer relevant. My suggestion doesn't work, so the original implementation is still the best.

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
dennisdoomen: What @jnyrup means is that if you're new `Satisfy` is wrapped in an `AssertionScope` (just like you do yourself), the `FailWith` call with _not_ throw an exception and return `false` instead.

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
```diff
@@ -416,6 +417,51 @@ public AndConstraint<TAssertions> NotBeAssignableTo(Type type, string because =
...
+    /// <summary>
+    /// Allows combining one or more assertions using the other assertion methods that this library offers on an instance of <typeparamref name="T"/>.
+    /// </summary>
+    /// <remarks>
+    /// If multiple assertions executed by the <paramref name="assertion"/> fail, they will be raised as a single failure.
+    /// </remarks>
+    /// <param name="assertion">The element inspector which must be satisfied by the <typeparamref name="TSubject" />.</param>
+    /// <returns>An <see cref="AndConstraint{T}" /> which can be used to chain assertions.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="assertion"/> is <see langword="null"/>.</exception>
+    public AndConstraint<TAssertions> Satisfy<T>(Action<T> assertion)
```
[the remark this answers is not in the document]
siewers: It might, but I don't know how to connect that to the nested assertions or how the resulting assertion message should look.
In practice, the nested assertions can have the `because`, but the parent would probably always have the message "because the assertions should be satisfied" or something like that.

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
siewers: I'm so sorry for the late reply. I will try to support that.

siewers: Does this test verify the expected behavior (it passes with the current implementation)?

```cs
[Fact]
public void Using_assertion_scope_with_null_subject()
{
    // Arrange
    object subject = null;

    // Act
    Action act = () =>
    {
        using (new AssertionScope())
        {
            subject.Should().Satisfy<object>(x => x.Should().NotBeNull());
        }
    };

    // Assert
        .WithMessage("Expected subject to be assignable to System.Object, but found <null>.");
}
```

If this is not what you mean, I need some help understanding the scenario you have in mind.

dennisdoomen (approved):

siewers:
@dennisdoomen I see you approved this PR, but it's blocked by a couple of checks. Is there anything further on my part I need to take care of?

IT-VBFK:
Rebase to current develop and force push

ITaluone:
@siewers To fix the current problems rebase again :)

siewers:
I've rebased, and all checks are passing. Can we merge this?

dennisdoomen:
> I've rebased, and all checks are passing. Can we merge this?

Waiting for a review from @jnyrup

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
jnyrup: Your `Using_assertion_scope_with_null_subject` reveals that using `Subject.GetType()` throws an NRE and needs to be `Subject?.GetType()`

If we also change `Typed_object_satisfied_against_incorrect_type_throws` to have 

```cs
Action act = () =>
{
    personDto.Should().Satisfy<AddressDto>(dto => dto.Should().NotBeNull());
};
```

that reveals that `assertion((T)Subject);` throws `System.InvalidCastException: 'Unable to cast object of type 'PersonDto' to type 'AddressDto'.'`

We usually handle this by return the return value of the assertion chain and check if it was successful.

```cs
    .WithDefaultIdentifier(Identifier)
    .FailWith("Expected {context:object} to be assignable to {0}{reason}, but found <null>.", typeof(T))
    .Then
    .ForCondition(Subject is T)
    .WithDefaultIdentifier(Identifier)
    .FailWith("Expected {context:object} to be assignable to {0}{reason}, but {1} is not.", typeof(T), Subject?.GetType());

if (success)
{
    string[] failuresFromInspector;

    using (var assertionScope = new AssertionScope())
    {
        assertion((T)Subject);
        failuresFromInspector = assertionScope.Discard();
    }

    if (failuresFromInspector.Length > 0)
    {
        string failureMessage = Environment.NewLine
            + string.Join(Environment.NewLine, failuresFromInspector.Select(x => x.IndentLines()));

        Execute.Assertion
            .WithDefaultIdentifier(Identifier)
            .WithExpectation("Expected {context:object} to match inspector, but the inspector was not satisfied:", Subject)
            .FailWithPreFormatted(failureMessage);
    }
}

```

siewers: I am unable to reproduce the exception with the suggested test.
Have you tested whether it throws an NRE?

I understand that the combination of `Execute.Assertion` and `AssertionScope` should behave in some specific way, but I don't know how that is.

In any case, I added the precondition check and the test, as suggested, even though there is no difference, regardless of what I do.

Could you please test it? I feel I'm wasting too much of your time because I don't understand the underlying problems you are trying to explain.

jnyrup: Run `Using_assertion_scope_with_null_subject` with a debugger attached and for a moment remove
https://github.com/fluentassertions/fluentassertions/blob/8656682fe60df5ebff2cd6cc7a0d3c140e96786b/Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs#L15

When using an `AssertionScope` failing calls to `ForCondition` doesn't immediately cause the entire assertion to stop.
This is usually used for doing multiple assertions on different subjects and get all failures in the output.
E.g. if both of these assertions fail we get two failure messages instead of only the one from the first assertion.
```cs
using (new AssertionScope())
{
    a.Should().BeFoo();
    b.Should().BeBar();
}
```

As `AssertionScope`s continues the execution, we also internally in FluentAssertions need to guard against the execution continuing.

So when wrapping the call to `Satisfy` in an `AssertionScope` the assertion continues executing after `Subject is not null` fails, which leads to the NRE when calling `Subject!.GetType()` on a null `Subject`.
image

The fix here is to change `Subject!.GetType()` to `Subject?.GetType()`

Your test `Using_assertion_scope_with_subject_satisfied_against_incorrect_type_throws` correctly exercises the case where  we continue after `Subject is T` fails and `if (success)` guards against casting `Subject` to `T`.
image

siewers: Ah, okay, now I'm starting to understand the problem.
The tests are not throwing the NREs, but internally, Fluent Assertions are not executing the assertion because it fails to handle the second precondition.

I mistakenly assumed `.Then` only got executed if the previous assertion succeeded.

Would it be better to split the checks instead of using `.Then`?

jnyrup (approved):

Src/FluentAssertions/Primitives/ReferenceTypeAssertions.cs
jnyrup: Your fix with replacing `!` with `?` is exactly what I would do.

PR #2607, opened 2024-03-17
Allow nested AssertionScopes to chain their context

dennisdoomen:
## IMPORTANT 


Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -48,25 +50,23 @@ public DeferredReportable(Func<string> valueFunc)
...
-    public AssertionScope(string context)
-        : this()
+    public AssertionScope()
+        : this(new CollectingAssertionStrategy())
     {
-        if (!string.IsNullOrEmpty(context))
-        {
-            Context = new Lazy<string>(() => context);
-        }
+        SetCurrentAssertionScope(this);
```
jnyrup, line 59: `SetCurrentAssertionScope` is already called by the chained constructor call.

Src/FluentAssertions/Execution/AssertionScope.cs
```diff
@@ -85,34 +85,42 @@ public AssertionScope(IAssertionStrategy assertionStrategy)
...
             FormattingOptions = parent.FormattingOptions.Clone();
+            Context = new Lazy<string>(() => JoinContext(parent.Context, context));
+        }
+        else
+        {
+            Context = context;
         }
     }
 
+    private static string JoinContext(params Lazy<string>[] contexts)
```
jnyrup, line 119: Suggestion on how to re-use context when possible.
```cs
private static Lazy<string> JoinContext(Lazy<string> parentContext, Lazy<string> context)
{
    if (parentContext is null)
    {
        return context;
    }

    if (context is null)
    {
        return parentContext;
    }

    return new(() => parentContext.Value + "/" + context.Value);
}
```

Tests/FluentAssertions.Specs/Execution/AssertionScope.MessageFormatingSpecs.cs
```diff
@@ -66,6 +66,22 @@ public void Message_should_use_the_lazy_name_of_the_scope_as_context()
             .WithMessage("Expected lazy foo to be equal to*");
     }
 
+    [Fact]
+    public void Nested_scopes_will_use_the_name_of_their_outer_scope_as_context()
```
jnyrup, line 70:
```suggestion
    public void Nested_scopes_use_the_name_of_their_outer_scope_as_context()
```

Src/FluentAssertions/Execution/AssertionScope.cs
dennisdoomen: Good catch (albeit harmless).

Src/FluentAssertions/Execution/AssertionScope.cs
dennisdoomen: Actually, they can both be `null`

jnyrup (approved):

PR #2669, opened 2024-06-10
Fix Qodana issue

IT-VBFK:
## IMPORTANT 


dennisdoomen:
> For the record: I would probably remove the `[StringSyntax("CompositeFormat")]` from the `FailReason`, but it could only be me..

Yes, I think that's the better option since it doesn't completely comply to that.

jnyrup:
Is it because we include our own `{context:yada}` in calls to `FailReason.Message` you believe it should be removed?
I'm curious what your reasoning is and how you found this 👀 

If yes and we agree that it should indeed be removed, then I guess `ContinuedAssertionScope` and `MessageBuilder` have the same problem.

IT-VBFK:
> I'm curious what your reasoning is and how you found this

That was pretty easy: [Latest develop build](https://github.com/fluentassertions/fluentassertions/actions/runs/9442792497/job/26005514554) (at time of opening this request) -> [Qodana upload link](https://qodana.cloud/projects/AgRgk/reports/Myo5J) -> See the error

> If yes and we agree that it should indeed be removed, then I guess ContinuedAssertionScope and MessageBuilder have the same problem.

It's up to you.. I just wanted to fix the red pipeline "x" which is caused by this qodana fail

dennisdoomen:
> It's up to you.. I just wanted to fix the red pipeline "x" which is caused by this qodana fail

I think we should remove the attribute.

IT-VBFK:
@jnyrup Do you see my point?

jnyrup:
Is your point "qodana reports it"?

As always I don't see qodana as the truth, but as a tool providing suggestions.

That said, I agree with what it points at here and think we should then also adjust the other places I point at.

IT-VBFK:
> Is your point "qodana reports it"?

Sort of. My point is I want to fix the pipeline. Either way is fine for me 😉

IT-VBFK:
> That said, I agree with what it points at here and think we should then also adjust the other places I point at.

Do you mean all methods in `ContinuedAssertionScope` and `MessageBuilder` which have this attribute?

dennisdoomen:
> Do you mean all methods in `ContinuedAssertionScope` and `MessageBuilder` which have this attribute?

Yes, anywhere where the attribute is used where the pattern doesn't comply.

IT-VBFK:
> Yes, anywhere where the attribute is used where the pattern doesn't comply.

Am I wrong, that this applies only to the public methods? If not, I would remove the attributes from all methods...

dennisdoomen:
> Am I wrong, that this applies only to the public methods? If not, I would remove the attributes from all methods...

Technically it would apply to any method.

IT-VBFK:
Qodana talking bullshit :)

Src/FluentAssertions/Execution/ContinuedAssertionScope.cs
```diff
@@ -99,7 +98,7 @@ public Continuation FailWith([StringSyntax("CompositeFormat")] string message, p
     }
 
     /// <inheritdoc/>
-    public IAssertionScope BecauseOf([StringSyntax("CompositeFormat")] string because, params object[] becauseArgs)
+    public IAssertionScope BecauseOf(string because, params object[] becauseArgs)
```
dennisdoomen, line 101: 🤔I'm pretty sure that `BecauseOf` complies to `string.Format` rules.

ITaluone, line 101:
```suggestion
    public IAssertionScope BecauseOf([StringSyntax("CompositeFormat")] string because, params object[] becauseArgs)
```

Src/FluentAssertions/Execution/ContinuedAssertionScope.cs
```diff
@@ -1,5 +1,4 @@
 using System;
-using System.Diagnostics.CodeAnalysis;
 
 namespace FluentAssertions.Execution;
```
ITaluone, line 3:
```suggestion
using System.Diagnostics.CodeAnalysis;

namespace FluentAssertions.Execution;
```

Src/FluentAssertions/Execution/ContinuedAssertionScope.cs
```diff
@@ -1,4 +1,5 @@
 using System;
+
```
ITaluone, line 2:
```suggestion
```

dennisdoomen (approved):

jnyrup (approved):

Issue #2668, opened 2024-06-09
[Feature]: Consider updating the ConfigurationManager transitive dependency version


ViktorHofer:
### Background and motivation

When referencing the FluentAssertions nuget package a five years old version of the System.Configuration.ConfigurationManager package gets referenced transitively. Those ancient 4.x.x versions of "System" nuget packages will eventually get marked as deprecated. Consider upgrading the ConfigurationManager version to something in support, i.e. 6.0.1.

This only impacts the net6.0 as it's the only TFM that references the package.

### Are you willing help with a pull-request?

Yes, please assign this issue to me.

dennisdoomen:
I think that should be okay, as long as we use the lowest version compatible with .NET 6.0. So indeed, 6.0.1. You agree @jnyrup ?

jnyrup:
Let's bump it. 
Thanks for the heads up @ViktorHofer

ITaluone:
What about bumping `System.Threading.Tasks.Extensions` to 4.5.4 in the same go?

Applies to netstandard2.0 and net47..

ViktorHofer:
Submitted https://github.com/fluentassertions/fluentassertions/pull/2673

jnyrup:
> What about bumping `System.Threading.Tasks.Extensions` to 4.5.4 in the same go?
> 
> Applies to netstandard2.0 and net47..

In general we want to use the lowest possible available libraries for the FA core library.

Imagine an application `app` having a dependency on `System.Threading.Tasks.Extensions@4.5.0`

Upgrading `System.Threading.Tasks.Extensions` in FA to 4.5.4 gives us this dependency graph.
```
tests -> app -> System.Threading.Tasks.Extensions@4.5.0
      -> FA -> System.Threading.Tasks.Extensions@4.5.4
```

Unit tests might now be testing a different behavior than what happens when `app` runs in production.
Especially tests-only libraries should be very aware of this.

That said, I'm fine with also bumping `System.Threading.Tasks.Extensions`

ITaluone:
> Imagine an application app having a dependency on System.Threading.Tasks.Extensions@4.5.0

What about the other way round? It's the same thing, no?

```
tests -> app -> System.Threading.Tasks.Extensions@4.5.4
      -> FA -> System.Threading.Tasks.Extensions@4.5.0
```

PR #2122, opened 2023-02-04
Declare System.Net.Http as a framework dependency

AArnott:
Declaring System.Net.Http as a nuget dependency for net47 is problematic for some consuming projects because it offers assembly version 4.1.1.3, whereas the FluentAssertions assembly itself references 4.2.0.0 of that same assembly. This causes the C# compiler to break the build for these projects.

The fix is to declare that System.Net.Http comes from an ordinary `Reference` item that is supplied by the target framework (which in fact it should as it does for FluentAssertions itself) instead of from the NuGet package.

It is interesting to call out that although this project previously declared `System.Net.Http` as a `PackageReference`, msbuild nevertheless chose to use the targeting framework version instead of the nuget version. This is ... not something I can explain. And many other projects have the same behavior, which is why it typically works for consuming projects. I don't know why exactly, but some projects do _not_ share this behavior, and those that actually use the nuget dependency that was declared here (before this change) would fail to compile.

Fixes #2121

## IMPORTANT 

* [x] If the PR touches the public API, the changes have been approved in a separate issue with the "api-approved" label.
* [x] The code complies with the [Coding Guidelines for C#](https://www.csharpcodingguidelines.com/).
* [x] The changes are covered by unit tests which follow the Arrange-Act-Assert syntax and the naming conventions such as is used [in these tests](../tree/develop/Tests/FluentAssertions.Equivalency.Specs/MemberMatchingSpecs.cs#L51-L430).
* [x] If the PR adds a feature or fixes a bug, please update [the release notes](../tree/develop/docs/_pages/releases.md) with a functional description that explains what the change means to consumers of this library, which are published on the [website](https://fluentassertions.com/releases).
* [x] If the PR changes the public API the changes needs to be included by running [AcceptApiChanges.ps1](../tree/develop/AcceptApiChanges.ps1) or [AcceptApiChanges.sh](../tree/develop/AcceptApiChanges.sh).
* [x] If the PR affects [the documentation](../tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).
    * [ ] Please also run `./build.sh --target spellcheck` or `.\build.ps1 --target spellcheck` before pushing and check the good outcome

PR #2673, opened 2024-06-17
Update FluentAssertions package dependencies

ViktorHofer:
Fixes https://github.com/fluentassertions/fluentassertions/issues/2668

Update the ConfigurationManager dependency to 6.0.1 and System.Threading.Tasks.Extensions to 4.5.4.

## IMPORTANT 

### DIFF
--- a/Tests/FluentAssertions.Extensibility.Specs/FluentAssertions.Extensibility.Specs.csproj
+++ b/Tests/FluentAssertions.Extensibility.Specs/FluentAssertions.Extensibility.Specs.csproj
@@ -31,4 +31,9 @@
   <ItemGroup>
     <ProjectReference Include="..\..\Src\FluentAssertions\FluentAssertions.csproj" />
   </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
+    <Reference Include="System.Net.Http" />
+  </ItemGroup>
+
 </Project>
--- a/Src/FluentAssertions/FluentAssertions.csproj
+++ b/Src/FluentAssertions/FluentAssertions.csproj
@@ -57,7 +57,7 @@
         <Compile Remove="Common/NullConfigurationStore.cs" />
       </ItemGroup>
       <ItemGroup>
-        <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.4.0" />
+        <PackageReference Include="System.Configuration.ConfigurationManager" Version="6.0.1" />
       </ItemGroup>
     </When>
     <When Condition="'$(TargetFramework)' == 'netstandard2.1'">
@@ -74,7 +74,7 @@
         <Compile Remove="EventRaisingExtensions.cs" />
       </ItemGroup>
       <ItemGroup>
-        <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.0" />
+        <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
       </ItemGroup>
     </When>
     <When Condition="'$(TargetFramework)' == 'net47'">
@@ -82,7 +82,7 @@
         <Compile Remove="Common/NullConfigurationStore.cs" />
       </ItemGroup>
       <ItemGroup>
-        <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.0" />
+        <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
         <TfmSpecificFrameworkAssemblyReferences Include="System.Net.Http">
           <TargetFramework>$(TargetFramework)</TargetFramework>
         </TfmSpecificFrameworkAssemblyReferences>
--- a/Tests/TestFrameworks/XUnit2.Specs/XUnit2.Specs.csproj
+++ b/Tests/TestFrameworks/XUnit2.Specs/XUnit2.Specs.csproj
@@ -18,4 +18,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
+    <Reference Include="System.Net.Http" />
+  </ItemGroup>
 </Project>
--- a/docs/_pages/releases.md
+++ b/docs/_pages/releases.md
@@ -79,7 +79,6 @@ sidebar:
 * Allow fluently calling `WithoutMatchingRules` and `WithoutSelectionRules` while using `BeEquivalentTo` - [#2457](https://github.com/fluentassertions/fluentassertions/pull/2457)
 * Removed `utcNow` overload for `.Monitor<T>()` - [#2629](https://github.com/fluentassertions/fluentassertions/pull/2629)
 * The semantics of `BeLowerCased`/`BeUpperCased` have been changed to align with the behavior of `ToLower`/`ToUpper` - [#2660](https://github.com/fluentassertions/fluentassertions/pull/2660)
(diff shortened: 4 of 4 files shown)


ViktorHofer:
Looking at the build failures...

Tests/FluentAssertions.Extensibility.Specs/FluentAssertions.Extensibility.Specs.csproj
```diff
@@ -31,4 +31,9 @@
   <ItemGroup>
     <ProjectReference Include="..\..\Src\FluentAssertions\FluentAssertions.csproj" />
   </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
+    <Reference Include="System.Net.Http" />
+  </ItemGroup>
```
jnyrup, line 37: For my edification, why is this necessary now?

Won't this break consumers of Fluent Assertions not having this set?

Context: we have had quite some trouble with `System.Net.Http`.
See e.g. #2122

ITaluone: Could most likely be because of the build failure. The log mentions somewhat related

dennisdoomen: What build failure are you referring to?

IT-VBFK: https://github.com/fluentassertions/fluentassertions/actions/runs/9546652380/job/26310522701

ViktorHofer: Yes, those references are now necessary as the two updated test projects don't automatically reference System.Net.Http.dll anymore. I assume you want the long answer, so here it is :)

The version bump of "System.Threading.Tasks.Extensions" from 4.5.0 to 4.5.4 caused this. With 4.5.4, the System.Threading.Tasks.Extensions.dll assembly doesn't depend on netstandard.dll anymore. The .NET SDK automatically adds netstandard facade assemblies for frameworks which support .NET Standard but which don't have the netstandard.dll assembly inbox (i.e. net461, net462, net47).

Back in 2020 the 4.5.4 package got updated to avoid binding redirects by not depending on netstandard.dll anymore. While this is a positive change (better perf and smaller output directory) and was requested by many customers, it's the reason why some projects now need to manually specify additional references.

Overall, this only affects customers on .NET Framework who target a version between 4.6.1 and 4.7.1 and who depend on types from assemblies that aren't referenced by default, i.e. `HttpResponseMessage` from `System.Net.Http.dll`. Our guidance here has been to add those reference items to the project file (or use VS's add reference UI).

https://github.com/fluentassertions/fluentassertions/pull/2122 lists great reasons why framework dependencies (called inbox) should be preferred and why the PR was submitted. cc @AArnott

dennisdoomen: Thanks for the extensive explanation. Do I understand you correctly that this may cause a breaking change for folks that target 4.6.1 - 4.7.1, use `HttpResponseMessage` and were relying on FluentAssertions to bring in that dependency?

ViktorHofer: That's my guess, yes. It should only affect projects that don't already reference other libraries that trigger the netstandard compat shims to be included. Might make sense to document that.

ITaluone: @jnyrup Or do you mean that the variable `TargetFrameworkIdentifier` isn't set per default?

What about adding it into the `<Choose><When>` construct, like so:

```xml
... 
<When Condition="'$(TargetFramework)' == 'net47'">
  <ItemGroup>
       <Compile Remove="Common/NullConfigurationStore.cs" />
  </ItemGroup>
  <ItemGroup>
       <Reference Include="System.Net.Http" />
  </ItemGroup>
....

```

lg2de: > That's my guess, yes. It should only affect projects that don't already reference other libraries that trigger the netstandard compat shims to be included. Might make sense to document that.

It could also be handled by the/a nuspec file to express this dependency when targeting e.g. net461.

ViktorHofer: > Or do you mean that the variable TargetFrameworkIdentifier isn't set per default?

The `TargetFrameworkIdentifier` property is set by the .NET SDK and doesn't need to be defined manually. It's available in .targets files or inside the project file for item groups and items. No need to set it yourself.

> It could also be handled by the/a nuspec file to express this dependency when targeting e.g. net461.

System.Net.Http.dll was just an example. There are other assemblies that aren't referenced by default on .NET Framework and need to be manually referenced. I don't think that FluentAssertions should do anything in regards to that.

dennisdoomen: > That's my guess, yes. It should only affect projects that don't already reference other libraries that trigger the netstandard compat shims to be included. Might make sense to document that.

In that case this should be added as a warning to the breaking changes section of the `releases.md`

dennisdoomen (approved):

ViktorHofer:
Hello @dennisdoomen, any update on when we can get this PR merged?

ITaluone:
It is waiting for @jnyrup's review.. As it seems, he is very busy these days :)

Tests/FluentAssertions.Extensibility.Specs/FluentAssertions.Extensibility.Specs.csproj
jnyrup: > > Or do you mean that the variable TargetFrameworkIdentifier isn't set per default?
> 
> The `TargetFrameworkIdentifier` property is set by the .NET SDK and doesn't need to be defined manually. It's available in .targets files or inside the project file for item groups and items. No need to set it yourself.
> 
> > It could also be handled by the/a nuspec file to express this dependency when targeting e.g. net461.
> 
> System.Net.Http.dll was just an example. There are other assemblies that aren't referenced by default on .NET Framework and need to be manually referenced. I don't think that FluentAssertions should do anything in regards to that.

It saddens me a bit if this library can't be used out-of-the-box without potentially fiddling with reference assemblies.
If this isn't needed for net472+ but only for net47 and net471, could we include a reference to System.Net.Http.dll only for those two TFMs and would that be problematic?

jnyrup: Great to hear.
And we just need to add reference assemblies explicitly in our test projects as we're using project references?

I notice that the generated nuspec file for the nupkg generated from this PR has
`<frameworkAssembly assemblyName="System.Net.Http" targetFramework=".NETFramework4.7, .NETFramework4.7" />`

where the one in 6.12 has
`<frameworkAssembly assemblyName="System.Net.Http" targetFramework=".NETFramework4.7" />`

Do you know what's going on?

ViktorHofer: > And we just need to add reference assemblies explicitly in our test projects as we're using project references?

Exactly. Consumers of FluentAssertions reference it via the package and automatically receive the necessary compiler references.

> Do you know what's going on?

No, not really. I don't know about this schema or which component exactly generates it. But I don't think it causes any issues, it's probably just more verbose.

jnyrup (approved):

PR #2677, opened 2024-06-19
Filter the list of events returned by RaisePropertyChangeFor

### LINKED ISSUE #393: ShouldRaisePropertyChangedFor followed by Count behaves unexpectedly (solution here)
I wanted to check the number of times `RaisePropertyChanged` had been raised for a particular property.

I expected to type:

```c#
myVm.ShouldRaisePropertyChangedFor(vm => vm.myProperty).Count().ShouldBeEquivalentTo(2);
```

This failed - the count returned was for ALL RaisePropertyChange events, not the specific one I am testing.

I have written a quick extension method:

```c#
public static int SpecificRaisePropertyChangedCount(this IEnumerable<RecordedEvent> value, string propertyName)
{
    return value.Select(r => r.Parameters)
        .SelectMany(s => s)
        .Count(p => p is PropertyChangedEventArgs && (p as PropertyChangedEventArgs).PropertyName == propertyName);
}
```

so you can now do:

```c#
myVm.ShouldRaisePropertyChangedFor(vm => vm.myProperty)
    .SpecificRaisePropertyChangedCount(nameof(myVm.myProperty))
    .ShouldBeEquivalentTo(2);
```

Sorry I haven't submitted this as a pull request - I'm up to my neck in it and no time to clone the repo etc, but feel free to add it or whatever (or tell me that I could have done it with the existing code easily ^_^ )

arocheleau:
The call to `RaisePropertyChangeFor` on `EventAssertions<T>` now returns a filtered list of events related to the property given as parameter.  The `NotRaisePropertyChangeFor` function has been revised as well to match the behavior of `RaisePropertyChangeFor` especially when the parameter provided is `null`.

This fixes #393.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Events/EventAssertions.cs
+++ b/Src/FluentAssertions/Events/EventAssertions.cs
@@ -123,7 +123,7 @@ public class EventAssertions<T> : ReferenceTypeAssertions<T, EventAssertions<T>>
                     Monitor.Subject, PropertyChangedEventName, propertyName, actualPropertyNames);
         }
 
-        return recording;
+        return recording.WithPropertyChanged(propertyName);
     }
 
     /// <summary>
--- a/Src/FluentAssertions/Events/EventAssertions.cs
+++ b/Src/FluentAssertions/Events/EventAssertions.cs
@@ -15,7 +15,7 @@ namespace FluentAssertions.Events;
 /// </summary>
 public class EventAssertions<T> : ReferenceTypeAssertions<T, EventAssertions<T>>
 {
-    private const string PropertyChangedEventName = "PropertyChanged";
+    private const string PropertyChangedEventName = nameof(INotifyPropertyChanged.PropertyChanged);
 
     protected internal EventAssertions(IMonitor<T> monitor)
         : base(monitor.Subject)
@@ -94,6 +94,9 @@ public class EventAssertions<T> : ReferenceTypeAssertions<T, EventAssertions<T>>
     /// <param name="becauseArgs">
     /// Zero or more objects to format using the placeholders in <paramref name="because"/>.
     /// </param>
+    /// <returns>
+    /// Returns only the events having arguments of type <see cref="PropertyChangedEventArgs"/> targeting the property.
+    /// </returns>
     public IEventRecording RaisePropertyChangeFor(Expression<Func<T, object>> propertyExpression,
         [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
     {
@@ -123,7 +126,7 @@ public class EventAssertions<T> : ReferenceTypeAssertions<T, EventAssertions<T>>
                     Monitor.Subject, PropertyChangedEventName, propertyName, actualPropertyNames);
         }
 
-        return recording.WithPropertyChanged(propertyName);
+        return recording.WithPropertyChangeFor(propertyName);
     }
 
     /// <summary>
@@ -146,7 +149,7 @@ public class EventAssertions<T> : ReferenceTypeAssertions<T, EventAssertions<T>>
 
         string propertyName = propertyExpression?.GetPropertyInfo().Name;
 
-        if (string.IsNullOrEmpty(propertyName))
+        if (propertyName is null)
         {
             Execute.Assertion
                 .BecauseOf(because, becauseArgs)
@@ -159,21 +162,11 @@ public class EventAssertions<T> : ReferenceTypeAssertions<T, EventAssertions<T>>
         {
             Execute.Assertion
                 .BecauseOf(because, becauseArgs)
-                .ForCondition(!recording.Any(@event =>
-                {
-                    var affectedPropertyName = GetAffectedPropertyName(@event);
-
-                    return string.IsNullOrEmpty(affectedPropertyName) || affectedPropertyName == propertyName;
-                }))
+                .ForCondition(!recording.Any(@event => @event.IsAffectingPropertyName(propertyName)))
                 .FailWith(
                     "Did not expect object {0} to raise the {1} event for property {2}{reason}, but it did.",
                     Monitor.Subject, PropertyChangedEventName, propertyName);
         }
-    }
-
-    private static string GetAffectedPropertyName(OccurredEvent @event)
-    {
-        return @event.Parameters.OfType<PropertyChangedEventArgs>().Single().PropertyName;
     }
 
     protected override string Identifier => "subject";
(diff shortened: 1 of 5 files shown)


Src/FluentAssertions/Events/EventAssertions.cs
```diff
@@ -144,13 +144,29 @@ public void NotRaise(string eventName, [StringSyntax("CompositeFormat")] string
     {
         IEventRecording recording = Monitor.GetRecordingFor(PropertyChangedEventName);
 
-        string propertyName = propertyExpression.GetPropertyInfo().Name;
+        string propertyName = propertyExpression?.GetPropertyInfo().Name;
 
-        if (recording.Any(@event => GetAffectedPropertyName(@event) == propertyName))
+        if (string.IsNullOrEmpty(propertyName))
```
dennisdoomen, line 149: 🤔 I don't think `propertyName` can ever be empty here. It can only be a valid name or `null`

Src/FluentAssertions/Events/EventAssertions.cs
```diff
@@ -144,13 +144,29 @@ public void NotRaise(string eventName, [StringSyntax("CompositeFormat")] string
...
+        }
+        else
         {
             Execute.Assertion
                 .BecauseOf(because, becauseArgs)
-                .FailWith("Did not expect object {0} to raise the {1} event for property {2}{reason}, but it did.",
+                .ForCondition(!recording.Any(@event =>
+                {
+                    var affectedPropertyName = GetAffectedPropertyName(@event);
+
```
dennisdoomen, line 165: ♻️ We can make this more readable by wrapping all of this and the `GetAffectedPropertyName` logic into `OccuredEvent`.

Tests/FluentAssertions.Specs/Events/EventAssertionSpecs.cs
```diff
@@ -540,6 +540,53 @@ public void
...
+        [Fact]
+        public void
+            The_number_of_property_changed_recorded_for_a_specific_property_matches_the_number_of_time_it_was_raised_including_agnostic_property()
+        {
+            // Arrange
+            var subject = new EventRaisingClass();
+            using var monitor = subject.Monitor();
+            subject.RaiseEventWithSenderAndPropertyName(nameof(EventRaisingClass.SomeProperty));
+            subject.RaiseEventWithSenderAndPropertyName(nameof(EventRaisingClass.SomeOtherProperty));
+            subject.RaiseEventWithSenderAndPropertyName(null);
```
dennisdoomen, line 568: 🤔`INotifyPropertyChanged` does not allow a `null` property. But if remove the line from this test that passes `null`, the test fails. I didn't expect that.

Tests/FluentAssertions.Specs/Events/EventAssertionSpecs.cs
```diff
@@ -540,6 +540,53 @@ public void
             act.Should().Throw<XunitException>()
                 .WithMessage("Expected*property*SomeProperty*but*OtherProperty1*OtherProperty2*");
         }
+
+        [Fact]
+        public void
+            The_number_of_property_changed_recorded_for_a_specific_property_matches_the_number_of_time_it_was_raised_specifically()
```
dennisdoomen, line 546: ⛏️ The number of time_s_

dennisdoomen:
@arocheleau just to be clear, if you have addressed the comments, please mark them as resolved and re-request a review. Then I'll know when to look at the rework.

Tests/FluentAssertions.Specs/Events/EventAssertionSpecs.cs
arocheleau: The remark section of the [documentation page](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.inotifypropertychanged.propertychanged?view=net-8.0#system-componentmodel-inotifypropertychanged-propertychanged) for `INotifyPropertyChanged.PropertyChanged` mentions that the property name of the `PropertyChangedEventArgs` could be `null` or `string.Empty` to target all properties.

dennisdoomen: Yeah, you're right. I'm at loss what I was thinking at that time. However, shouldn't the assertion then not expect 4 events?

ITaluone: It makes sense for me as it is, because he's checking for `x.SomeProperty` which isn't matching for `SomeOtherProperty`, but for everything else in this case.

arocheleau: The purpose of this test was to make sure all events related to a single property is being returned as expected.  3 out of 4 events are raised for this purpose ("SomeProperty", `null` and `string.Empty`).  The last event is targeting "SomeOtherProperty" and should not be returned or count in the final result.  The reason it was added to the test was to validate that the function is actually returning a filtered list of events and not the full list.  To achieve this, I added an event, an intruder, that is not supposed to be returned.

dennisdoomen (approved):
Great work

Tests/FluentAssertions.Specs/Events/EventAssertionSpecs.cs
```diff
@@ -540,6 +540,53 @@ public void
...
+        {
+            // Arrange
+            var subject = new EventRaisingClass();
+            using var monitor = subject.Monitor();
+            subject.RaiseEventWithSenderAndPropertyName(nameof(EventRaisingClass.SomeProperty));
+            subject.RaiseEventWithSenderAndPropertyName(nameof(EventRaisingClass.SomeProperty));
+            subject.RaiseEventWithSenderAndPropertyName(nameof(EventRaisingClass.SomeOtherProperty));
+
+            // Act
+            monitor.Should().RaisePropertyChangeFor(x => x.SomeProperty).Count().Should().Be(2);
```
jnyrup, line 556:
```suggestion
            monitor.Should().RaisePropertyChangeFor(x => x.SomeProperty).Should().HaveCount(2);
```

jnyrup (approved):

Src/FluentAssertions/Events/EventAssertions.cs
```diff
@@ -15,8 +15,6 @@ namespace FluentAssertions.Events;
 /// </summary>
 public class EventAssertions<T> : ReferenceTypeAssertions<T, EventAssertions<T>>
 {
-    private const string PropertyChangedEventName = "PropertyChanged";
```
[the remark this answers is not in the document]
dennisdoomen: > r just to replace the string literal "PropertyChanged" with nameof(INotifyPropertyChanged.PropertyChanged)?

Just this part.

dennisdoomen:
@arocheleau Let me know if you want to do something with the suggestions from @jnyrup so we know when to merge this PR.

arocheleau:
@dennisdoomen I have applied the suggestions from @jnyrup

jnyrup (approved):

PR #2679, opened 2024-06-20
Improve code coverage in `...Formatter`s

ITaluone:
### `TaskFormatter`
Removed unreachable path.
I think we can rely on the return value of `CanHandle()` that we are dealing with a `Task`.
I tried to get into the `Format` method using `Formatter.Format((Task)null)`, but I always ended up in the `NullFormatter`.

### `GuidValueFormatter` / `StringValueFormatter`
An interpolated string has no missed branches for whatever reason.

### `FormatterObjectGraph`
Removed an unused method, which was introduced in https://github.com/fluentassertions/fluentassertions/pull/2144

### `DateTimeOffsetValueFormatter`
Since we first check in `HasTime` that the value also could have milli seconds, the `if` branch can never be entered

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
+++ b/Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
@@ -92,14 +92,7 @@ public class DateTimeOffsetValueFormatter : IValueFormatter
 
         if (!hasDate && !hasTime)
         {
-            if (HasMilliSeconds(dateTimeOffset))
-            {
-                formattedGraph.AddFragment("0001-01-01 00:00:00." + dateTimeOffset.ToString("fff", CultureInfo.InvariantCulture));
-            }
-            else
-            {
-                formattedGraph.AddFragment("0001-01-01 00:00:00.000");
-            }
+            formattedGraph.AddFragment("0001-01-01 00:00:00.000");
         }
 
         formattedGraph.AddFragment(">");
--- a/Src/FluentAssertions/Formatting/FormattedObjectGraph.cs
+++ b/Src/FluentAssertions/Formatting/FormattedObjectGraph.cs
@@ -101,7 +101,7 @@ public class FormattedObjectGraph
     {
         if (lineBuilder.Length > 0)
         {
-            AppendSafely(lineBuilderWhitespace + lineBuilder);
+            AppendSafely($"{lineBuilderWhitespace}{lineBuilder}");
 
             lineBuilder.Clear();
             lineBuilderWhitespace = Whitespace;
@@ -247,24 +247,6 @@ public class FormattedObjectGraph
             else
             {
                 parentGraph.AddFragmentOnNewLine(fragment);
-            }
-        }
-
-        /// <summary>
-        /// Write the fragment.  If more lines have been added since this instance was
-        /// created then also flush the line and indent the next line.
-        /// </summary>
-        internal void AddEndingLineOrFragment(string fragment)
-        {
-            if (FormatOnSingleLine)
-            {
-                parentGraph.AddFragment(fragment);
-            }
-            else
-            {
-                parentGraph.AddFragment(fragment);
-                parentGraph.FlushCurrentLine();
-                parentGraph.lineBuilderWhitespace += MakeWhitespace(1);
             }
         }
 
--- a/Src/FluentAssertions/Formatting/TaskFormatter.cs
+++ b/Src/FluentAssertions/Formatting/TaskFormatter.cs
@@ -3,7 +3,7 @@
 namespace FluentAssertions.Formatting;
 
 /// <summary>
-/// Provides a human readable version of a generic or non-generic <see cref="Task"/>
+/// Provides a human-readable version of a generic or non-generic <see cref="Task"/>
 /// including its state.
 /// </summary>
 public class TaskFormatter : IValueFormatter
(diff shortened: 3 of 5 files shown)


Src/FluentAssertions/Formatting/TaskFormatter.cs
```diff
@@ -15,14 +15,10 @@ public bool CanHandle(object value)
...
-        if (value is Task task)
-        {
-            formatChild("type", task.GetType(), formattedGraph);
-            formattedGraph.AddFragment($" {{Status={task.Status}}}");
-        }
-        else
-        {
-            formattedGraph.AddFragment("<null>");
-        }
+        var task = value as Task;
```
jnyrup, line 18: Let's use a hard cast. That should also remove the need for silencing ReSharper

dennisdoomen (approved):

Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
```diff
@@ -92,14 +92,7 @@ public void Format(object value, FormattedObjectGraph formattedGraph, Formatting
 
         if (!hasDate && !hasTime)
         {
-            if (HasMilliSeconds(dateTimeOffset))
-            {
-                formattedGraph.AddFragment("0001-01-01 00:00:00." + dateTimeOffset.ToString("fff", CultureInfo.InvariantCulture));
-            }
-            else
-            {
-                formattedGraph.AddFragment("0001-01-01 00:00:00.000");
```
ITaluone, line 101: You _most definitely_ cannot reach this `if` body, if you have milliseconds set.
Thus, checking inside this `if` whether you have milliseconds set is a bit redundant..

Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
```diff
@@ -92,14 +92,7 @@ public void Format(object value, FormattedObjectGraph formattedGraph, Formatting
...
-            }
+            formattedGraph.AddFragment("0001-01-01 00:00:00.000");
         }
```
ITaluone, line 96: You most definitely cannot reach this if body, if you have milliseconds set.
Thus, checking inside this if whether you have milliseconds set is a bit redundant..

Src/FluentAssertions/Formatting/DateTimeOffsetValueFormatter.cs
```diff
@@ -92,14 +92,7 @@ public void Format(object value, FormattedObjectGraph formattedGraph, Formatting
...
         if (!hasDate && !hasTime)
         {
-            if (HasMilliSeconds(dateTimeOffset))
-            {
-                formattedGraph.AddFragment("0001-01-01 00:00:00." + dateTimeOffset.ToString("fff", CultureInfo.InvariantCulture));
-            }
-            else
-            {
-                formattedGraph.AddFragment("0001-01-01 00:00:00.000");
-            }
```
ITaluone, line 102: You most definitely cannot reach this if body, if you have milliseconds set.
Thus, checking inside this if whether you have milliseconds set is a bit redundant..

IT-VBFK: In short `!hasTime == true` when `dateTimeOffset` has all zeros in the time parts

Src/FluentAssertions/Formatting/FormattedObjectGraph.cs
```diff
@@ -101,7 +101,7 @@ private void FlushCurrentLine()
     {
         if (lineBuilder.Length > 0)
         {
-            AppendSafely(lineBuilderWhitespace + lineBuilder);
+            AppendSafely($"{lineBuilderWhitespace}{lineBuilder}");
```
jnyrup, line 104: Why start using string interpolation?

IT-VBFK: Because we don't have any missed coverage branches anymore

dennisdoomen: It seems that the [general consensus](https://stackoverflow.com/a/72928810/253961) is that string interpolation is more efficient than simple concatenation.

dennisdoomen (approved):

Src/FluentAssertions/Formatting/FormattedObjectGraph.cs
jnyrup: Performance doesn't care about consensus 😉

I didn't immediately spot that using string interpolation moves the null check of `lineBuilder` inside `DefaultInterpolatedStringHandler.AppendFormatted`, which increases the code coverage.
After understanding that, I'm fine with this change.

```cs
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using System.Text;

BenchmarkRunner.Run<StringConcat>();

[ReturnValueValidator]
[MemoryDiagnoser]
public class StringConcat
{
    private readonly StringBuilder lineBuilder = new("foo");

    [Params("    ")]
    public string LineBuilderWhitespace { get; set; }

    [Benchmark(Baseline = true)]
    public string Concat() => LineBuilderWhitespace + lineBuilder;

    [Benchmark]
    public string Interpolate() => $"{LineBuilderWhitespace}{lineBuilder}";

    [Benchmark]
    public string Opt() => string.Create(LineBuilderWhitespace.Length + lineBuilder.Length, (LineBuilderWhitespace, lineBuilder), (c, state) =>
    {
        state.LineBuilderWhitespace.CopyTo(c);
        state.lineBuilder.CopyTo(0, c[state.LineBuilderWhitespace.Length..], state.lineBuilder.Length);
    });
}
```

| Method      | Mean     | Error    | StdDev   | Ratio | Gen0   | Allocated | Alloc Ratio |
|------------ |---------:|---------:|---------:|------:|-------:|----------:|------------:|
| Concat      | 15.71 ns | 0.055 ns | 0.046 ns |  1.00 | 0.0115 |      72 B |        1.00 |
| Interpolate | 47.03 ns | 0.149 ns | 0.132 ns |  2.99 | 0.0114 |      72 B |        1.00 |
| Opt         | 14.83 ns | 0.185 ns | 0.164 ns |  0.94 | 0.0063 |      40 B |        0.56 |

Issue #2589, opened 2024-02-25
Add feature to check if an XElement or XAttribute is absent within the XDocument


skukshaus:
### Background and motivation

Currently, there is an implementation to check whenever an Attribute or Element is within the Document: https://fluentassertions.com/xml/
However, sometimes it is necessary to check if those attributes are absent. Currently, there is no possibility to do this. Maybe it would be possible to use the HaveElement() Method and catch the Exception, but this feels dirty.

I think it could also be a good idea, to add a constraint if the attribute/element does not have a specific value

### API Proposal

```C#
public class XElementAssertions
{
    public AndConstraint<XElementAssertions> NotHaveAttribute(string unexpectedAttribute, string because = "", params object[] becauseArgs);
    public AndConstraint<XElementAssertions> NotHaveAttribute(string unexpectedAttribute, string unexpectedValue, string because = "", params object[] becauseArgs);

    public AndWhichConstraint<XElementAssertions, XElement> NotHaveElement(string unexpectedElement, string because = "", params object[] becauseArgs);
    public AndWhichConstraint<XElementAssertions, XElement> NotHaveElement(string unexpectedElement, object unexpectedValue, string because = "", params object[] becauseArgs);
}
```

### API Usage

```C#
var element = XElement.Parse("<user><forename>john</forename></user>");
element.Should().NotHaveElement("surname");


var element = XElement.Parse("<user><forename>john</forename><surname>doe</surname></user>");
element.Should().NotHaveElement("surname", "smith");



var element = XElement.Parse("<user forename='john' />");
element.Should().NotHaveAttribute("surname");


var element = XElement.Parse("<user forename='john' surname='doe' />");
element.Should().NotHaveAttribute("surname", "smith");
```

### Are you willing to help with a proof-of-concept (as PR in that or a separate repo) first and as pull-request later on?

Yes, please assign this issue to me.

dennisdoomen:
Sure, makes sense. What do you think @jnyrup ?

jnyrup:
It certainly makes sense to have more negated assertions for XML types 👍 

You suggest adding an overload of `NotHaveElement` having an `object unexpectedValue`, but we don't currently have an affirmative `HaveElement(string expectedElement, object expectedValue)`.
So I'm not sure we should add that overload.

We currently have `HaveElement(string expectedName)` on 
* `XDocumentAssertions`
* `XElementAssertions`
* `XmlElementAssertions`

[...]

PR #2588, opened 2024-02-24
Added Support for "NotHaveElement" and "NotHaveAttribute" on XElement…

skukshaus:
It is important not only to check if a XElement or XAttribute is present within the XDocument, but it is also important to check, if these elements or attributes are absent.

PR #2690, opened 2024-07-08
Add a bunch of assertions on `XElement` and `XDocument`

ITaluone:
- [x] `HaveElementWithValue(string expectedName, string expectedValue)` - The element exists and has a particular value
  - [x] `XElementAssertions`
  - [x] `XDocumentAssertions`
- [x] `NotHaveElement(string unexpectedName)` - The element does not exist
  - [x] `XElementAssertions`
  - [x] `XDocumentAssertions`
- [x] `NotHaveElementWithValue(string unexpectedName, string unexpectedValue)` - The element either does not exist or doesn't have the particular value
  - [x] `XElementAssertions`
  - [x] `XDocumentAssertions`
- [x] `HaveAttribute(string expectedName)` - The attribute exists and can have any value
- [x] `HaveAttributeWithValue(string expectedName, string expectedValue)` - The attribute exists and has a particular value
- [x] `NotHaveAttribute(string unexpectedName)` - The attribute does not exist
- [x] `NotHaveAttributeWithValue(string unexpectedName, string unexpectedValue)` - The attribute either does not exist or doesn't have the particular value

Closes: #2589 
Supersedes: #2588 

## IMPORTANT 


Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -1231,6 +1231,221 @@ public void When_asserting_a_document_has_an_element_with_a_null_xname_it_should
...
+
+            // Act
+            Action act = () => element.Should().HaveElementWithValue("child", "b");
+
+            // Assert
+            act.Should().Throw<XunitException>().WithMessage("*child*b*element itself is <null>*");
+        }
+
+        [Fact]
+        public void Has_element_with_specified_value()
```
dennisdoomen, line 1250: 🔧This doesn't represent the behavior you expect.

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -1231,6 +1231,221 @@ public void When_asserting_a_document_has_an_element_with_a_null_xname_it_should
...
+                    <child>b</child>
+                </parent>
+                """);
+
+            // Act / Assert
+            element.Should().HaveElementWithValue("child", "b");
+        }
+
+        [Fact]
+        public void Throws_when_element_not_found()
```
dennisdoomen, line 1266:
```suggestion
        public void Throws_when_element_is_not_found()
```

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -1231,6 +1231,221 @@ public void When_asserting_a_document_has_an_element_with_a_null_xname_it_should
         }
     }
 
+    public class HaveElementWithValue
+    {
+        [Fact]
+        public void Throws_when_null()
```
dennisdoomen, line 1237:
```suggestion
        public void The_document_cannot_be_null()
```

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -1231,6 +1231,221 @@ public void When_asserting_a_document_has_an_element_with_a_null_xname_it_should
...
+        {
+            // Arrange
+            XDocument element = null;
```
dennisdoomen, line 1240: ⛏️ It's `document`, not `element`

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
dennisdoomen: 🔧  Although we prefer short names, they should still be proper English. Same for all the other tests.

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -1231,6 +1231,221 @@ public void When_asserting_a_document_has_an_element_with_a_null_xname_it_should
...
+
+            // Act
+            Action act = () => element.Should().HaveElementWithValue(null, "a");
+
+            // Assert
+            act.Should().Throw<ArgumentNullException>().WithMessage("*expectedElement*");
+        }
+
+        [Fact]
+        public void Throws_when_expected_element_is_null_with_ns()
```
dennisdoomen, line 1323: 🤔I don't understand this test.

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
dennisdoomen: 🔧 We prefer to use fact-based names for tests. Same for the other tests.

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
```diff
@@ -1409,4 +1624,333 @@ public void When_asserting_a_null_document_to_have_an_element_count_it_should_fa
...
+            var element = XDocument.Parse(
+                """
+                <parent>
+                    <child>a</child>
+                    <child>b</child>
+                </parent>
+                """);
+
+            // Act / Assert
+            element.Should().NotHaveElement("grandchild");
```
dennisdoomen, line 1656: 🤔 I think it would be clearer if you would use `c` instead of `grandchild`.

Src/FluentAssertions/Xml/XDocumentAssertions.cs
```diff
@@ -344,6 +344,219 @@ public XDocumentAssertions(XDocument document)
         return new AndWhichConstraint<XDocumentAssertions, IEnumerable<XElement>>(this, xElements);
     }
 
+    /// <summary>
+    /// Asserts that the <see cref="XDocument.Root"/> of the current <see cref="XDocument"/> doesn't have the specified child elements.
```
dennisdoomen, line 348: 🤔 Why are you consistently using "child elements" (as in plural)?

Src/FluentAssertions/Xml/XDocumentAssertions.cs
```diff
@@ -344,6 +344,219 @@ public XDocumentAssertions(XDocument document)
...
+        Guard.ThrowIfArgumentIsNull(unexpectedElement, nameof(unexpectedElement));
+
+        bool success = Execute.Assertion
+            .ForCondition(Subject is not null)
+            .BecauseOf(because, becauseArgs)
+            .FailWith(
+                "Did not expect {context:subject} to have an element {0}{reason}, but the element itself is <null>.",
+                unexpectedElement.ToString());
+
+        if (success)
```
dennisdoomen, line 393: ♻️You can avoid the separate `if (succes)` by chaining the two assertions using `Then`

Src/FluentAssertions/Xml/XDocumentAssertions.cs
```diff
@@ -344,6 +344,219 @@ public XDocumentAssertions(XDocument document)
...
+        {
+            Execute.Assertion
+                .WithExpectation("Did not expect {context:subject} to have an element {0}{reason}, ",
```
dennisdoomen, line 396: ♻️ You don't need to use `WithExpectation` unless you want to chain multiple `FailWith` calls.

dennisdoomen (commented):
It's a great start. Thanks for that.

I left an initial set of comments for some of the files, but which also need to be applied on the rest. Please re-request the review when you feel the rest also addresses the comments.

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
ITaluone: I don't get, what you mean. This test passes, when the document has both an element and a particular value for this element.

dennisdoomen: These days, we try to write the test names as a fact, preferably without using `when` and `should` (like we did in the past).  In this case, I probably would have named it `The_expected_element_with_the_expected_value_is_valid`. But I'm nitpicking here.

Src/FluentAssertions/Xml/XDocumentAssertions.cs
ITaluone: I implemented this how it was implemented previously to protect against Subject is `null` and wrapped into an `AssertionScope`.

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
ITaluone: Just checking the path with an added `XNamespace`

Src/FluentAssertions/Xml/XDocumentAssertions.cs
ITaluone: Good catch! :)

Tests/FluentAssertions.Specs/Xml/XDocumentAssertionSpecs.cs
ITaluone: Ah ok, changed that :)

ITaluone:
I revised several changed code and I think I got everything 🤞

ITaluone:
Does the qodana key is expired?

`2024/08/08 06:52:01 License request: token was declined by Qodana Cloud server`

dennisdoomen:
> Is the qodana key expired?

Yes, and I already have a nice license, but somehow it's not working yet. Working with JetBrains to resolve this ASAP.

dennisdoomen:
> Is the qodana key expired?

Fixed

dennisdoomen (approved):

ITaluone:
(rebased onto current develop)

Src/FluentAssertions/Xml/XDocumentAssertions.cs
```diff
@@ -344,7 +344,201 @@ public AndWhichConstraint<XDocumentAssertions, IEnumerable<XElement>> HaveElemen
...
+        assertionChain
+            .ForCondition(Subject is not null)
+            .BecauseOf(because, becauseArgs)
+            .FailWith(
+                "Expected {context:subject} to have an element {0} with value {1}{reason}, but the element itself is <null>.",
+                expectedElement.ToString(), expectedValue);
+
+        IEnumerable<XElement> xElements = [];
+
+        if (assertionChain.Succeeded)
```
ITaluone, line 461: The new `AssertionChain` construct allows to write this like that:

```csharp
assertionChain
    .WithExpectation("Expected {context:subject} to have an element {0} with value {1}{reason}, ",
        expectedElement.ToString(), expectedValue, chain => chain
            .FailWith(
                "but the element itself is <null>.",
                expectedElement.ToString(), expectedValue))
    .Then
    .WithExpectation("Expected {context:subject} to have an element {0} with value {1}{reason}, ",
        expectedElement, expectedValue, chain => chain
            .Given(() =>
            {
                xElements = Subject!.Root!.Elements(expectedElement).ToList();

                return xElements;
            })
            .ForCondition(collection => collection.Any())
            .FailWith("but the element {0} isn't found.", expectedElement)
            .Then
            .ForCondition(collection => collection.Any(e => e.Value == expectedValue))
            .FailWith("but the element {0} does not have such a value.", expectedElement));
```
@dennisdoomen Should I go for that (and everywhere else in this PR)?

dennisdoomen: Totally up to you. I'm fine merging it like this.

IT-VBFK: Me too.. I like such chainings :)

jnyrup:
> @dennisdoomen Is everything alright with Jonas, or is he that much occupied? 🤔

@IT-VBFK 
Thanks for checking in with me ❤️ 
I'm hopelessly behind on FA stuff - much more active on restaurant, bar and theater visits.

PR #2721, opened 2024-08-12
Fix `ArgumentOutOfRangeException` when using global tracing

### LINKED ISSUE #2619: ArgumentOutOfRangeException in StringBuilderTraceWriter.ToString()
### Description

I'm getting random `ArgumentOutOfRangeException` from an internal `StringBuilder.ToString()` call when using object equivalency (`Should().BeEquivalentTo`) with large object graphs during concurrent test runs with xUnit:

```
Error Message:
   System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than or equal to the size of the collection. (Parameter 'chunkLength')
  Stack Trace:
     at System.Text.StringBuilder.ToString()
   at FluentAssertions.Equivalency.Tracing.StringBuilderTraceWriter.ToString()
   at FluentAssertions.Equivalency.EquivalencyValidator.AssertEquality(Comparands comparands, EquivalencyValidationContext context)
   at FluentAssertions.Collections.GenericCollectionAssertions`3.BeEquivalentTo[TExpectation](IEnumerable`1 expectation, Func`2 config, String because, Object[] becauseArgs)
   at FluentAssertions.Collections.GenericCollectionAssertions`3.BeEquivalentTo[TExpectation](IEnumerable`1 expectation, String because, Object[] becauseArgs)
```

This only seems to occur _sometimes_ when multiple tests are running concurrently. When run individually or sequentially, this error does not occur.

Is there any static or shared StringBuilder internally that might be used in overlapping test runs?

### Reproduction Steps

This is difficult to reproduce in a controlled environment, since it only happens occasionally during concurrent test runs. It seems more likely to happen when evaluating large object graphs (depth > 15 or so) and only during concurrent test runs (xUnit concurrency settings `"parallelizeAssembly": false`)

### Expected behavior

Not expecting an exception from StringBuilder.ToString()

### Actual behavior

`ArgumentOutOfRangeException`

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

.NET 7
Fluent Assertions 6.12.0
xUnit 2.4.1

### Other information

_No response_

No

ITaluone:
Closes #2619 

## IMPORTANT 

### DIFF
--- a/docs/_pages/releases.md
+++ b/docs/_pages/releases.md
@@ -50,6 +50,7 @@ sidebar:
 * One overload of the `AssertionScope` constructor would not create an actual scope associated with the thread - [#2607](https://github.com/fluentassertions/fluentassertions/pull/2607)
 * Fixed `ThrowWithinAsync` not respecting `OperationCanceledException` - [#2614](https://github.com/fluentassertions/fluentassertions/pull/2614)
 * Fixed using `BeEquivalentTo` with an `IEqualityComparer` targeting nullable types - [#2648](https://github.com/fluentassertions/fluentassertions/pull/2648)
+* Fixed `ArgumentOutOfRangeException` when using globally enabled tracing - [#2721](https://github.com/fluentassertions/fluentassertions/pull/2721)
 
 ### Breaking Changes (for users)
 * Moved support for `DataSet`, `DataTable`, `DataRow` and `DataColumn` into a new package `FluentAssertions.DataSet` - [#2267](https://github.com/fluentassertions/fluentassertions/pull/2267)
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyOptions.cs
@@ -100,7 +100,7 @@ public abstract class SelfReferenceEquivalencyOptions<TSelf> : IEquivalencyOptio
         matchingRules.AddRange(defaults.MatchingRules);
         OrderingRules = new OrderingRuleCollection(defaults.OrderingRules);
 
-        TraceWriter = defaults.TraceWriter;
+        TraceWriter = defaults.TraceWriter?.NewInstance();
 
         RemoveSelectionRule<AllPropertiesSelectionRule>();
         RemoveSelectionRule<AllFieldsSelectionRule>();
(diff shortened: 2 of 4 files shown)


docs/_pages/releases.md
```diff
@@ -50,6 +50,7 @@ sidebar:
 * One overload of the `AssertionScope` constructor would not create an actual scope associated with the thread - [#2607](https://github.com/fluentassertions/fluentassertions/pull/2607)
 * Fixed `ThrowWithinAsync` not respecting `OperationCanceledException` - [#2614](https://github.com/fluentassertions/fluentassertions/pull/2614)
 * Fixed using `BeEquivalentTo` with an `IEqualityComparer` targeting nullable types - [#2648](https://github.com/fluentassertions/fluentassertions/pull/2648)
+* Fixed `ArgumentOutOfRangeException` when using globally enabled tracing - [#2721](https://github.com/fluentassertions/fluentassertions/pull/2721)
```
lg2de, line 53: The breaking change (`StringBuilderTraceWriter` => `StringBuilderTraceWriterPool`) should be mentioned!?

ITaluone: Hmm.. Or leave the original visibility?

jnyrup:
Before rushing to implementation, we should redesign the feature first.

The snippet I gave in the issue was meant as a workaround.

ITaluone:
Ou.. ok :)

jnyrup:
When designing a feature, I recommend that you forget about code a moment - that's why we request that all design go into issues rather than PRs.
Developers tend to implement the perfect code for the wrong feature (I'm certainly not exception to this).

ITaluone:
Ok, but is this redesign really required, since this bug is "fixed"? Or do I misunderstand something here?

jnyrup:
> Ok, but is this redesign really required, since this bug is "fixed"? Or do I misunderstand something here?

As I read the changes in this PR (by skimming them) they are breaking the current behavior in at least two ways:
* Any custom trace writer is overwritten by `StringBuilderTraceWriter`
*  if one _doesn't_  want a new instance they will still get a new instance.

ITaluone:
Ah I see.. yeah that's a problem..

Unfortunately, I'm not into this `IEquivalencyOptions` thing, so this could probably take a long time, so I think this should be tackled by someone who actually is pretty familiar with this particular code.

dennisdoomen:
> Unfortunately, I'm not into this `IEquivalencyOptions` thing, so this could probably take a long time, so I think this should be tackled by someone who actually is pretty familiar with this particular code.

Then let's close this PR if you don't plan to continue with it.

PR #2732, opened 2024-08-18
Upmerge from master

dennisdoomen:
Upmerge the documentation changes that happened on `master` to `develop`

### DIFF
--- a/cSpell.json
+++ b/cSpell.json
@@ -4,6 +4,7 @@
     "words": [
         "browsable",
         "comparands",
+        "Faqt",
         "formattable",
         "unformattable",
         "Guids",
--- a/docs/_data/navigation.yml
+++ b/docs/_data/navigation.yml
@@ -75,6 +75,8 @@ sidebar:
         url: /executiontime
       - title: HTTP response messages
         url: /httpresponsemessages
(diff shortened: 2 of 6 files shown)


jn-foreflight (approved):

jnyrup (approved):

cSpell.json
```diff
@@ -79,5 +81,7 @@
         "Frontmatter",
         "Liquid Includes"
     ],
-    "ignorePaths": []
+    "ignorePaths": [
+        "docs/_pages/releases.md"
+    ]
```
IT-VBFK, line 86: I think this should be reverted, because on develop it is empty and no spell error is thrown.

Note, that I have added a few more ignore patterns back than

IT-VBFK: For reference: https://github.com/fluentassertions/fluentassertions/pull/2361

cSpell.json
```diff
@@ -4,6 +4,8 @@
     "words": [
         "browsable",
         "comparands",
+        "Doomen",
```
IT-VBFK, line 7: This is also covered by the `Contributors` ignore pattern

PR #2741, opened 2024-08-25
Update to `GitVersion` 6.0.2 and fix PR version generation

IT-VBFK:
~**Note**: This issue is possibly related to NUKE, because the `GitVersion` class have some fields which the actual response of `dotnet gitversion` doesn't have (because of breaking changes), so this should also be fixed in NUKE itself.~

~But until then, we do this generation manually...~

## IMPORTANT 

### DIFF
--- a/Build/Build.cs
+++ b/Build/Build.cs
@@ -117,8 +117,8 @@ class Build : NukeBuild
         .Executes(() =>
         {
             ReportSummary(s => s
-                .WhenNotNull(GitVersion, (v, o) => v
-                    .AddPair("Version", o.SemVer)));
+                .WhenNotNull(GitVersion, (summary, gitVersion) => summary
+                    .AddPair("Version", gitVersion.SemVer)));
 
             DotNetBuild(s => s
                 .SetProjectFile(Solution)
--- a/Build/Build.cs
+++ b/Build/Build.cs
@@ -113,12 +113,13 @@ class Build : NukeBuild
 
     Target Compile => _ => _
         .DependsOn(Restore)
+        .DependsOn(CalculateNugetVersion)
         .OnlyWhenDynamic(() => RunAllTargets || HasSourceChanges)
         .Executes(() =>
         {
             ReportSummary(s => s
-                .WhenNotNull(GitVersion, (summary, gitVersion) => summary
-                    .AddPair("Version", gitVersion.SemVer)));
+                .WhenNotNull(SemVer, (summary, semVer) => summary
+                    .AddPair("Version", semVer)));
 
             DotNetBuild(s => s
                 .SetProjectFile(Solution)
@@ -128,6 +129,7 @@ class Build : NukeBuild
                 )
                 .EnableNoLogo()
                 .EnableNoRestore()
+                .SetVersion(SemVer)
                 .SetAssemblyVersion(GitVersion.AssemblySemVer)
                 .SetFileVersion(GitVersion.AssemblySemFileVer)
                 .SetInformationalVersion(GitVersion.InformationalVersion));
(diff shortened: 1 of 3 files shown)


Build/Build.cs
```diff
@@ -92,7 +92,9 @@ class Build : NukeBuild
                     "Branch spec {branchspec} is a pull request. Adding build number {buildnumber}",
                     BranchSpec, BuildNumber);
 
-                SemVer = string.Join('.', GitVersion.SemVer.Split('.').Take(3).Union([BuildNumber]));
+                var prNumber = BranchSpec.Split("/").First(token => token.All(c => char.IsDigit(c)));
```
IT-VBFK, line 95: My thoughts: we cannot rely on the number always on index 2 after split, but we can rely on the PR number is always "a number"

Build/Build.cs
```diff
@@ -81,7 +81,7 @@ class Build : NukeBuild
         });
 
     Target CalculateNugetVersion => _ => _
-        .OnlyWhenDynamic(() => RunAllTargets || HasSourceChanges)
+        .OnlyWhenDynamic(() => RunAllTargets || HasSourceChanges || IsLocalBuild)
         .Executes(() =>
         {
             SemVer = GitVersion.SemVer;
```
IT-VBFK, line 87: I testet this with 5.12 and 6.0.2 and both generate the same `SemVer` when on `develop`, so I think this works as intended...

Build/Build.cs
```diff
@@ -128,6 +131,7 @@ class Build : NukeBuild
                 )
                 .EnableNoLogo()
                 .EnableNoRestore()
+                .SetVersion(SemVer)
```
IT-VBFK, line 134: Does this have any advantage when compiled with a pre-release version instead of 7.0.0 in this case?

I know that the actual nuget version have to be versioned correctly, but does this also apply to the .dll itself? Or doesn't this mater in any way?

GitVersion.yml
```diff
@@ -1,13 +1,11 @@
...
   pull-request:
     mode: ContinuousDeployment
-    regex: ((pull|pull\-requests|pr)[/-]|[/-](merge))
-    tag: pr
-    tag-number-pattern: '[/-]?(?<number>\d+)'
-    prevent-increment-of-merged-branch-version: false
+    prevent-increment:
+      of-merged-branch: false
 ignore:
   sha: []
```
ITaluone, line 11:
```suggestion
workflow: GitHubFlow/v1
branches:
  develop:
    regex: ^dev(elop)?(ment)?$
    label: alpha
    increment: Minor
  release:
    regex: releases?[/-]
    label: rc
    increment: Patch
  pull-request:
    mode: ContinuousDelivery
    regex: ((pull|pull\-requests|pr)[/-]|[/-](merge))
    label: pr
    label-number-pattern: '[/-]?(?<number>\d+)'
    prevent-increment:
      of-merged-branch: false
ignore:
  sha: []

```

ITaluone: This should be it, no need to change anything in `Build.cs`

Build/Build.cs
dennisdoomen: I'd like to keep the DLL versions in sync with the package versions, yes.

dennisdoomen:
> This issue is possibly related to NUKE, because the GitVersion class have some fields which the actual response of dotnet gitversion doesn't have (because of breaking changes), so this should also be fixed in NUKE itself.
> But until then, we do this generation manually...

I don't get that rationale.

IT-VBFK:
Ah.. nevermind 😉

Build/Build.cs
IT-VBFK: Not needed...

Build/Build.cs
IT-VBFK: So.. with v6 there are a lot of breaking changes as it seems.. but I figured out (with help from [here](https://github.com/GitTools/GitVersion/discussions/4181) what the problem was.

dennisdoomen (approved):
Looks good to me. I've also checked the contents of the NuGet package.

image

PR #2745, opened 2024-08-26
Various naming improvements 

dennisdoomen:
Extracted from #2539

* The name of `EquivalencyResult.AssertionCompleted` wasn't clear what it tried to accomplish. This has been improved. Also renamed the interface that equivalency steps can use for recursive assertions.
* Made the node, field and property internal
* Fix the unit testing template to work with xUnit
* Suppress unnecessary Roslyn complaints about spacing
* Various internal naming improvements

**Note** Can be reviewed commit-by-commit


Src/FluentAssertions/Equivalency/Field.cs
```diff
@@ -8,7 +8,7 @@ namespace FluentAssertions.Equivalency;
 /// <summary>
 /// A specialized type of <see cref="INode  "/> that represents a field of an object in a structural equivalency assertion.
 /// </summary>
-public class Field : Node, IMember
+internal class Field : Node, IMember
```
ITaluone, line 11: _Technically_ these are breaking changes. Should we mention this?

Src/FluentAssertions/Equivalency/IValidateChildNodeEquivalency.cs
```diff
@@ -1,9 +1,9 @@
 namespace FluentAssertions.Equivalency;
 
-public interface IEquivalencyValidator
+public interface IValidateChildNodeEquivalency
```
jnyrup, line 3: Only info, nothing to change.
I know this is one point where we differ, as I don't see this renaming as an improvement.

Src/FluentAssertions/Equivalency/IValidateChildNodeEquivalency.cs
```diff
@@ -1,9 +1,9 @@
...
 {
     /// <summary>
     /// Runs a deep recursive equivalency assertion on the provided <paramref name="comparands"/>.
     /// </summary>
-    void RecursivelyAssertEquality(Comparands comparands, IEquivalencyValidationContext context);
+    void AssertEquivalencyOf(Comparands comparands, IEquivalencyValidationContext context);
```
jnyrup, line 8: I do like this renaming.

jnyrup (approved):
We should document _all_ breaking changes.

Src/FluentAssertions/Equivalency/IValidateChildNodeEquivalency.cs
dennisdoomen: > I know this is one point where we differ, as I don't see this renaming as an improvement.

I know, but I'm trying to apply Interface Seggregation here. It comes up every time as feedback from my students every time I run my TDD/Clean Code workshop ;-)

Src/FluentAssertions/Equivalency/Field.cs
dennisdoomen: Yes, see also the comment from @jnyrup.

Src/FluentAssertions/Equivalency/IValidateChildNodeEquivalency.cs
IT-VBFK: Consider let out the re-renaming of `...Equality` to `...Equivalency`, because this more correct, I think...

PR #2746, opened 2024-08-27
Increase code coverage

ITaluone:
#### First commit:
This applies to places where removal is not an option. E.g. custom `Exceptions`.

Thus: exclude from coverage report

#### Second commit
Using interpolation instead of concatenation should also increase CC

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/Execution/AssertionFailedException.cs
+++ b/Src/FluentAssertions/Execution/AssertionFailedException.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.Serialization;
 
 namespace FluentAssertions.Execution;
@@ -16,6 +17,7 @@ public class AssertionFailedException : Exception
     {
     }
 
+    [ExcludeFromCodeCoverage]
     protected AssertionFailedException(SerializationInfo info, StreamingContext context)
         : base(info, context)
     {
--- a/Src/FluentAssertions/Execution/AssertionFailedException.cs
+++ b/Src/FluentAssertions/Execution/AssertionFailedException.cs
@@ -5,6 +5,7 @@ namespace FluentAssertions.Execution;
 /// <summary>
 /// Represents the default exception in case no test framework is configured.
 /// </summary>
+/// <param name="message">The <em>mandatory</em> exception message</param>
 #pragma warning disable CA1032, RCS1194 // AssertionFailedException should never be constructed with an empty message
 public class AssertionFailedException(string message) : Exception(message)
 #pragma warning restore CA1032, RCS1194
--- a/Src/FluentAssertions/Execution/AssertionFailedException.cs
+++ b/Src/FluentAssertions/Execution/AssertionFailedException.cs
@@ -1,25 +1,12 @@
 using System;
-using System.Diagnostics.CodeAnalysis;
-using System.Runtime.Serialization;
 
 namespace FluentAssertions.Execution;
 
 /// <summary>
 /// Represents the default exception in case no test framework is configured.
 /// </summary>
-[Serializable]
 #pragma warning disable CA1032, RCS1194 // AssertionFailedException should never be constructed with an empty message
-public class AssertionFailedException : Exception
+public class AssertionFailedException(string message) : Exception(message)
 #pragma warning restore CA1032, RCS1194
 {
-    public AssertionFailedException(string message)
-        : base(message)
-    {
-    }
-
-    [ExcludeFromCodeCoverage]
-    protected AssertionFailedException(SerializationInfo info, StreamingContext context)
-        : base(info, context)
-    {
-    }
 }
--- a/Src/FluentAssertions/Execution/DefaultAssertionStrategy.cs
+++ b/Src/FluentAssertions/Execution/DefaultAssertionStrategy.cs
@@ -1,8 +1,10 @@
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using FluentAssertions.Common;
 
 namespace FluentAssertions.Execution;
 
+[ExcludeFromCodeCoverage]
 internal class DefaultAssertionStrategy : IAssertionStrategy
 {
     /// <summary>
--- a/Src/FluentAssertions/Formatting/MaxLinesExceededException.cs
+++ b/Src/FluentAssertions/Formatting/MaxLinesExceededException.cs
@@ -1,7 +1,9 @@
 ﻿using System;
+using System.Diagnostics.CodeAnalysis;
 
 namespace FluentAssertions.Formatting;
 
+[ExcludeFromCodeCoverage]
 public class MaxLinesExceededException : Exception
 {
     public MaxLinesExceededException(string message, Exception innerException)
--- a/Src/FluentAssertions/Formatting/MaxLinesExceededException.cs
+++ b/Src/FluentAssertions/Formatting/MaxLinesExceededException.cs
@@ -1,22 +1,7 @@
 ﻿using System;
-using System.Diagnostics.CodeAnalysis;
 
 namespace FluentAssertions.Formatting;
 
-[ExcludeFromCodeCoverage]
-public class MaxLinesExceededException : Exception
-{
-    public MaxLinesExceededException(string message, Exception innerException)
-        : base(message, innerException)
-    {
-    }
-
-    public MaxLinesExceededException(string message)
-        : base(message)
-    {
-    }
-
-    public MaxLinesExceededException()
-    {
-    }
-}
+#pragma warning disable RCS1194, CA1032 // Add constructors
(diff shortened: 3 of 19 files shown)


Src/FluentAssertions/Execution/AssertionFailedException.cs
```diff
@@ -16,6 +17,7 @@ public AssertionFailedException(string message)
     {
     }
 
+    [ExcludeFromCodeCoverage]
     protected AssertionFailedException(SerializationInfo info, StreamingContext context)
         : base(info, context)
     {
```
ITaluone, line 23: This was introduced in https://github.com/fluentassertions/fluentassertions/pull/268, but since the corresponding test was removed somewhere around the binary serializer issue, this is neither referenced nor tested against. And since `XmlSerializer` cannot serialize such an `Exception` (throws, because of some `IDictionary` which cannot be serialized), we could even consider to remove it, or leave it with exclude of CC.

Src/FluentAssertions/Execution/DefaultAssertionStrategy.cs
```diff
@@ -1,8 +1,10 @@
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using FluentAssertions.Common;
 
 namespace FluentAssertions.Execution;
 
+[ExcludeFromCodeCoverage]
 internal class DefaultAssertionStrategy : IAssertionStrategy
 {
```
ITaluone, line 9: Maybe there are a few other places with default implementations, which could also be excluded?

Src/FluentAssertions/Primitives/StringValidatorSupportingNull.cs
```diff
@@ -17,11 +17,6 @@ public StringValidatorSupportingNull(IStringComparisonStrategy comparisonStrateg
 
     public void Validate(string subject, string expected)
     {
-        if (expected is null && subject is null)
-        {
-            return;
-        }
-
```
ITaluone, line 24: Since we are guarding against `null` in `ContainsEquivalentOf` this path cannot be reached

dennisdoomen:
Please rebase on `develop`. We had lots of recent changes.

Src/FluentAssertions/Primitives/StringWildcardMatchingStrategy.cs
```diff
@@ -41,11 +41,9 @@ private bool IsMatch(string subject, string expected)
 
     private static string ConvertWildcardToRegEx(string wildcardExpression)
     {
-        return "^"
-            + Regex.Escape(wildcardExpression)
-                .Replace("\\*", ".*", StringComparison.Ordinal)
-                .Replace("\\?", ".", StringComparison.Ordinal)
-            + "$";
+        return $"^{Regex.Escape(wildcardExpression)
```
dennisdoomen, line 44: 🤔 I don't think this will make it necessarily better.

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.EndWith.cs
```diff
@@ -83,6 +83,21 @@ public void When_string_ending_is_compared_with_string_that_is_longer_it_should_
                 "\"ABC\" is too short.");
         }
 
+        [Fact]
+        public void Throws_when_wrapped_into_an_assertion_scope_with_both_messages()
```
dennisdoomen, line 87: 🤔 I _think_ you intended to cover the scenario where both assertion messages should be included in the final message, but that has been changed recently. The last one isn't even evaluated anymore.

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.StartWith.cs
```diff
@@ -89,6 +90,22 @@ public void When_string_start_is_compared_with_string_that_is_longer_it_should_t
                 "Expected string to start with \"ABCDEF\", but \"ABC\" is too short.");
         }
 
+        [Fact]
+        [SuppressMessage("ReSharper", "StringLiteralTypo")]
+        public void When_wrapped_in_an_assertion_scope_it_throws_with_both_messages()
```
dennisdoomen, line 95: Same comment as the other test

Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.Excluding.cs
```diff
@@ -666,5 +666,26 @@ public void When_excluding_virtual_or_abstract_property_exclusion_works_properly
                 .Excluding(o => o.VirtualProperty)
                 .Excluding(o => o.DerivedProperty2));
         }
+
+        [Fact]
+        public void Selecting_with_an_unsupported_expression_throws_the_appropriate_exception()
```
dennisdoomen, line 671: 🔧"Abstract properties cannot be excluded"

dennisdoomen (changes requested):

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.EndWith.cs
ITaluone: Ah.. true, actually the test with both messages failed :) I just removed the second expected message part.. but I had overlooked, that then this test could be removed completely.

Src/FluentAssertions/Execution/AssertionFailedException.cs
```diff
@@ -16,6 +17,7 @@ public AssertionFailedException(string message)
...
     }
 
+    [ExcludeFromCodeCoverage]
```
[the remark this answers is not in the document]
ITaluone: How can I test this? Since the exception message cannot be serialized at all, because of some issues in the internals (as I stated [here](https://github.com/fluentassertions/fluentassertions/pull/2746#discussion_r1732494992)

Src/FluentAssertions/Formatting/MaxLinesExceededException.cs
```diff
@@ -1,7 +1,9 @@
 ﻿using System;
+using System.Diagnostics.CodeAnalysis;
 
 namespace FluentAssertions.Formatting;
 
+[ExcludeFromCodeCoverage]
```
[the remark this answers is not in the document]
ITaluone: How can I write a test, when this type of exception is catched everywhere (and therefore isn't throwing at test level)?

Src/FluentAssertions/Primitives/StringWildcardMatchingStrategy.cs
ITaluone: True, hadn't looked at the actual coverage

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.EndWith.cs
IT-VBFK: Actually, this test is not completely useless: see the now missed lines (which previously weren't): https://coveralls.io/builds/70549123/source?filename=Src%2FFluentAssertions%2FPrimitives%2FStringEndStrategy.cs

Src/FluentAssertions/Formatting/MaxLinesExceededException.cs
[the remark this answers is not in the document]
ITaluone: I have found an (pretty ugly) workaround to get a test to throw a `MaxLinesExceededException`, but I cannot imagine that this ist the way to go:

```csharp
try
{
    Format(value, output, context,
        (path, childValue, childOutput)
            => FormatChild(path, childValue, childOutput, context, options, graph));
}
catch (MaxLinesExceededException) when (Environment.GetEnvironmentVariable("DO_NOT_CATCH_MAX_LINES_EXCEPTION") is null)
{
}
```

and setting this env var in the respective test...

Please don't tell me that we have to do such things 🙏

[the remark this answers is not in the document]
dennisdoomen: I guess those two constructors can just be removed.

Tests/FluentAssertions.Specs/Primitives/StringAssertionSpecs.EndWith.cs
dennisdoomen: My point was that the title was a bit misleading.

Src/FluentAssertions/Execution/AssertionFailedException.cs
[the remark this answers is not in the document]
ITaluone: Is this a breaking change then?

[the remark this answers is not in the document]
dennisdoomen: Technically yes, so we should mention it in the `releases.md`, but in reality nobody will be affected by this.

dennisdoomen (approved):

Issue #2759, opened 2024-09-16
BeEquivalentTo fails with default interface methods (6.12.1)


dusk0r:
### Description

Comparing the same Interface with BeEquivalentTo() fails if the interface contains default interface methods. 

### Reproduction Steps

```csharp
    public void BeEquivalentTo_Default_Interface_Method()
    {
        List<ITest> lista = [new Test { Name = "Test" }];
        List<ITest> listb = [new Test { Name = "Test" }];

        lista.Should().BeEquivalentTo(listb); // Fails with: Expectation has property lista[0].NameLength that the other object does not have.
    }
    
    public class Test : ITest
    {
        public string Name { get; set; }
    }

    public interface ITest
    {
        public string Name { get; }
        public int NameLength => Name.Length;
    }
```

### Expected behavior

Doesn't thorw an exception.

### Actual behavior

Fails with: Expectation has property lista[0].NameLength that the other object does not have.

### Regression?

This worked in 6.12.0 and before but fails in 6.12.1.

### Configuration

Fluent Assertion 6.12.1
.net 8.0.401

### Are you willing to help with a pull-request?

No

ITaluone:
I don't think this is the exception message (`Expectation has property lista[0].NameLength that the other object does not have.`), because this works only with v7 (meaning the PR isn't even merged yet)..

ITaluone:
Sounds like #2754

omarjuul:
I also ran into this issue when updating to v6.12.1 today. I just finished creating what I thought was a minimal repro when I saw this issue with a smaller repro.
I don't see the correlation with #2754 as that issue seems to be about priority, whereas this issue is about being able to find the property at all.

dennisdoomen:
I'm struggling to fix this problem and wondering whether we should at all. 

The problem is that during the nested equivalency check, FA doesn't actually now the compile-time type of the `Test` instance and only sees the `Test` class. Default interface properties are only visible through the interface and not the type that implements the interface. This kind of worked because the 6.12 implementation was a bit naive. Fixing this will be hard, since we changed the implementation considerably to fix #2146. 

The only way to fix this is to enhance the mechanism that scans for the members of a type, to also include interfaces they implement. But that will break some of the more complicated scenarios around explicitly implemented properties.

MartinDembergerR9:
Sorry, but I don't understand your argumentation.
Isn't the case here that the members of the elements of lista are determined by the compile time type and the members of listb by the runtime type?

[...]

PR #2794, opened 2024-10-28
Better support for default interface and explicitly implemented properties

### LINKED ISSUE #2754: BeEquivalentTo doesn't always choose normal properties over explicitly implemented properties
### Description

I use my own implementation of Exception which implements a interface with a Message-Property. This property is of another type.

### Reproduction Steps

```csharp
[Fact]
public void Test()
{
    var instance = new MyClass()
    {
        MyError = 42,
    };
    var other = new MyClass()
    {
        MyError = 42,
    };

    instance.Should().BeEquivalentTo(other);
}

public class MyClass : Exception, IMyInterface
{
    public int MyError { get; set; }
    int IMyInterface.Message => MyError;
}

public interface IMyInterface
{
    int Message { get; }
}
```

### Expected behavior

The test passes.

### Actual behavior

The test failes with the message:

```
Message: 
Expected property instance.MyNamespace.IMyInterface.Message to be System.String, but found System.Int32.

With configuration:
- Compare enums by value


  Stack Trace: 
XUnit2TestFramework.Throw(String message)
TestFrameworkProvider.Throw(String message)
CollectingAssertionStrategy.ThrowIfAny(IDictionary`2 context)
AssertionScope.Dispose()
EquivalencyValidator.AssertEquality(Comparands comparands, EquivalencyValidationContext context)
ObjectAssertions`2.BeEquivalentTo[TExpectation](TExpectation expectation, Func`2 config, String because, Object[] becauseArgs)
ObjectAssertions`2.BeEquivalentTo[TExpectation](TExpectation expectation, String because, Object[] becauseArgs)
MyTestClass.Test() line 426
RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

### Regression?

In 6.12.0 it worked by with 6.12.1 it is failing.

### Known Workarounds

_No response_

### Configuration

net8.0
FluentAssertions 6.12.1

### Other information

_No response_

No

dennisdoomen:
Reimplemented the Reflection process using a content-only open-source library needed to find all types of properties and fields including default interface properties.

Fixes #2754 and #2759


ITaluone:
🤔 Missing release notes and/or documentation?

ITaluone:
image

Do we really want this?

dennisdoomen:
> image
> 
> Do we really want this?

Nope. You're right. I need to fix that.

Build/Build.cs
```diff
@@ -259,6 +259,7 @@ void ReportTestOutcome(params string[] globFilters)
                     ReportTypes.lcov,
                     ReportTypes.HtmlInline_AzurePipelines_Dark)
                 .AddFileFilters("-*.g.cs")
+                .AddFileFilters("-**/.nuget/**")
```
ITaluone, line 262: Maybe I am wrong, but I think this should be: `-*.nuget*.cs`, since it supports wildcard but not glob filters.

dennisdoomen: I think you're right.

ITaluone: Hmm.. apparently I wasn't.. I don't how to handle this then, because this behind the scenes is converted to a regex and and a `Regex.IsMatch()` is performed:

https://github.com/danielpalme/ReportGenerator/blob/15c6fc1004070c26689d4c30110b81d05d256544/src/ReportGenerator.Core/Parser/Filtering/DefaultFilter.cs#L73-L99

dennisdoomen: I think it's fixed now.

ITaluone (approved):

Src/FluentAssertions/Equivalency/MemberVisibilityExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    private static readonly ConcurrentDictionary<MemberVisibility, MemberKind> Cache = new();
+
+    public static MemberKind ToMemberKind(this MemberVisibility visibility)
+    {
+        return Cache.GetOrAdd(visibility, v =>
+        {
+            MemberKind result = MemberKind.None;
+
+            foreach (MemberVisibility flag in Enum.GetValues(typeof(MemberVisibility)))
```
jnyrup, line 17: We could polyfill `Enum.GetValues<T>` for older frameworks like we've done for other APIs in e.g. `SystemExtensions` and `StringBuilderExtensions` that PolySharp doesn't already provide.

| Method     | Job                  | Runtime              | Mean      | Error     | StdDev    | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|----------- |--------------------- |--------------------- |----------:|----------:|----------:|------:|--------:|-------:|----------:|------------:|
| NonGeneric | .NET 8.0             | .NET 8.0             | 101.05 ns |  0.896 ns |  0.794 ns |  1.00 |    0.01 | 0.0076 |      48 B |        1.00 |
| Generic    | .NET 8.0             | .NET 8.0             |  32.72 ns |  0.272 ns |  0.241 ns |  0.32 |    0.00 | 0.0076 |      48 B |        1.00 |
|            |                      |                      |           |           |           |       |         |        |           |             |
| NonGeneric | .NET Framework 4.7.2 | .NET Framework 4.7.2 | 744.11 ns | 14.475 ns | 15.488 ns |  1.00 |    0.03 | 0.0267 |     169 B |        1.00 |
| Generic    | .NET Framework 4.7.2 | .NET Framework 4.7.2 | 738.11 ns |  3.947 ns |  3.692 ns |  0.99 |    0.02 | 0.0267 |     169 B |        1.00 |

```cs
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Jobs;
using BenchmarkDotNet.Running;
using System;

BenchmarkRunner.Run<GetValuesBenchmark>();

[MemoryDiagnoser]
[SimpleJob(RuntimeMoniker.Net472)]
[SimpleJob(RuntimeMoniker.Net80)]
public class GetValuesBenchmark
{
    [Benchmark(Baseline = true)]
    public MemberVisibility[] NonGeneric() => (MemberVisibility[])Enum.GetValues(typeof(MemberVisibility));

    [Benchmark]
#if NET6_0_OR_GREATER
    public MemberVisibility[] Generic() => Enum.GetValues<MemberVisibility>();
#else
    public MemberVisibility[] Generic() => (MemberVisibility[])Enum.GetValues(typeof(MemberVisibility));
#endif
}

[Flags]
public enum MemberVisibility
{
    None = 0,
    Internal = 1,
    Public = 2,
    ExplicitlyImplemented = 4,
    DefaultInterfaceProperties = 8
}
```

Src/FluentAssertions/Equivalency/MemberVisibilityExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace FluentAssertions.Equivalency;
+
+internal static class MemberVisibilityExtensions
+{
+    private static readonly ConcurrentDictionary<MemberVisibility, MemberKind> Cache = new();
+
+    public static MemberKind ToMemberKind(this MemberVisibility visibility)
+    {
+        return Cache.GetOrAdd(visibility, v =>
```
jnyrup, line 13: To have the compiler help enforce that we don't capture anything
```suggestion
        return Cache.GetOrAdd(visibility, static v =>
```

dennisdoomen: Cool. Didn't know about this one.

Src/FluentAssertions/Equivalency/MemberVisibilityExtensions.cs
dennisdoomen: I just used an #ifdef for now.

jnyrup (approved):

PR #2819, opened 2024-11-11
The expectation node identified as a cyclic reference is still compared to the subject node using simple equality.

### LINKED ISSUE #2787: BeEquivalentTo with IgnoringCyclicReferences does not detect that actual model does not have cyclic reference.
### Description

When asserting `Should().BeEquivalentTo` on a model that has 2-way references, with the `IgnoringCyclicReferences`, the assertion doesn't detect that the actual model objects don't have the cyclic reference, if the expected model has a cyclic or back-reference. While the equivalency assertion the other way around does detect the difference.

See the code example, but in short:
- Expected model: Parent references Child, and Child references parent
- Actual model: Parent references Child, but Child doesn't reference parent.

The weird thing is, when doing 
`actualParent.Should().BeEquivalentTo(expectedParent, options => options.IgnoringCyclicReferences())`
it does not detect the difference between both models, 
while when doing 
`expectedParent.Should().BeEquivalentTo(actualParent, options => options.IgnoringCyclicReferences())`
it does detect the difference.
I would expect that equivalency works both ways.

### Reproduction Steps

```csharp
using FluentAssertions;
using Xunit;
using Xunit.Sdk;

namespace FluentAssertionsExperiment;

public class EquivalencyOptionsIgnoreCyclicTestsSimplified
{
    [Fact]
    public void IgnoringCyclicReferences_BackReferenceIsNull_ActualEquivalentToExpected_DoesNotThrow()
    {
        // Arrange expected model
        var expectedChild = new Child() {Id = 1};
        var expectedParent = new Parent(){ Id = 100, Children = new List<Child>(){expectedChild}};
        expectedChild.Parent = expectedParent;

        // Arrange actual model
        var actualChild = new Child() { Id = 1 };
        var actualParent = new Parent() { Id = 100, Children = new List<Child>() { actualChild } };
        //actualChild.Parent = actualParent; // => this is cleary different between both models

        // Assert actual model be equivalent to expected model => Does not throw.
        var action1 = () => actualParent.Should().BeEquivalentTo(expectedParent, options => options.IgnoringCyclicReferences());
        action1.Should().NotThrow();

        // Assert expected model be equivalent to actual model => Does throw.
        var action2 = () => expectedParent.Should().BeEquivalentTo(actualParent, options => options.IgnoringCyclicReferences());
        action2.Should().Throw<XunitException>();
    }

    private class Parent
    {
        public int Id { get; set; }
        public List<Child> Children { get; set; }
    }

    private class Child
    {
        public int Id { get; set; }
        public Parent Parent { get; set; }
    }
}
```

### Expected behavior

Should.BeEquivalent() should throw in both directions.

### Actual behavior

Should.BeEquivalent() throws only when the expected model doesn't have the cyclic reference, and does NOT throw when the expected model has the cyclic reference.

### Regression?

Don't know.

### Known Workarounds

Do the Should().BeEquivalent both ways in your test.

### Configuration

Tested with .Net 8, and version 6.12.1 of FluentAssertions.

### Other information

I tested this with a few different setups:
- Parent has collection of Children as reference
- Parent has single item reference to Child
- Similar behavior on the 2nd level from the root (added GrandChild level)
- Setting the Child.Parent property to a different Parent, instead of leaving empty.
I can provide testcode for these scenarios if needed.

I am willing to help test solutions, or doing some code-analysis, if I get some pointers to where to start.

No

dennisdoomen:
When the equivalency validation engine detects a cyclic reference, it excludes the node from further comparisons. But if the expectation is non-null and the subject is null (or vice versa), it should still trigger an assertion failure.

Fixes #2787

### DIFF
--- a/Src/FluentAssertions/Equivalency/EquivalencyValidator.cs
+++ b/Src/FluentAssertions/Equivalency/EquivalencyValidator.cs
@@ -35,7 +35,11 @@ public class EquivalencyValidator : IEquivalencyValidator
         {
             TrackWhatIsNeededToProvideContextToFailures(scope, comparands, context.CurrentNode);
 
-            if (!context.IsCyclicReference(comparands.Expectation))
+            if (context.IsCyclicReference(comparands.Expectation))
+            {
+                AssertComparandsPointToActualObjects(comparands);
+            }
+            else
             {
                 TryToProveNodesAreEquivalent(comparands, context);
             }
@@ -60,6 +64,19 @@ public class EquivalencyValidator : IEquivalencyValidator
         scope.Context = new Lazy<string>(() => currentNode.Description);
 
         scope.TrackComparands(comparands.Subject, comparands.Expectation);
+    }
+
+    private static void AssertComparandsPointToActualObjects(Comparands comparands)
+    {
+        if (ReferenceEquals(comparands.Subject, comparands.Expectation))
+        {
+            return;
+        }
+
+        if (comparands.Subject is null)
+        {
+            comparands.Subject.Should().BeSameAs(comparands.Expectation);
+        }
     }
 
     private void TryToProveNodesAreEquivalent(Comparands comparands, IEquivalencyValidationContext context)
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -23,7 +23,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.507">
+    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.556">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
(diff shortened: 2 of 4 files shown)


Src/FluentAssertions/Equivalency/EquivalencyValidator.cs
```diff
@@ -62,6 +66,19 @@ private static void TrackWhatIsNeededToProvideContextToFailures(AssertionScope s
...
+        if (ReferenceEquals(comparands.Subject, comparands.Expectation))
+        {
+            return;
+        }
+
+        if (comparands.Subject is null || comparands.Expectation is null)
+        {
+            comparands.Subject.Should().BeSameAs(comparands.Expectation);
+        }
+    }
```
jnyrup, line 80: On my phone, but I assume the CC complains about missing tests of these branches.

dennisdoomen: Yeah, I was too quick with the review request. Trying to create a test for the first branch.

Src/FluentAssertions/Equivalency/EquivalencyValidator.cs
```diff
@@ -62,6 +66,19 @@ private static void TrackWhatIsNeededToProvideContextToFailures(AssertionScope s
...
     }
 
+    private static void AssertComparandsPointToActualObjects(Comparands comparands)
+    {
+        if (ReferenceEquals(comparands.Subject, comparands.Expectation))
+        {
+            return;
+        }
+
+        if (comparands.Subject is null || comparands.Expectation is null)
```
jnyrup, line 76: The ` || comparands.Expectation is null` part is not covered by a test.
Can we write one or is this path uncoverable?

jnyrup (approved):

PR #2825, opened 2024-11-17
Upmerge the changes from master and bumped some dependencies

dennisdoomen:


Tests/UWP.Specs/UWP.Specs.csproj
```diff
@@ -87,7 +87,7 @@
       <Version>6.2.14</Version>
     </PackageReference>
     <PackageReference Include="MSTest.TestAdapter">
-      <Version>3.3.1</Version>
+      <Version>3.6.3</Version>
```
jnyrup, line 90: FYI, I had problems running this project with newer versions of MStest. Haven't found out if it the problem is on my side or the updated package. 
But since we don't run this automatic, I'm fine bumping the version to silence nuget warnings. 
Alternatively we can silence nuget warnings from test projects?

dennisdoomen: I don't have the UWP tools installed myself, so I never had that problem.

dennisdoomen: Just to be sure, I'll role this back.

PR #2845, opened 2024-11-26
Fix minor syntax error in objectgraphs.md

rklec:
It's still a function and needs to be called, otherwise it shows a syntax error.

(Removed the checklist as it is not relevant to this change.)


ITaluone:
This should target `main` instead of `develop`

rklec:
Urg I thought it's one click on GitHub to change this but nooo, obvioously it brings alll other changes too damn...

rklec:
Okay cherry-picking is harder than it should be, but it's done in https://github.com/fluentassertions/fluentassertions/pull/2847

PR #2851, opened 2024-11-26
You can exclude explicitly implemented properties from BeEquivalentTo

### LINKED ISSUE #2811: BeEquivalentTo compares explicitly implemented methods of internal interfaces
### Description

I am comparing objects that inherit from a base class from a library assembly that I don't control. That base class implements properties of an internal interface explicitly. Since #2794 / 6.12.2 these explicitly implemented properties are now compared.

This is problematic because 
1. I cannot exclude these properties using `.Excluding` (or `.For(..).Exclude(..)`) because I cannot name the property. I cannot name the property because I cannot name the interface (it's internal).
2. `IncludingAllDeclaredProperties` doesn't seem to have an effect. Contrary to what the doc comment on the method says, the explicitly implemented property of the internal interface seems to be treated as a public property declared on my type. Sure, interface implementations are technically public, but I would expect the visibility of the interface to "win" here.

### Reproduction Steps

```csharp
// Assembly A (not under my control)
internal interface IEntity {
  bool IsValid { get; }
}

public class Base(bool v) : IEntity {
  bool IEntity.IsValid { get ; } = v;
}

// Assembly B (my assembly)
// references assembly A
class MyClass(bool v) : Base(v) {
}
```

```csharp
// Objects should be equivalent because internal interfaces are ignored.
new MyClass(true).Should().BeEquivalentTo(new MyClass(false));
```

### Expected behavior

Internal interfaces are ignored for `BeEquivalentTo`.

### Actual behavior

Explicit property implementations of internal interfaces are considered for `BeEquivalentTo`.

### Regression?

This worked as expected in 6.12.1. It is broken in 6.12.2.

### Known Workarounds

Use the `IMemberInfo` overload of `.Excluding`.

### Configuration

.NET 8

### Other information

_No response_

No

dennisdoomen:
You can exclude explicitly implemented properties from `BeEquivalentTo` via `ExcludingExplicitlyImplementedProperties` 

Closes #2811

* [X] If the PR touches the public API, the changes have been approved in a separate issue with the "api-approved" label.
* [X] If the PR changes the public API the changes needs to be included by running [AcceptApiChanges.ps1](../tree/develop/AcceptApiChanges.ps1) or [AcceptApiChanges.sh](../tree/develop/AcceptApiChanges.sh).
* [X] If the PR affects [the documentation](../tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).
    * [X] Please also run `./build.sh --target spellcheck` or `.\build.ps1 --target spellcheck` before pushing and check the good outcome

### DIFF
--- a/docs/_pages/releases.md
+++ b/docs/_pages/releases.md
@@ -7,7 +7,7 @@ sidebar:
   nav: "sidebar"
 ---
 
-## 7.0 Alpha X
+## 8.0 Alpha X
 
 ### What's new
 
@@ -41,6 +41,7 @@ sidebar:
 * All assertions that support chaining using the `.Which` construct will now amend the caller identifier - [2539](https://github.com/fluentassertions/pull/2539)
 * Introduced a `MethodInfoFormatter` and improved the `PropertyInfoFormatter` - [2539](https://github.com/fluentassertions/pull/2539)
 * `Excluding()` / `For().Exclude()` and `Including()` on `BeEquivalentTo()` now also accepts an anonymous object to include/exclude multiple members at once - [#2488](https://github.com/fluentassertions/fluentassertions/pull/2488)
+* You can exclude explicitly implemented properties from `BeEquivalentTo` via `ExcludingExplicitlyImplementedProperties` - [9999](https://github.com/fluentassertions/pull/9999)
 
 ### Fixes
 * Fixed formatting error when checking nullable `DateTimeOffset` with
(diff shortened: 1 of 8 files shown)


docs/_pages/releases.md
```diff
@@ -41,6 +41,7 @@ sidebar:
 * All assertions that support chaining using the `.Which` construct will now amend the caller identifier - [2539](https://github.com/fluentassertions/pull/2539)
 * Introduced a `MethodInfoFormatter` and improved the `PropertyInfoFormatter` - [2539](https://github.com/fluentassertions/pull/2539)
 * `Excluding()` / `For().Exclude()` and `Including()` on `BeEquivalentTo()` now also accepts an anonymous object to include/exclude multiple members at once - [#2488](https://github.com/fluentassertions/fluentassertions/pull/2488)
+* You can exclude explicitly implemented properties from `BeEquivalentTo` via `ExcludingExplicitlyImplementedProperties` - [9999](https://github.com/fluentassertions/pull/9999)
```
jnyrup, line 44: 9999 -> 2851

dennisdoomen: Stupid me. Thanks.

PR #2855, opened 2024-11-29
Downgrade `ConfigurationManager` from 6.0.1 to 6.0.0

jnyrup:
#2673 bumped from 4.4.0 -> 6.0.1, but 6.0.0 should be enough. 
As a testing-only library I think we should strive for picking lower versions of dependencies to avoid that tests use a newer package than what is used in production due to included FA.

## IMPORTANT 

### DIFF
--- a/Src/FluentAssertions/FluentAssertions.csproj
+++ b/Src/FluentAssertions/FluentAssertions.csproj
@@ -64,7 +64,7 @@
         <Compile Remove="Common/NullConfigurationStore.cs" />
       </ItemGroup>
       <ItemGroup>
-        <PackageReference Include="System.Configuration.ConfigurationManager" Version="6.0.1" />
+        <PackageReference Include="System.Configuration.ConfigurationManager" Version="6.0.0" />
       </ItemGroup>
     </When>
     <When Condition="'$(TargetFramework)' == 'netstandard2.1'">


jnyrup:
`System.Threading.Tasks.Extensions` (STTE) still seems beneficial to upgrade to 4.5.4.
#2673 has great explanations, here are some more.

Currently our net47 TFM uses the net45 variant of STTE even though a netstandard2.0 exists, as net45 is [closer](https://nugettools.azurewebsites.net/6.12.1/get-nearest-framework?project=net47&package=net45%0D%0Anetstandard2.0) to net47.

STTE 4.5.4 added a net461 TFM, which our net47 TFM will use since that's even [closer](https://nugettools.azurewebsites.net/6.12.1/get-nearest-framework?project=net47&package=net45%0D%0Anet461%0D%0Anetstandard2.0).

I compared the output of building the net47 tests with STTE 4.5.0 and 4.5.4
With STTE `bin/` now contains different versions of STTE and System.Runtime.CompilerServices.Unsafe.

image

Comparing the packages using NuGet Package Explorer the net47 dll now also has a lot fewer metadata references (I assume this is better, but honestly don't know)
image

dennisdoomen:
I don't really get that either, but if you look in STTE 4.5.0 and 4.5.3, you'll see that they contain the same (albeit a newer version) DLLs, and only as a PCL. 

image

dennisdoomen (approved):

PR #2890, opened 2024-12-17
Downgrade System.Collections.Immutable to 1.5.0

jnyrup:
This partly reverts #2872.

This is to avoid pulling in dependencies that are newer than what end users might be using.

## IMPORTANT 


dennisdoomen:
Why? #2872 only affected test projects.

jnyrup:
> Why? #2872 only affected test projects.

Because `System.Collections.Immutable` 9.0.0 brings in these  changes to dependencies.
* Adds `System.Buffers` 4.5.1
* Adds `System.Memory` 4.5.5
* Adds `System.Numerics.Vectors` 4.5.0
* Updates `System.Runtime.CompilerServices.Unsafe` 4.5.3 -> 6.0.0

Perhaps a conservative view, but by keeping the versions we use in our test projects low, we ensure that end users also don't need to pull in newer versions of these dependencies.

dennisdoomen:
> Perhaps a conservative view, but by keeping the versions we use in our test projects low, we ensure that end users also don't need to pull in newer versions of these dependencies.

Personally, I find that a bit too farfetched. But if you care about it that much, I'm fine too.

PR #2901, opened 2024-12-20
Redesign Configuration, Services and AssertionOptions

dennisdoomen:
## IMPORTANT 


Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -44,6 +44,10 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.0" />
```
jnyrup, line 50: Wondering if we could put this in a separate package to avoid adding dependencies to the core lib? 
E.g. Everything can be manually configured via code, or one can pull in
`FluentAssertions.Configuration` and do configuration from json or environment variables.

dennisdoomen: Technically, this should be possible, but I seriously doubt that those extra dependencies are going to be an issue. Every modern .NET application will have those packages installed anyway. The only risk is that our version is too high. We could go for version 6.0.0, which is the lowest still supported.

jnyrup: While many new applications add those dependencies, FA is also used to test libraries, e.g. other OSS libraries that don't pull in additional dependencies. 

At least I would rather configure FA via code than pulling in extra dependencies.

lg2de: Version 6.x will get our of support in about 11 months. 
I'm here with Jonas: FA (and all packages with high usage):always should have as few as possible dependencies.

dennisdoomen: Hmm, I could ship that as a separate package that uses the new `AssertionEngineInitializerAttribute`, but that won't guarantee it loads at the right time. If another package like `FluentAssertions.DataSets` gets initialized first and updates the equivalency configuration, that new package would overwrite all those changes. Unless, that package would also smart enough to combine the existing configuration with what it retrieved through the `ConfigurationBuilder`.

And what would this package be named? `FluentAssertions.Configuration` implies it's an extension to assert on configurations.

dennisdoomen: In other words, not using `Microsoft.Extensions.Configuration` at all and letting the consumer handle it themselves?

jnyrup: Yes. At least in the core library.

I had a look at the xunit and mstest packages and they both have little to no dependencies from other namespaces.

Tests/FluentAssertions.Specs/Configuration/EquivalencyOptionsSpecs.cs
```diff
@@ -6,45 +6,43 @@
...
         }
 
         [Fact]
         public void It_should_throw()
         {
             Result.Should().ThrowExactly<ArgumentNullException>()
-                .WithParameterName("defaultsConfigurer");
+                .WithParameterName("configureOptions");
         }
     }
```
jnyrup, line 42: Not necessarily for this PR, but with the restructurings done in this PR, I think these tests are now be more readable _without_ Chill.

E.g. the current 
```cs
public class When_injecting_a_null_configurer : GivenSubject<EquivalencyOptions, Action>
{
    public When_injecting_a_null_configurer()
    {
        When(() => () => AssertionConfiguration.Current.Equivalency.Modify(configureOptions: null));
    }

    [Fact]
    public void It_should_throw()
    {
        Result.Should().ThrowExactly<ArgumentNullException>()
            .WithParameterName("configureOptions");
    }
}
```

becomes

```cs
[Collection("ConfigurationSpecs")]
public sealed class EquivalencyOptionsSpecs : IDisposable

...

[Fact]
public void When_injecting_a_null_configurer_it_should_throw()
{
    var action = () => AssertionConfiguration.Current.Equivalency.Modify(configureOptions: null);

    action.Should().ThrowExactly<ArgumentNullException>()
            .WithParameterName("configureOptions");
}

...

public void Dispose() => AssertionEngine.ResetToDefaults();
```

jnyrup (commented):
Really nice restructuring 💪 
It's _much_ easier to reason about global vs local options/configuration 🧠 
No more fiddling with which TFMs have `System.Configuration` available 😌 
One dependency less 🚀 
(Two if we say goodbye to Chill 😉)

Src/FluentAssertions/AssertionEngine.cs
```diff
@@ -1,47 +1,61 @@
...
+
             lock (Lockable)
             {
-                return configuration ??= new Configuration(ConfigurationStore);
+                configuration = new GlobalConfiguration();
+
+                return configuration;
             }
         }
     }
```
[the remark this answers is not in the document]
dennisdoomen: You're right. Initially this code was using the `ConfigurationBuilder` to get the settings from environment variables or `appsettings.json` files, but I've removed that.

Tests/FluentAssertions.Specs/Execution/FallbackTestFrameworkTests.cs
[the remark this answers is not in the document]
dennisdoomen: Me neither ;-)

Tests/FluentAssertions.Specs/Formatting/FormatterSpecs.cs
```diff
@@ -13,8 +11,7 @@
 
 namespace FluentAssertions.Specs.Formatting;
 
-[Collection("FormatterSpecs")]
```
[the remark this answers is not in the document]
dennisdoomen: Since I've moved all the tests that do something with the global formatting options to `FormattingOptionsSpecs`, I don't think this is needed anymore. Do you?

Src/FluentAssertions/Execution/TestFrameworkFactory.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        // Keep TUnitFramework and XUnitTestFramework last as they use a try/catch approach
+        [TestFramework.TUnit] = new TUnitFramework(),
+        [TestFramework.XUnit2] = new XUnitTestFramework("xunit.assert"),
+        [TestFramework.XUnit3] = new XUnitTestFramework("xunit.v3.assert"),
+    };
+
+#pragma warning disable AV1553
+    public static ITestFramework GetFramework(TestFramework? testFrameWork = null)
+#pragma warning restore AV1553
```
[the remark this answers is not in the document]
jnyrup: `TestFrameworkFactory.GetFramework` is not part of the public API, so I'm looking at internal usages.
a) The only place where we call `GetFramework` from Core lib we already pass in an enum.

So if we make it mandatory, we can remove the suppression of `AV1553` and just have to pass in a null in b).

image

Tests/FluentAssertions.Specs/Configuration/EquivalencyOptionsSpecs.cs
dennisdoomen: Let's keep that for a future PR.

Src/FluentAssertions/Execution/TestFrameworkFactory.cs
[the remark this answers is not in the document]
dennisdoomen: But `Configuration.TestFramework` is `null` by default, so we _do_ pass `null`.

Tests/FluentAssertions.Specs/Formatting/FormatterSpecs.cs
```diff
@@ -1254,3 +1064,5 @@ public Node()
 
     public List<Node> Children { get; set; }
 }
+
+// Due to the tests that call Configuration.Current
```
jnyrup, line 1068: I guess this should be placed somewhere more appropriate

Tests/FluentAssertions.Specs/Formatting/FormatterSpecs.cs
```diff
@@ -13,8 +11,8 @@
 
 namespace FluentAssertions.Specs.Formatting;
 
-[Collection("FormatterSpecs")]
-public class FormatterSpecs
+[Collection("FormattingSpecs")]
```
jnyrup, line 14: To match the name of the `[CollectionDefinition]` on `FormatterSpecsDefinition`
```suggestion
[Collection("FormatterSpecs")]
```

Src/FluentAssertions/Configuration/TestFramework.cs
```diff
@@ -0,0 +1,14 @@
...
+
+/// <summary>
+/// The test frameworks supported by Fluent Assertions.
+/// </summary>
+public enum TestFramework
+{
+    XUnit2,
+    XUnit3,
+    TUnit,
+    MsTest,
```
jnyrup, line 11: Suggestion: `MsTest` -> `MSTest`
This aligns it with the name of the [MSTest](https://www.nuget.org/profiles/MSTestFramework) nugets

dennisdoomen: Technically that doesn't follow the .NET naming conventions (because of the three consecutive capitals), so I prefer to keep it as-is.

jnyrup (approved):

Tests/FluentAssertions.Specs/Configuration/FormattingOptionsSpecs.cs
```diff
@@ -0,0 +1,216 @@
...
+
+        // Act
+        string result = Formatter.ToString(subject);
+
+        // Assert
+        result.Should().Be("Property is SomeValue", "it should use my custom formatter");
+    }
+
+    [Fact]
+    public void When_no_custom_formatter_exists_in_the_specified_assembly_it_should_use_the_default()
```
jnyrup, line 99: It seems some behavior changed.
This test used to cover the `return null` in `AttributeBasedFormatter.GetFormatter`

dennisdoomen: I'm not sure what you mean. This is what it looked like before my merge. https://github.com/fluentassertions/fluentassertions/blob/3d9334af01c211749f8113341646731be0b544e3/Tests/FluentAssertions.Specs/Formatting/FormatterSpecs.cs#L1022

jnyrup: [Before](https://coveralls.io/builds/71581506/source?filename=Src%2FFluentAssertions%2FFormatting%2FAttributeBasedFormatter.cs#L59):
image
[After](https://coveralls.io/builds/71582076/source?filename=Src%2FFluentAssertions%2FFormatting%2FAttributeBasedFormatter.cs#L57):
image

Before this PR `When_no_custom_formatter_exists_in_the_specified_assembly_it_should_use_the_default` was one of the tests exercising that `return null`.

PR #2307, opened 2023-09-18
Improve failure message for string assertions when checking for equality

vbreuss:
When asserting that two strings are equal, and one of the strings is longer than 8 characters or contains a NewLine, instead of simply stating at which index the first difference occurs, this difference is made visible using arrows, e.g.
```csharp
var subject = "this is a long text that differs in between";
var expected = "this was too short";
subject.Should().Be(expected, "because we use arrows now");
```
throws with message
```
Expected subject to be the same string because we use arrows now, but they differ at index 5:
        ↓ (actual)
  "this is a long text…"
  "this was too short"
        ↑ (expected)."
```

For multi-line strings, instead of only providing information about the index, the failure message also contains the line and column information, e.g. `…but they differ on line 5 and column 1 (index 93):`

For short single-line strings the behaviour remains unchanged!

This closes #2050.

## IMPORTANT 

* [ ] If the PR touches the public API, the changes have been approved in a separate issue with the "api-approved" label.
* [x] The code complies with the [Coding Guidelines for C#](https://www.csharpcodingguidelines.com/).
* [ ] The changes are covered by unit tests which follow the Arrange-Act-Assert syntax and the naming conventions such as is used [in these tests](../tree/develop/Tests/FluentAssertions.Equivalency.Specs/MemberMatchingSpecs.cs#L51-L430).
* [x] If the PR adds a feature or fixes a bug, please update [the release notes](../tree/develop/docs/_pages/releases.md) with a functional description that explains what the change means to consumers of this library, which are published on the [website](https://fluentassertions.com/releases).
* [ ] If the PR changes the public API the changes needs to be included by running [AcceptApiChanges.ps1](../tree/develop/AcceptApiChanges.ps1) or [AcceptApiChanges.sh](../tree/develop/AcceptApiChanges.sh).
* [ ] If the PR affects [the documentation](../tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).
    * [ ] Please also run `./build.sh --target spellcheck` or `.\build.ps1 --target spellcheck` before pushing and check the good outcome

PR #2972, opened 2025-01-17
Backport "Improve failure message for string assertions when checking for equality"

vbreuss:
Backport #2307 to v7:

When asserting that two strings are equal, and one of the strings is longer than 8 characters or contains a NewLine, instead of simply stating at which index the first difference occurs, this difference is made visible using arrows, e.g.
```csharp
var subject = "this is a long text that differs in between";
var expected = "this was too short";
subject.Should().Be(expected, "because we use arrows now");
```
throws with message
```
Expected subject to be the same string because we use arrows now, but they differ at index 5:
        ↓ (actual)
  "this is a long text…"
  "this was too short"
        ↑ (expected)."
```

For multi-line strings, instead of only providing information about the index, the failure message also contains the line and column information, e.g. `…but they differ on line 5 and column 1 (index 93):`

For short single-line strings the behaviour remains unchanged!

---

**Let's see, how open you are for community commits on v7, or if we should consider it dead...**


docs/_pages/releases.md
```diff
@@ -10,6 +10,7 @@ sidebar:
 ## 7.1.0
 
 ### Improvements
+* Improve failure message for string assertions when checking for equality - [#2307](https://github.com/fluentassertions/fluentassertions/pull/2307)
```
jnyrup, line 13: We already released 7.1.0
So this will be in a newer release.

Issue #2977, opened 2025-01-20
Release Notes missing for 7.1.0


cremor:
They are on GitHub (https://github.com/fluentassertions/fluentassertions/releases/tag/7.1.0), but not on the page the NuGet package links to (https://fluentassertions.com/releases/).

dennisdoomen:
Thanks for reporting it. I was preparing a fix last night. Will push it later today.

BTW, you can always find them in de GH release notes as well. See https://github.com/fluentassertions/fluentassertions/releases/tag/7.1.0

Issue #2959, opened 2025-01-15
Make license change more clear


burnchar:
### Description

Congratulations on getting some compensation after years of serving the community essentially for free. I get it. The new license allows you and Xceed to get a share from for-profit companies that use your work. (The work of contributers is outside the scope of this bug report).

However, the very significant change is unclear. The change needs clear notice in plain language that FluentAssertions usage is now very different from most NuGet packages, and different from its past. Something like:
"**NOTICE: FluentAssertions is no longer free for commercial use, and is no longer under the Apache 2.0 license**."
It needs a direct link to the new license.

For example, the first bullet in the release notes is:

- Update ownership and license - [#2943](https://github.com/fluentassertions/fluentassertions/pull/2943)
That link doesn't point to a license, and explains nothing without each reader digging in.

The main web page (fluentassertions.com) does casually mention the Xceed partnership, but little about it. The "Learn more" link does not help users learn more -- it just points them to the Xceed corporate web page.

### Reproduction Steps

```csharp
// Arrange
string input = "MyString";

// Act
char result = input[0];

// Assert
result.Should().Be('M');
```

### Expected behavior

Honest, clear language about the change with direct links to actual information.

### Actual behavior

Hidden, casual, borderline misleading language.

### Are you willing to help with a pull-request?

No

ms6073:
> The new license allows you and Xceed to get a share from for-profit companies that use your work.
Would be nice to also clarify the how this applied to not-for-profit/non-profit entities.

klemmchr:
> "NOTICE: FluentAssertions is no longer free for commercial use, and is no longer under the Apache 2.0 license."

This would be a blatant lie. All work done prior to the license change is in fact licensed under Apache 2.0. This essentially affects everything done so far since there is not a single commit contributing any code under the new license. The old code will remain under Apache 2.0 **forever**. You cannot retroactively change a license on existing code.

PR #2978, opened 2025-01-20
Various textual and styling fixes as well as some clarification of v7 and v8.

dennisdoomen:
Addresses #2977 and #2959

### DIFF
--- a/docs/_sass/minimal-mistakes/_base.scss
+++ b/docs/_sass/minimal-mistakes/_base.scss
@@ -31,13 +31,13 @@ h6 {
   line-height: 1.2;
   font-family: $header-font-family;
   font-weight: bold;
-  color: #007edf;
+  color: #000000;
 }
 
 h1 {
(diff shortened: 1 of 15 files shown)


docs/_sass/minimal-mistakes/_base.scss
```diff
@@ -31,14 +31,14 @@ h6 {
...
+  color: #000000;
 }
 
 h1 {
   margin-top: 0;
   font-size: $type-size-3;
-  color: #ed1c24;
-}
+  color: #ff8c00;
+}ff8c00
```
jnyrup, line 41: I've never written scss before, so this looks like an error to me?

dennisdoomen: You're right.

dennisdoomen: Pressed CTRL-C once too many

PR #2984, opened 2025-01-20
Provide a toggle to suppress the soft warning about commercial use

### LINKED ISSUE #2963: No Documentation on Installing Commercial License
### Description

So, I purchased a commercial license from Xceed.  As a software developer, I wanted to support continued development as I use FluentAssertions extensively in my tests.  However, there is no documentation on how to install the license and disable the Console.WriteLine for the softWarning which I don't want to be part of my builds

Steps:
1) Go to Xceed. https://xceed.com/products/unit-testing/fluent-assertions/
2) Purchase license
3) Wait for receipt
4) Wait for license.  You may need to email sales@xceed.com as I'm not sure the licenses are automatically generated and emailed out.
5) Explore the documentation to find there is no mention of how to attach a license to a NuGet downloaded version of FluentAssertions, v8.

### Reproduction Steps

```csharp
// Arrange
string input = "MyString";

// Act
char result = input[0];

// Assert
result.Should().Be('M');
```

### Expected behavior

Expect:
* Documentation on how to disable the soft warning in my builds with a purchased license.
* More than just an email with a license key that doesn't do anything.  The commercial license should at least allow me to disable all of the warnings.

### Actual behavior

No changes.

### Regression?

I would call changing the licensing for v8 a regression.

### Known Workarounds

Stick with v7 or just ignore the warnings for v8.

### Configuration

_No response_

### Other information

_No response_

No

dennisdoomen:
Resolves #2963 by providing a toggle to suppress the license warning that happens during each test run.

### DIFF
--- a/Src/FluentAssertions/AssertionEngine.cs
+++ b/Src/FluentAssertions/AssertionEngine.cs
@@ -78,26 +78,29 @@ public static class AssertionEngine
         {
             if (!isInitialized)
             {
-                const string softWarning =
-                    """
-                         Warning:
-                         The component "Fluent Assertions" is governed by the rules defined in the Xceed License Agreement and
-                         the Xceed Fluent Assertions Community License. You may use Fluent Assertions free of charge for
-                         non-commercial use only. An active subscription is required to use Fluent Assertions for commercial use.
-
-                         Please contact Xceed Sales mailto:sales@xceed.com to acquire a subscription at a very low cost.
-
-                         A paid commercial license supports the development and continued increasing support of
-                         Fluent Assertions users under both commercial and community licenses. Help us
-                         keep Fluent Assertions at the forefront of unit testing.
-
-                         For more information, visit https://xceed.com/products/unit-testing/fluent-assertions/
-                    """;
-
-                Console.WriteLine(softWarning);
-                Trace.WriteLine(softWarning);
-
                 ExecuteCustomInitializers();
+
+                if (!License.Accepted)
+                {
+                    const string softWarning =
+                        """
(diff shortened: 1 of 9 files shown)


Src/FluentAssertions/Licensing.cs
```diff
@@ -0,0 +1,12 @@
...
+
+/// <summary>
+/// Provides access to the licensing state of Fluent Assertions
+/// </summary>
+public static class Licensing
+{
+    /// <summary>
+    /// Allows to disable or enable the warning that reminds consumers about the license requirements for commercial use.
+    /// </summary>
+    public static bool AcknowledgeSoftWarning { get; set; } = true;
```
jnyrup, line 11: Shouldn't this be `false` by default?

jnyrup, line 11: As an end user I'm not sure I would know what "soft warning" refers to.

I find "Acknowledge" too vague as it means "accept or admit the existence or truth of something".

What if one could call `License.Accept = true`?

jnyrup (changes requested):

Src/FluentAssertions/Licensing.cs
dennisdoomen: But accept what?

jnyrup: > But accept what?

The license.

dennisdoomen: Alright. I can live with that.

jnyrup (approved):

PR #2987, opened 2025-01-22
Backport "Add [NotNull] attribute on the Should() method for object assertions"

0xced:
(backported from commit ca87a816e50659cad9a3f1cb32756bce84527ed1)

### DIFF
--- a/Src/FluentAssertions/AssertionExtensions.cs
+++ b/Src/FluentAssertions/AssertionExtensions.cs
@@ -300,7 +300,7 @@ public static class AssertionExtensions
     /// current nullable <see cref="bool"/>.
     /// </summary>
     [Pure]
-    public static NullableBooleanAssertions Should(this bool? actualValue)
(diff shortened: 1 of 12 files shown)


jnyrup:
Would it make sense to also add this for nullable value types?

0xced:
Absolutely! Addressed in 7541ea86b9555da8693d7f9bb072bfa8e9595774.

dennisdoomen:
@0xced We appreciate the contribution and want to let you know that by accepting the modifications to your original PR we make this code _also_ available to the V8 version which is no longer under Apache 2.0 license. Do you agree with that?

sdepouw:
Sounds like code that is getting put under the pay-for version should get a cut of the revenue to me.

0xced:
I'm fine if it goes into both v7 and v8.

jnyrup (approved):

Issue #2704, opened 2024-07-26
Struct record types generate format exception in dictionary key


Renaudyes:
### Description

When using a `record struct` in the generic `TKey` of a Dictionary, when an error of fluent assertion is generated, the `.ToString()` method is called which generates the following output:
`MyStruct{ MyProperty1 = "MyProperty1", MyProperty2 = "MyProeprty2", ... }`

One can see that using a `ToString()` on a record generates the characters `{` and `}` in the returned string.
The returned string is then used in turn in the `string.Format()` method but will crash because it considers `{` and `}` as parameter injection.
I mean, a message like this : `MyStruct{ MyProperty1 = "MyProperty1", MyProperty2 = "MyProeprty2", ... }` is considered as `MyStruct{ paramForFormatMethod }`.
So fluent assertion tries to inject a param in the generated string from the record struct.
(the example will help clarify)

### Reproduction Steps

```csharp
public readonly record struct MyRecordStruct
{
    public string MyProperty { get; init; }
}

[Fact]
public void myTest()
{
    var dict = new Dictionary<MyRecordStruct, MyRecordStruct>();
    var dict2 = new Dictionary<MyRecordStruct, MyRecordStruct>();

    var key = new MyRecordStruct() { MyProperty = "MyKey" };

    dict.Add(key, new MyRecordStruct());
    dict2.Add(key, new MyRecordStruct() { MyProperty = ""});

    dict2.Should().BeEquivalentTo(dict);
}
```

### Expected behavior

```
Expected property dict2[MyRecordStruct { MyProperty = MyKey }].MyProperty to be null, but found "".
```

### Actual behavior

It generates the following error message : 
```
 **WARNING** failure message 'Expected property dict2[MyRecordStruct { MyProperty = MyKey }].MyProperty to be {0}, but found {1}.' could not be formatted with string.Format
   at System.Text.ValueStringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ReadOnlySpan`1 args)
   at System.String.FormatHelper(IFormatProvider provider, String format, ReadOnlySpan`1 args)
   at System.String.Format(IFormatProvider provider, String format, Object[] args)
   at FluentAssertions.Execution.MessageBuilder.FormatArgumentPlaceholders(String failureMessage, Object[] failureArgs)
```

### Configuration

- FluentAssertions Version="6.12.0" 
- .NET8

jnyrup:
[...]

PR #3029, opened 2025-02-21
Fix a formatting exception when {} is used as a dictionary key.

dennisdoomen:
Especially in `BeEquivalentTo` calls we often pass the path from the root of the graph to the position in the graph that is being compared as a string into the `FailWith` calls. Since it's (apparently) possible that dictionary keys contain `{` and/or `}`, this caused a crash in the `MessageBuilder` while building the final failure message. We couldn't use normal `string.Format` placeholders like `{0}`, `{1}` and such because that caused the `Formatter` class to enclose strings with quotes. We solved that by introducing a special internal extension method `AsNonFormattable` that wraps the object to be formatted in such a way that the `Formatter` will not try to apply any formatting. 

This solves #2704

### DIFF
--- a/Src/FluentAssertions/Equivalency/Steps/EnumEqualityStep.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/EnumEqualityStep.cs
@@ -29,7 +29,8 @@ public class EnumEqualityStep : IEquivalencyStep
                 string expectationName = GetDisplayNameForEnumComparison(comparands.Expectation, expectationsUnderlyingValue);
 
                 return new FailReason(
-                    $"Expected {{context:enum}} to be equivalent to {expectationName}{{reason}}, but found {{0}}.",
+                    "Expected {context:enum} to be equivalent to {0}{reason}, but found {1}.",
+                    expectationName.AsNonFormattable(),
                     comparands.Subject);
             });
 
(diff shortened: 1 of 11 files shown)


Src/FluentAssertions/Equivalency/Steps/EnumEqualityStep.cs
```diff
@@ -88,7 +90,7 @@ private static void HandleByName(AssertionChain assertionChain, Comparands compa
                 string expectationName = GetDisplayNameForEnumComparison(comparands.Expectation, expectationsUnderlyingValue);
 
                 return new FailReason(
-                    $"Expected {{context:enum}} to equal {expectationName} by name{{reason}}, but found {subjectsName}.");
+                    "Expected {context:enum} to equal {0} by name{reason}, but found {1}.", expectationName.AsNonFormattable(), subjectsName.AsNonFormattable());
```
jnyrup, line 93:
```suggestion
                    "Expected {context:enum} to equal {0} by name{reason}, but found {1}.",
                    expectationName.AsNonFormattable(), subjectsName.AsNonFormattable());
```

Tests/FluentAssertions.Specs/Execution/AssertionChainSpecs.MessageFormating.cs
```diff
@@ -386,5 +386,53 @@ public void Message_should_contain_the_reason_as_defined_with_arguments()
...
+            var expectation = new Dictionary<string, string> { ["{}"] = null };
+
+            // Act
+            var act = () => expectation.Should().BeEquivalentTo(subject);
+
+            // Assert
+            act.Should().Throw<XunitException>()
+                .WithMessage("Expected expectation[{}] to be \"\", but found <null>.*");
+        }
+
```
jnyrup, line 404: covered by `Can_handle_more_braces_in_dictionary_keys`

```suggestion
```

jnyrup (approved):

Issue #3021, opened 2025-02-15
Fluentassertions launches browser without consent and possibility to opt out when I start my Visual Studio 2022


gabor-repos:
### Description

I am using `<PackageReference Include="FluentAssertions" Version="8.0.1" />` in my `<TargetFramework>net9.0</TargetFramework>` project.

Whenever I start my Visual Studio 2022 with this project Fluentassertions launches my default browser without consent and possibility to opt out. The launched url is: https://xceed.com/fluent-assertions/

Although I completely understand and support the concept that a community should support software projects, I am sure this is definitely not the way, and never was an option to promote any product or project.

**Think about if this become mainstream fashion and all vendors/projects adopts this attitude, dozens for page launches could happen when a developer opens her/his project in Visual Studio 2022, that will render unusable nuget as package manager.**

Nuget is a package manager, please do not use its automation infrastructure to automate advertisements.

### Reproduction Steps

- Create a .NET (9?) project in VS 2022, add "FluentAssertions" Version="8.0.1".
- Save
- Close VS 2022
- Open VS 2022, with the option, that VS 2022 automatically opens last project
- Observe that your default browser opens with the promotional link https://xceed.com/fluent-assertions/, and this new window takes the focus on the UI.
 

### Expected behavior

VS 2022 starts and opens the last project, and no automatic browser launch happens.

### Actual behavior

VS 2022 starts and opens the last project, and browser launch happens without consent and possibility to opt out. The page is: https://xceed.com/fluent-assertions/

### Are you willing to help with a pull-request?

No

dennisdoomen:
Wait, it does that _every_ time you launch Visual Studio?

njrardin:
I'm having the same issue. This is incredibly annoying. Is there any known workaround yet?

dennisdoomen:
We're investigating this. It doesn't happen on Rider though, but it definitely isn't supposed to do that. 

Cc @xceedsoftware

dennisdoomen:
Fixed in 8.1.1

PR #3030, opened 2025-02-21
Removed the PS script that opens the FA website because it misbehaves.

dennisdoomen:
Solves #3021


Src/FluentAssertions/FluentAssertions.csproj
```diff
@@ -44,7 +44,6 @@ Check out the [license page](LICENSE) for more information.
   <ItemGroup Label="Package files">
     <None Include="..\FluentAssertions.png" Pack="true" Visible="false" PackagePath="" />
     <None Include="..\..\LICENSE" Pack="true" PackagePath=""/>
-    <None Include="..\init.ps1" Pack="true" PackagePath="tools"/>
```
jnyrup, line 47: Should we also _remove_ `init.ps1`?
`init.ps1` seems overly complicated to open an url.
See https://github.com/fluentassertions/fluentassertions/pull/2943#discussion_r1918909066

jnyrup (approved):

PR #3063, opened 2025-06-22
Revisit the .editorconfig rules

dennisdoomen:


Src/FluentAssertions/CallerIdentifier.cs
```diff
@@ -212,7 +214,7 @@ private static string[] GetCallerIdentifiersFrom(StackFrame frame)
 
         if (string.IsNullOrEmpty(fileName) || expectedLineNumber == 0)
         {
-            return null;
+            return [];
```
jnyrup, line 213: Great!
This could cause an NRE in `ExtractCallersFrom`.

Tests/.editorconfig
```diff
@@ -197,6 +208,10 @@ dotnet_diagnostic.RCS1213.severity = none
 # Implement IComparable when implementing IComparable<T><T>. Disabled since we don't want to be so strict in tests.
 dotnet_diagnostic.RCS1241.severity = none
 
+# Purpose: Use an overload that has a IEqualityComparer<string> or IComparer<string> parameter
+# Reason: Not important in testsG
```
jnyrup: `testsG` -> `tests`

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -1593,12 +1593,13 @@ public AndConstraint<TAssertions> NotContain(string unexpected,
     public AndConstraint<TAssertions> NotContainAll(IEnumerable<string> values,
         [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
     {
-        ThrowIfValuesNullOrEmpty(values);
+        string[] strings = values?.ToArray();
```
jnyrup: Can we use `ConvertOrCastToCollection()` instead of `ToArray()`?

Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
```diff
@@ -68,8 +68,8 @@ public AndConstraint<TAssertions> Equal<T>(T expected,
 
         if (assertionChain.Succeeded)
         {
-            IEnumerable<TKey> subjectKeys = GetKeys(Subject);
-            IEnumerable<TKey> expectedKeys = GetKeys(expected);
+            IEnumerable<TKey> subjectKeys = GetKeys(Subject).ToArray();
```
jnyrup: Use `ConvertOrCastToCollection` over `ToArray`.
`GetKeys` will most likely return `IDictionary<TKey, TValue>.Keys`, which is of type `ICollection<TKey>`.

Src/FluentAssertions/Common/TypeExtensions.cs
jnyrup, line 1: Removing the `inherit` argument from the calls to BCL `GetCustomAttributes` methods changes behavior as then implicitly call with `inherit: true`.

I guess we should only remove the `inherit` parameter from line 102 and 112 and then pass `inherit: false` on line 108 and 116

Src/FluentAssertions/Execution/LateBoundTestFramework.cs
```diff
@@ -7,14 +7,15 @@ namespace FluentAssertions.Execution;
...
-    {
-        this.loadAssembly = loadAssembly;
-        exceptionFactory = _ => throw new InvalidOperationException($"{nameof(IsAvailable)} must be called first.");
-    }
+    /// <summary>
+    /// When set to <c>true</c>, the assembly specified by the <c>AssemblyName</c> property will
+    /// be dynamically loaded if it is not already loaded in the application domain.
+    /// When set to <c>false</c>, the framework will not attempt to load the assembly dynamically.
+    /// </summary>
+    protected bool LoadAssembly { get; set; }
```
jnyrup:
```suggestion
    protected bool LoadAssembly { get; init; }
```

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -1593,12 +1593,13 @@ public AndConstraint<TAssertions> NotContain(string unexpected,
...
+        ThrowIfValuesNullOrEmpty(strings);
 
-        var matches = values.Count(v => Contains(Subject, v, StringComparison.Ordinal));
+        var matches = strings!.Count(v => Contains(Subject, v, StringComparison.Ordinal));
```
jnyrup, line 1599: Instead of adding `!`, we could do like the methods in `Guard` and add `[ValidatedNotNull]`.

Src/FluentAssertions/Equivalency/Execution/ObjectReference.cs
jnyrup, line 1: `compareByMembers` is now always initialized with a non-null value, so the property `CompareByMembers` never has to lazy initialize it and could be a regular get-only auto-property.

Src/FluentAssertions/Equivalency/Execution/ObjectReference.cs
```diff
@@ -16,11 +16,11 @@ internal class ObjectReference
     private readonly bool? compareByMembers;
     private string[] pathElements;
 
-    public ObjectReference(object @object, string path, bool? compareByMembers = null)
+    public ObjectReference(object @object, string path, EqualityStrategy equalityStrategy = EqualityStrategy.Members)
     {
         this.@object = @object;
         this.path = path;
-        this.compareByMembers = compareByMembers;
+        compareByMembers = equalityStrategy == EqualityStrategy.Members || equalityStrategy == EqualityStrategy.ForceMembers;
```
jnyrup:
```suggestion
        compareByMembers = equalityStrategy is EqualityStrategy.Members or EqualityStrategy.ForceMembers;
```

Src/FluentAssertions/CallerIdentifier.cs
dennisdoomen: Is that a good thing or not?

Src/FluentAssertions/Primitives/StringAssertions.cs
dennisdoomen: No, but `ConvertOrCastToList` does work

Src/FluentAssertions/Primitives/StringAssertions.cs
dennisdoomen: Tried it, but doesn't seem to work in Rider.

Src/FluentAssertions/Common/TypeExtensions.cs
dennisdoomen: Good catch!

Src/FluentAssertions/CallerIdentifier.cs
jnyrup: Good!
`main` has the potential NRE and this PR fixes it.

dennisdoomen:
I didn't make any changes, but tried to split the modifications in a bunch of commits for clarity.

Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
```diff
@@ -68,8 +68,8 @@ public AndConstraint<TAssertions> Equal<T>(T expected,
 
         if (assertionChain.Succeeded)
         {
-            IEnumerable<TKey> subjectKeys = GetKeys(Subject);
-            IEnumerable<TKey> expectedKeys = GetKeys(expected);
+            IEnumerable<TKey> subjectKeys = GetKeys(Subject).ConvertOrCastToCollection();
+            IEnumerable<TKey> expectedKeys = GetKeys(expected).ConvertOrCastToCollection();
```
jnyrup: Suggestion:
```suggestion
            ICollection<TKey> subjectKeys = GetKeys(Subject).ConvertOrCastToCollection();
            ICollection<TKey> expectedKeys = GetKeys(expected).ConvertOrCastToCollection();
```

Src/FluentAssertions/Collections/GenericDictionaryAssertions.cs
```diff
@@ -139,8 +139,8 @@ public AndConstraint<TAssertions> NotEqual<T>(T unexpected,
                     .FailWith("Expected dictionaries not to be equal{reason}, but they both reference the same object.");
             }
 
-            IEnumerable<TKey> subjectKeys = GetKeys(Subject);
-            IEnumerable<TKey> unexpectedKeys = GetKeys(unexpected);
+            IEnumerable<TKey> subjectKeys = GetKeys(Subject).ConvertOrCastToCollection();
+            IEnumerable<TKey> unexpectedKeys = GetKeys(unexpected).ConvertOrCastToCollection();
```
jnyrup, line 143: Suggestion:
```suggestion
            ICollection<TKey> subjectKeys = GetKeys(Subject).ConvertOrCastToCollection();
            ICollection<TKey> unexpectedKeys = GetKeys(unexpected).ConvertOrCastToCollection();
```

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -1593,12 +1593,13 @@ public AndConstraint<TAssertions> NotContain(string unexpected,
     public AndConstraint<TAssertions> NotContainAll(IEnumerable<string> values,
         [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
     {
-        ThrowIfValuesNullOrEmpty(values);
+        IList<string> strings = values?.ConvertOrCastToList();
```
jnyrup:
```suggestion
        ICollection<string> strings = values?.ConvertOrCastToCollection();
```

Src/FluentAssertions/Equivalency/Execution/ObjectReference.cs
```diff
@@ -69,5 +72,14 @@ public override string ToString()
...
-    public bool CompareByMembers => compareByMembers ?? @object?.GetType().OverridesEquals() == false;
+    /// <summary>
+    /// Indicates whether the equality comparison for the object should be based on its members rather than its default
+    /// implementation of <c>Equals</c>.
+    /// </summary>
+    /// <remarks>
+    /// This property returns <c>true</c> if the equality strategy is set to <see cref="EqualityStrategy.Members"/> or
+    /// <see cref="EqualityStrategy.ForceMembers"/>, and the object does not override its <c>Equals</c> method.
+    /// Otherwise, it returns <c>false</c>.
+    /// </remarks>
```
jnyrup:
```suggestion
    /// <summary>
    /// Indicates whether the equality comparison for the object should be based on its members rather than its default
    /// implementation of <see cref="object.Equals(object)" />.
    /// </summary>
    /// <remarks>
    /// This property returns <see langword="true" /> if the equality strategy is set to <see cref="EqualityStrategy.Members"/> or
    /// <see cref="EqualityStrategy.ForceMembers"/>, and the object does not override its <see cref="object.Equals(object)" /> method.
    /// Otherwise, it returns <see langword="false" />.
    /// </remarks>
```

Src/FluentAssertions/Execution/LateBoundTestFramework.cs
```diff
@@ -7,14 +7,15 @@ namespace FluentAssertions.Execution;
...
-    protected LateBoundTestFramework(bool loadAssembly = false)
-    {
-        this.loadAssembly = loadAssembly;
-        exceptionFactory = _ => throw new InvalidOperationException($"{nameof(IsAvailable)} must be called first.");
-    }
+    /// <summary>
+    /// When set to <c>true</c>, the assembly specified by the <c>AssemblyName</c> property will
+    /// be dynamically loaded if it is not already loaded in the application domain.
+    /// When set to <c>false</c>, the framework will not attempt to load the assembly dynamically.
+    /// </summary>
```
jnyrup: Found by [MA0154](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0154.md)

```suggestion
    /// <summary>
    /// When set to <see langword="true" />, the assembly specified by the <see cref="AssemblyName"/> property will
    /// be dynamically loaded if it is not already loaded in the application domain.
    /// When set to <see langword="false" />, the framework will not attempt to load the assembly dynamically.
    /// </summary>
```

Src/FluentAssertions/Common/KeyValuePairCollectionExtensions.cs
jnyrup, line 1: I see the changes here are due to Qodana's
`Local function has the same name as a method and hides it`

I don't see why it's a problem 🤔 

How about we keep them as local functions, but rename them?

Otherwise, how about we also extract `ContainsKey` and `GetValue` into static methods such that we only call `ObjectExtensions.GetComparer` in there.

Src/FluentAssertions/Equivalency/Execution/ObjectReference.cs
```diff
@@ -13,14 +13,15 @@ internal class ObjectReference
...
 
-    public ObjectReference(object @object, string path, bool? compareByMembers = null)
+    public ObjectReference(object @object, string path, EqualityStrategy equalityStrategy = EqualityStrategy.Members)
     {
         this.@object = @object;
         this.path = path;
-        this.compareByMembers = compareByMembers;
+
+        bool compareByMembers = equalityStrategy is EqualityStrategy.Members or EqualityStrategy.ForceMembers;
+        CompareByMembers = compareByMembers && @object?.GetType().OverridesEquals() == false;
```
jnyrup: When an `ObjectReference` is constructed from `EquivalencyValidationContext.IsCyclicReference` the call to  `Options.GetEqualityStrategy` have called `OverridesEquals` if necessary and cached it.
That's also how the current code works.

On the other hand, when constructing an `ObjectReference` from `ObjectGraph.ObjectGraph` we immediate call `trackerIsCyclicReference` which calls `ObjectReference.CompareByMembers` so there's no longer a point in lazy-evaluating `CompareByMembers`.

https://github.com/fluentassertions/fluentassertions/blob/cc607f1d3d539f246ddb239cc3decb06d3807d41/Src/FluentAssertions/Formatting/Formatter.cs#L218-L219

We can satisfy both things and remove the Boolean parameter by having two constructors.
```cs
public ObjectReference(object @object, string path, EqualityStrategy equalityStrategy)
{
    this.@object = @object;
    this.path = path;

    CompareByMembers = equalityStrategy is EqualityStrategy.Members or EqualityStrategy.ForceMembers;
}

public ObjectReference(object @object, string path)
{
    this.@object = @object;
    this.path = path;

    CompareByMembers = @object?.GetType().OverridesEquals() == false;
}
```

jnyrup (commented):
Good changes!

I often look in analyzer suggestions and it's now flooded with suggestions that are currently disabled.
Could we keep those disabled if we don't care about them?

AV1551
AV1706
AV2305
SA1615
SA1623

Src/FluentAssertions/Equivalency/Execution/ObjectReference.cs
dennisdoomen: `<see langword="true" />` is not recognized by Rider and doesn't do anything either.

Src/FluentAssertions/Common/KeyValuePairCollectionExtensions.cs
dennisdoomen: They were already fixed in the last commit before your review.

jnyrup: My comment was to also introduce `ContainsKeyLocal` and `GetValueLocal` like how you did with `TryGetValueLocal`.
Such that `ObjectExtensions.GetComparer` is only called on the fallback path.

dennisdoomen: If it wasn't caused by my changes, let's do that at a later point .

jnyrup: It's caused by inlining the local methods `ContainsKey` and `GetValue`, but I'll handle it in another PR to get this one merged.
It's a nitpick from my side and certainly not blocking.

dennisdoomen: Then approve the PR. By the way, making it clear that it's a nitpick can also help with the frustration of having to deal with multiple batches of comments.

PR #3067, opened 2025-06-29
Ensured WithTracing is safe when used with BeEquivalentTo globally

dennisdoomen:
Solves #2619

### DIFF
--- a/Src/FluentAssertions/Equivalency/EquivalencyValidationContext.cs
+++ b/Src/FluentAssertions/Equivalency/EquivalencyValidationContext.cs
@@ -82,6 +82,20 @@ public class EquivalencyValidationContext : IEquivalencyValidationContext
 
     public ITraceWriter TraceWriter { get; set; }
 
+    /// <summary>
+    /// This method ensures that tracing starts with a fresh state when invoked.
+    /// </summary>
+    internal void ResetTracing()
+    {
+        // SMELL: We need to ensure that if tracing is enabled using the built-in internal writer,
+        // we start with a fresh instance of InternalTraceWriter. We can't add extend ITraceWriter
+        // as that would be a breaking change.
+        if (TraceWriter is InternalTraceWriter)
+        {
+            TraceWriter = new InternalTraceWriter();
+        }
+    }
+
     public override string ToString()
     {
         return Invariant($"{{Path=\"{CurrentNode.Subject.PathAndName}\"}}");
--- a/Src/FluentAssertions/Equivalency/EquivalencyValidator.cs
+++ b/Src/FluentAssertions/Equivalency/EquivalencyValidator.cs
@@ -13,6 +13,8 @@ internal class EquivalencyValidator : IValidateChildNodeEquivalency
 
     public void AssertEquality(Comparands comparands, EquivalencyValidationContext context)
     {
+        context.ResetTracing();
+
         using var scope = new AssertionScope();
 
         RecursivelyAssertEquivalencyOf(comparands, context);
--- a/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyOptions.cs
+++ b/Src/FluentAssertions/Equivalency/SelfReferenceEquivalencyOptions.cs
@@ -747,7 +747,7 @@ public abstract class SelfReferenceEquivalencyOptions<TSelf> : IEquivalencyOptio
     /// </summary>
     public TSelf WithTracing(ITraceWriter writer = null)
     {
-        TraceWriter = writer ?? new StringBuilderTraceWriter();
+        TraceWriter = writer ?? new InternalTraceWriter();
         return (TSelf)this;
     }
 
--- a/Src/FluentAssertions/Equivalency/Tracing/InternalTraceWriter.cs
+++ b/Src/FluentAssertions/Equivalency/Tracing/InternalTraceWriter.cs
@@ -1,0 +1,5 @@
+namespace FluentAssertions.Equivalency.Tracing;
+
+internal sealed class InternalTraceWriter : StringBuilderTraceWriter
+{
+}
--- a/Tests/FluentAssertions.Equivalency.Specs/TracingSpecs.cs
+++ b/Tests/FluentAssertions.Equivalency.Specs/TracingSpecs.cs
@@ -1,0 +1,34 @@
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Sdk;
+
+namespace FluentAssertions.Equivalency.Specs;
+
+[Collection("ConfigurationSpecs")]
+public class TracingSpecs
+{
+    [Fact]
+    public void Tracing_must_be_safe_when_executed_concurrently()
+    {
+        try
+        {
+            // Arrange
+            AssertionConfiguration.Current.Equivalency.Modify(e => e.WithTracing());
+
+            Parallel.For(1, 10_000, (_, _) =>
+            {
+                try
+                {
+                    new { A = "a" }.Should().BeEquivalentTo(new { A = "b" });
+                }
+                catch (XunitException)
+                {
+                }
+            });
+        }
+        finally
+        {
+            AssertionEngine.ResetToDefaults();
+        }
+    }
+}
--- a/docs/_pages/releases.md
+++ b/docs/_pages/releases.md
@@ -13,6 +13,10 @@ sidebar:
 
 * Add `WithStrictTyping` and `WithStrictTypingFor` to `BeEquivalentTo` to ensure types also match - [#3066](https://github.com/fluentassertions/fluentassertions/pull/3066)
 
+## Fixes
+
+* Ensured `WithTracing` is safe when used with `BeEquivalentTo` globally - [#3067](https://github.com/fluentassertions/fluentassertions/pull/3067)
+
 ## 8.4.0
 
 ## Enhancements


Src/FluentAssertions/Equivalency/Tracing/StringBuilderTraceWriter.cs
```diff
@@ -17,25 +17,41 @@ public IDisposable AddBlock(string trace)
...
-        return builder.ToString();
+        return GetStateSafely().Builder.ToString();
+    }
+
+    private State GetStateSafely()
+    {
+        stateProvider.Value ??= new();
+
+        return stateProvider.Value;
+    }
```
jnyrup: We need to ensure that each test run has a new or clean `State`.

test1: gets a fresh `State`
test1: writes a line to its `State`
test1: passes, so `ITraceWriter.ToString` is not called
test2: gets the `State` which test1 had previously used
test2: writes a line to its `State`
test2: fails, so `ITraceWriter.ToString` writes lines from both test1 and test2

I ran the test below to capture traces.
See how 4 out of 19 traces have double entries because the `State` is reused.

```cs
[Fact]
public void Tracing_must_be_safe_when_executed_concurrently()
{
    var logs = new ConcurrentBag<string>();
    try
    {
        // Arrange
        AssertionConfiguration.Current.Equivalency.Modify(e => e.WithTracing());

        Parallel.For(1, 20, (_, _) =>
        {
            try
            {
                new { A = "a" }.Should().BeEquivalentTo(new { A = "b" });
            }
            catch (XunitException e)
            {
                logs.Add(e.Message);
            }
        });
    }
    finally
    {
        AssertionEngine.ResetToDefaults();
    }

    Assert.Fail(new StringBuilder().AppendJoin(Environment.NewLine, logs).ToString());
}
```

```
Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

Expected property A to be "b", but "a" differs near "a" (index 0).

With trace:
  {
    property A
    {
      Equivalency was proven by StringEqualityEquivalencyStep
    }
    Equivalency was proven by StructuralEqualityEquivalencyStep
  }

```

dennisdoomen: That's a good point. Basically every time somebody calls BeEquivalentTo, we need to clear the tracing. That shouldn't be too difficult.

dennisdoomen: That's a good point. Basically every time somebody calls BeEquivalentTo, we need to clear the tracing. That shouldn't be too difficult.

Src/FluentAssertions/Equivalency/Tracing/StringBuilderTraceWriter.cs
```diff
@@ -17,25 +17,41 @@ public IDisposable AddBlock(string trace)
...
+
+    private class State
+    {
+        public StringBuilder Builder { get; } = new();
+
+        public int Depth { get; set; }
```
jnyrup:
```suggestion
        public int Depth { get; set; } = 1;
```

Src/FluentAssertions/Equivalency/Tracing/StringBuilderTraceWriter.cs
jnyrup: Reduces from two to a single call to  `AsyncLocal<T>.Value` getter
```suggestion
    private State GetStateSafely() => stateProvider.Value ??= new();
```

Src/FluentAssertions/Equivalency/Tracing/StringBuilderTraceWriter.cs
jnyrup, line 1: Suggestion: Call `GetStateSafely()` a single time from the three public methods `AddSingle`, `AddBlock` and `ToString` and pass as a parameter to `WriteLine`.

example:
```cs
public IDisposable AddBlock(string trace)
{
    State state = GetStateSafely();
    WriteLine(state, trace);
    WriteLine(state, "{");
    state.Depth++;

    return new Disposable(() =>
    {
        state.Depth--;
        WriteLine(state, "}");
    });
}
```

Src/FluentAssertions/Equivalency/Tracing/StringBuilderTraceWriter.cs
jnyrup: I haven't thought it through, but here goes. 
Perhaps the tracer should live with the context or scope we pass along, such that it also works with chained assertions or when using an `AssertionScope` from a unit test.

dennisdoomen: Then it will never be assigned to `stateProvider.Value`

Src/FluentAssertions/Equivalency/Tracing/StringBuilderTraceWriter.cs
dennisdoomen: I've renamed the method to a property. I don't think the rest will matter much, unless you strongly believe that calling `stateProvider.Value` is noticeably expensive.

Src/FluentAssertions/Equivalency/Tracing/StringBuilderTraceWriter.cs
dennisdoomen: I came up with a solution that is not that elegant but will work. Any other proper solutions would require changing the `ITraceWriter` interface or changing the `WithTracing` method to take a factory to an `ITraceWriter`. Either way, it would be a breaking change.

Src/FluentAssertions/Equivalency/Tracing/StringBuilderTraceWriter.cs
jnyrup, line 1: With the introduction of `ResetTracing` it seems the extra hoops added `StringBuilderTraceWriter` are no longer needed as we don't share instances 🎉

Src/FluentAssertions/Equivalency/EquivalencyValidationContext.cs
```diff
@@ -80,6 +80,20 @@ public bool IsCyclicReference(object expectation)
...
 
+    /// <summary>
+    /// This method ensures that tracing starts with a fresh state when invoked.
+    /// </summary>
+    internal void ResetTracing()
+    {
+        // SMELL: We need to ensure that if tracing is enabled using the built-in writer,
+        // we start with a fresh instance of StringBuilderTraceWriter. We can't add extend ITraceWriter
+        // as that would be a breaking change.
+        if (TraceWriter is StringBuilderTraceWriter)
```
jnyrup: Should we use `if (TraceWriter?.GetType() == typeof(StringBuilderTraceWriter))` in case a user has subclassed `StringBuilderTraceWriter`? 🤔

dennisdoomen: Nah, even if they do, they can't override the methods of `ITraceWriter`, unless they use `new`, but then all bets are off anyway.

Src/FluentAssertions/Equivalency/Tracing/StringBuilderTraceWriter.cs
dennisdoomen: Yeah, you're right. Good catch!

Tests/FluentAssertions.Equivalency.Specs/DictionarySpecs.cs
```diff
@@ -514,7 +515,7 @@ public void
         // Arrange
         object object1 = new NonGenericDictionary { ["greeting"] = "hello" };
         object object2 = new NonGenericDictionary { ["greeting"] = "hello" };
-        var traceWriter = new StringBuilderTraceWriter();
+        var traceWriter = new CustomTraceWriter();
```
jnyrup: This is a breaking change if people are currently providing their own instance of `StringBuilderTraceWriter`.

We should be able to handle this by having a "marker" to tell us whether the instance was provided by the user or created by us.

e.g. introduce `InternalWriter` in FA
```cs
internal sealed class InternalWriter : StringBuilderTraceWriter;
```

such that we can differentiate between who created the `StringBuilderTraceWriter` instance.
This allows us to revert the changes to this file.

```cs
public TSelf WithTracing(ITraceWriter writer = null)
{
    TraceWriter = writer ?? new InternalWriter();
    return (TSelf)this;
}
```

```cs
internal void ResetTracing()
{
    // SMELL: We need to ensure that if tracing is enabled using the built-in writer,
    // we start with a fresh instance of StringBuilderTraceWriter. We can't add extend ITraceWriter
    // as that would be a breaking change.
    if (TraceWriter is InternalWriter)
    {
        TraceWriter = new InternalWriter();
    }
}
```

dennisdoomen: I like that.

PR #3076, opened 2025-07-20
Add support for inline assertions using Value.ThatMatches and Value.ThatSatisfies

dennisdoomen:
Closes #3070 

## IMPORTANT 

* [ ] If the PR touches the public API, the changes have been approved in a separate issue with the "api-approved" label.
* [ ] The code complies with the [Coding Guidelines for C#](https://www.csharpcodingguidelines.com/).
* [ ] The changes are covered by unit tests which follow the Arrange-Act-Assert syntax and the naming conventions such as is used [in these tests](../tree/develop/Tests/FluentAssertions.Equivalency.Specs/MemberMatchingSpecs.cs#L51-L430).
* [ ] If the PR adds a feature or fixes a bug, please update [the release notes](../tree/develop/docs/_pages/releases.md) with a functional description that explains what the change means to consumers of this library, which are published on the [website](https://fluentassertions.com/releases).
* [ ] If the PR changes the public API the changes needs to be included by running [AcceptApiChanges.ps1](../tree/develop/AcceptApiChanges.ps1) or [AcceptApiChanges.sh](../tree/develop/AcceptApiChanges.sh).
* [ ] If the PR affects [the documentation](../tree/develop/docs/_pages), please include your changes in this pull request so the documentation will appear on the [website](https://www.fluentassertions.com/introduction).
    * [ ] Please also run `./build.sh --target spellcheck` or `.\build.ps1 --target spellcheck` before pushing and check the good outcome

## CONTRIBUTOR LICENSE GRANT

By submitting this contribution, the contributor hereby irrevocably grants to the project owners and maintainers a perpetual, worldwide, royalty-free, irrevocable license to use, reproduce, modify, distribute, sublicense, and create derivative works of the contribution for any purpose and under any terms, including proprietary licensing.

The contributor waives any moral rights in the contribution to the extent permitted by law and agrees not to assert any claim of authorship or control over the contribution. The contributor represents that they are the sole author of the contribution and that it is provided free of any third-party claims.

The contributor understands and agrees that the maintainers may, at their sole discretion, use, license, or redistribute the contribution as part of any work and under any terms they choose, without further permission or attribution.

* [ ] I have read the Contributor License Grant and accept the conditions
* [ ] I'm interested in a free license for Fluent Assertions and will share my email address through sales@xceed.com

PR #3079, opened 2025-08-09
Follow-up to #3076

jnyrup:
## IMPORTANT 

## CONTRIBUTOR LICENSE GRANT

* [ ] I have read the Contributor License Grant and accept the conditions
* [ ] I'm interested in a free license for Fluent Assertions and will share my email address through sales@xceed.com

### DIFF
--- a/Src/FluentAssertions/Equivalency/Inlining/ConditionBasedInlineAssertion.cs
+++ b/Src/FluentAssertions/Equivalency/Inlining/ConditionBasedInlineAssertion.cs
@@ -22,9 +22,13 @@ internal class ConditionBasedInlineAssertion<T>(Expression<Func<T, bool>> condit
 
         assertionChain
             .ForCondition(comparands.Subject is T)
-            .FailWith("Expected {context:subject} to be of type {0}, but found {1}.", typeof(T), comparands.Subject?.GetType())
-            .Then
-            .ForCondition(condition.Compile()((T)comparands.Subject))
-            .FailWith("Expected {context:subject} to meet condition {0}, but it did not.", condition);
+            .FailWith("Expected {context:subject} to be of type {0}, but found {1}.", typeof(T), comparands.Subject?.GetType());
+
+        if (assertionChain.Succeeded)
+        {
+            assertionChain
+                .ForCondition(condition.Compile()((T)comparands.Subject))
+                .FailWith("Expected {context:subject} to meet condition {0}, but it did not.", condition);
+        }
     }
 }
--- a/Src/FluentAssertions/Equivalency/Inlining/ConditionBasedInlineAssertion.cs
+++ b/Src/FluentAssertions/Equivalency/Inlining/ConditionBasedInlineAssertion.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Diagnostics.CodeAnalysis;
 using System.Linq.Expressions;
 using FluentAssertions.Execution;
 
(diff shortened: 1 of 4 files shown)


Src/FluentAssertions/Equivalency/Inlining/ConditionBasedInlineAssertion.cs
```diff
@@ -9,22 +8,20 @@ namespace FluentAssertions.Equivalency.Inlining;
...
-
         assertionChain
             .ForCondition(comparands.Subject is T)
-            .FailWith("Expected {context:subject} to be of type {0}, but found {1}.", typeof(T), comparands.Subject?.GetType())
-            .Then
-            .ForCondition(condition.Compile()((T)comparands.Subject))
-            .FailWith("Expected {context:subject} to meet condition {0}, but it did not.", condition);
+            .FailWith("Expected {context:subject} to be of type {0}, but found {1}.", typeof(T), comparands.Subject?.GetType());
+
+        if (assertionChain.Succeeded)
```
dennisdoomen: Good catch. Instead of a separate `if`, what I should have done is 

```
assertionChain
    .ForCondition(comparands.Subject is T)
    .FailWith("Expected {context:subject} to be of type {0}, but found {1}.", typeof(T), comparands.Subject?.GetType())
    .Then
    .Given(() => (T)comparands.Subject)
    .ForCondition(subject => condition.Compile()(subject))
    .FailWith("Expected {context:subject} to meet condition {0}, but it did not.", condition);
```

Tests/FluentAssertions.Equivalency.Specs/InlineAssertionsSpecs.cs
```diff
@@ -75,24 +80,11 @@ public void Succeeds_for_a_matching_condition()
...
-        var expectation = new
-        {
-            Name = "John",
-            Age = Value.ThatMatches<int>(null)
-        };
-
-        // Act
-        var act = () => actual.Should().BeEquivalentTo(expectation);
+        // Arrange / Act
+        var act = () => Value.ThatMatches<int>(null);
```
dennisdoomen: Although this is technically correct, the point is that `Value.ThatMatches` is supposed to be used in combination with `BeEquivalentTo`. The fact that you moved the `null` check is an implementation detail and should not affect the "surface area" of the test.  Same for the other test.

dennisdoomen (changes requested):

Tests/FluentAssertions.Equivalency.Specs/InlineAssertionsSpecs.cs
jnyrup: Not sure I understand what you'd like.

Since the null guard moved to `Value.ThatMatches`, it's now evaluated before calling `BeEquivalentTo`.
So the test cannot be left be completely untouched, as only the call to `BeEquivalentTo` is currently wrapped in the `act` delegate.

To keep using `Value.ThatMatches` in the context an end-user would use it, I could write it as e.g.

```cs
// Arrange
var actual = new
{
    Name = "John",
    Age = 30
};

// Act
var act = () =>
{
    var expectation = new
    {
        Name = "John",
        Age = Value.ThatMatches<int>(null)
    };
    return actual.Should().BeEquivalentTo(expectation);
};

// Assert
act.Should().Throw<ArgumentNullException>().WithParameterName("condition");
```

Or am I completely missing what you're pointing at?

dennisdoomen: Yeah, that's what I meant as well, although I didn't realize it would mean wrapping the entire test in a lambda. But in the end, this more closely resembles how consumers use it.

PR #3094, opened 2025-09-14
Added System.Text.Json assertion APIs and BeEquivalentTo support

### LINKED ISSUE #2205: Assertions on System.Text.Json

dennisdoomen:
This PR adds comprehensive support for JSON assertion APIs in FluentAssertions along with `BeEquivalentTo` functionality for JSON nodes.

The PR introduces new assertion methods for `System.Text.Json.Nodes.JsonNode` and `JsonArray` types, including type checking methods like `BeNumeric()`, `BeString()`, `BeBool()`, `BeLocalDate()`, and `BeUtcDate()`. It also adds property verification methods such as `HaveProperty()` and array checking with `BeAnArray()`.

A key feature is the `BeEquivalentTo()` method that allows comparing JSON nodes directly to C# objects, with automatic conversion of JSON values to appropriate .NET types (including ISO 8601 date parsing). This comparison supports case-insensitive property matching through the `IgnoringJsonPropertyCasing()` option.

The implementation includes new matching rules specifically for JSON properties that work alongside existing member matching, and a JSON conversion step that automatically handles type conversions during equivalency comparisons. The system seamlessly integrates with the existing equivalency framework, allowing JSON nodes to be compared against complex object graphs with nested properties and collections.

Closes #2205


Src/FluentAssertions/Equivalency/IEquivalencyOptions.cs
```diff
@@ -121,4 +121,9 @@ public interface IEquivalencyOptions
     /// Enabling this option will replace all occurrences of <c>\r\n</c> and <c>\r</c> with <c>\n</c> in the strings before comparing them.
     /// </remarks>
     bool IgnoreNewlineStyle { get; }
+
+    /// <summary>
+    /// When set to <c>true</c>, the comparison will ignore the casing of JSON property names when comparing objects to JSON trees.
+    /// </summary>
+    bool IgnoreJsonPropertyCasing { get; }
```
jnyrup, line 128: Note: technically a breaking change as we're adding members to a public interface.
But searching github and grepp.app I couldn't find any implementers of this besides our own [`DataEquivalencyAssertionOptions`](https://github.com/fluentassertions/fluentassertions.datasets/blob/main/Src/FluentAssertions.DataSets/DataEquivalencyAssertionOptions.cs) but that class also derives from `EquivalencyOptions<TExpectation>` which implements `IEquivalencyOptions`.

So I'm good with this change.

Tests/FluentAssertions.Extensibility.Specs/HttpResponseMessageAssertions.cs
```diff
@@ -0,0 +1,27 @@
+#if NETCOREAPP3_1_OR_GREATER
```
jnyrup:
```suggestion
#if NET6_0_OR_GREATER
```

Tests/FluentAssertions.Extensibility.Specs/JsonWithInlineAssertionsSpecs.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            Id = "TestPackage",
+            Versions = new[]
+            {
+                new
+                {
+                    Version = "1.0.0",
+                    Description = Value.ThatSatisfies<string>(s => s.Should().Contain("description")),
+                    RepositoryUrl = "https://github.com/test/package",
+                    Owner = "testowner",                }
```
jnyrup:
```suggestion
                    Owner = "testowner"
                }
```

Tests/FluentAssertions.Extensibility.Specs/JsonWithInlineAssertionsSpecs.cs
```diff
@@ -0,0 +1,91 @@
+#if NET8_0_OR_GREATER
```
jnyrup:
```suggestion
#if NET6_0_OR_GREATER
```

Src/FluentAssertions/AssertionExtensions.cs
```diff
@@ -903,6 +909,25 @@ public static TaskCompletionSourceAssertions Should(this TaskCompletionSource tc
...
+    }
+
+    /// <summary>
+    /// Return an object that provides various assertion APIs that treat a <see cref="JsonArray"/>
+    /// as a collection of <see cref="JsonNode"/>s.
+    /// </summary>
+    [Pure]
+    public static GenericCollectionAssertions<JsonNode> Should([NotNull] this JsonArray jsonArray)
+    {
+        return new GenericCollectionAssertions<JsonNode>(jsonArray.ToArray(), AssertionChain.GetOrCreate());
```
jnyrup: Note: will throw NRE if `jsonArray` is null

Tests/ExampleExtensions/ExampleExtensions.csproj
```diff
@@ -1,15 +1,15 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFramework>netstandard2.0</TargetFramework>
+        <TargetFrameworks>netstandard2.0;net8.0</TargetFrameworks>
```
jnyrup:
```suggestion
        <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
```

Tests/FluentAssertions.Equivalency.Specs/FluentAssertions.Equivalency.Specs.csproj
```diff
@@ -27,6 +27,7 @@
   <ItemGroup>
     <PackageReference Include="JetBrains.Annotations" Version="2024.3.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
+    <PackageReference Include="System.Text.Json" Version="8.0.5" />
```
jnyrup: Shouldn't be necessary when only enabling these assertions for net6+
```suggestion
```

Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
```diff
@@ -0,0 +1,481 @@
...
+    /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+    /// </param>
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+    /// </param>
+    public AndWhichConstraint<JsonNodeAssertions, JsonNode> HaveProperty(string code,
+        [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
+    {
+        CurrentAssertionChain
+            .ForCondition(Subject[code] is not null)
```
jnyrup: If `Subject` is null, this throws NRE

Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
```diff
@@ -0,0 +1,481 @@
...
+    /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+    /// </param>
+    public AndConstraint<JsonNodeAssertions> NotHaveProperty(string code,
+        [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
+    {
+        CurrentAssertionChain
+            .ForCondition(Subject[code] is null)
+            .BecauseOf(because, becauseArgs)
+            .FailWith("Did not expect {context:JSON node} to have property {0}{reason}.", code);
+
```
jnyrup: If `Subject` is null, this throws NRE

Tests/FluentAssertions.Extensibility.Specs/JsonWithInlineAssertionsSpecs.cs
```diff
@@ -0,0 +1,92 @@
...
+                {
+                    Version = "1.0.0",
+                    Description = Value.ThatMatches<string>(s => s.Contains("Test")),
+                    RepositoryUrl = "https://github.com/test/package",
+                    Owner = "testowner",
+                }
+            }
+        });
+
+        // Assert
```
jnyrup: empty assert section?

also applies to test below

Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
```diff
@@ -0,0 +1,488 @@
...
+    /// </param>
+    public AndWhichConstraint<JsonNodeAssertions, JsonNode> HaveProperty(string code,
+        [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
+    {
+        CurrentAssertionChain
+            .BecauseOf(because, becauseArgs)
+            .ForCondition(Subject is not null)
+            .FailWith("Cannot assert the existence of a property on a <null> JSON node{reason}.")
+            .Then
+            .ForCondition(Subject is not null && Subject[code] is not null)
```
jnyrup:
```suggestion
            .ForCondition(Subject?[code] is not null)
```

Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
```diff
@@ -0,0 +1,488 @@
...
+            .FailWith("Expected {context:JSON node} to have property {0}{reason}.", code);
+
+        return new AndWhichConstraint<JsonNodeAssertions, JsonNode>(this, Subject[code]);
```
jnyrup:
```suggestion
        return new AndWhichConstraint<JsonNodeAssertions, JsonNode>(this, Subject?[code]);
```

Src/FluentAssertions/Equivalency/JsonProperty.cs
```diff
@@ -0,0 +1,145 @@
...
+    /// <remarks>
+    /// Whether the name matching is case-sensitive or not is determined by <paramref name="nameComparisonMode"/>.
+    /// </remarks>
+    /// <returns>
+    /// An instance of <see cref="JsonProperty"/> if a matching property was found, or <c>null</c> otherwise.
+    /// </returns>
+    public static JsonProperty Find(IMember expectedMember, INode parent, object subject, StringComparison nameComparisonMode)
+    {
+        var jsonNode = subject as JsonObject;
+        if (jsonNode is null)
```
jnyrup:
```suggestion
        if (subject is not JsonObject jsonNode)
```

Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
```diff
@@ -0,0 +1,488 @@
...
+    /// </param>
+    public AndWhichConstraint<JsonNodeAssertions, IEnumerable<JsonNode>> BeAnArray(
+        [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
+    {
+        CurrentAssertionChain
+            .ForCondition(Subject is JsonArray)
+            .BecauseOf(because, becauseArgs)
+            .FailWith("Expected {context:JSON node} to be an array{reason}, but {0} is not.", Subject);
+
+        return new AndWhichConstraint<JsonNodeAssertions, IEnumerable<JsonNode>>(this, (Subject as JsonArray)!.AsEnumerable());
```
jnyrup: Otherwise we throw an NRE if wrapped inside an `AssertionScope`

```suggestion
        return new AndWhichConstraint<JsonNodeAssertions, IEnumerable<JsonNode>>(this, (Subject as JsonArray)?.AsEnumerable());
```

dennisdoomen: Oh, good catch.

Tests/FluentAssertions.Equivalency.Specs/JsonNodeSpecs.cs
```diff
@@ -0,0 +1,551 @@
...
+    {
+        // Arrange
+        var node = JsonNode.Parse("{}");
+
+        // Act
+        var act = () => node.Should().BeEquivalentTo(new
+        {
+        });
+
+        // Assert - Currently throws InvalidOperationException for empty objects
```
jnyrup: Is this a todo note?

Src/FluentAssertions/Equivalency/JsonProperty.cs
```diff
@@ -0,0 +1,144 @@
...
+    /// <inheritdoc />
+    public Pathway Subject
+    {
+        get;
+        set;
+    }
+
+    /// <inheritdoc />
+    // ReSharper disable once UnassignedGetOnlyAutoProperty
+    public Pathway Expectation { get; }
```
jnyrup, line 47: If this property is only there to implement the interface, could we make it an explicit interface implementation then?
Or somehow else indicate why we have a read-only property.

Src/FluentAssertions/Equivalency/JsonProperty.cs
```diff
@@ -0,0 +1,144 @@
...
+    /// <inheritdoc />
+    public bool RootIsCollection
+    {
+        get => false;
+    }
+
+    /// <inheritdoc />
+    public void AdjustForRemappedSubject(IMember subjectMember)
+    {
+    }
```
jnyrup, line 70: Double check: is this method intentionally empty?

dennisdoomen: Same reason as above.

Issue #3093, opened 2025-09-14
[API Proposal]: Allow `WithoutMessage` when using Should().Throw()


dennisdoomen:
### Background and motivation

Sometimes you want to ensure a particular exception is thrown, but where the message does _not_ contain a particular wildcard phrase.

### API Proposal

Affects `ExceptionAssertions` and `ExceptionAssertionsExtensions`

### API Usage

Instead of 

```C#
act.Should().Throw<XunitException>().WithMessage("*existingProperty*wrongvalue*")
            .Which.Message.Should().NotContain("missingProperty");
```

it would allow

```C#
act.Should().Throw<XunitException>()
      .WithMessage("*existingProperty*wrongvalue*")
      .WithoutMessage("*missingProperty*");
```

### Are you willing to help with a proof-of-concept (as PR in that or a separate repo) first and as pull-request later on?

No

jnyrup:
Sounds good - approval from here.

PR #3100, opened 2025-09-20
Allow WithoutMessage when using Should().Throw()

dennisdoomen:
#3093

### DIFF
--- a/Tests/FluentAssertions.Specs/Exceptions/AggregateExceptionSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Exceptions/AggregateExceptionSpecs.cs
@@ -4,7 +4,7 @@ using Xunit.Sdk;
 
 namespace FluentAssertions.Specs.Exceptions;
 
-public class ExceptionAssertionSpecs
+public class AggregateExceptionSpecs
 {
     [Fact]
     public void When_method_throws_an_empty_AggregateException_it_should_fail()
--- a/Tests/FluentAssertions.Specs/Exceptions/ExceptionMessageSpecs.cs
+++ b/Tests/FluentAssertions.Specs/Exceptions/ExceptionMessageSpecs.cs
@@ -1,0 +1,18 @@
+using Xunit;
+
+namespace FluentAssertions.Specs.Exceptions;
+
+public class ExceptionMessageSpecs
+{
+    [Fact]
+    public void Can_assert_the_exception_message_does_not_include_a_specific_string()
+    {
+        // Arrange
+
+
+        // Act
+
+
+        // Assert
+    }
+}


Src/FluentAssertions/Specialized/ExceptionAssertions.cs
```diff
@@ -84,24 +84,74 @@ public virtual ExceptionAssertions<TException> WithMessage(string expectedWildca
...
+    /// </list>
+    /// </remarks>
+    public virtual ExceptionAssertions<TException> WithoutMessage(string wildcardPattern,
+        [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
+    {
+        assertionChain
+            .BecauseOf(because, becauseArgs)
+            .UsingLineBreaks
+            .ForCondition(Subject.Any())
+            .FailWith("Expected exception with message {0}{reason}, but no exception was thrown.", wildcardPattern);
```
jnyrup: With -> without

docs/_pages/releases.md
```diff
@@ -12,6 +12,7 @@ sidebar:
 ### What's new
 
 * Added support for `System.Text.Json.JsonNode` and `JsonArray` through new assertions as well as the `BeEquivalentTo` API - [#3094](https://github.com/fluentassertions/fluentassertions/pull/3094)
+* Added `WithoutMessage` allow asserting an exception message does not contain a wildcard pattern - [#3100](https://github.com/fluentassertions/fluentassertions/pull/3100)
```
jnyrup: "Allow" -> "to allow"

Src/FluentAssertions/Specialized/ExceptionAssertions.cs
```diff
@@ -84,24 +84,74 @@ public virtual ExceptionAssertions<TException> WithMessage(string expectedWildca
...
+    /// </item>
+    /// </list>
+    /// </remarks>
+    public virtual ExceptionAssertions<TException> WithoutMessage(string wildcardPattern,
+        [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
+    {
+        assertionChain
+            .BecauseOf(because, becauseArgs)
+            .UsingLineBreaks
+            .ForCondition(Subject.Any())
```
jnyrup, line 132: I can't recall how we do for other APIs. 
In logic:
If no exception is thrown it doesn't contain any message and hence it should pass.
(I'm on mobile right now, so can't conveniently look into it right now)

dennisdoomen: Because of the `AssertionChain`, if no exception is thrown and that was expected, this assertion will not even be executed. `WithoutMessage`

Src/FluentAssertions/Specialized/ExceptionAssertions.cs
dennisdoomen: Oh, good catch.

Src/FluentAssertions/Specialized/ExceptionAssertions.cs
```diff
@@ -84,24 +84,74 @@ public virtual ExceptionAssertions<TException> WithMessage(string expectedWildca
...
+            .FailWith("Expected exception without message matching {0}{reason}, but no exception was thrown.", wildcardPattern);
+
+        AssertExceptionMessage(message =>
+            message.Should().NotMatchEquivalentOf(wildcardPattern, because, becauseArgs));
```
jnyrup, line 136: This snippet throws an `InvalidOperationException` because we continue executing `AssertExceptionMessage` even though the assertion above failed.

```cs
Action notThrowException = () => { };
notThrowException.Should().Throw<InvalidOperationException>()
    .WithoutMessage("*bad*");
```

The same applies to `WithMessage`.

jnyrup (approved):
As missing handling of AssertionScopes also exists in `main` it's not blocking for merging this PR.

Src/FluentAssertions/Specialized/ExceptionAssertions.cs
dennisdoomen: Am I missing something? If I execute this, I get

```
Expected a <System.InvalidOperationException> to be thrown, but no exception was thrown.
```

jnyrup: Ahh. You need to remove the DebugNonUserCode attribute and attach a debugger

Issue #3088, opened 2025-08-30
Improve reporting of strings differences


dennisdoomen:
### Background and motivation

While using `BeEquivalentTo` to compare two graphs containing string properties, I got the following error:

```
Expected property actual.Versions[0].Description to be the same string, but they differ at index 0:
   ↓ (actual)
  "Fluently building and…"
  ""
   ↑ (expected).
Expected property actual.Versions[1].Description to be the same string, but they differ at index 0:
   ↓ (actual)
  "Fluently building and…"
  ""
   ↑ (expected).
```

This is quite annoying as I can't see the full string. Let's see if we can display the entire string if it's under a certain length. Also, the message itself is a bit weird from a grammatical point of view. 

### Are you willing help with a pull-request?

No

dennisdoomen:
Will be released through 8.8

PR #3101, opened 2025-09-21
Improve reporting of the differences between long strings

dennisdoomen:
Given the following test:

```csharp
[Fact]
public void Will_include_the_full_comparands_for_long_text()
{
    var subject = """
    The streets were quiet except for the faint hum of neon signs flickering in the night. A cat darted across the alley, chasing shadows only it could see. Somewhere, a radio whispered a half-forgotten song, mixing with the distant rumble of a train. The air carried the scent of rain on warm asphalt, heavy and restless. Each step felt like moving between moments—time stretching, folding, and twisting—while the city itself seemed to wait, holding its breath for what might come next.
    """;

    var expected = """
    The streets were quiet except for the faint hum of neon signs flicering in the night. A cat darted across the alley, chasing shadows only it could see. Somewhere, a radio whispered a half-forgotten song, mixing with the distant rumble of a train. The air carried the scent of rain on warm asphalt, heavy and restless. Each step felt like moving between moments—time stretching, folding, and twisting—while the city itself seemed to wait, holding its breath for what might come next.
    """;

    // Act
    subject.Should().Be(expected);
}
```

Before this PR, the result would be

```
Expected subject to be the same string, but they differ at index 66:
                   ↓ (actual)
  "…neon signs flickering in…"
  "…neon signs flicering in…"
                   ↑ (expected).
```

With the line length extended to 80 and the full subject and expectation included when it was truncated, this will be reported now

```
Expected subject to be a match with the expectation, but it differs at index 66:
                   ↓ (actual)
  "…neon signs flickering in the night. A cat darted across the alley, chasing shadows…"
  "…neon signs flicering in the night. A cat darted across the alley, chasing shadows…"
                   ↑ (expected)

Full expectation:

    "The streets were quiet except for the faint hum of neon signs flicering in the night. A cat darted across the alley, chasing shadows only it could see. Somewhere, a radio whispered a half-forgotten song, mixing with the distant rumble of a train. The air carried the scent of rain on warm asphalt, heavy and restless. Each step felt like moving between moments—time stretching, folding, and twisting—while the city itself seemed to wait, holding its breath for what might come next.",

Full subject:

    "The streets were quiet except for the faint hum of neon signs flickering in the night. A cat darted across the alley, chasing shadows only it could see. Somewhere, a radio whispered a half-forgotten song, mixing with the distant rumble of a train. The air carried the scent of rain on warm asphalt, heavy and restless. Each step felt like moving between moments—time stretching, folding, and twisting—while the city itself seemed to wait, holding its breath for what might come next."
```

**Note** If `BeEquivalentTo` runs a string comparison, we will _not_ include the full comparands, unless the new option `IncludingFullStringsInDifference` is used.

#3088


Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -14,7 +14,6 @@ namespace FluentAssertions.Primitives;
 /// <summary>
 /// Contains a number of methods to assert that a <see cref="string"/> is in the expected state.
 /// </summary>
-[DebuggerNonUserCode]
```
jnyrup, line 17: Intentional removal?

Src/FluentAssertions/Equivalency/IEquivalencyOptions.cs
```diff
@@ -122,6 +122,12 @@ public interface IEquivalencyOptions
     /// </remarks>
     bool IgnoreNewlineStyle { get; }
 
+    /// <summary>
+    /// Gets a value indicating whether the differences between string properties should
+    /// include the full values of the subject and expectation instead of just the fragment that differs.
+    /// </summary>
+    bool IncludeFullStringsInDifference { get; }
+
```
jnyrup, line 130: Approving breaking change, same reason as in https://github.com/fluentassertions/fluentassertions/pull/3094#discussion_r2352503339

Src/FluentAssertions/Primitives/StringAssertions.cs
```diff
@@ -2004,23 +2003,29 @@ private static bool HasMixedOrNoCase(string value)
...
     {
         Guard.ThrowIfArgumentIsNull(config);
 
         EquivalencyOptions<string> options = config(AssertionConfiguration.Current.Equivalency.CloneDefaults<string>());
 
-        var expectation = new StringValidator(assertionChain,
-            new StringEqualityStrategy(options.GetStringComparerOrDefault(), "be"),
-            because, becauseArgs);
+        StringEqualityStrategy strategy = new(options.GetStringComparerOrDefault(), "be");
+        strategy.IncludeFullDetails = options.IncludeFullStringsInDifference;
```
jnyrup: I guess we also need to set `IncludeFullDetails = options.IncludeFullStringsInDifference` on line 155

Src/FluentAssertions/Primitives/StringEqualityStrategy.cs
```diff
@@ -18,6 +19,12 @@ public StringEqualityStrategy(IEqualityComparer<string> comparer, string predica
         this.predicateDescription = predicateDescription;
     }
 
+    /// <summary>
+    /// Gets a value indicating whether the differences between string properties should
+    /// include the full values of the subject and expectation instead of just the fragment that differs.
+    /// </summary>
+    public bool IncludeFullDetails { get; set; } = true;
```
jnyrup, line 26: Shouldn't this default to false, such that one has to call `IncludingFullStringsInDifference()`?

Src/FluentAssertions/Primitives/StringAssertions.cs
dennisdoomen: Oops

Src/FluentAssertions/Primitives/StringEqualityStrategy.cs
dennisdoomen: I want the new behavior to be the default for normal assertions, with the exception of string assertions within a `BeEquivalentTo` call.

Src/FluentAssertions/Primitives/StringAssertions.cs
dennisdoomen: No, because there, the default is correct. A `BeEquivalentTo` directly on a `string` will need to show the full detais.

jnyrup (approved):

PR #3105, opened 2025-09-28
JSON assertions did not properly handle floats, doubles and unsigned …

dennisdoomen:
This PR fixes a bug in JSON assertions where the library wasn't properly handling certain numeric types during comparisons. The core issue was that the JSON conversion logic only checked for `int` and `double` values, but missed other important numeric types like unsigned integers (`uint`, `ulong`), signed long integers (), and floats (`float`). `long`

The fix adds support for these missing numeric types by extending the conversion step to handle `uint`, , `ulong`, and `float` values. Additionally, it introduces a new `IsNumeric()` helper method that properly identifies all numeric JSON values by checking for , `ulong`, or `double` types. `long``long`

Thanks to @jnyrup for the discovery.

### DIFF
--- a/Src/FluentAssertions/Equivalency/Steps/JsonConversionStep.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/JsonConversionStep.cs
@@ -14,9 +14,13 @@ public class JsonConversionStep : IEquivalencyStep
     {
         if (comparands.Subject is JsonValue json)
         {
-            if (json.TryGetValue(out int intValue))
+            if (json.TryGetValue(out long longValue))
             {
-                comparands.Subject = intValue;
+                comparands.Subject = longValue;
+            }
+            else if (json.TryGetValue(out ulong ulongValue))
+            {
+                comparands.Subject = ulongValue;
             }
             else if (json.TryGetValue(out double doubleValue))
             {
--- a/Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
+++ b/Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
@@ -122,7 +122,7 @@ public class JsonNodeAssertions<T> : ReferenceTypeAssertions<T, JsonNodeAssertio
     }
 
     /// <summary>
-    /// Asserts that the current <see cref="JsonNode"/> represents a 32-bit signed integer.
+    /// Asserts that the current <see cref="JsonNode"/> represents a numeric value.
     /// </summary>
     /// <param name="because">
     /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
@@ -131,20 +131,43 @@ public class JsonNodeAssertions<T> : ReferenceTypeAssertions<T, JsonNodeAssertio
     /// <param name="becauseArgs">
     /// Zero or more objects to format using the placeholders in <paramref name="because" />.
     /// </param>
-    public AndWhichConstraint<JsonNodeAssertions<T>, long> BeNumeric(
+    public AndWhichConstraint<JsonNodeAssertions<T>, string> BeNumeric(
         [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
     {
         CurrentAssertionChain
-            .ForCondition(Subject is JsonValue value && value.TryGetValue<int>(out _))
+            .ForCondition(Subject is JsonValue value && value.IsNumeric())
             .BecauseOf(because, becauseArgs)
-            .FailWith("Expected {context:JSON node} to be an Int32{reason}, but {0} is not.", Subject);
+            .FailWith("Expected {context:JSON node} to be a numeric value{reason}, but {0} is not.", Subject);
 
-        var actualValue = Subject is JsonValue jsonValue && jsonValue.TryGetValue<long>(out var result) ? result : 0;
-        return new AndWhichConstraint<JsonNodeAssertions<T>, long>(this, actualValue);
+        return new AndWhichConstraint<JsonNodeAssertions<T>, string>(this, (Subject as JsonValue)?.ToString() ?? string.Empty);
     }
 
     /// <summary>
-    /// Asserts that the current <see cref="JsonNode"/> does not represent a 32-bit signed integer.
+    /// Asserts that the current <see cref="JsonNode"/> represents a numeric value and returns
+    /// the actual value provided it can be converted to <typeparamref name="TValue"/>.
+    /// </summary>
+    /// <param name="because">
+    /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+    /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+    /// </param>
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+    /// </param>
+    public AndWhichConstraint<JsonNodeAssertions<T>, TValue> BeNumeric<TValue>(
+        [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
+    {
+        TValue actualValue = default;
+
+        CurrentAssertionChain
+            .ForCondition(Subject is JsonValue value && value.IsNumeric() && value.TryGetValue(out actualValue))
+            .BecauseOf(because, becauseArgs)
+            .FailWith("Expected {context:JSON node} to be a numeric value{reason}, but {0} is not.", Subject);
+
+        return new AndWhichConstraint<JsonNodeAssertions<T>, TValue>(this, actualValue);
+    }
+
+    /// <summary>
+    /// Asserts that the current <see cref="JsonNode"/> does not represent a numeric value.
     /// </summary>
     /// <param name="because">
     /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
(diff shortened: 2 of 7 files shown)


Src/FluentAssertions/Equivalency/Steps/JsonConversionStep.cs
```diff
@@ -18,6 +18,22 @@ public EquivalencyResult Handle(Comparands comparands, IEquivalencyValidationCon
             {
                 comparands.Subject = intValue;
             }
+            else if (json.TryGetValue(out uint uintValue))
+            {
+                comparands.Subject = uintValue;
+            }
```
jnyrup: All `int` and `uint` values can be represented as a `long`, so we can delete the `int` and `uint` cases.

Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
jnyrup: This will return the wrong value if the `JsonValue` is a `double` or an `ulong` not representable as a `long`.

I think we'll need to change this from `AndWhichContraint<, long>` to `AndConstraint<>`, or add a generic constraint to pass on to `TryGetValue` 

e.g.
```cs
/// <summary>
/// Asserts that the current <see cref="JsonNode"/> represents a numeric value.
/// </summary>
/// <param name="because">
/// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
/// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
/// </param>
/// </param>
public AndConstraint<JsonNodeAssertions<T>> BeNumeric(
    [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
{
    CurrentAssertionChain
        .ForCondition(Subject is JsonValue value && value.IsNumeric())
        .FailWith("Expected {context:JSON node} to be a numeric value{reason}, but {0} is not.", Subject);

    return new AndConstraint<JsonNodeAssertions<T>>(this);
}

/// <summary>
/// Asserts that the current <see cref="JsonNode"/> represents a numeric value.
/// </summary>
/// <param name="because">
/// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
/// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
/// </param>
/// </param>
public AndWhichConstraint<JsonNodeAssertions<T>, TValue> BeNumeric<TValue>(
    [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
{
    TValue actualValue = default;

    CurrentAssertionChain
        .ForCondition(Subject is JsonValue jsonValue && jsonValue.TryGetValue(out actualValue))
        .FailWith("Expected {context:JSON node} to be a numeric value{reason}, but {0} is not.", Subject);

    return new AndWhichConstraint<JsonNodeAssertions<T>, TValue>(this, actualValue);
}
```

Src/FluentAssertions/Equivalency/Steps/JsonConversionStep.cs
```diff
@@ -18,6 +18,22 @@ public EquivalencyResult Handle(Comparands comparands, IEquivalencyValidationCon
...
+                comparands.Subject = longValue;
+            }
+            else if (json.TryGetValue(out ulong ulongValue))
+            {
+                comparands.Subject = ulongValue;
+            }
+            else if (json.TryGetValue(out float floatValue))
+            {
+                comparands.Subject = floatValue;
+            }
```
jnyrup: Not sure why we need float here?
As far as I know, every `float` is representable as a `double`.

Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
jnyrup:
```suggestion
    /// Asserts that the current <see cref="JsonNode"/> represents a numeric value.
```

jnyrup (changes requested):

Src/FluentAssertions/Equivalency/Steps/JsonConversionStep.cs
dennisdoomen: My thoughts were that I wanted to keep the "original" type in the object graph that results from the conversion, in such a way that you can compare floats with floats. But after your comment, I realize that that is not important at all.

Src/FluentAssertions/Equivalency/Steps/JsonConversionStep.cs
dennisdoomen: Same reason as I mentioned in your comment about ints/units.

Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
dennisdoomen: Or change the return type to a `AndWhichConstraint<JsonNodeAssertions<T>, decimal>`. That should fit every other numeric type. What do you think?

jnyrup: `decimal` can't represent all other numeric values.

Here's a counter example.
```cs
var doubleValue = "79228162514264337593543950336";
var canParseDouble = double.TryParse(doubleValue, CultureInfo.InvariantCulture, out _); // true
var canParseDecimal = decimal.TryParse(doubleValue, CultureInfo.InvariantCulture, out _); // false
```

We can use `decimal` in `EnumEqualityStep` because enums can only have integer types as their base type, and decimal covers all integer types.

dennisdoomen: OK, then your suggestion seems to be the best.

What did you mean with `EnumEqualityStep` though?

jnyrup: I thought you had might be inspired to use `decimal` by the code in `EnumEqualityStep`.

Src/FluentAssertions/Specialized/JsonValueExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System.Text.Json.Nodes;
+
+namespace FluentAssertions.Specialized;
+
+internal static class JsonValueExtensions
+{
+    public static bool IsNumeric(this JsonValue value)
+    {
+        return value.TryGetValue<long>(out _) || value.TryGetValue<ulong>(out _) || value.TryGetValue<double>(out _);
```
jnyrup: From https://stackoverflow.com/a/71590219
```suggestion
        return value.GetValue<JsonElement>().ValueKind == JsonValueKind.Number;
```

Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
dennisdoomen: I think I'm going to settle on returning the original string through the `AndWhichConstraint` and add the generic overload you mentioned. All of this is technically a breaking change, but since we just release 8.7 and not a lot of people know about the `Which` construct, I'm pretty sure nobody will notice.

jnyrup: Sounds good 👍

Src/FluentAssertions/Specialized/JsonValueExtensions.cs
dennisdoomen: How did you figure that out? 🫣

jnyrup: Good ol' Stack Overflow.

Since `JsonValueKind.Number` is available in .NET 6 I thought there had to be a way to get it.

Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
```diff
@@ -131,16 +131,39 @@ public AndConstraint<JsonNodeAssertions<T>> NotBeAnArray(
...
     /// </param>
-    public AndWhichConstraint<JsonNodeAssertions<T>, long> BeNumeric(
+    public AndWhichConstraint<JsonNodeAssertions<T>, string> BeNumeric(
         [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
     {
         CurrentAssertionChain
-            .ForCondition(Subject is JsonValue value && value.TryGetValue<int>(out _))
+            .ForCondition(Subject is JsonValue value && value.IsNumeric())
             .BecauseOf(because, becauseArgs)
             .FailWith("Expected {context:JSON node} to be an Int32{reason}, but {0} is not.", Subject);
```
jnyrup:
```suggestion
            .FailWith("Expected {context:JSON node} to be a numeric value{reason}, but {0} is not.", Subject);
```

Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
```diff
@@ -131,16 +131,39 @@ public AndConstraint<JsonNodeAssertions<T>> NotBeAnArray(
...
+    /// </param>
+    public AndWhichConstraint<JsonNodeAssertions<T>, TValue> BeNumeric<TValue>(
+        [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
+    {
+        TValue actualValue = default;
+
+        CurrentAssertionChain
+            .ForCondition(Subject is JsonValue value && value.IsNumeric() && value.TryGetValue(out actualValue))
+            .BecauseOf(because, becauseArgs)
+            .FailWith("Expected {context:JSON node} to be an Int32{reason}, but {0} is not.", Subject);
```
jnyrup:
```suggestion
            .FailWith("Expected {context:JSON node} to be a numeric value{reason}, but {0} is not.", Subject);
```

Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
jnyrup:
```suggestion
    /// Asserts that the current <see cref="JsonNode"/> does not represent a numeric value.
```

Src/FluentAssertions/Specialized/JsonNodeAssertions.cs
```diff
@@ -157,7 +180,7 @@ public AndConstraint<JsonNodeAssertions<T>> NotBeNumeric(
         [StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
     {
         CurrentAssertionChain
-            .ForCondition(!(Subject is JsonValue value && value.TryGetValue<long>(out _)))
+            .ForCondition(Subject is not JsonValue value || !value.IsNumeric())
             .BecauseOf(because, becauseArgs)
             .FailWith("Did not expect {context:JSON node} to be an Int32{reason}, but {0} is.", Subject);
```
jnyrup:
```suggestion
            .FailWith("Did not expect {context:JSON node} to be a numeric value{reason}, but {0} is.", Subject);
```

PR #3107, opened 2025-10-01
Added support for excluding members marked [XmlIgnore], and updated BeXmlSerializable to use it

logiclrd:
This PR:

- Adds support for ignoring members marked `[XmlIgnore]`.
- Adds support for ignoring members marked `[IgnoreDataMember]`.
- Adds support for ignoring fields marked `[NonSerialized]`.
- Adds tests that exercise these kinds of types with respect to `.Should().BeXmlSerializable()` and `.Should().BeDataContractSerializable()`.
- Makes the tests pass by correcting the behaviour of `.BeXmlSerializable()` and `.BeDataContractSerializable()` using the new infrastructure.
- Documents everything.
- Updates the API surface description.

The first three changes are very similar to one another.

I haven't yet created an issue for the API changes.

## IMPORTANT 

## CONTRIBUTOR LICENSE GRANT

* [x] I have read the Contributor License Grant and accept the conditions
* [x] I'm interested in a free license for Fluent Assertions and will share my email address through sales@xceed.com


logiclrd:
Rebased to resolve merge conflict in releases.md. Took the opportunity to restructure and streamline those changes at the same time.

docs/_pages/objectgraphs.md
```diff
@@ -293,6 +293,167 @@ orderDto.Should().BeEquivalentTo(order, options => options
 
 Notice that you can also map properties to fields and vice-versa.
 
+### Ignored Members
```
dennisdoomen: 🔧This whole section doesn't belong here as it's not part of the built-in API. Instead, we should move it to a separate page under the "Tips" section.

docs/_pages/basicassertions.md
```diff
@@ -110,3 +110,11 @@ We’ve also added the possibility to assert that an object can be serialized an
 theObject.Should().BeXmlSerializable();
 theObject.Should().BeDataContractSerializable();
 ```
+
```
dennisdoomen, line 113: 🔧This does not belong here and can be removed. See also the other comment

docs/_pages/releases.md
```diff
@@ -7,6 +7,13 @@ sidebar:
   nav: "sidebar"
 ---
 
+## Unreleased
+
+### What's new
+
+* New member selection rules: `ExcludeXmlIgnoredMembersRule`, `ExcludeNonSerializedFieldsRule` and `ExcludeIgnoredDataMembersRule`. - [#3107](https://github.com/fluentassertions/fluentassertions/pull/3107)
```
dennisdoomen: 🔧Old note

docs/_pages/releases.md
```diff
@@ -7,6 +7,13 @@ sidebar:
...
+* New overload of `ObjectAssertions.BeXmlSerializable` that allow `EquivalencyOptions` to be configured, allowing the use of `ExcludeXmlIgnoredMembersRule`. (`BeDataContractSerializable` already has a comparable overload.) - [#3107](https://github.com/fluentassertions/fluentassertions/pull/3107)
```
dennisdoomen: 🔧The reference to `ExcludeXmlIgnoredMembersRule` can be removed as it doesn't exist
⛏️(nitpick) The `ObjectAssertions` prefix is not necessary.

Src/FluentAssertions/Equivalency/Field.cs
```diff
@@ -1,6 +1,8 @@
 using System;
 using System.ComponentModel;
 using System.Reflection;
+using System.Runtime.Serialization;
```
dennisdoomen: 🔧Must be removed

Src/FluentAssertions/ObjectAssertionsExtensions.cs
```diff
@@ -96,13 +121,59 @@ private static object CreateCloneUsingDataContractSerializer(object subject)
...
+    /// <see cref="AssertionConfiguration"/> class.
+    /// </param>
+    /// <param name="because">
+    /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+    /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+    /// </param>
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+    /// </param>
+    public static AndConstraint<ObjectAssertions> BeXmlSerializable(this ObjectAssertions assertions,
```
dennisdoomen, line 143: 🤔Why didn't you use a generic definition instead of using `object`? `BeDataContractSerializable` is using the same idea.

Src/FluentAssertions/ObjectAssertionsExtensions.cs
```diff
@@ -28,6 +28,31 @@ public static AndConstraint<ObjectAssertions> BeDataContractSerializable(this Ob
...
+    /// </param>
+    /// <param name="because">
+    /// A formatted phrase as is supported by <see cref="string.Format(string,object[])" /> explaining why the assertion
+    /// is needed. If the phrase does not start with the word <i>because</i>, it is prepended automatically.
+    /// </param>
+    /// <param name="becauseArgs">
+    /// Zero or more objects to format using the placeholders in <paramref name="because" />.
+    /// </param>
+    /// <exception cref="ArgumentNullException"><paramref name="options"/> is <see langword="null"/>.</exception>
+    public static AndConstraint<ObjectAssertions> BeDataContractSerializable(this ObjectAssertions assertions,
```
dennisdoomen, line 49: 🤔Why do we need this extra non-generic overload?

logiclrd: Because without it:

image

> The type arguments for method 'ObjectAssertionsExtensions.BeDataContractSerializable<T>(ObjectAssertions, string, params object[])' cannot be inferred from the usage. Try specifying the type arguments explicitly.

Src/FluentAssertions/ObjectAssertionsExtensions.cs
logiclrd: Without the non-generic definition, the full type name must always be specified in every call.

> The type arguments for method 'ObjectAssertionsExtensions.BeXmlSerializable<T>(ObjectAssertions, string, params object[])' cannot be inferred from the usage. Try specifying the type arguments explicitly.

image

docs/_pages/basicassertions.md
logiclrd: It makes sense that the information about ignored members be moved to `Tips` but it won't be very discoverable there. Perhaps a brief reference to the `Tips` section here, in place of providing specific details (especially if said details are out-of-date)?

dennisdoomen: As we've already discussed, your problem is very unique, so there's no point to mention it anywhere then in the tips section.

docs/_pages/tips.md
```diff
@@ -39,3 +39,164 @@ If you see something missing, please consider submitting a pull request.
 {% include assertion-comparison.html header1="MSTest" header2="Fluent Assertions" idPrefix="mstest-" caption="CollectionAssert"  examples=site.data.mstest-migration.collectionAssert %}
 {% include assertion-comparison.html header1="MSTest" header2="Fluent Assertions" idPrefix="mstest-" caption="StringAssert"      examples=site.data.mstest-migration.stringAssert %}
 {% include assertion-comparison.html header1="MSTest" header2="Fluent Assertions" idPrefix="mstest-" caption="Exceptions"        examples=site.data.mstest-migration.exceptions %}
+
```
dennisdoomen: 🔧 (required change) In my original comment, I said "separate page", so please do that. You also need to update the `navigation.yml` file.

logiclrd: Ohh 🙂 Coming right up.

logiclrd: Does the latest update look about right?

docs/_pages/basicassertions.md
```diff
@@ -110,3 +110,5 @@ We’ve also added the possibility to assert that an object can be serialized an
...
+See [Tips](tips.md) for additional information about the `BeXmlSerializable` and `BeDataContractSerializable` assertions and ignored members.
```
dennisdoomen: 🤔Don't you want to refer to that new page?

Tests/FluentAssertions.Specs/Primitives/ObjectAssertionSpecs.BeDataContractSerializable.cs
```diff
@@ -62,22 +62,37 @@ public void When_an_object_is_data_contract_serializable_but_doesnt_restore_all_
         }
 
         [Fact]
-        public void When_a_data_contract_serializable_object_doesnt_restore_an_ignored_property_it_should_succeed()
+        public void When_an_object_is_not_binary_serializable_and_has_properties_marked_IgnoreDataMember_and_ExcludeIgnoredDataMembers_is_not_used_it_should_fail()
```
dennisdoomen: 🔧This and the other test tiles don't make sense anymore.

logiclrd: Sorted 🙂

dennisdoomen: Sorry, I don't get those titles. There's no support for those attributes, so why would you mention them in the test title.
Also, we don't explicitly use technical names in test titles. The titles need to remain functional.

logiclrd: The attributes are processed by XML serialization and Data Contract serialization themselves. Their support isn't within FluentAssertions but within the framework.

```
BeDataContractSerializable.When_an_object_is_not_binary_serializable_and_has_properties_marked_IgnoreDataMember_it_should_fail()
```

This test verifies the expectation that `BeDataContractSerializable` fails _specifically_ when you hand an object that is not binary-serializable and which has properties marked `[IgnoreDataMember]`. This behaviour is baked into the definition of what `BeDataContractSerializable` is, because of how the underlying DataContract serializer system works. If you present it with that and the assertion _doesn't_ fail, then something is wrong with `BeDataContractSerializable`. If you give it an object that is not binary-serializable, but doesn't have `[IgnoreDataMember]`, or an object that has fields marked `[IgnoreDataMember]` but also the type is marked `[Serializable]`, then the assertion isn't expected to fail for that reason.

The names make sense in my brain, but I don't object to changing them. I'm just not sure what would be better names.

dennisdoomen: Ah yes, I wasn't fully in context when I was reviewing. In this case, you can use a more generic term like "and has explicitly excluded properties" or something like that.

Issue #378, opened 2016-03-22
Limited failure info from comparing collections of objects


Dm1tri:
I compare two collections of objects with .ShouldAllBeEquivalentTo
The collections has different object count and some of the properties of the objects has different values.
The test results only report about different object count between the collections but no info on the mismatches between the property values.
Is there a syntax to be able to achieve a report with the property value mismatches besides the object count difference?
Thanks

dennisdoomen:
Unfortunately that's a design limitation. It first verifies whether both objects are collections and then ensures their lengths are equal. See the implementation details [here](https://github.com/dennisdoomen/FluentAssertions/blob/master/Src/Core/Equivalency/EnumerableEquivalencyValidator.cs#L53).

johnddx:
This issue hinders me frequently.  If I'm comparing {a, b, d} and {a, b, c, d}, the useful difference is "c" (or first mismatch is "c" != "d").  Telling me that 3 != 4 is fairly unhelpful.

I sometimes wind up writing wrappers for the assert in order to get more verbose logging, which winds up being non-fluent.

PR #3133, opened 2025-12-14
Improve reporting the differences between differently sized collections in BeEquivalentTo

dennisdoomen:
This commit significantly improves how FluentAssertions reports differences when comparing collections of different sizes using `BeEquivalentTo`.

Previously, the error messages would simply state that collections had different counts. Now, the messages explicitly identify which items are **missing** from the actual collection and which items are **extraneous** (unexpected). The commit refactors the collection comparison logic by extracting two strategies: one for strictly-ordered comparisons and another for loosely-ordered comparisons that uses permutation analysis to find the best matches between items.

Additionally, it introduces a new `WithFullDump()` option that includes the complete contents of the subject collection in failure messages, making it easier to diagnose complex comparison failures. The implementation adds helper classes to track indexed items and uses caching to avoid redundant comparisons during the matching process.

An example of what it can report now is

```
Expected property actual[0].Name to be "Dennis" with a length of 6, but "Jits" has a length of 4, differs near "Jit" (index 0).
Expected property actual[0].Age to be 52, but found 13.
Expected actual to contain exactly one item, but found one extraneous item FluentAssertions.Equivalency.Specs.Customer
{
    Age = 16,
    Birthdate = <0001-01-01 00:00:00.000>,
    Id = 0L,
    Name = "Teddie"
}

Full dump of actual: {
    FluentAssertions.Equivalency.Specs.Customer
    {
        Age = 13,
        Birthdate = <0001-01-01 00:00:00.000>,
        Id = 0L,
        Name = "Jits"
    },
    FluentAssertions.Equivalency.Specs.Customer
    {
        Age = 16,
        Birthdate = <0001-01-01 00:00:00.000>,
        Id = 0L,
        Name = "Teddie"
    }
}

With configuration:
- Prefer the declared type of the members
- Compare enums by value
- Exclude member Id
- Match (JSON) member by name (or throw)
- Without automatic conversion
```

#378


Src/FluentAssertions/Common/StringExtensions.cs
```diff
@@ -178,4 +178,10 @@ public static string RenderAsIndentedBlock(this string message)
 
         return "    \"" + string.Join(Environment.NewLine + "    ", wrappedLines) + "\"";
     }
+
+    /// <summary>
+    /// Simplifies setting the invariant culture when using interpolated strings.
+    /// </summary>
+    public static string Invariant(FormattableString formattable)
+        => FormattableString.Invariant(formattable);
```
jnyrup: ❓ This reverts the work I did in #3130 to use `string.Create` for more efficient culture invariant string interpolation.
Could we continue using that?

Src/FluentAssertions/Equivalency/Steps/IndexedItem{T}.cs
```diff
@@ -0,0 +1,14 @@
+﻿#pragma warning disable CA1305
+
```
jnyrup: This seems unneeded?

Here and other places CA1305 is disabled

Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
```diff
@@ -1,219 +1,128 @@
...
-            }
-        }
+        return OrderingRules.IsOrderingStrictFor(new ObjectInfo(new Comparands(subjects, expectations, typeof(T[])),
+            currentNode));
     }
 
-    private void AssertElementGraphEquivalencyWithLooseOrdering<T>(object[] subjects, T[] expectations)
-    {
-        int failedCount = 0;
+#pragma warning disable CA1305
```
jnyrup, line 84: 🙃 Suggestion: 
Instead of disabling CA1305, we could add a polyfill of `Append` to `StringBuilderExtensions` like we have for `AppendLine`.

Src/FluentAssertions/Equivalency/Steps/IndexedItemCollection{T}.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// <inheritdoc />
+    IEnumerator IEnumerable.GetEnumerator() => ((IEnumerable)items).GetEnumerator();
+
+    public int Count => items.Count;
+
+    /// <inheritdoc />
+    public IndexedItem<T> this[int index]
+    {
+        get => items[index];
+    }
```
jnyrup: 🙃 syntax suggestion:
```suggestion
    public IndexedItem<T> this[int index] => items[index];
```

Src/FluentAssertions/Equivalency/Steps/IndexedItemCollection{T}.cs
```diff
@@ -0,0 +1,60 @@
...
+        foreach ((int Index, T Item) iterator in source.Index())
+        {
+            items.Add(new IndexedItem<T>(iterator.Item, iterator.Index));
+        }
+    }
+
+    public IndexedItemCollection(IEnumerable<IndexedItem<T>> source)
+    {
+        items.AddRange(source);
+    }
```
jnyrup, line 24: 🙃 Suggestion: Create the list directly instead of mutating it.

```suggestion
    private readonly List<IndexedItem<T>> items;

    public IndexedItemCollection(IEnumerable<T> source)
    {
        items = source.Select((item, index) => new IndexedItem<T>(item, index)).ToList();
    }

    public IndexedItemCollection(IEnumerable<IndexedItem<T>> source)
    {
        items = source.ToList();
    }
```

Tests/FluentAssertions.Equivalency.Specs/CollectionSpecs.cs
```diff
@@ -373,19 +373,25 @@ public void When_a_collection_contains_less_items_than_expected_it_should_throw(
         };
 
         // Act
-        Action act = () => subject.Should().BeEquivalentTo(expected);
+        Action act = () => subject.Should().BeEquivalentTo(expected, opt => opt.Excluding(x => x.Customers[0].Id));
```
jnyrup: ❓Why the added options delegate?

Same question for line 406

Tests/FluentAssertions.Equivalency.Specs/CollectionSpecs.cs
```diff
@@ -197,18 +197,18 @@ public void When_the_expectation_is_an_array_of_interface_type_it_should_respect
...
 
         // Act
         Action act = () => actual.Should().BeEquivalentTo(expected);
 
         // Assert
         act.Should().Throw<XunitException>()
-            .WithMessage("*be a collection with 2 item(s)*contains 1 item(s) less than*",
+            .WithMessage("*actual[0] to be of type System.String, but found {\"A\", \"B\"} of type System.String[]*",
                 "adding a `params object[]` overload cannot distinguish 'an array of objects' from 'an element which is an array of objects'");
     }
```
jnyrup, line 213: ❓Why the change of this test?

Src/FluentAssertions/Equivalency/Steps/LooselyOrderedEquivalencyStrategy.cs
```diff
@@ -0,0 +1,202 @@
...
+        List<object> remainingSubjects, IndexedItemCollection<TExpectation> expectationsWithIndexes,
+        Func<TExpectation, object, int, string[]> getFailures)
+    {
+        var bestScore = int.MaxValue;
+        List<(IndexedItem<TExpectation> ExpectationWithIndex, object, string[] Failures)> bestSet = null;
+
+        const int maxPermutations = 200_000;
+        int seen = 0;
+
+        foreach (IReadOnlyList<object> assignment in remainingSubjects.Permute())
```
jnyrup, line 146: ❓ 
Iterating through up to 200,000 sounds potentially quite costly to me?
How large can `failuresCache` grow?

To get some "numbers", I ran the snippet below in release mode with v8.8.0 and with this PR
It went from 76ms to 766ms.
I didn't compare memory usage.

```cs
var subject = Enumerable.Range(0, 50).Select(i => new { Value = i }).ToArray();
var expected = Enumerable.Range(51, 50).Select(i => new { Value = i }).ToArray();

subject.Should().BeEquivalentTo(expected, e => e.WithoutStrictOrdering());
```

These concerns might not actual be relevant, it just reminds me of all the measures we taken over the years to limit the time we spent generating failure messages.
#898
#1469
#1575

dennisdoomen:
> Just to let your know, I've begun reviewing this, but haven't finished yet.

No worries. I know it's a complicated change. Took me a couple of weeks.

Src/FluentAssertions/Equivalency/Steps/IndexedItem{T}.cs
jnyrup:
```suggestion
```

Src/FluentAssertions/Equivalency/Steps/IndexedItemCollection{T}.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions.Common;
+
+#pragma warning disable CA1305
+
```
jnyrup:
```suggestion
```

Src/FluentAssertions/Equivalency/Steps/LooselyOrderedEquivalencyStrategy.cs
```diff
@@ -0,0 +1,202 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using FluentAssertions.Common;
+using FluentAssertions.Equivalency.Tracing;
+using FluentAssertions.Execution;
+
+#pragma warning disable CA1305
+
```
jnyrup:
```suggestion
```

Src/FluentAssertions/Common/StringExtensions.cs
dennisdoomen: Used your `string.Create` inside `Invariant`.

Tests/FluentAssertions.Equivalency.Specs/CollectionSpecs.cs
dennisdoomen: The old implementation didn't even look at the items in the collection if the collections didn't have the same size. The new one does do that, and thus complains that `object` doesn't have any members to compare.

Tests/FluentAssertions.Equivalency.Specs/CollectionSpecs.cs
dennisdoomen: No clue. Maybe it failed in an intermediate version of my changes.

Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
dennisdoomen: I'm struggling to get that to work with the `ref AppendInterpolatedStringHandler`. Maybe you can try separately.

Src/FluentAssertions/Equivalency/Steps/LooselyOrderedEquivalencyStrategy.cs
dennisdoomen: As discussed, you will do some investigation.

jnyrup: I tried to look more into this to see if I could find a pathalogical example, that would make the runtime go completely nuts.
I used the test below and compiled using our CI profile such that FA is compiled in release mode and the test project in debug mode.
The time measurements are unscientifically by running the test a few times and picking the smallest value.
The `opt` case in the measurements is from applying the adjusted caching described in the end of this comment.

```cs
[Fact]
public void Expensive_test()
{
    int n = 50;
    var subject = Enumerable.Range(0, n).Select(i => Create(i)).ToArray();
    var expected = Enumerable.Range(n, n).Select(i => Create(i)).ToArray();

    subject.Should().BeEquivalentTo(expected, e => e.WithoutStrictOrdering());
}

private object Create(int i) => new { Value = i };
```

```
n=50: 
  main:  85ms
  pr:   900ms
  opt:  592ms
	
n=500: 
  main  226ms
  pr:  9700ms
  opt: 8800ms
```

I then changed `Create` to create a more complex object that could make `TryToMatch` more expensive.
Both the time to calculate the hashcode of `cacheKey` and in case of a cache miss, the time to call `AssertEquivalencyOf`.

```cs
private object Create(int i) => new
{
    Value = i,
    Name = $"Item-{i}-{new string((char)('A' + (i % 26)), 50)}",
    Description = $"Description for item number {i} with extra padding {new string('x', 100)}",
    Category = $"Category-{i % 10}",
    Nested = new
    {
        Id = i * 100,
        Label = $"Nested-{i}-{new string('z', 50)}",
        SubNested = new
        {
            Code = i * 1000,
            Data = $"Data-{i}-{new string('y', 80)}"
        }
    },
    Meta = new { K1 = i, K2 = i * 2, K3 = i * 3, K4 = i * 4, K5 = i * 5 }
};
```

```
n=50
  main: 195ms
  pr:   6.6s
  opt:  1.4s
	
n=500
  main  1.9s
  pr:  1.6min
  opt: 39.4s
```


It seems we can optimize the caching some more by creating a targeted `IEqualityComparer`.
Instead of computing hash code based on the _contents_ of `subject` and `expectation` it simply uses their memory references.
This assumes that different object references also have different content.
Additionally the snippet below also pre-allocates `failuresCache` to avoid dynamic resizing.

```cs
public void FindAndRemoveMatches(List<object> subjects, List<TExpectation> expectations)
{
    failuresCache = new(subjects.Count * expectations.Count, new ReferentialComparer());

...

private sealed class ReferentialComparer : IEqualityComparer<(object Subject, object Expectation, int ExpectationIndex)>
{
    public bool Equals((object Subject, object Expectation, int ExpectationIndex) x,
        (object Subject, object Expectation, int ExpectationIndex) y)
    {
        return ReferenceEquals(x.Subject, y.Subject)
            && ReferenceEquals(x.Expectation, y.Expectation)
            && x.ExpectationIndex == y.ExpectationIndex;
    }

    public int GetHashCode((object Subject, object Expectation, int ExpectationIndex) obj)
    {
        int hashCode = RuntimeHelpers.GetHashCode(obj.Subject);
        hashCode = (hashCode * 397) + RuntimeHelpers.GetHashCode(obj.Expectation);
        hashCode = (hashCode * 397) + obj.ExpectationIndex;
        return hashCode;
    }
}
```

Conclusion:
With 500 items and a complex object the runtime is ~40s if we apply the cache adjustment described above.
That's fine by me.
If we get reports with examples that takes way longer, we can then look into using a more efficient algorithm.

dennisdoomen: I had to remove the initial sizing of the `failuresCache` as it caused an out-of-memory exception in some of the other tests. But in my tests, it didn't make any difference.

I use dotTrace to profile your example tests and saw some opportunities to improve the performance using more caching in various places. I'll prepare a separate PR for that.

jnyrup: For later: We could also look into using a better algorithm.

The problem we're trying to solve is called [the unbalanced assignment problem](https://en.wikipedia.org/wiki/Assignment_problem#Unbalanced_assignment).
It's quite related to #1500

jnyrup (approved):

PR #3152, opened 2026-01-23
Clean-up the stack trace when throwing an assertion failure

### LINKED ISSUE #512: Confusing CallStack when using Fluent Assertions
Using MsTest Assert I have this call stack.
   Assert.AreEqual failed. Expected:<4>. Actual:<5>. 
   at Foo.Bar() in c:\Foo\Bar.cs:line 38

Using Fluent Assertions I see this

Expected value to be 4, but found 5.
   at FluentAssertions.Execution.LateBoundTestFramework.Throw(String message) in z:\Builds\work\eed5b735bfe0fb8d\Shared\Execution\LateBoundTestFramework.cs:line 30
   at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args) in z:\Builds\work\eed5b735bfe0fb8d\FluentAssertions.Core\Execution\AssertionScope.cs:line 197
   at FluentAssertions.Numeric.NumericAssertions`1.Be(T expected, String because, Object[] reasonArgs) in z:\Builds\work\eed5b735bfe0fb8d\FluentAssertions.Core\Numeric\NumericAssertions.cs:line 49
   at Foo.Bar() in c:\Foo\Bar.cs:line 36

Is there a way where we don't get the complete call stack that includes FluentAssertions since it is confusing compared to simple MSTest Asserts?

dennisdoomen:
This change marks a large portion of the library’s public and internal surface with the runtime’s `StackTraceHidden` attribute (only when running on .NET 6+), so failures and exceptions show cleaner stack traces that focus on user code instead of framework plumbing.

As a result, assertion failures should be easier to read and debug because internal helper frames are suppressed by the debugger/stack trace renderer.

Alongside that, there are a few formatting-only tweaks (indentation/line breaks) that don’t affect behavior.
The public API baseline for .NET 6 reflects these added attributes so the API “shape” stays in sync.

**Note** The changes were a mix of AI and manual work, hence there's differences how the attribute was applied

Fixes #512

### DIFF
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
@@ -36,7 +36,6 @@ public class GenericCollectionAssertions<TCollection, T>
 
 #pragma warning disable CS0659, S1206 // Ignore not overriding Object.GetHashCode()
 #pragma warning disable CA1065 // Ignore throwing NotSupportedException from Equals
-
 [DebuggerNonUserCode]
 public class GenericCollectionAssertions<TCollection, T, TAssertions> : ReferenceTypeAssertions<TCollection, TAssertions>
     where TCollection : IEnumerable<T>
@@ -1186,28 +1185,28 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> : Referenc
             switch (actualItems.Count)
             {
                 case 0: // Fail, Collection is empty
-                {
-                    assertionChain
-                        .BecauseOf(because, becauseArgs)
-                        .FailWith("Expected {context:collection} to contain a single item{reason}, but the collection is empty.");
+                    {
+                        assertionChain
+                            .BecauseOf(because, becauseArgs)
+                            .FailWith("Expected {context:collection} to contain a single item{reason}, but the collection is empty.");
 
-                    break;
-                }
+                        break;
+                    }
 
                 case 1: // Success Condition
-                {
-                    match = actualItems.Single();
-                    break;
-                }
+                    {
+                        match = actualItems.Single();
+                        break;
+                    }
 
                 default: // Fail, Collection contains more than a single item
-                {
-                    assertionChain
-                        .BecauseOf(because, becauseArgs)
-                        .FailWith("Expected {context:collection} to contain a single item{reason}, but found {0}.", Subject);
+                    {
+                        assertionChain
+                            .BecauseOf(because, becauseArgs)
+                            .FailWith("Expected {context:collection} to contain a single item{reason}, but found {0}.", Subject);
 
-                    break;
-                }
+                        break;
+                    }
             }
         }
 
--- a/Src/FluentAssertions/Common/DateTimeExtensions.cs
+++ b/Src/FluentAssertions/Common/DateTimeExtensions.cs
@@ -18,3 +18,4 @@ public static class DateTimeExtensions
         return new DateTimeOffset(DateTime.SpecifyKind(dateTime, DateTimeKind.Unspecified), offset);
     }
 }
+
--- a/Tests/FluentAssertions.Specs/StackTraceHiddenSpecs.cs
+++ b/Tests/FluentAssertions.Specs/StackTraceHiddenSpecs.cs
@@ -1,0 +1,88 @@
+#if NET6_0_OR_GREATER
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using FluentAssertions.Equivalency;
+using FluentAssertions.Types;
+using Reflectify;
+using Xunit;
+
+namespace FluentAssertions.Specs;
+
+public class StackTraceHiddenSpecs
+{
+    [Fact]
+    public void All_non_assertion_classes_have_their_stack_trace_hidden()
+    {
+        // Arrange / Act
+        var allFluentAssertionsClasses = GetAnnotatableTypes()
+            .Except(ShouldNotBeAnnotated())
+            .Types();
+
+        // Assert
+        allFluentAssertionsClasses.Should().BeDecoratedWith<StackTraceHiddenAttribute>(
+            "because all \"internal\" code should be hidden from the stacktrace to improve the debugging experience");
+    }
+
+    [Fact]
+    public void Assertion_classes_do_not_have_their_stack_trace_hidden()
+    {
+        // Arrange / Act
+        var allFluentAssertionsClasses = ShouldNotBeAnnotated();
+
+        // Assert
+        allFluentAssertionsClasses.Should().NotBeDecoratedWith<StackTraceHiddenAttribute>(
+            "because we do want to see the actual assertion method being used");
+    }
+
+    private static TypeSelector GetAnnotatableTypes()
+    {
+        return AllTypes
+            .From(typeof(FluentAssertions.AssertionExtensions).Assembly)
+            .ThatAreUnderNamespace("FluentAssertions")
+            .ThatAreClasses()
+            .ThatSatisfy(t => !t.IsNested)
+            .ThatSatisfy(t => !t.IsSameOrInherits<Delegate>())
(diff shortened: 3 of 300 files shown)


Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -37,6 +43,9 @@ public GenericCollectionAssertions(TCollection actualValue, AssertionChain asser
 #pragma warning disable CS0659, S1206 // Ignore not overriding Object.GetHashCode()
 #pragma warning disable CA1065 // Ignore throwing NotSupportedException from Equals
 
+#if NET6_0_OR_GREATER
+[StackTraceHidden]
+#endif
```
jnyrup: I think adding `StackTraceHidden` on the outermost `Assertions.cs` classes hides too much of the stack trace.

E.g. if I make `When_one_of_the_types_does_not_match_the_generic_type_it_should_throw_with_a_clear_explanation` fail I get these two stack traces.

I can't see that `AllBeAssignableTo` is the failing call - I'd like to keep that information visible in the stack trace.

main:
```
GenericCollectionAssertions`3.<AllBeAssignableTo>b__4_0[TExpectation](AssertionChain chain) line 91
GenericCollectionAssertions`3.AllBeAssignableTo[TExpectation](String because, Object[] becauseArgs) line 89
AllBeAssignableTo.When_one_of_the_types_does_not_match_the_generic_type_it_should_throw_with_a_clear_explanation() line 130
RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```


PR:
```
AllBeAssignableTo.When_one_of_the_types_does_not_match_the_generic_type_it_should_throw_with_a_clear_explanation() line 130
RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

Src/FluentAssertions/Collections/MaximumMatching/Element.cs
jnyrup, line 1: Let's polyfill `StackTraceHidden` instead of adding hundreds of `#if NET6_0_OR_GREATER`.
PolySharp can polyfill for `StackTraceHidden` by adding `<PolySharpIncludeRuntimeSupportedAttributes>true</PolySharpIncludeRuntimeSupportedAttributes>` to `FluentAssertions.csproj`.

It won't make the stack traces pretty when running on older frameworks, but it'll keep our source code a lot cleaner.

dennisdoomen: Great suggestion 🥳🥳

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
dennisdoomen: Yeah, I agree. Was in doubt myself.

Tests/FluentAssertions.Specs/StackTraceHiddenSpecs.cs
```diff
@@ -0,0 +1,69 @@
+#if NET6_0_OR_GREATER
```
jnyrup, line 1: With the poly filled attribute, we can remove this directive.

Tests/FluentAssertions.Specs/StackTraceHiddenSpecs.cs
```diff
@@ -0,0 +1,69 @@
...
+
+    private string GetNonGenericName(Type type)
+    {
+        string name = type.Name;
+        int backtickIndex = name.IndexOf('`');
+        return backtickIndex > 0 ? name.Substring(0, backtickIndex) : name;
+    }
+
+    [Fact]
+    public void Assertion_classes_do_not_have_their_stack_trace_hidden()
```
jnyrup, line 29: Assertion classes like `EnumAssertions` do not derive from `ReferenceTypeAssertions`.

dennisdoomen: Good catch.

Tests/FluentAssertions.Specs/StackTraceHiddenSpecs.cs
jnyrup, line 1: We should ensure that all annotatable types are included in either the positive or negative test.
One way to do this is to start out with defining all annotatable types `AnnotatableTypes`.
Then we find the ones that _shouldn't_ be annotated `ShouldNotBeAnnotated`.
The remaining types "`AnnotatableTypes` except `ShouldNotBeAnnotated`" are then the types that _should_ be annotated.

With this method there's 10 classes that didn't pass.

```cs
using System;
using System.Diagnostics;
using System.Linq;
using System.Runtime.CompilerServices;
using FluentAssertions.Types;
using Reflectify;
using Xunit;

namespace FluentAssertions.Specs;

public class StackTraceHiddenSpecs
{
    [Fact]
    public void All_non_assertion_classes_have_their_stack_trace_hidden()
    {
        // Arrange / Act
        var allFluentAssertionsClasses = AnnotatableTypes()
            .Except(ShouldNotBeAnnotated())
            .Types();

        // Assert
        allFluentAssertionsClasses.Should().BeDecoratedWith<StackTraceHiddenAttribute>(
            "because all \"internal\" code should be hidden from the stacktrace to improve the debugging experience");
    }

    [Fact]
    public void Assertion_classes_do_not_have_their_stack_trace_hidden()
    {
        // Arrange / Act
        var allFluentAssertionsClasses = ShouldNotBeAnnotated();

        // Assert
        allFluentAssertionsClasses.Should().NotBeDecoratedWith<StackTraceHiddenAttribute>(
            "because we do want to see the actual assertion method being used");
    }

    private static TypeSelector AnnotatableTypes()
    {
        return AllTypes
            .From(typeof(FluentAssertions.AssertionExtensions).Assembly)
            .ThatAreUnderNamespace("FluentAssertions")
            .ThatAreClasses()
            .ThatSatisfy(t => !t.IsNested)
            .ThatSatisfy(t => !t.IsSameOrInherits<Delegate>())
            .ThatSatisfy(t => !t.IsCompilerGenerated() && !t.HasAttribute<CompilerGeneratedAttribute>());
    }

    private static string GetNonGenericName(Type type)
    {
        string name = type.Name;
        int backtickIndex = name.IndexOf('`');
        return backtickIndex > 0 ? name.Substring(0, backtickIndex) : name;
    }

    private static TypeSelector ShouldNotBeAnnotated()
    {
        string[] typeNameSuffixes =
        [
            "Assertions",
            "AssertionsBase",
            "AssertionExtensions",
            "AssertionsExtensions",
        ];

        string[] namespaces =
        [
            "FluentAssertions.CallerIdentification",
            "FluentAssertions.Collections.MaximumMatching",
            "FluentAssertions.Equivalency.Matching",
            "FluentAssertions.Equivalency.Ordering",
            "FluentAssertions.Equivalency.Selection",
            "FluentAssertions.Equivalency.Tracing",
            "FluentAssertions.Formatting",
        ];

        return AnnotatableTypes()
              .ThatSatisfy(t =>
                  t == typeof(EquivalencyOptions) ||
                  typeNameSuffixes.Any(bt => GetNonGenericName(t).EndsWith(bt, StringComparison.Ordinal)) ||
                  namespaces.Any(ns => t.Namespace.StartsWith(ns, StringComparison.Ordinal))
              );
    }
}
```

Src/FluentAssertions/Common/DateTimeExtensions.cs
jnyrup, line 1: There's a few public extension methods that are often invoked _outside_ the assertions, so they shouldn't be annotated.

```
DateTimeExtensions
EventRaisingExtensions
FluentDateTimeExtensions
FluentTimeSpanExtensions
OccurrenceConstraintExtensions
TypeEnumerableExtensions
TypeExtensions
```

jnyrup (changes requested):
Great work!

If we want to take this a step further _later_, we could even annotate internal _methods_ like `GenericCollectionAssertions.BeInOrder`.

Tests/FluentAssertions.Specs/StackTraceHiddenSpecs.cs
dennisdoomen: Yeah, sounds like a good idea. I just wish you came up with that suggestion in your first review 🫣

PR #3154, opened 2026-01-29
Fail when applying member-level Excluding/Including to value-semantic types (fixes #2571)

### LINKED ISSUE #2571: Excluding option in BeEquivalentTo ignored unless ComparingByMembers also specified for types that implement IEquatable<T>
### Description

I am updating the FluentAssertions package on a project from a very old version that would always do member-wise comparisons for Should().BeEquivalentTo(), even if the objects implemented IEquatable<T>. In general, I am ok with the new behavior, but here is a consequence I think is unexpected and undesirable: Explicit member exclusions are being completely ignored.

IMO, FluentAssertions should consider the fact that the implementation of IEquatable<T>.Equals is opaque and hence cannot be assumed to work in any particular way, so it should automatically fall back to perform member-wise comparisons when the assertion is configured with any type of explicit member exclusion.

In practice with the current behavior (which repros on 6.12 and 7.0.0 preview), I am forced to add ComparingByMembers<T>() specifying an explicit T (because it cannot be inferred by the compiler, which adds to the friction) for BeEquivalentTo() call that uses Excluding().

### Reproduction Steps

```csharp
// Paste this in a console app:

using FluentAssertions;

var actual = new SomeEquatableClass { Id = 1, Etag = "x" };
var expected = new SomeEquatableClass { Id = 1, Etag = "y" };

// This assertion fails, even if only Etag property has different values and it's being explicitly excluded:
actual.Should().BeEquivalentTo(expected, opt => opt.Excluding(o => o.Etag));

// This assertion passes, but only because I call ComparingByMembers which I think shouldn't be necessary:
actual.Should().BeEquivalentTo(expected, opt => opt.ComparingByMembers<SomeEquatableClass>().Excluding(o => o.Etag));

// This assertion fails, with custom assertion workaround:
actual.Should().BeEquivalentTo(expected, opt => opt.ExcludingByMembers("Etag"));

public class SomeEquatableClass : IEquatable<SomeEquatableClass>
{
    public int Id { get; set; }
    public string? Etag { get; set; }

    public override bool Equals(object? obj)
    {
        return Equals(obj as SomeEquatableClass);
    }
    public bool Equals(SomeEquatableClass? other)
    {
        if (other == null)
            return false;

        return Id == other.Id && StringComparer.OrdinalIgnoreCase.Equals(Etag, other.Etag);
    }

    {
        return HashCode.Combine(Id, Etag);
    }
}

public static class UnitTestExtensions
{
    public static EquivalencyAssertionOptions<TExpectation> ExcludingByMembers<TExpectation>(this EquivalencyAssertionOptions<TExpectation> options, params string[] exclusions)
    {
        return options.ComparingByMembers<TExpectation>().Excluding(memberInfo => exclusions.Any(e => memberInfo.Path.Equals(e, StringComparison.OrdinalIgnoreCase)));
    }
}

```

### Expected behavior

Explicit exclusions should never be ignored, and this assertion should pass:

```csharp
var actual = new SomeEquatableClass { Id = 1, Etag = "x" };
var expected = new SomeEquatableClass { Id = 1, Etag = "y" };

actual.Should().BeEquivalentTo(expected, opt => opt.Excluding(o => o.Etag));
```

### Actual behavior

Exclusions are ignored so this assertion fails:

```csharp
actual.Should().BeEquivalentTo(expected, opt => opt.Excluding(o => o.Etag));
```

### Regression?

This is a regression but only from very old FluentAssertions versions like 3.5.0.

### Known Workarounds

- Always add ComparingByMembers<T>. This is particularly painful because T has to be specified explictly. 
- Come up with your own exclusion extension that automates this and leverages type inference, like:
```csharp
   public static EquivalencyAssertionOptions<TExpectation> ExcludingByMembers<TExpectation>(this EquivalencyAssertionOptions<TExpectation> options, params string[] exclusions)
    {
        return options.ComparingByMembers<TExpectation>().Excluding(memberInfo => exclusions.Any(e => memberInfo.Path.Equals(e, StringComparison.OrdinalIgnoreCase)));
    }
}
```  

### Configuration

Repros on FA 6.12 and 7.0.0 preview. 
Used to work fine with FA 3.5.0.
Tested on both .NET Framework 4.x and .NET 8.

### Other information

Apart from this, the way FluentAssertions now leverages IEquatable<T>.Equals forced us to debug a bunch of incorrect equality evaluation code in our domain classes. While this is ultimately a good thing and I am thankful to FA team for this :smile:, it feels like a different concern from what the tests were initially designed to cover. I know that the behavior can be customized, so this is just a comment on the choice of default behavior. 

Note: I am a newbie using FluentAssertions so I will be the first to admit that my mental model might be lacking. I apologize if I am being naive or if I missed any discussion in which it was decided that this was by design (I did a search on the repo, but I did not find anything that seemed relevant).

The answer to the question on if I would be able to contribute a fix, is "maybe" :smile: I am not sure how long it can take me to understand the FluentAssertions codebase well enough.

No

WillRayAtPropertyMe:
**Summary**

This change enforces a clear failure when callers attempt to apply member/path-level selection rules (e.g., `Excluding(...)`, `Including(...)` by path) to types that are compared using value semantics (i.e., `Equals` override or explicit `ComparingByValue`/`ForceEquals`). Previously such rules could be silently ignored or produce inconsistent behavior across overloads.

**What I changed**

- Equivalency logic
  - `Src/FluentAssertions/Equivalency/EquivalencyValidator.cs`
    - Added checks so that when a path-based selection rule (currently `ExcludeMemberByPathSelectionRule`) targets a node whose type is compared by value semantics, the assertion fails with an actionable message explaining how to resolve the issue (for example, `ComparingByMembers<T>()`). The check is conservative: it ignores primitive/value types and `string`, and it only fails for non-primitive reference types which either override `Equals` or are explicitly forced to use `Equals`.

- Tests
  - `Tests/FluentAssertions.Equivalency.Specs/SelectionRulesSpecs.Excluding.cs`
    - Added a spec that asserts `Excluding(...)` on a type that overrides `Equals` results in a helpful failure instead of being silently ignored.
  - `Tests/FluentAssertions.Equivalency.Specs/TestTypes.cs`
    - Added a small `ValueSemanticWithProperty` helper type used in the spec.

**Rationale**

- The behavior described in GitHub issue #2571 is confusing: users expect that when they specify member-level exclusions or inclusions, those rules are honored or reported; silently ignoring them is a poor developer experience. The maintainers' preferred approach is to fail with a clear message (rather than implicitly switching equality strategy).

**Notes & limitations**

- This initial implementation enforces the rule for path-based selection rules (exclude-by-path). It does not yet enforce for predicate-based selection rules (`Excluding(ctx => ...)`) because those can apply at arbitrary depths and require more nuanced runtime analysis to decide whether they definitely target a specific type/member. I can extend enforcement to include predicate-based rules in a follow-up change.

- The check ignores primitive types and `string` because excluding members that are primitive or strings is harmless and expected.

- I ran the full test suite locally. All test cases executed, and 18610 tests passed; a single build error was emitted by one legacy test target due to an outdated VSTest MSBuild target on .NET 10 (this is unrelated to the change).

**How to review**

- Focus on `EquivalencyValidator.cs` for enforcement logic and the small test added to the existing specs.
- If acceptable, I can extend the enforcement to include `IncludeMemberByPathSelectionRule` and consider predicate-based rules.

### DIFF
--- a/Src/FluentAssertions/Equivalency/EquivalencyValidator.cs
+++ b/Src/FluentAssertions/Equivalency/EquivalencyValidator.cs
@@ -1,4 +1,5 @@
 using System;
+using FluentAssertions.Common;
 using FluentAssertions.Equivalency.Tracing;
 using FluentAssertions.Execution;
 
@@ -16,7 +17,7 @@ internal class EquivalencyValidator : IValidateChildNodeEquivalency
         context.ResetTracing();
 
         using var scope = new AssertionScope();
-
+        
         RecursivelyAssertEquivalencyOf(comparands, context);
 
         if (context.TraceWriter is not null)
@@ -38,6 +39,50 @@ internal class EquivalencyValidator : IValidateChildNodeEquivalency
 
         if (ShouldContinueThisDeep(context.CurrentNode, context.Options, assertionChain))
         {
+            // Fail when a path-based ExcludeMemberByPathSelectionRule targets this node but its type is compared using value semantics.
+            foreach (var rule in context.Options.SelectionRules)
+            {
+                if (rule is FluentAssertions.Equivalency.Selection.ExcludeMemberByPathSelectionRule excludeRule)
+                {
+                    try
+                    {
+                        try
+                        {
+                            var nodePath = new MemberPath(context.CurrentNode.Subject.PathAndName);
+
+                            if (nodePath.IsParentOrChildOf(excludeRule.CurrentPath) || nodePath.IsSameAs(excludeRule.CurrentPath))
+                            {
+                                Type typeToCheck = context.CurrentNode.Type;
+
+                                // Ignore primitive/value types and strings — excluding these is harmless and expected.
+                                if (typeToCheck.IsValueType || typeToCheck == typeof(string))
+                                {
+                                    continue;
+                                }
+
+                                EqualityStrategy strategy = context.Options.GetEqualityStrategy(typeToCheck);
+
+                                if ((strategy is EqualityStrategy.Equals or EqualityStrategy.ForceEquals)
+                                    && (typeToCheck.OverridesEquals() || strategy == EqualityStrategy.ForceEquals))
+                                {
+                                    assertionChain.FailWith(
+                                        "Cannot apply member-level inclusion/exclusion to path {0} because the type {1} is compared using value semantics (Equals). Configure it to be compared by members instead (for example, using ComparingByMembers<{2}>()) before applying member-level rules.",
+                                        excludeRule.CurrentPath, typeToCheck, typeToCheck);
+                                }
+                            }
+                        }
+                        catch
+                        {
+                            // Swallow any unexpected errors during inspection to avoid changing behavior in unknown cases.
+                        }
+                    }
+                    catch
+                    {
+                        // Swallow any unexpected errors during inspection to avoid changing behavior in unknown cases.
+                    }
+                }
+            }
+
             if (!context.IsCyclicReference(comparands.Expectation))
             {
                 TryToProveNodesAreEquivalent(comparands, context);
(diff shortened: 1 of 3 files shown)


dennisdoomen:
Why did you close it?

PR #3198, opened 2026-04-22
BeEmpty for IEnumerable<T> assertions now lists the first 10 items in…

### LINKED ISSUE #2782: `BeEmpty()` no longer lists all items in the collection in the error message
### Description

First off: thanks for the awesome library! We've been using it for a few years now and it's been a huge boon to our tests :)

This is kind of a https://xkcd.com/1172/ situation, but I thought it'd be at least worth writing down to see what other people think.

Previously, using `foo.Should().BeEmpty()` on a non-empty collection would produce a result like "expected foo to be empty, but was {1, 2, 3}". Now it produces an error like "expected foo to be empty, but found at least one item {1}". 

This is annoying in the case where the test needs to be re-run multiple times to discover more unexpected items in the collection, especially if the test is slow to run or can only be run on the CI server. An example in our case is checking for dlls that are part of the installed product which aren't authenticode-signed.

### Reproduction Steps

```csharp
[Test]
public async Task Check_things_are_valid()
{
    var things = new List<string> { "fine", "PROBLEM", "other_fine", "OTHER_PROBLEM" };

    var problems = new List<string>();
    foreach (var thing in things)
    {
        if (!await SomeLongWindedProcessToCheckIfThingIsValid(thing))
        {
            problems.Add(thing);
        }
    }
    problems.Should().BeEmpty();
}

private Task<bool> SomeLongWindedProcessToCheckIfThingIsValid(string thing) =>
    Task.FromResult(thing == thing.ToLowerInvariant());
```

### Expected behavior

The error message from the test should list all of the problems so they can be fixed in one go.

As a potential compromise in the case of infinite collections, it could maybe list up to some max number of items?

### Actual behavior

The error message from the test only lists one problem at a time, meaning we have to fix one at a time, push and wait for a CI build to see if there's another problem or not

### Regression?

This appears to have been an intentional change in https://github.com/fluentassertions/fluentassertions/pull/2530

### Known Workarounds

`problems.Should().Equal([]);` prints out all of the problems (but looks wrong: someone will come along and say "why don't you just use BeEmpty?")

### Configuration

using FluentAssertions 6.12.1 and .NET 8.0.403

### Other information

_No response_

Maybe? I don't really know the code that well but can try if we agree on a change

Xceed-DelvaJB:
… the collection instead of only the first one.

Resolves #2782.

## IMPORTANT 

## CONTRIBUTOR LICENSE GRANT

* [x] I have read the Contributor License Grant and accept the conditions
* [ ] I'm interested in a free license for Fluent Assertions and will share my email address through sales@xceed.com

### DIFF
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -285,17 +285,17 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> : Referenc
     /// </param>
     public AndConstraint<TAssertions> BeEmpty([StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
     {
-        var singleItemArray = Subject?.Take(1).ToArray();
+        var tenItemArray = Subject?.Take(10).ToArray();
 
         assertionChain
             .BecauseOf(because, becauseArgs)
             .WithExpectation("Expected {context:collection} to be empty{reason}, ", chain => chain
-                .Given(() => singleItemArray)
+                .Given(() => tenItemArray)
                 .ForCondition(subject => subject is not null)
                 .FailWith("but found <null>.")
                 .Then
                 .ForCondition(subject => subject.Length == 0)
-                .FailWith("but found at least one item {0}.", singleItemArray));
+                .FailWith("but found at least {0}.", tenItemArray));
 
         return new AndConstraint<TAssertions>((TAssertions)this);
     }
--- a/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
+++ b/Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
@@ -42,6 +42,7 @@ public class GenericCollectionAssertions<TCollection, T, TAssertions> : Referenc
     where TCollection : IEnumerable<T>
     where TAssertions : GenericCollectionAssertions<TCollection, T, TAssertions>
 {
+    private const int MaxItemsShownOnBeEmptyFailure = 10;
     private readonly AssertionChain assertionChain;
 
     public GenericCollectionAssertions(TCollection actualValue, AssertionChain assertionChain)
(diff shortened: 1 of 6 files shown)


dennisdoomen:
@Xceed-DelvaJB did you see the review comments and failing builds?

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
```diff
@@ -285,17 +285,17 @@ public AndConstraint<TAssertions> AllBeOfType(Type expectedType, [StringSyntax("
     /// </param>
     public AndConstraint<TAssertions> BeEmpty([StringSyntax("CompositeFormat")] string because = "", params object[] becauseArgs)
     {
-        var singleItemArray = Subject?.Take(1).ToArray();
+        var tenItemArray = Subject?.Take(10).ToArray();
```
dennisdoomen: 🔧 The magic number `10` is hardcoded inline. Consider extracting it to a named constant, e.g. `private const int MaxItemsShownOnBeEmptyFailure = 10;`. That makes the intent self-documenting and ensures the variable name, the `Take()` call, and any future tests all stay in sync from a single place.

dennisdoomen: ⛏ The name `tenItemArray` embeds the current limit in the identifier. If the limit is ever adjusted, the name becomes misleading. A more durable name would be `firstItems` or `itemSample`.

docs/_pages/releases.md
```diff
@@ -7,6 +7,11 @@ sidebar:
   nav: "sidebar"
 ---
 
+## Unreleased
+
+### Enhancements
+* `BeEmpty` for `IEnumerable<T>` assertions now lists the first 10 items in the collection instead of only the first one - [#2782](https://github.com/fluentassertions/fluentassertions/pull/2782)
```
dennisdoomen: 🔧 The link target points at `pull/2782`, but #2782 is an **issue**, not a pull request — the URL as written will 404. Please change it to: `[#2782](https://github.com/fluentassertions/fluentassertions/issues/2782)`

Src/FluentAssertions/Collections/GenericCollectionAssertions.cs
Xceed-DelvaJB: Of course! No problem.

docs/_pages/releases.md
dennisdoomen: Sorry, it needs to be the number of the PR

dennisdoomen (approved):

jnyrup (approved):

PR #3203, opened 2026-05-03
Include original index in extraneous item failure messages

### LINKED ISSUE #985: collection Should().BeEquivalentTo() vs position of first difference
### Description

I use the collection `.Should().BeEquivalent()` to ensure that two collections are equivalent. When my two collections do not have the same size, it is very hard to know at which index the first difference was found.

Here's my test code:

```csharp
class MyPair
{
    public int x { get; set; }
    public int y { get; set; }
    public override string ToString() => $"{x}-{y}";
}

[Fact]
public void TestBeEquivalent()
{
    var expected = Enumerable.Range(1, 100).Select(i => new MyPair { x = i, y = i + 20 }).ToList();

    // Make 'actual' almost identical to 'expected'
    var actual = new List<MyPair>(expected);
    actual.Insert(95, new MyPair { x = 0, y = 0 });

}
```

### Expected behavior:

I would expect the test to fail with a message similar to this that pin points where the two collections start to be non-equivalent:

```
The collections are identical until position 95 (0 based) where "0-0" is not equivalent to "96-116". 
```

### Actual behavior:

The test fails with this message:

```
Expected actual to be a collection with 100 item(s), but {1-21, 2-22, 3-23, 4-24, 5-25, 6-26, 7-27, 8-28, 9-29, 10-30, 11-31, 12-32, 13-33, 14-34, 15-35, 16-36, 17-37, 18-38, 19-39, 20-40, 21-41, 22-42, 23-43, 24-44, 25-45, 26-46, 27-47, 28-48, 29-49, 30-50, 31-51, 32-52, …69 more…}"
"contains 1 item(s) more than"
"{1-21, 2-22, 3-23, 4-24, 5-25, 6-26, 7-27, 8-28, 9-29, 10-30, 11-31, 12-32, 13-33, 14-34, 15-35, 16-36, 17-37, 18-38, 19-39, 20-40, 21-41, 22-42, 23-43, 24-44, 25-45, 26-46, 27-47, 28-48, 29-49, 30-50, 31-51, 32-52, …68 more…}.
```

As you can see, it is very hard to know "why" the collections are not equivalent. Just looking at the output, it is impossible to know the error. I have to run the debugger, watch variables, copy paste the two collections in a diff tool and find the differences from there. It would be very hepfull if FluentAssertions could help identify where the collections start to diverge.

### Versions

* FluentAssertions: 5.5.3
* .NET Core 2.1

dennisdoomen:
## Summary

Fixes #985.

When `BeEquivalentTo` fails because the subject collection has more items than expected, the error message now includes the **original index** of each extraneous item, making it easier to diagnose where collections diverge.

### Before

```
Expected actual to contain exactly one item, but found one extraneous item Customer { Age = 16, ... }
```

### After

**Single extra item:**
```
Expected actual to contain exactly one item, but found one extraneous item at index 1: Customer { Age = 16, ... }
```

**Multiple extra items:**
```
Expected actual to contain exactly 2 items, but found extraneous items: Customer { Age = 16, ... } (at index 2), Customer { Age = 21, ... } (at index 4)
```

## Implementation

- `EnumerableEquivalencyValidator`: subjects are now wrapped in `List<IndexedItem<object>>` so original indices survive through the matching pipeline to the reporting stage.
- `StrictlyOrderedEquivalencyStrategy` / `LooselyOrderedEquivalencyStrategy`: updated to pass `List<IndexedItem<object>>` instead of `List<object>` for subjects; raw objects are accessed via `.Item`.
- When only extra items exist, each item is reported with `at index N:`. When both missing and extra items exist, the compact fallback format is used to avoid expensive per-item formatting.
- Braces in formatted item strings (e.g. `Customer { Age = 16 }`) are escaped before embedding into the `FailWith` format string to prevent `string.Format` treating them as placeholders.

### DIFF
--- a/Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
--- a/Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
@@ -152,7 +152,7 @@ internal class EnumerableEquivalencyValidator(
 
             object subjectsArg = remainingSubjects.Count == 1
                 ? remainingSubjects.Single().Item
-                : remainingSubjects.Select(s => s.Item).ToList<object>();
+                : remainingSubjects.Select(s => s.Item).ToList();
 
             assertionChain.FailWith(message.ToString(), remainingExpectations, subjectsArg, allSubjects);
         }
--- a/Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
@@ -151,7 +151,7 @@ internal class EnumerableEquivalencyValidator(
             }
 
             object subjectsArg = remainingSubjects.Count == 1
-                ? remainingSubjects.Single().Item
+                ? remainingSubjects[0].Item
                 : remainingSubjects.Select(s => s.Item).ToList();
 
             assertionChain.FailWith(message.ToString(), remainingExpectations, subjectsArg, allSubjects);
--- a/Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
+++ b/Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
@@ -4,6 +4,7 @@ using System.Text;
 using FluentAssertions.Equivalency.Execution;
 using FluentAssertions.Equivalency.Tracing;
 using FluentAssertions.Execution;
+using FluentAssertions.Formatting;
 using static FluentAssertions.Common.StringExtensions;
 
 namespace FluentAssertions.Equivalency.Steps;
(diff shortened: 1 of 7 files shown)


Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
```diff
@@ -111,8 +116,31 @@ private void ReportRemainingOrMissingItems<T>(List<object> remainingSubjects, Li
...
+                    // Show each extraneous item with its original index to help pinpoint the divergence.
+                    if (remainingSubjects.Count == 1)
+                    {
+                        message.Append($"found one extraneous item at index {remainingSubjects[0].Index}: {{1}}");
+                    }
+                    else
+                    {
+                        string formattedItems = string.Join(", ",
+                            remainingSubjects.Select(s =>
+                                $"{Formatter.ToString(s.Item).Replace("{", "{{", StringComparison.Ordinal).Replace("}", "}}", StringComparison.Ordinal)} (at index {s.Index})"));
```
jnyrup: We have `EscapePlaceholders` for replacing curly braces.

If I remove the string replacement no tests fail, so it seems it's either unneeded or we're missing a test? 🤔

Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
```diff
@@ -122,8 +150,11 @@ private void ReportRemainingOrMissingItems<T>(List<object> remainingSubjects, Li
                 message.AppendLine("Full dump of {context:subject}: {2}");
             }
 
-            assertionChain.FailWith(message.ToString(), remainingExpectations,
-                remainingSubjects.Count == 1 ? remainingSubjects.Single() : remainingSubjects, allSubjects);
+            object subjectsArg = remainingSubjects.Count == 1
+                ? remainingSubjects.Single().Item
```
jnyrup:
```suggestion
                ? remainingSubjects[0].Item
```

jnyrup (changes requested):

Tests/FluentAssertions.Equivalency.Specs/CollectionSpecs.cs
```diff
@@ -2222,7 +2222,60 @@ public void When_the_subject_contains_more_items_than_expected_it_should_throw()
...
+        };
+
+        // Act
+        Action action =
+            () => subject.Should().BeEquivalentTo(expectation);
+
+        // Assert
+        action.Should().Throw<XunitException>()
+            .WithMessage("*extraneous items*Age = 24*at index 1*Age = 32*at index 2*")
+            .And.Message.Should().NotContain("**WARNING**");
```
jnyrup:
```suggestion
            .WithoutMessage("**WARNING**");
```

Src/FluentAssertions/Equivalency/Steps/EnumerableEquivalencyValidator.cs
jnyrup:
```suggestion
                                $"{Formatter.ToString(s.Item).EscapePlaceholders()} (at index {s.Index})"));
```

jnyrup (approved):
Nice!

PR #3217, opened 2026-05-19
Prevent pwn-request RCE via the Qodana workflow

dennisdoomen:
### The problem

`.github/workflows/code_quality.yml` used `pull_request_target` (which runs on the **base repo** with access to repository secrets) together with `actions/checkout` using `ref: github.event.pull_request.head.sha`, which checks out **untrusted contributor code** in a privileged context. Because `QODANA_TOKEN` was injected via `env:`, any PR from an external contributor could exfiltrate or abuse it, and any other secret available to the job — a classic [pwn request](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) RCE pattern.

### The fix: split trusted vs. untrusted triggers, and scope the PR token

The single vulnerable workflow is replaced with two workflows plus a dedicated GitHub Environment:

| File | Trigger | Secrets available | Runs contributor code? |
|---|---|---|---|
| `code_quality.yml` | `push` to `main`/`develop`, `workflow_dispatch` | Repo-wide `QODANA_TOKEN` | No (trusted) |
| `code_quality_pr.yml` | `pull_request` | Only `QODANA_TOKEN`, scoped to the `qodana-pr` GitHub Environment | Yes (untrusted, but blast radius contained) |

**Workflow 1 (`code_quality.yml`)** — unchanged behaviour for branch pushes and manual triggers. Checks out trusted code and uploads results to Qodana Cloud with the repo-wide `QODANA_TOKEN`.

**Workflow 2 (`code_quality_pr.yml`)** — triggered by `pull_request` (not `pull_request_target`), so it runs with the contributor's untrusted code. Since the Qodana cloud linter requires a token to run at all, this workflow still needs `QODANA_TOKEN` — but that secret now lives in a dedicated `qodana-pr` GitHub Environment rather than as a repo/org-wide secret. This means:
- Untrusted PR code can only ever reach `QODANA_TOKEN`, never any other repo or org secret, even if this workflow is edited later to add more steps.
- The `qodana-pr` environment can have **required reviewers** configured, so a maintainer must manually approve the run for first-time or fork contributors before their code executes with the token.
- `GITHUB_TOKEN` permissions for this job are explicitly locked down to `contents: read` and `pull-requests: write` (needed only for Qodana to post its PR comment), so the auto-token can't be used to push commits, modify workflows, or touch repo settings.

### Why two separate workflow files instead of one combined file

It's tempting to merge these into a single workflow with `on: [push, pull_request, workflow_dispatch]` and a conditional `environment:`/`if:` per job. We deliberately keep them split, for future maintainers:

- **Auditability of the trust boundary.** With two files, the security-relevant question — "does this code run with elevated trust or not?" — is answered just by reading the `on:` trigger of the file. A combined file requires carefully reading conditional expressions every time to reach the same conclusion, which is much easier to get wrong during review.
- **Blast radius of future edits.** In a merged workflow, a later change (e.g. adding a step, tweaking an `if:`) can accidentally let a step that needs broader secrets run in the PR-triggered branch. Separate files make that class of mistake structurally harder to introduce.
- **This matches GitHub's own guidance.** The [GitHub Security Lab advisory](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) this fix is based on explicitly recommends separating trusted-trigger workflows from untrusted-trigger workflows rather than branching inside one file.
- **The permission/environment shape genuinely differs.** `code_quality.yml` needs no `environment:`/scoped token and broad `contents: read`; `code_quality_pr.yml` needs the `qodana-pr` environment plus `pull-requests: write`. Combining them doesn't save much complexity, it just costs readability of the trust boundary.

**Please keep this split when touching Qodana CI in the future** — resist the urge to consolidate them back into one file for "DRY" reasons; the duplication here is intentional and load-bearing for security.

### References

- [GitHub Security Lab: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)
- [Qodana GitHub Actions integration](https://www.jetbrains.com/help/qodana/github.html)

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -1,0 +1,24 @@
+name: Qodana
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - develop
+
+jobs:
+  qodana-scan:
+    name: Qodana Scan
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - name: 'Qodana Scan'
+        uses: JetBrains/qodana-action@v2026.1
+        with:
+          upload-result: true
+          args: --baseline qodana.sarif.json --linter qodana-dotnet --within-docker false
+        env:
+          QODANA_TOKEN: ${{ secrets.QODANA_TOKEN }}
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -2,7 +2,6 @@ name: Qodana
 
 on:
   workflow_dispatch:
-  pull_request_target:
   push:
     branches:
       - main
@@ -15,13 +14,11 @@ jobs:
     steps:
       - uses: actions/checkout@v7
         with:
-          ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
       - name: 'Qodana Scan'
         uses: JetBrains/qodana-action@v2026.1
         with:
-          upload-result: ${{ github.ref_name == 'main' || github.ref_name == 'develop' }}
+          upload-result: true
           args: --baseline qodana.sarif.json --linter qodana-dotnet --within-docker false
-          pr-mode: ${{ github.event_name == 'pull_request_target' }}
         env:
           QODANA_TOKEN: ${{ secrets.QODANA_TOKEN }}
--- a/.github/workflows/code_quality_pr.yml
+++ b/.github/workflows/code_quality_pr.yml
@@ -13,6 +13,7 @@ on:
   pull_request:
 
 permissions:
+  checks: write
   contents: read
   pull-requests: write
 
--- a/.github/workflows/code_quality_pr.yml
+++ b/.github/workflows/code_quality_pr.yml
@@ -1,0 +1,37 @@
+name: Qodana (PR)
+
+# Triggered by pull_request (NOT pull_request_target), so this job runs with
+# the PR contributor's untrusted code. It still needs QODANA_TOKEN to run the
+# cloud linter, so that secret is scoped to the "qodana-pr" GitHub Environment
+# instead of being a repo/org-wide secret. That keeps this job's blast radius
+# limited to QODANA_TOKEN only: no other repository or organization secret is
+# exposed to code checked out from a pull request, no matter how the job
+# steps are edited later. Add required reviewers on the "qodana-pr"
+# environment to also gate first-time/fork contributors behind manual
+# approval before their code runs with the token.
+on:
+  pull_request:
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  qodana-scan:
+    name: Qodana Scan
+    runs-on: windows-latest
+    environment: qodana-pr
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+      - name: 'Qodana Scan'
+        uses: JetBrains/qodana-action@v2026.1
+        with:
+          upload-result: true
+          args: --baseline qodana.sarif.json --linter qodana-dotnet --within-docker false
+          pr-mode: true
+        env:
+          QODANA_TOKEN: ${{ secrets.QODANA_TOKEN }}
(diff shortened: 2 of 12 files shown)


.github/workflows/code_quality.yml
```diff
@@ -13,15 +12,13 @@ jobs:
     name: Qodana Scan
     runs-on: windows-latest
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@v6
```
jnyrup: Can we keep using v7?

.github/workflows/code_quality_pr.yml
```diff
@@ -0,0 +1,38 @@
...
+# the safe half of the two-workflow "pwn-request" mitigation pattern.
+on:
+  pull_request:
+
+jobs:
+  qodana-scan:
+    name: Qodana Scan
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v6
```
jnyrup: Can we use v7?

.github/workflows/code_quality_pr.yml
```diff
@@ -0,0 +1,38 @@
...
+          results-dir: ${{ runner.temp }}/qodana-results
+
+      # Persist the PR number alongside the results so the upload workflow can
+      # associate the report with the correct pull request.
+      - name: Save PR number
+        shell: bash
+        run: echo "${{ github.event.pull_request.number }}" > "${{ runner.temp }}/qodana-results/pr-number.txt"
+
+      - name: Upload Qodana results artifact
+        uses: actions/upload-artifact@v4
```
jnyrup: Can we use v7?

.github/workflows/code_quality_upload.yml
```diff
@@ -0,0 +1,36 @@
...
+      - completed
+
+jobs:
+  upload:
+    name: Upload to Qodana Cloud
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion == 'success'
+    steps:
+      - name: Download Qodana results artifact
+        uses: actions/download-artifact@v4
```
jnyrup: Can we use v8?

jnyrup (changes requested):
The new "Qodana (PR)" pipeline fails to run.
I guess it still requires the Qodana token to run?

.github/workflows/code_quality.yml
dennisdoomen: Updated this to `actions/checkout@v7` as suggested.


_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

.github/workflows/code_quality_pr.yml
dennisdoomen: Updated this to `actions/checkout@v7` as suggested.


_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

.github/workflows/code_quality_pr.yml
dennisdoomen: Updated this to `actions/upload-artifact@v7` as suggested.


_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

.github/workflows/code_quality_upload.yml
dennisdoomen: Updated this to `actions/download-artifact@v8` as suggested.


_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

dennisdoomen:
> The new "Qodana (PR)" pipeline fails to run.
> I guess it still requires the Qodana token to run?

I had to revisit the solution. The PR description now properly explains the approach.

.github/workflows/code_quality_pr.yml
```diff
@@ -0,0 +1,37 @@
...
+# instead of being a repo/org-wide secret. That keeps this job's blast radius
+# limited to QODANA_TOKEN only: no other repository or organization secret is
+# exposed to code checked out from a pull request, no matter how the job
+# steps are edited later. Add required reviewers on the "qodana-pr"
+# environment to also gate first-time/fork contributors behind manual
+# approval before their code runs with the token.
+on:
+  pull_request:
+
+permissions:
```
jnyrup, line 15: The pipeline would like more permissions. Should we grant those?

> Not able to publish annotations with Checks API – Resource not accessible by integration - https://docs.github.com/rest/checks/runs#update-a-check-run, 
    using limited (10 problems per level) output instead. Check job permissions (checks: write, pull-requests: write needed)

jnyrup (approved):

.github/workflows/code_quality_pr.yml
dennisdoomen: @copilot add the write permissions for checks

PR #3172, opened 2026-03-07
Add support for Span<T>, ReadOnlySpan<T>, Memory<T> and ReadOnlyMemory<T>

dennisdoomen:
This pull request adds comprehensive support for assertion extensions for `Span<T>`, `ReadOnlySpan<T>`, `Memory<T>`, and `ReadOnlyMemory<T>` types in FluentAssertions for .NET 6.0 and .NET Standard 2.1 or greater. It introduces new assertion classes and methods, updates the public API, and ensures proper annotation and validation in tests. These changes enhance the ability to write fluent assertions for modern .NET types, improving test readability and coverage.

### Assertion Extensions for Span and Memory Types

* Added new `Should` extension methods for `Span<T>`, `ReadOnlySpan<T>`, `Memory<T>`, and `ReadOnlyMemory<T>`, returning appropriate assertion types (`SpanAssertions<T>`, `GenericCollectionAssertions<T>`, `StringCollectionAssertions`). These methods are conditionally compiled for .NET 6.0 and .NET Standard 2.1 or greater. (`Src/FluentAssertions/AssertionExtensions.cs`)
* Introduced the `SpanAssertions<T>` class, enabling assertions on `Span<T>` and `ReadOnlySpan<T>`, including a `BeEqualTo` method for equality checks. (`Src/FluentAssertions/Collections/SpanAssertions.cs`)
* Added string-specific assertion extension for `SpanAssertions<char>`, allowing direct string equality assertions. (`Src/FluentAssertions/Collections/SpanAssertionsExtensions.cs`)

Closes #902

PR #3225, opened 2026-05-27
Fix CS0453 regression for reference types caused by OverloadResolutionPriority on Enum overloads

### LINKED ISSUE #3224: Possible regression in 8.9.0: OverloadResolutionPriority causes CS0453 on reference types
### Description

I’ve noticed a change in overload resolution behavior starting with version 8.9.0 that affects various reference types when using the .NET 9+ SDK. The compiler appears to incorrectly prefer specialized overloads (like Enum) over the general object-based fallback due to prioritization changes.

### Reproduction Steps

```csharp
using FluentAssertions;
using Xunit;

public class MyClass { }

public class Repro
{
    [Fact]
    public void Test_Object()
    {
        object obj = new object();
        obj.Should().NotBeNull(); // Resolves correctly
    }

    [Fact]
    public void Test_CustomClass()
    {
        MyClass item = new MyClass();
        item.Should().NotBeNull(); // Fails to compile in 8.9.0
    }
}
```

### Expected behavior

Both tests should compile successfully, resolving to `AssertionExtensions.Should(object)`.

### Actual behavior

The second test fails with CS0453:
"The type 'MyClass' must be a non-nullable value type in order to use it as parameter 'TEnum' in the generic type or method 'EnumAssertionsExtensions.Should<TEnum>(TEnum)'"

### Regression?

Yes. This worked correctly in version 8.8.0.

### Known Workarounds

- Downgrading to version 8.8.0.
- Explicitly casting the subject to object: `((object)item).Should().NotBeNull()`.

### Configuration

- FluentAssertions Version: 8.9.0, 8.10.0
- Target Framework: .NET 9 / .NET 10
- SDK: .NET 9.0.100+ (C# 13)

### Other information

This appears to be related to the introduction of `[OverloadResolutionPriority]` in PR #3172. 

In `AssertionExtensions.cs`, the catch-all `Should(object)` overload is now decorated with `[OverloadResolutionPriority(-2)]`. However, specialized extensions like `EnumAssertionsExtensions.Should<TEnum>(TEnum)` retain the default priority of 0.

When calling `.Should()` on a reference type other than `object` itself, the C# 13 compiler prefers the higher priority (0) Enum overload despite its `where TEnum : struct, Enum` constraint. 

I’d appreciate any guidance on whether this prioritization gap is intentional or if specialized extensions should eventually have their priorities synchronized with the base object overload.

No

dennisdoomen:
## Problem

Fixes #3224.

Since 8.9.0, calling `.Should()` on any custom reference type that doesn't have a dedicated overload causes a **CS0453** compile error with the .NET 9 SDK (Roslyn 4.9):

```
The type 'MyClass' must be a non-nullable value type in order to use it as parameter 'TEnum'
in the generic type or method 'EnumAssertionsExtensions.Should(TEnum)'
```

## Root Cause

PR #3172 introduced `[OverloadResolutionPriority(-2)]` on `AssertionExtensions.Should(this object)` to make it a last-resort fallback. However, `EnumAssertionsExtensions.Should<TEnum>` was left at the default priority (0).

C# 13's `OverloadResolutionPriority` performs priority filtering *before* fully validating generic constraints. When resolving `myClass.Should()`:

1. Type inference for `Should<TEnum>` succeeds (TEnum = MyClass)
2. Priority filtering removes the lower-priority (`-2`) object overload from the candidate set
3. Constraint checking then fails (`MyClass` is not a `struct`) -> **CS0453**

## Fix

Add `[OverloadResolutionPriority(-2)]` to both `Should<TEnum>` overloads in `EnumAssertionsExtensions`. This places them at the same priority level as `Should(this object)`, so no priority filtering occurs between them. Normal overload resolution then correctly selects:

- The Enum overload for actual enum types (more specific than `object`) ✅
- The object overload for reference types (struct constraint fails -> only object applies) ✅

**Priority levels after fix:**

| Priority | Overloads |
|----------|-----------|
| 0 (default) | All specific typed overloads (`string`, `bool`, `Stream`, etc.) |
| -1 | `Span<T>`, `ReadOnlySpan<T>`, `Memory<T>`, `ReadOnlyMemory<T>` |
| -2 | Enum overloads + `Should(this object)` catch-all |


jnyrup:
Were you able to reproduce the reported issue? I'm a bit curious how we didn't caught this 🤔

L34T:
@jnyrup
Reproduction repository: https://github.com/L34T/fluentassertions-3224/
Build instructions: run `./scripts/build-matrix.sh`
Build log: https://github.com/L34T/fluentassertions-3224/blob/master/build.log

dennisdoomen:
> **GitHub Copilot summary** (posted on behalf of @dennisdoomen):
>
> I was wondering the same thing, so I dug into it.
>
> **Why the existing tests did not catch this:**
>
> The test projects target `net47;net6.0;net8.0`, which default to C# 7.3, 10, and 12 respectively. The `[OverloadResolutionPriority]` attribute only affects overload resolution in **C# 13+**, so the attribute is completely ignored by the existing test compiler and the CS0453 error never surfaces.
>
> **What actually causes the regression:**
>
> `AssertionExtensions.Should(object)` already carries `[OverloadResolutionPriority(-2)]`. The enum `Should<TEnum>()` overloads were introduced without the attribute, so they defaulted to priority 0. In **Roslyn 4.9 (.NET 9 SDK / C# 13)**, when a higher-priority overload is in the candidate set for a reference-type argument, the compiler tries to bind it first and fails the `where TEnum : struct, Enum` constraint check — emitting CS0453 instead of falling back to the lower-priority `Should(object)` overload.
>
> **Net 10 SDK note:**
>
> Roslyn 4.14 (.NET 10 SDK) changed the constraint-fallback behavior, so the bug does **not** manifest under the .NET 10 toolchain even without the fix. The fix is still correct and necessary for users on .NET 9 SDK. Because the CS0453 is Roslyn 4.9-specific and any SDK capable of building a net10.0 project uses a newer Roslyn, adding an automated regression test that catches the original error in CI is not straightforward. The external reproduction at https://github.com/L34T/fluentassertions-3224/ demonstrates the issue on .NET 9 SDK.

jnyrup:
I tried reproducing this locally and I managed to reproduce it when using SDK 9.0.117, while SDK 9.0.200 can compile the project.
[TestProject20.zip](https://github.com/user-attachments/files/28387448/TestProject20.zip)

It seems this problem was fixed by https://github.com/dotnet/roslyn/pull/75878.

In my local project adding `[OverloadResolutionPriority(-2)]` _didn't_ resolve the problem for me.
Things that did work:
* upgrade the compiler to 9.0.200
* add a .net9 TFM to the library project.

I also tried opening FluentAssertions and do the following edits:
* Added `[OverloadResolutionPriority(-2)]` to two `Should` overloads in `EnumAssertionsExtensions`
* Added a .net9 target to `FluentAssertions.Specs` 
* Removed the net47 target to `FluentAssertions.Specs` 
* Fixed the SDK to 9.0.117
* Changed `LangVersion` to `preview`.

`dotnet build .\Tests\FluentAssertions.Specs\FluentAssertions.Specs.csproj` fails with CS0453.

Lastly I tried building this branch into a nuget package and installing that nupkg into the local project.
I still get CS0453 when compiling with SDK 9.0.117

So I'm still unsure if this PR solves the problem.
Am I missing something?

L34T:
| FluentAssertions | 8.0.126 | 9.0.116/117     | 10.0.107 |
| ---------------- | ------- | --------------- | -------- |
| 8.7.0            | SUCCESS | SUCCESS         | SUCCESS  |
| 8.7.1            | SUCCESS | SUCCESS         | SUCCESS  |
| 8.8.0            | SUCCESS | SUCCESS         | SUCCESS  |
| 8.9.0            | SUCCESS | **FAILED (CS0453)** | SUCCESS  |
| 8.10.0           | SUCCESS | **FAILED (CS0453)** | SUCCESS  |

| FluentAssertions | 8.0.421 | 9.0.314 | 10.0.300 | 11.0.100-preview.4.26230.115 |
| ---------------- | ------- | ------- | -------- | ---------------------------- |
| 8.7.0            | SUCCESS | SUCCESS | SUCCESS  | SUCCESS |
| 8.7.1            | SUCCESS | SUCCESS | SUCCESS  | SUCCESS |
| 8.8.0            | SUCCESS | SUCCESS | SUCCESS  | SUCCESS |
| 8.9.0            | SUCCESS | **SUCCESS** | SUCCESS  | SUCCESS |
| 8.10.0           | SUCCESS | **SUCCESS** | SUCCESS  | SUCCESS |

dennisdoomen:
## Reproduction & Verification

> 🤖 *This comment was written by GitHub Copilot on behalf of @dennisdoomen.*

Copilot reproduced the CS0453 bug and verified the fix using a standalone console project targeting `net9.0` with **SDK 9.0.117** (Roslyn 4.9 / C# 13).

### Repro project (`Program.cs`)

```csharp
using System;
using FluentAssertions;

// Scenario 1: .Should() on a custom reference type without a dedicated overload
var subject = new MyClass { Name = "hello" };
ObjectAssertions assertions = subject.Should();   // <-- CS0453 on FA 8.9.0 + SDK 9
assertions.Subject.Should().BeSameAs(subject);

// Scenario 2: .Should() on an actual enum must still work
var status = Status.Active;
status.Should().Be(Status.Active);

// Scenario 3: .Should() on a nullable enum must still work
Status? nullableStatus = Status.Inactive;
nullableStatus.Should().Be(Status.Inactive);

Console.WriteLine("All scenarios passed.");

class MyClass { public string Name { get; set; } = ""; }
enum Status { Active, Inactive }
```

### Results

| FA version | SDK 9.0.117 result |
|---|---|
| **8.9.0** (NuGet) | ❌ Two CS0453 errors — `'MyClass'` on the direct `.Should()` call, and `'object'` on the chained `.Should().BeSameAs(...)`. Roslyn 4.9 filters out the `-2` priority `object` overload *before* checking the `struct` constraint, so even `assertions.Subject` (typed as `object`) triggers a second error. |
| **PR #3225 fix** (local build with `[OverloadResolutionPriority(-2)]` on both enum overloads) | ✅ Compiles and prints `All scenarios passed.` |

### Existing test suite

All 70 `EnumAssertionSpecs` and all 112 `AssertionExtensionsSpecs` pass on `net47`, `net6.0`, and `net8.0` with the fix applied. The new regression test (`Calling_should_on_a_reference_type_without_a_specific_overload_returns_object_assertions`) also passes on all three targets.

LGTM 👍

dennisdoomen:
## ⚠️ Correction: this fix does **not** resolve #3224 for the reported environment

> 🤖 *This comment was written by GitHub Copilot on behalf of @dennisdoomen.*

My earlier verification was a **false positive** — I referenced a locally-built `netstandard2.1` assembly, which happens to dodge the bug. After building the actual scenario (a `net9.0` console app referencing the **NuGet package produced by this PR**, `8.10.1-...-fix-overload-resolution-priority-enum.1`), it **still fails** with the original two CS0453 errors.

### Evidence

The user's repro project (`net9.0`, package reference) across SDKs:

| SDK | Result |
|---|---|
| **9.0.117** | ❌ CS0453 (unchanged) |
| 9.0.314 | ✅ OK |
| 10.0.300 | ✅ OK |

A `net9.0` consumer resolves the package's **`net6.0`** asset. Per-asset under **SDK 9.0.117**:

| Referenced asset | Result |
|---|---|
| **net6.0** ← what net9 apps pick | ❌ CS0453 |
| netstandard2.1 | ✅ |
| netstandard2.0 | ✅ |
| net47 | ✅ |

I confirmed via reflection that the shipped `net6.0` assembly **does** carry the fix correctly: `Should(object)` = `-2`, and both enum `Should<TEnum>` overloads = `-2`. So the priorities are right — the compiler just isn't honoring them.

### Root cause: it's a Roslyn bug in the SDK 9.0.1xx band

This is **not** something the priority annotations can fix. Demonstrated by reduction:

- A **minimal** library with the *exact same* priorities (`object`=-2, `enum<T>`=-2) compiled for `net6.0` (with the PolySharp polyfill of `OverloadResolutionPriorityAttribute`) is honored correctly under 9.0.117 → no error.
- Adding more of FluentAssertions' overload shape (the `-1` `Span`/`Memory` group + generic priority-`0` overloads) makes the **same** SDK 9.0.117 mis-resolve the call (it flips between CS0453 and CS1061 depending on the exact candidate set).
- SDK **9.0.2xx+** (9.0.314 tested) and SDK **10** honor the priorities for *all* assets, including `net6.0`.

In other words: with FA's full overload set, the 9.0.1xx compiler mis-handles `OverloadResolutionPriority` regardless of the (correct) annotations. This matches the note already in the repro's `global.json`: `// works if upgrading to 9.0.200`.

### Packaging mitigation — investigated, not a reliable fix

- The polyfill's accessibility is **not** the differentiator (the `internal` polyfill is honored fine for the netstandard/net47 assets; only `net6.0` trips under 9.0.117).
- Shipping a `net9.0` target (which would use the *real* BCL attribute, no polyfill) does make a minimal repro pass under 9.0.117 — but making FA actually target `net9.0` is non-trivial (the source currently hits unrelated `net9` conflicts: `System.Threading.ITimer` vs `FluentAssertions.Common.ITimer`, `Enumerable.Index`, etc.). A `net8.0` target would still rely on the polyfill and so likely wouldn't help.

### Recommendation

1. Keep this PR — the priority model is correct and prevents other breakages on conforming compilers.
2. But it **cannot** close #3224 on its own for users on **SDK 9.0.1xx**. The practical fix is a toolchain bump to **SDK 9.0.200+** (or SDK 10), exactly as hinted in the repro.
3. Consider documenting this in the release notes / issue: "If you still hit CS0453 on the .NET 9.0.1xx SDK, upgrade to the 9.0.2xx SDK or newer — it's a Roslyn overload-resolution-priority bug, fixed there."

If we want a true library-side fix for 9.0.1xx users, it would require restructuring the `Should` overload set to avoid the pattern that trips the older compiler (and/or adding a real net9 target), which is a larger change than this PR.

jnyrup:
How about we do no source code changes, but instead document that for a net9 test project at least SDK 9.0.200 is required when using FA 8.9.0?

dennisdoomen:
I've updated https://github.com/fluentassertions/fluentassertions/releases/tag/8.9.0 with a warning

PR #3297, opened 2026-08-09
Add HaveLineCount()/NotHaveLineCount() and ContainLine()/NotContainLine() to StringAssertions

dennisdoomen:
Implements the API proposed in #3291.

### What's new

Adds four methods to `StringAssertions<TAssertions>`, matching the proposal in the issue:

```csharp
public AndConstraint<TAssertions> HaveLineCount(int expected, string because = "", params object[] becauseArgs);
public AndConstraint<TAssertions> NotHaveLineCount(int unexpected, string because = "", params object[] becauseArgs);
public AndConstraint<TAssertions> ContainLine(string expectedLine, string because = "", params object[] becauseArgs);
public AndConstraint<TAssertions> NotContainLine(string unexpectedLine, string because = "", params object[] becauseArgs);
```

```csharp
string log = "Starting up\nConnected to database\nReady";

log.Should().HaveLineCount(3);
log.Should().ContainLine("Connected to database");
log.Should().NotContainLine("Error");
```

- Lines are split using `\r\n`, `\n`, and `\r` as line separators (all three treated as equivalent), consistent with line-ending handling already used internally elsewhere in the library.
- `ContainLine`/`NotContainLine` match whole lines only (not substrings), and guard against a `null` expected/unexpected line via `Guard.ThrowIfArgumentIsNull` (empty string is a valid line to search for).
- `HaveLineCount`/`NotHaveLineCount` fail explicitly with a `<null>`-aware message when the subject string is null, following the same pattern as `HaveLength`.

### Tests

Added spec files under `Tests/FluentAssertions.Specs/Primitives/`:
- `StringAssertionSpecs.HaveLineCount.cs`
- `StringAssertionSpecs.NotHaveLineCount.cs`
- `StringAssertionSpecs.ContainLine.cs`
- `StringAssertionSpecs.NotContainLine.cs`

Covering success/failure paths, null subjects, mixed line endings, and argument validation.

### Housekeeping

- Updated `Tests/Approval.Tests/ApprovedApi/FluentAssertions/*.verified.txt` via `AcceptApiChanges.ps1` for the new public API surface.
- Updated `docs/_pages/releases.md` and `docs/_pages/strings.md`.
- Full `FluentAssertions.Specs` suite (6003 tests) and approval tests pass locally.

Closes #3291 (pending maintainer API approval).

PR #3301, opened 2026-08-10
Support asserting on delegates that return a value task

### LINKED ISSUE #3269: Direct `Should()` support for `Func<ValueTask>` / `Func<ValueTask<T>>`
### Background and motivation

Today, the only way to get async assertion support (`Throw`/`NotThrow`/`CompleteWithinAsync`, etc.) for a `ValueTask`-returning delegate is to go through the `Awaiting` adapters (`AssertionExtensions.cs`), which wrap the `ValueTask` into a `Task` via `.AsTask()`:

```csharp
public static Func<Task> Awaiting<T>(this T subject, Func<T, ValueTask> action)
{
    return () => action(subject).AsTask();
}

public static Func<Task<TResult>> Awaiting<T, TResult>(this T subject, Func<T, ValueTask<TResult>> action)
{
    return () => action(subject).AsTask();
}
```

There is no direct `Should()` overload for `Func<ValueTask>` / `Func<ValueTask<T>>`, unlike the existing overloads for `Func<Task>` and `Func<Task<T>>`:

```csharp
public static NonGenericAsyncFunctionAssertions Should([NotNull] this Func<Task> action);
public static GenericAsyncFunctionAssertions<T> Should<T>([NotNull] this Func<Task<T>> action);
```

`ValueTask` is increasingly common in modern, performance-sensitive APIs (e.g. `IAsyncEnumerable`-adjacent code, ASP.NET Core, and general hot-path async code), so tests that assert against such members currently need the extra `Awaiting` indirection, or an explicit `.AsTask()` call, purely to reach parity with `Task`-based assertions. Adding direct `Should()` overloads for `Func<ValueTask>` / `Func<ValueTask<T>>` would remove that friction and give `ValueTask` the same first-class treatment as `Task`.

### API Proposal

```C#
public static class AssertionExtensions
{
    public static NonGenericAsyncFunctionAssertions Should(this Func<ValueTask> action);

    public static GenericAsyncFunctionAssertions<T> Should<T>(this Func<ValueTask<T>> action);
}
```

### API Usage

```C#
Func<ValueTask> act = () => ThrowValueTaskAsync();
await act.Should().ThrowAsync<InvalidOperationException>();

Func<ValueTask<int>> act2 = () => ComputeValueTaskAsync();
await act2.Should().NotThrowAsync();

Func<ValueTask> act3 = () => DelayValueTaskAsync();
await act3.Should().CompleteWithinAsync(500.Milliseconds());
```

### Alternative Designs

- Keep relying on the existing `Awaiting` extension methods combined with `.AsTask()` conversion (status quo); this works today but is less discoverable and inconsistent with the `Task` overloads.
- Instead of adding new `Should()` overloads, internally convert the `ValueTask` to a `Task` inside `NonGenericAsyncFunctionAssertions`/`GenericAsyncFunctionAssertions<T>` by accepting `Func<ValueTask>` as an additional constructor overload, so both `Should()` overloads funnel into the same assertions classes without duplicating assertion logic.

### Risks

- `Func<ValueTask>` currently does not resolve to any existing `Should()` overload, so this is purely additive and should not be a breaking change.
- `ValueTask`/`ValueTask<T>` instances must not be awaited multiple times or accessed after being awaited; the implementation needs to ensure the delegate is invoked exactly once per assertion (mirroring how `Awaiting` already handles this via `.AsTask()`).
- Slight increase in the public API surface and overload resolution complexity for `Should()`.

dennisdoomen:
## Summary

Asserting on a delegate that returns a value task previously meant going through the `Awaiting` adapter or calling `.AsTask()` by hand, purely to reach the assertions that task-returning delegates already get directly. This gives value tasks the same first-class treatment, so `ThrowAsync`, `NotThrowAsync`, `CompleteWithinAsync` and friends work straight off a `Func<ValueTask>` or `Func<ValueTask<T>>`.

Closes #3269.

## Reviewer note: this is a source-breaking change

Issue #3269 states that these delegate shapes "currently do not resolve to any existing `Should()` overload, so this is purely additive". That is not correct, and I want to flag it clearly rather than let it slip through.

`AssertionExtensions` already has a catch-all overload:

```csharp
public static FunctionAssertions<T> Should<T>(this Func<T> func)
```

A `Func<ValueTask>` binds to it with `T = ValueTask`, and a `Func<ValueTask<int>>` binds with `T = ValueTask<int>`. So today this compiles:

```csharp
Func<ValueTask> act = () => DoSomethingAsync();

act.Should().NotThrow();   // binds to FunctionAssertions<ValueTask>
```

It compiles, but it does not do what it looks like it does. `FunctionAssertions<T>` invokes the delegate and treats the returned value task as an ordinary return value. It never awaits it, so any exception thrown after the first suspension point goes unobserved and the assertion passes regardless. In other words, the existing binding is close to useless — but it is there, and people may have written it.

### What changes

C# prefers a non-generic applicable overload over a generic one, so the new `Should(this Func<ValueTask>)` now wins. Two consequences:

**1. Compile errors (`CS1061`).** The synchronous members disappear from the returned type:

| Was available | Replacement |
|---|---|
| `Throw<TException>()` | `ThrowAsync<TException>()` |
| `ThrowExactly<TException>()` | `ThrowExactlyAsync<TException>()` |
| `NotThrow()` | `NotThrowAsync()` |
| `NotThrowAfter(...)` | `NotThrowAfterAsync(...)` |

Anyone hitting this gets a loud compiler error and a one-line fix that also makes their test correct for the first time, so I think the break is worth taking.

**2. A silent behaviour change on `Subject`.** Members shared by both types — `Subject`, `BeSameAs`, `Match`, `Satisfy` — still compile but can now behave differently, because `Subject` no longer returns the delegate you passed in:

```csharp
Func<ValueTask> act = () => DoSomethingAsync();

act.Should().Subject.Should().BeSameAs(act);   // passed before, fails now
```

This one I could not avoid. The asynchronous assertions derive from `AsyncFunctionAssertions<TTask, TAssertions> where TTask : Task`, so `Subject` is a `Func<TTask>` and a `Func<ValueTask>` can never be it. Preserving the original identity would need a whole new assertions type duplicating the existing async logic, which the issue explicitly rules out in favour of reusing the existing classes. Instead the adapter is documented in `<remarks>` on both overloads and pinned down by specs, so the behaviour is intentional rather than incidental.

Note this is **source**-breaking only. No existing type, member or signature is removed, so already-compiled assemblies keep working; only recompilation is affected.

### How I verified it

I built `origin/main` in Release, compiled a scratch console app against the resulting `FluentAssertions.dll`, then swapped the reference to this branch's build and recompiled the same call sites. `act.Should().NotThrow()` compiles against `main` and fails with `CS1061` against this branch. Reflection over `FunctionAssertions<ValueTask>` versus `NonGenericAsyncFunctionAssertions` produced the member table above.

### What I would like you to decide

The `api-approved` label on #3269 was granted on the assumption that the change is additive. Given it is not, please confirm you are happy to take the break in the next release — the signatures themselves are unchanged from what was approved.

## Design notes

Value tasks are common in modern, performance-sensitive code, but there was no direct entry point for them, which made the assertions harder to discover and inconsistent with their task-based counterparts.

Two new entry points accept a value-task-returning delegate and reuse the existing asynchronous assertions, so behaviour, chaining and failure messages stay identical to the task-based ones. The returned value task is converted exactly once per invocation, which respects the rule that a value task must not be awaited more than once. A null delegate is still reported as `<null>` rather than throwing.

## Verification

- 15 new specs cover throwing, not throwing, completing within a time limit, generic results, null delegates, the delegate being invoked exactly once, and the adapted subject.
- `FluentAssertions.Specs` (6036 tests on net8.0/net6.0, 111 on net47), `FluentAssertions.Equivalency.Specs` and the API approval tests all pass.
- The release notes record the break under "Breaking Changes (for users)".
- Rebased onto the latest `upstream/main`; the only conflict was in the release notes, where this PR's "What's new" entry collided with a newly landed entry for `HaveLineCount`/`ContainLine` (#3297) — resolved by keeping both.

### DIFF
--- a/docs/_pages/releases.md
+++ b/docs/_pages/releases.md
@@ -14,6 +14,7 @@ sidebar:
 * Introduced new collection assertion methods `BeSupersetOf`, `BeProperSubsetOf` and `BeProperSupersetOf`, where an empty expected subset is treated as a valid (trivially satisfied) case for `BeSupersetOf` and `BeProperSupersetOf` instead of throwing - [#3271](https://github.com/fluentassertions/fluentassertions/pull/3271)
 * Added `BeJsonSerializable<T>()` for .NET 6+ to assert `System.Text.Json` round-tripping, with overloads for equivalency options and `JsonSerializerOptions` - [#3293](https://github.com/fluentassertions/fluentassertions/pull/3293)
 * Added `HaveLineCount`/`NotHaveLineCount` and `ContainLine`/`NotContainLine` to `StringAssertions` for asserting on the number of lines in a string and whether it contains a specific line, normalizing `\r\n`, `\n` and `\r` line endings. `ContainLine` returns an `AndWhichConstraint` exposing the matched line for further chaining - [#3297](https://github.com/fluentassertions/fluentassertions/pull/3297)
+* Added `Should()` overloads for `Func<ValueTask>` and `Func<ValueTask<T>>`, so `ValueTask`-returning delegates support the same assertions as `Task`-returning ones - [#3301](https://github.com/fluentassertions/fluentassertions/pull/3301)
 
 ### Enhancements
 * `BeEmpty` for `IEnumerable<T>` assertions now lists the first 10 items in the collection instead of only the first one - [#3198](https://github.com/fluentassertions/fluentassertions/pull/3198)
@@ -21,6 +22,21 @@ sidebar:
 ### Fixes
 * `JsonNodeAssertions.HaveProperty`/`NotHaveProperty` now correctly distinguish an absent property from one with an explicit `null` value - [#3282](https://github.com/fluentassertions/fluentassertions/pull/3282)
 * `JsonNodeAssertions.HaveProperty` no longer throws an `InvalidOperationException` when wrapped in an `AssertionScope` and invoked on a non-`JsonObject` subject - [#3295](https://github.com/fluentassertions/fluentassertions/pull/3295)
+
+### Breaking Changes (for users)
+* `Func<ValueTask>` and `Func<ValueTask<T>>` used to bind to the synchronous `Should<T>(Func<T>)` overload, which treated the value task as an ordinary return value. They now bind to the new asynchronous overloads instead. Replace `Throw`, `ThrowExactly`, `NotThrow` and `NotThrowAfter` with their `ThrowAsync`, `ThrowExactlyAsync`, `NotThrowAsync` and `NotThrowAfterAsync` counterparts, and note that `Subject` now exposes a task-based adapter rather than the original delegate - [#3301](https://github.com/fluentassertions/fluentassertions/pull/3301)
+
+## 8.10.0
+
+### What's new
+
+* Add `ComparingNullCollectionsAsEmpty` and `ComparingNullStringsAsEmpty` options to `BeEquivalentTo` - [#3202](https://github.com/fluentassertions/fluentassertions/pull/3202)
+
+### Enhancements
+
+* Fail with a descriptive error when path-based rules are used on value-semantic types - [#3187](https://github.com/fluentassertions/fluentassertions/pull/3187)
+* Significantly speed up `BeEquivalentTo` for large unordered collections - [#3189](https://github.com/fluentassertions/fluentassertions/pull/3188)
+* Include original index in extraneous item failure messages - [#3203](https://github.com/fluentassertions/fluentassertions/pull/3203)
 
 ## 8.9.0
 
--- a/Src/FluentAssertions/AssertionExtensions.cs
+++ b/Src/FluentAssertions/AssertionExtensions.cs
@@ -916,6 +916,40 @@ public static class AssertionExtensions
     }
 
     /// <summary>
+    /// Returns a <see cref="NonGenericAsyncFunctionAssertions"/> object that can be used to assert the
+    /// current <see><cref>System.Func{ValueTask}</cref></see>.
+    /// </summary>
+    /// <remarks>
+    /// Because the assertions operate on tasks, the delegate is wrapped in a <see cref="Func{Task}"/> that converts the
+    /// value task returned by every invocation into a task. The <c>Subject</c> property therefore exposes that wrapper
+    /// instead of the original delegate.
+    /// </remarks>
+    [Pure]
+    public static NonGenericAsyncFunctionAssertions Should([NotNull] this Func<ValueTask> action)
+    {
+        Func<Task> taskFunc = action is null ? null : () => action().AsTask();
+
+        return new NonGenericAsyncFunctionAssertions(taskFunc, Extractor, AssertionChain.GetOrCreate());
+    }
+
+    /// <summary>
+    /// Returns a <see cref="GenericAsyncFunctionAssertions{T}"/> object that can be used to assert the
+    /// current <see><cref>System.Func{ValueTask{T}}</cref></see>.
+    /// </summary>
+    /// <remarks>
+    /// Because the assertions operate on tasks, the delegate is wrapped in a <see cref="Func{Task}"/> that converts the
+    /// value task returned by every invocation into a task. The <c>Subject</c> property therefore exposes that wrapper
+    /// instead of the original delegate.
+    /// </remarks>
+    [Pure]
+    public static GenericAsyncFunctionAssertions<T> Should<T>([NotNull] this Func<ValueTask<T>> action)
+    {
+        Func<Task<T>> taskFunc = action is null ? null : () => action().AsTask();
+
+        return new GenericAsyncFunctionAssertions<T>(taskFunc, Extractor, AssertionChain.GetOrCreate());
+    }
+
+    /// <summary>
     /// Returns a <see cref="FunctionAssertions{T}"/> object that can be used to assert the
     /// current <see cref="Func{T}"/>.
     /// </summary>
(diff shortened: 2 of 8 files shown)


docs/_pages/releases.md
```diff
@@ -13,6 +13,7 @@ sidebar:
 * Added `ThatSatisfy` to `MethodInfoSelector` and `PropertyInfoSelector` - [#3257](https://github.com/fluentassertions/fluentassertions/pull/3257)
 * Introduced new collection assertion methods `BeSupersetOf`, `BeProperSubsetOf` and `BeProperSupersetOf`, where an empty expected subset is treated as a valid (trivially satisfied) case for `BeSupersetOf` and `BeProperSupersetOf` instead of throwing - [#3271](https://github.com/fluentassertions/fluentassertions/pull/3271)
 * Added `BeJsonSerializable<T>()` for .NET 6+ to assert `System.Text.Json` round-tripping, with overloads for equivalency options and `JsonSerializerOptions` - [#3293](https://github.com/fluentassertions/fluentassertions/pull/3293)
+* Added `Should()` overloads for `Func<ValueTask>` and `Func<ValueTask<T>>`, so `ValueTask`-returning delegates support the same assertions as `Task`-returning ones - [#3269](https://github.com/fluentassertions/fluentassertions/issues/3269)
```
jnyrup: Reference PR

docs/_pages/releases.md
```diff
@@ -22,6 +23,21 @@ sidebar:
...
+## 8.10.0
+
+### What's new
+
+* Add ComparingNullCollectionsAsEmpty and ComparingNullStringsAsEmpty options to BeEquivalentTo - [#3202](https://github.com/fluentassertions/fluentassertions/pull/3202)
+
+### Enhancements
+
+* Fail with a descriptive error when path-based rules are used on value-semantic types - [#3187](https://github.com/fluentassertions/fluentassertions/pull/3187)
+* Significantly speed up BeEquivalentTo for large unordered collections - [3189](https://github.com/fluentassertions/fluentassertions/pull/3188)
```
```suggestion
* Significantly speed up BeEquivalentTo for large unordered collections - [#3189](https://github.com/fluentassertions/fluentassertions/pull/3188)
```

docs/_pages/releases.md
```diff
@@ -22,6 +23,21 @@ sidebar:
...
 * `JsonNodeAssertions.HaveProperty` no longer throws an `InvalidOperationException` when wrapped in an `AssertionScope` and invoked on a non-`JsonObject` subject - [#3295](https://github.com/fluentassertions/fluentassertions/pull/3295)
 
+### Breaking Changes (for users)
+* `Func<ValueTask>` and `Func<ValueTask<T>>` used to bind to the synchronous `Should<T>(Func<T>)` overload, which treated the value task as an ordinary return value. They now bind to the new asynchronous overloads instead. Replace `Throw`, `ThrowExactly`, `NotThrow` and `NotThrowAfter` with their `ThrowAsync`, `ThrowExactlyAsync`, `NotThrowAsync` and `NotThrowAfterAsync` counterparts, and note that `Subject` now exposes a task-based adapter rather than the original delegate - [#3301](https://github.com/fluentassertions/fluentassertions/issues/3301)
+
+## 8.10.0
+
+### What's new
+
+* Add ComparingNullCollectionsAsEmpty and ComparingNullStringsAsEmpty options to BeEquivalentTo - [#3202](https://github.com/fluentassertions/fluentassertions/pull/3202)
```
```suggestion
* Add `ComparingNullCollectionsAsEmpty` and `ComparingNullStringsAsEmpty` options to `BeEquivalentTo` - [#3202](https://github.com/fluentassertions/fluentassertions/pull/3202)
```

docs/_pages/releases.md
```suggestion
* Significantly speed up `BeEquivalentTo` for large unordered collections - [3189](https://github.com/fluentassertions/fluentassertions/pull/3188)
```

docs/_pages/releases.md
```diff
@@ -22,6 +23,21 @@ sidebar:
 * `JsonNodeAssertions.HaveProperty`/`NotHaveProperty` now correctly distinguish an absent property from one with an explicit `null` value - [#3282](https://github.com/fluentassertions/fluentassertions/pull/3282)
 * `JsonNodeAssertions.HaveProperty` no longer throws an `InvalidOperationException` when wrapped in an `AssertionScope` and invoked on a non-`JsonObject` subject - [#3295](https://github.com/fluentassertions/fluentassertions/pull/3295)
 
+### Breaking Changes (for users)
+* `Func<ValueTask>` and `Func<ValueTask<T>>` used to bind to the synchronous `Should<T>(Func<T>)` overload, which treated the value task as an ordinary return value. They now bind to the new asynchronous overloads instead. Replace `Throw`, `ThrowExactly`, `NotThrow` and `NotThrowAfter` with their `ThrowAsync`, `ThrowExactlyAsync`, `NotThrowAsync` and `NotThrowAfterAsync` counterparts, and note that `Subject` now exposes a task-based adapter rather than the original delegate - [#3301](https://github.com/fluentassertions/fluentassertions/issues/3301)
```
```suggestion
* `Func<ValueTask>` and `Func<ValueTask<T>>` used to bind to the synchronous `Should<T>(Func<T>)` overload, which treated the value task as an ordinary return value. They now bind to the new asynchronous overloads instead. Replace `Throw`, `ThrowExactly`, `NotThrow` and `NotThrowAfter` with their `ThrowAsync`, `ThrowExactlyAsync`, `NotThrowAsync` and `NotThrowAfterAsync` counterparts, and note that `Subject` now exposes a task-based adapter rather than the original delegate - [#3301](https://github.com/fluentassertions/fluentassertions/pull/3301)
```

jnyrup (approved):

PR #3323, opened 2026-08-22
Fix potential NREs when using AssertionScope (#3178)

### LINKED ISSUE #3178: Potential NREs when using AssertionScope
### Description

Searching for this pattern `ForCondition\(([a-z]+) is not null\)\r\n(.*\r\n){1,11} +\.Then(\r\n *)?\.ForCondition\(\!?\1\!?\.` 
approximately reveals places on the form
```cs
.ForCondition(foo is not null)
...
.Then
.ForCondition(foo.)
```

When running the assertion method inside an `AssertionScope`, the `AssertionChain` doesn't stop on first error but continues until the surrounding `AssertionScope` is disposed.

So in the snippet above `foo.` might throw a `NullReferenceException`.

* `PropertyInfoAssertions`
  * `BeVirtual`
  * `NotBeVirtual`
  * `BeWritable`
  * `BeWritable(CSharpAccessModifier)`
  * `NotBeWritable`
  * `BeReadable`
  * `BeReadable(CSharpAccessModifier`
  * `NotBeReadable`
  * `Return`
  * `NotReturn`
* `TypeAssertions`
  * `HaveProperty`
  * `HaveIndexer`

### Reproduction Steps

Edit the existing `When_asserting_a_type_has_a_property_which_it_does_not_it_fails` to wrap the call to `HaveProperty` in an `AssertionScope`

```csharp
[Fact]
public void When_asserting_a_type_has_a_property_which_it_does_not_it_fails()
{
    // Arrange
    var type = typeof(ClassWithNoMembers);

    // Act
    Action act = () =>
    {
        type.Should().HaveProperty(typeof(string), "PublicProperty", "we want to test the failure {0}", "message");
    };

    // Assert
        .WithMessage("Expected ClassWithNoMembers to have a property PublicProperty of type String because we want to test the failure message, but it does not.");
}
```

### Expected behavior

No NRE should be thrown

### Actual behavior

An NRE is thrown

### Regression?

No

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

No

dennisdoomen:
Fixes #3178

## Problem

Several assertion methods followed this pattern:

```csharp
assertionChain
    .ForCondition(subject is not null)
    .FailWith("... is <null>.")
    .Then
    .ForCondition(subject.Member)   // eagerly evaluated!
    .FailWith(...);
```

C# evaluates method arguments before invoking the method, so `subject.Member` is evaluated even when `subject` is `null` and the first condition already failed. Outside of an `AssertionScope`, the first failure throws immediately, so the chain never reaches the second line. Inside an `AssertionScope`, failures are only recorded (not thrown), so execution continues to the second `.ForCondition(...)`, and evaluating `subject.Member` throws a `NullReferenceException` instead of producing a proper assertion failure message.

## Fix

Replaced the eager dereferences with null-conditional comparisons (e.g. `subject?.Member == true`), following the pattern already used in `TimeOnlyAssertions.cs`. Also made a failure message in `TypeAssertions.HaveProperty` lazy, since it referenced `propertyInfo.Name` even when `propertyInfo` could be `null`.

### Affected members
- `PropertyInfoAssertions`: `BeVirtual`, `NotBeVirtual`, `BeWritable` (both overloads), `NotBeWritable`, `BeReadable` (both overloads), `NotBeReadable`, `Return`, `NotReturn`
- `TypeAssertions`: `HaveProperty`, `HaveIndexer`

## Testing

- Added regression tests that wrap each affected assertion in an `AssertionScope`, reproducing the exact scenario from the issue.
- Full `FluentAssertions.Specs` suite passes: 5963/5963 (net8.0).
- Targeted `PropertyInfoAssertionSpecs`/`TypeAssertionSpecs` pass on net47, net6.0, and net8.0 (402/402 each).

No public API changes, so no API-approval step is required.


Tests/FluentAssertions.Specs/Types/TypeAssertionSpecs.HaveProperty.cs
```diff
@@ -60,6 +61,25 @@ public void When_asserting_a_type_has_a_property_which_it_does_not_it_fails()
...
+            // Act
+            Action act = () =>
+            {
+                using var _ = new AssertionScope();
+                type.Should().HaveProperty(typeof(string), "PublicProperty", "we want to test the failure {0}", "message");
+            };
+
+            // Assert
+            act.Should().Throw<XunitException>()
+                .WithMessage("Expected ClassWithNoMembers to have a property PublicProperty of type String because we want to test the failure message, but it does not.");
```
jnyrup: The same failure message is thrown when an NRE occurs, so having `WithMessage` is over-asserting.
Same comment for the other tests.

```suggestion
            act.Should().Throw<XunitException>();
```

docs/_pages/releases.md
```diff
@@ -23,6 +23,7 @@ sidebar:
 ### Fixes
 * `JsonNodeAssertions.HaveProperty`/`NotHaveProperty` now correctly distinguish an absent property from one with an explicit `null` value - [#3282](https://github.com/fluentassertions/fluentassertions/pull/3282)
 * `JsonNodeAssertions.HaveProperty` no longer throws an `InvalidOperationException` when wrapped in an `AssertionScope` and invoked on a non-`JsonObject` subject - [#3295](https://github.com/fluentassertions/fluentassertions/pull/3295)
+* Fixed several `PropertyInfoAssertions` methods (`BeVirtual`, `NotBeVirtual`, `BeWritable`, `NotBeWritable`, `BeReadable`, `NotBeReadable`, `Return`, `NotReturn`) and `TypeAssertions.HaveProperty`/`HaveIndexer` throwing a `NullReferenceException` instead of failing with a proper message when used inside an `AssertionScope` - [#3178](https://github.com/fluentassertions/fluentassertions/issues/3178)
```
```suggestion
* Fixed several `PropertyInfoAssertions` methods (`BeVirtual`, `NotBeVirtual`, `BeWritable`, `NotBeWritable`, `BeReadable`, `NotBeReadable`, `Return`, `NotReturn`) and `TypeAssertions.HaveProperty`/`HaveIndexer` throwing a `NullReferenceException` instead of failing with a proper message when used inside an `AssertionScope` - [#3323](https://github.com/fluentassertions/fluentassertions/pull/3323)
```

jnyrup: > throwing a `NullReferenceException` instead of failing with a proper message when used inside an `AssertionScope`

is slightly incorrect. The assertions would fail with the correct failure message, but an _intermediate_ NRE is no longer thrown.

jnyrup (changes requested):

Tests/FluentAssertions.Specs/Types/TypeAssertionSpecs.HaveProperty.cs
dennisdoomen: Good point, done — removed `.WithMessage(...)` from this test and the other 11 regression tests, keeping only `Throw<XunitException>()` since these tests are only about confirming no NRE occurs, not about the message content.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

docs/_pages/releases.md
dennisdoomen: Updated the link to reference this PR (#3323) instead of the issue.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

dennisdoomen: Reworded to clarify that the assertions always produced the correct failure message — what's fixed is an *intermediate* `NullReferenceException` that was thrown before that message could be produced.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

jnyrup (approved):

